PR #3, opened 2025-03-11
feat: add runtime object module

Skarlso:

PR #5, opened 2025-03-11
feat: add config v1

Skarlso:

After https://github.com/open-component-model/open-component-model/pull/3 is merged.


Skarlso:
Might actually need to make this 2 prs so I can merge and tag config.

Skarlso:
Will point to a hash instead.

PR #10, opened 2025-03-13
feat: runtime package

jakobmoellerdev:

A base runtime package that contains groupable, versioned types and registries that are able to be embedded in any Struct which can be used for type registrations. It uses reflection instead of code generation right now.

supercedes https://github.com/open-component-model/open-component-model/pull/3


bindings/go/runtime/type.go
```diff
@@ -0,0 +1,166 @@
...
+	var namePart, versionPart string
+	if len(parts) == 1 {
+		namePart = parts[0] // Unversioned format
+	} else {
+		namePart, versionPart = parts[0], parts[1] // Versioned format
+	}
+
+	// Split name part to extract group (if present)
+	nameParts := strings.Split(namePart, ".")
+	if len(nameParts) < 1 {
```
Skarlso, line 63: split always returns a element, even if there are no elements, it returns one empty element.

bindings/go/runtime/registry.go
```diff
@@ -0,0 +1,221 @@
...
+	for _, opt := range opts {
+		opt(reg)
+	}
+	return reg
+}
+
+type SchemeOption func(*Scheme)
+
+// WithAllowUnknown allows unknown types to be created.
+func WithAllowUnknown(allowUnknown bool) SchemeOption {
```
Skarlso, line 37: FWIW I would just remove the allow since this is a bool field, convention is that `WithAllowUnknown` sets it to true.

bindings/go/runtime/registry.go
```diff
@@ -0,0 +1,221 @@
...
+	// construct by full type
+	proto, exists := r.types[typ]
+	if exists {
+		t := reflect.TypeOf(proto)
+		for t.Kind() == reflect.Ptr {
+			t = t.Elem()
+		}
+		object = reflect.New(t).Interface()
+
+		return object.(Typed), nil
```
Skarlso, line 156: should we type check here? Or is this just always true... ?

jakobmoellerdev: I think its a bug. It will return any now because the Typed interface is not used anymore as binding (I changed to any and reflect)

bindings/go/runtime/type.go
jakobmoellerdev: mind telling me a case which I can test against that fails? I thought I covered most cases X)

bindings/go/runtime/registry.go
jakobmoellerdev: removed type assertion

bindings/go/runtime/registry.go
```diff
@@ -0,0 +1,221 @@
...
+	return func(registry *Scheme) {
+		registry.allowUnknown = allowUnknown
+	}
+}
+
+func (r *Scheme) Clone() *Scheme {
+	r.mu.RLock()
+	defer r.mu.RUnlock()
+	clone := NewScheme(WithAllowUnknown(r.allowUnknown))
+	maps.Copy(r.types, clone.types)
```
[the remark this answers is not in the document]
jakobmoellerdev: fixed clone and tested

bindings/go/runtime/registry.go
jakobmoellerdev: removed bool

bindings/go/runtime/type.go
Skarlso: Well, you could say if it's `len(nameParts) == 1 && nameParts[0] == ""` if that happens then there is definitely a problem. :)

Skarlso: Oh you mean an actual case. :D

To trigger the above, you are missing a testcase like this: `/v1`. If someone forgets to write the name part. :)

This would result here in an empty namePart but with the length of `1`.

jakobmoellerdev: Ah that case is actually covered by checking if the Name is filled (line 80). I added a test case to make sure.

Skarlso: Ah good catch, okay dokay.

Skarlso (approved):
Nice work!

PR #12, opened 2025-03-14
chore(ci): introduce linting via golangci-lint

jakobmoellerdev:

We already have some CI so lets also setup linting!


.github/workflows/ci.yml
```diff
@@ -0,0 +1,166 @@
...
+            ${{ matrix.module }}
+            .github/config/golangci.yml
+      - name: Setup Go ${{ matrix.go-version }}
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: '${{ github.workspace }}/${{ matrix.module }}/go.mod'
+      - name: golangci-lint ${{ matrix.modules }}
+        uses: golangci/golangci-lint-action@v6
+        with:
+          version: ${{ env.GOLANGCI_LINT_VERSION }}
```
hilmarf, line 110: why not using the new go 1.24 feature: `tool`s?
e.g. https://github.com/open-component-model/ocm/blob/main/go.mod#L532

jakobmoellerdev: https://golangci-lint.run/welcome/install/#install-from-sources

There are various caveats and golangci-lint does not recommend installing from source for this reason.

actually the env variable right now isnt used because it uses the latest version of golangci lint

Skarlso:
31 commits? :D

Skarlso (approved):

jakobmoellerdev:
Pshhhhhhh

PR #15, opened 2025-03-17
feat: create generation boilerplate coding

jakobmoellerdev:
This adds a generator cli for GetType as well as the kubernetes deep copy generator running automatically on all modules.

This:

```go
// +k8s:deepcopy-gen=true
//
//go:generate ocmtypegen $GOFILE
type SampleType struct {
}

```

generates `zz_generated.deepcopy.go` and `zz_generated.type.ocm.software.go`

code generators used to prepare for the descriptor package as part of https://github.com/open-component-model/ocm-project/issues/443


Skarlso (dismissed):

Taskfile.yml
```diff
@@ -11,10 +11,84 @@ includes:
...
+    vars:
+      TARGET_VERSION: 1.64.7
+      CURRENT_VERSION:
+        sh: 'tmp/bin/golangci-lint version --format short || echo "0.0.0"'
+    status:
+      - '{{ contains .TARGET_VERSION .CURRENT_VERSION }}'
     cmds:
-      - task: bindings/go/descriptor:test
-      - task: bindings/go/runtime:test
+      - curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/HEAD/install.sh | sh -s -- -b {{ .ROOT_DIR }}/tmp/bin v{{ .TARGET_VERSION }}
```
frewilhelm, line 94: The link points to the latest version of the script and is directly piped into a shell. If the script is updated with malicious code we wouldn't notice. 

Usually this should not be a problem, but it increases the attack surface

bindings/go/generator/ocmtypegen/main.go
```diff
@@ -0,0 +1,123 @@
...
+	"go/token"
+	"log"
+	"os"
+	"path/filepath"
+	"strings"
+)
+
+// Type struct definition
+// It must be embedded in other structs to generate GetType
+
```
frewilhelm, line 16:
```suggestion
```

Is this intended?

Taskfile.yml
jakobmoellerdev: I would like to take this as a calculated risk since its for local testing and CI on isolated runners only. Acceptable to you?

frewilhelm: Maybe we can meet in between

```suggestion
      - curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/refs/tags/v{{ .TARGET_VERSION }}/install.sh | sh -s -- -b {{ .ROOT_DIR }}/tmp/bin v{{ .TARGET_VERSION }}
```

This would at least decrease the attack surface a bit, as the version is pinned (and by this, also an official release).

jakobmoellerdev: Works for me 👍 :)

bindings/go/generator/ocmtypegen/main.go
jakobmoellerdev: I removed it, it was a leftover

Taskfile.yml
```diff
@@ -11,10 +13,40 @@ includes:
...
+    cmds:
+      - task: 'bindings/go/generator:ocmtypegen/generate'
+      - task: 'tools:controller-gen/generate-deepcopy'
+
+  ensure-clean:
+    desc: "Ensure the project is clean by checking for uncommitted changes"
     cmds:
-      - task: bindings/go/descriptor:test
-      - task: bindings/go/runtime:test
+      - git diff --exit-code
```
frewilhelm, line 51: I do not understand the purpose of `git diff --exit-code` here. If we want to check for uncommitted changes `git status --porcelain` should be enough.

frewilhelm: I don't really understand what the purpose of this task is.. 

If it is returning an exit-code, if files were modified - then yes, `git diff --exit-code` will do that for you.
If you want to ensure that "the project is clean" and has no "uncommitted changes", then `git diff --exit-code` won't be enough as it does not consider untracked files. 

You could also consider such an approach
```sh
gitStatus="$(git status --porcelain)"
if [[ -z "${gitStatus}" ]]; then
   exit 0
fi
echo "${gitStatus}"
exit 1
```

frewilhelm: Is this task in any automated programs or is it used by humans to get some info about the project? Is the exit-code actually required for something?

jakobmoellerdev: Im not sure if you read fully through the PR now because it is quite easy to see that the github action for checking code generation uses this task. 

Also regarding the purpose of the task: I think it should discover diffs introduced by code generators. Looking at that again im not sure if I can omit procelain because the code generation can generates new files potentially.

That said, this is something that is likely only relevant in PRs and in the submission, so I think I will kill it from the taskfile and introduce a script in the github action only.

jakobmoellerdev: Changed so that only the github action contains the diff check and the task is only used for generation (because thats something we want to do locally too)

frewilhelm: > Im not sure if you read fully through the PR now because it is quite easy to see that the github action for checking code generation uses this task.

I overlooked that - sorry. Thanks for pointing it out.

Yes, as previously said `git diff` does not track untracked files and `git status` does not return an exit code. I think there is no pretty solution for that. Either go with both commands or check if `git status --porcelain` returns anything and then return an exit code yourself :/

bindings/go/generator/ocmtypegen/test/type.go
```diff
@@ -0,0 +1,12 @@
+package test
+
+import (
+	"ocm.software/open-component-model/bindings/go/runtime"
+)
+
+// +k8s:deepcopy-gen=true
+//
+//go:generate ocmtypegen $GOFILE
```
frewilhelm, line 9: What is `$GOFILE` about?

Initially I thought `$GOFILE` should actually be a go-file, e.g. `filename.go` (based on the first `log.Fatal` from the `main` method). But seeing it here as well confuses me. 

Isn't obvious that it is a `$GOFILE` through the file extension `.go`?

frewilhelm: Nevermind, `$GOFILE` is a variable set by `go generate` and contains the base name of the file

frewilhelm (approved):
lgtm

PR #17, opened 2025-03-17
feat: introduce runtime identities for usage in OCM

jakobmoellerdev:

An Identity is a map that represents a set of attributes that uniquely identity arbitrary resources. It is used in various places in Open Component Model to uniquely identity objects such as resources or components.

depends on https://github.com/open-component-model/open-component-model/pull/15

used for descriptor+config package as part of https://github.com/open-component-model/ocm-project/issues/443


bindings/go/runtime/identity_match.go
```diff
@@ -0,0 +1,93 @@
...
+// Match delegates to the IdentityMatchingFn.
+func (f IdentityMatchingFn) Match(a, b Identity) bool {
+	return f(a, b)
+}
+
+// IdentityMatchesPath returns true if the identity a matches the subpath of the identity b.
+// If the path attribute is not set in either identity, it returns true.
+// If the path attribute is set in both identities,
+// it returns true if the path attribute of b contains the path attribute of a.
+// For more information, check path.Match
```
Skarlso, line 27: This should outline why it's deleting attributes.

bindings/go/runtime/identity_match.go
```diff
@@ -0,0 +1,93 @@
+package runtime
+
+import (
+	"maps"
+	"path"
+)
+
+// IdentityMatchingFn is a function that takes two identities and returns if they match.
+// It is expected that the function can modify the identities in place.
+// If a comparison is absolute, the function can choose to delete attributes.
```
Skarlso, line 10: This is not a good enough description for me, sorry. :D Why does an IDENTITY MATCH have to delete attributes or _could_? This is like saying, `Equal` might delete some keys while it's checking. I'm not sure why this is a possibility. This seems like an error waiting to happen.

jakobmoellerdev: Explained eeven further. You conveniently left out the last comment line that was also hinting at the chaining :D

Skarlso (approved):

PR #18, opened 2025-03-17
feat: blob + filesystem abstraction

jakobmoellerdev:
```
// Package blob provides various interfaces and types for working with Binary Large Object's (BLOBs).
//
// When working with BLOBs through this package, it is important to understand the following concepts:
//   - ReadOnlyBlob: An interface that represents a read-only BLOB.
//   - Blob: An interface that represents a BLOB that can be both read and written to.
//   - SizeAware: An interface that represents any arbitrary object that can be sized.
//   - DigestAware: An interface that represents any arbitrary object that can be digested.
//   - MediaTypeAware: An interface that represents any arbitrary object that can have a media type.
//
// These core concepts make it possible to describe any arbitrary data without actually introspecting it.
// The Blob is the fundamental interface that can be used by implementations to provide abstractions over
// dynamic content-ware, size-aware and content-proof data.
//
// Additionally, the package provides convenience implementations of typical blob scenarios:
//   - EagerBufferedReader: A reader that eagerly buffers the data it reads, to make it easier to
//     introspect the data. This is useful for cases where the size and digest of the data are not known
//     in advance but still need to be computed (through buffering in-memory instead of directly streaming).
//
// Note that filesystem-backed blobs are located in a separated sub-package.
package blob
```

```
// Package filesystem provides various interfaces and types for working with Binary Large Object's (BLOBs) based on a FileSystem.
//
// When working with BLOBs through this package, the main interaction point will be the Blob.
//
// Additionally, the package provides convenience implementations of typical blob scenarios:
//   - GetBlobFromOSPath
//   - CopyBlobToOSPath
package filesystem
```

used for descriptor+config package as part of https://github.com/open-component-model/ocm-project/issues/443


bindings/go/blob/buffered_reader.go
```diff
@@ -0,0 +1,137 @@
...
+	if newSize := uint64(b.buf.Len()); newSize > b.size { //nolint:gosec // G115 can never be negative
+		b.size = newSize
+	}
+	b.loaded = true
+	return nil
+}
+
+func (b *EagerBufferedReader) Loaded() bool {
+	b.mu.RLock()
+	defer b.mu.RUnlock()
```
frewilhelm, line 70: Why is the lock necessary? We neither load nor alter the data

bindings/go/blob/buffered_reader.go
```diff
@@ -0,0 +1,137 @@
...
+	if err := b.LoadEagerly(); err != nil {
+		return -1, err
+	}
+	return b.buf.Read(p)
+}
+
+func (b *EagerBufferedReader) Close() error {
+	closeable, ok := b.reader.(io.Closer)
+	if !ok {
+		return nil
```
frewilhelm, line 84: Shouldn't we return an error if the object cannot be closed?
It seems counter-intuitive to return no error, when a `Close()` method is called on an object that does not provide the `io.Closer` interface

bindings/go/blob/buffered_reader.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+	return closeable.Close()
+}
+
+func (b *EagerBufferedReader) Digest() (string, bool) {
+	if b.LoadEagerly() != nil {
+		return "", false
```
frewilhelm, line 91: Why don't we return the error from `b.LoadEagerly()`?

bindings/go/blob/buffered_reader.go
frewilhelm: I guess to make sure that the object is not processed by `b.LoadEagerly()` while calling `Loaded()`

bindings/go/blob/filesystem/filesystem.go
```diff
@@ -0,0 +1,130 @@
...
+		return nil, fmt.Errorf("path is not a directory: %s", base)
+	}
+	return &osFileSystem{base: base, flag: flag}, nil
+}
+
+type osFileSystem struct {
+	base string
+
+	flagMu sync.RWMutex
+	flag   int
```
frewilhelm, line 66: I would expect some comment what kind of flag is meant here. From reading the code, I got the impression that `flag` represents conditions.

frewilhelm (commented):
Very nice documentation! It makes the code a lot easier to understand.

If I am not mistaken, this PR introduces changes, like `CopyBlobToOSPath()`, that probably should be tested on different OS. What do you think?

bindings/go/blob/buffered_reader.go
jakobmoellerdev: If the underlying reader is not closeable, then this is the same as not closing (NopCloser comes to mind). I think it is fine to not error here. WDYT?

bindings/go/blob/buffered_reader.go
jakobmoellerdev: because the Digest Interface is not designed to have dynamic error returns. Meaning that if the loading failed the Digest is not known

bindings/go/blob/filesystem/filesystem.go
jakobmoellerdev: I think I can rename to mode to make it clear that this is about the filesystem mode. However its not only mode because its a bitmask so it can also contain other attributes.

jakobmoellerdev:
> Very nice documentation! It makes the code a lot easier to understand.
> 
> If I am not mistaken, this PR introduces changes, like `CopyBlobToOSPath()`, that probably should be tested on different OS. What do you think?

I do have bindings/go/blob/filesystem/file_unix_test.go for the Testing on unix, I am not sure if windows is supported without changes, would have to add stuff here I believe (like binding to windows named pipes). How do you think this should be separated?

bindings/go/blob/buffered_reader.go
frewilhelm: Yes, that is true. This means that we trust the user to know what he's doing :D I think thats fair

bindings/go/blob/buffered_reader.go
frewilhelm: I am "worried" about users that use `Digest()` without calling `LoadEagerly()` beforehand, because they would never see an error - just an empty digest that is not known. One could argue that `b.LoadEagerly()` will always be called beforehand, when calling `Read()`. hach, I am unsure, but I guess this works and we can adjust, if it is too unclear for users.

bindings/go/blob/filesystem/filesystem.go
frewilhelm: A comment what `flag` should contains would be enough for me

frewilhelm:
> How do you think this should be separated?

I am a bit worried that we have to check for the OS and provide the respective methods for each OS (Windows/unix should be enough for the moment). However, I never developed for windows before, so I cannot estimate if and what is necessary.

We could start with running our GHA CI on windows and ubuntu. This should give us an idea, if any test is failing on another OS.

(This is assuming that the library should also run on windows. However, I don't know if we want to support windows at this point. We could argue that we focus on unix first.)

jakobmoellerdev:
I think in the previous library we did have code on windows but it was never ran in CI (for obvious complication reasons). I think we can keep windows as a second class citizen but it will eventually bite us with regressions if we are not careful.

I think this is the most low level code we have, so for me it is fine if this is platform specific. We always have the option to work with files, which would be OS-agnostic. Only if we want to leverage kernel buffers and pipes will os-dependency become an issue.

bindings/go/blob/buffered_reader.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+	if b.reader != nil {
+		b.reader = io.MultiReader(b.buf, b.reader)
+	}
+	dig, err := digest.FromReader(io.TeeReader(b.reader, b.buf))
+	if err != nil {
+		return err
+	}
+	b.digest = dig.String()
+	if newSize := int64(b.buf.Len()); newSize > b.size {
```
frewilhelm, line 61:
```suggestion
	if newSize := int64(b.buf.Len()); newSize != b.size {
```

Why is `newSize` only set when it is bigger then the current `b.size`. What happens when the underlying blob changes and the blob is now smaller? Is such a scenario realistic?

jakobmoellerdev: I think the size of a buffered reader should never decrease, only increase. This is because buffers are by definition allocation based and it would be strange to suddenly have a shrinking buffer. Thats also why I wrote

```
	// SetPrecalculatedSize sets the size of the blob if it is known in advance.
	// It MUST be safe for concurrent use, serializing as necessary.
	// It MUST be safe to call multiple times, with the same or greater different size.
	// Once called, HasPrecalculatedSize MUST return true.
	// If called with a size less than the current size, it MUST ignore this size.
	SetPrecalculatedSize(size int64)
```

frewilhelm: okay, that makes sense

frewilhelm (approved):

bindings/go/blob/filesystem/filesystem.go
jakobmoellerdev: Added a comment with a reference to OpenFile which is where I got the name of the parameter from

PR #20, opened 2025-03-17
feat: add descriptor package

Skarlso:


Skarlso:
fixing the linter and unit test issues

bindings/go/descriptor/descriptor.go
```diff
@@ -0,0 +1,177 @@
...
+}
+
+type SignatureSpec struct {
+	Algorithm string `json:"algorithm"`
+	Value     string `json:"value"`
+	MediaType string `json:"mediaType"`
+	Issuer    string `json:"issuer,omitempty"`
+}
+
+type Label struct {
```
jakobmoellerdev, line 174: This is missing the signature relevance bool

bindings/go/descriptor/descriptor.go
```diff
@@ -0,0 +1,177 @@
...
+	return json.Marshal(a.Typed)
+}
+
+type Reference struct {
+	Name          string            `json:"name"`
+	ExtraIdentity map[string]string `json:"extraIdentity,omitempty"`
+	Component     string            `json:"componentName"`
+	Version       string            `json:"version"`
+	Digest        Digest            `json:"digest,omitempty"`
+	Labels        []*Label          `json:"labels,omitempty"`
```
jakobmoellerdev, line 127: I think slices of pointers are usually a bad idea

bindings/go/descriptor/descriptor.go
```diff
@@ -0,0 +1,177 @@
...
+func (r *Source) GetIdentity() map[string]string {
+	m := maps.Clone(r.ExtraIdentity)
+	if m == nil {
+		m = make(map[string]string)
+	}
+	m["name"] = r.Name
+	return m
+}
+
+type Access struct {
```
jakobmoellerdev, line 108: I think this is no longer needed with the new runtime package

bindings/go/descriptor/descriptor.go
```diff
@@ -0,0 +1,177 @@
...
+	return runtime.Identity{
+		"name":    c.Name,
+		"version": c.Version,
+	}
+}
+
+type Resource struct {
+	ObjectMeta    `json:",inline"`
+	ExtraIdentity map[string]string `json:"extraIdentity,omitempty"`
+	SourceRefs    []*SourceRef      `json:"sourceRefs,omitempty"`
```
jakobmoellerdev, line 75: Pointer in slice is not nice, it rhymes! :D

bindings/go/descriptor/descriptor.go
```diff
@@ -0,0 +1,177 @@
...
+	ComponentIdentity  `json:",inline"`
+	Labels             []*Label                `json:"labels,omitempty"`
+	RepositoryContexts []*runtime.Unstructured `json:"repositoryContexts,omitempty"`
+	Provider           string                  `json:"provider"`
+	Resources          []*Resource             `json:"resources,omitempty"`
+	Sources            []*Source               `json:"sources,omitempty"`
+	References         []*Reference            `json:"componentReferences,omitempty"`
+}
+
+type ComponentIdentity struct {
```
jakobmoellerdev, line 42: This looks like its not a component Identity. Rather a NameVersion.

bindings/go/descriptor/descriptor.go
```diff
@@ -0,0 +1,177 @@
...
+	IdentitySelector map[string]string `json:"identitySelector,omitempty"`
+	Labels           map[string]string `json:"labels,omitempty"`
+}
+
+type Meta struct {
+	Version string `json:"schemaVersion"`
+}
+
+type ObjectMeta struct {
+	ComponentIdentity `json:",inline"`
```
jakobmoellerdev, line 140: This surely should nto be a component identity for every object

bindings/go/descriptor/descriptor.go
```diff
@@ -0,0 +1,177 @@
...
+	ExtraIdentity map[string]string `json:"extraIdentity,omitempty"`
+	Component     string            `json:"componentName"`
+	Version       string            `json:"version"`
+	Digest        Digest            `json:"digest,omitempty"`
+	Labels        []*Label          `json:"labels,omitempty"`
+}
+
+type SourceRef struct {
+	IdentitySelector map[string]string `json:"identitySelector,omitempty"`
+	Labels           map[string]string `json:"labels,omitempty"`
```
jakobmoellerdev, line 132: This is not the label struct we use everywhere else

bindings/go/descriptor/descriptor.go
```diff
@@ -0,0 +1,177 @@
...
+	Name          string            `json:"name"`
+	ExtraIdentity map[string]string `json:"extraIdentity,omitempty"`
+	Component     string            `json:"componentName"`
+	Version       string            `json:"version"`
+	Digest        Digest            `json:"digest,omitempty"`
+	Labels        []*Label          `json:"labels,omitempty"`
+}
+
+type SourceRef struct {
+	IdentitySelector map[string]string `json:"identitySelector,omitempty"`
```
jakobmoellerdev, line 131: I think this is not a proper selector. we probably need to add that

bindings/go/descriptor/extensions/local_access.go
```diff
@@ -0,0 +1,42 @@
...
+	// to optionally determine a globally referenceable access according
+	// to the OCI distribution spec. The result will be stored
+	// by the repository in the field ImageReference.
+	// The value is typically an OCI repository name optionally
+	// followed by a colon ':' and a tag
+	ReferenceName string `json:"referenceName,omitempty"`
+}
+
+const (
+	LocalBlobUploadType = "uploader.localblob"
```
jakobmoellerdev, line 37: We should leave this type out because the uploader type is specific to the technology now.

bindings/go/descriptor/extensions/scheme.go
```diff
@@ -0,0 +1,24 @@
...
+var Scheme = runtime.NewScheme()
+
+func init() {
+	MustAddToScheme(Scheme)
+}
+
+func MustAddToScheme(scheme *runtime.Scheme) {
+	obj := &LocalBlob{}
+	scheme.MustRegister(obj, "v1")
+	scheme.MustRegister(obj, "")
```
jakobmoellerdev, line 14: This looks off. We should register it as ungroupedunversioned here probably

bindings/go/descriptor/extensions/scheme.go
```diff
@@ -0,0 +1,24 @@
...
+	scheme.MustRegisterWithAlias(
+		obj,
+		runtime.NewUngroupedUnversionedType(LocalBlobAccessType),
+		runtime.NewUngroupedUnversionedType(LocalBlobAccessTypeV1),
+	)
+	upload := &LocalBlobUpload{}
+	scheme.MustRegister(upload, "v1")
+	scheme.MustRegister(upload, "")
```
jakobmoellerdev, line 22: Same here

bindings/go/descriptor/descriptor.go
Skarlso: What is that? :O :D

bindings/go/descriptor/descriptor.go
Skarlso: Everything in here is slices of pointers. Do you want to change everything?

bindings/go/descriptor/descriptor.go
Skarlso: Indeed. Do we even need this? 🤔 Can't this just be `runtime.Identity`?

bindings/go/descriptor/extensions/scheme.go
Skarlso: I will just delete this entire extension, not?

bindings/go/descriptor/descriptor.go
Skarlso: What do you mean? How does a proper selector look like? Do you mean like a matcher interface?

bindings/go/descriptor/descriptor.go
```diff
@@ -0,0 +1,177 @@
...
+
+// cli needs sample of transfer (source repo - target repo + parsing)
+// once the command starts up, load plugins, get capabilities of plugins, match repository type from the command
+// download component version -> hands me descriptor
+// use descriptor to parse, look for the target plugin and use transfer component version
+//
+
+type Descriptor struct {
+	Meta       *Meta        `json:"meta"`
+	Component  *Component   `json:"component"`
```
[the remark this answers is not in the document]
Skarlso: Yah. Agreed.

bindings/go/descriptor/descriptor.go
jakobmoellerdev: A bool indicating if a label should or should not be part of a potential hash for the component version signature

bindings/go/descriptor/descriptor.go
jakobmoellerdev: I think the purpose was originally that it must be exactly name and version and can be converted to an identity. And then the rest of the fields would be extra identity

bindings/go/descriptor/descriptor.go
jakobmoellerdev: I double checked, the Spec indeed only accounts for a string map, so ignore that :D I thought it needs some more complex logic

bindings/go/descriptor/extensions/scheme.go
jakobmoellerdev: I think a LocalBlob is still necessary as an access type no?

bindings/go/descriptor/descriptor.go
jakobmoellerdev: See https://github.com/open-component-model/pocm/pull/37/files#diff-d1709647f6bb5514e00c6b06d47ffb39e60f22a1996b5938127df5009f121f15

bindings/go/descriptor/extensions/scheme.go
Skarlso: Doesn't it feel wrong to have that in the descriptor package? 🤔

bindings/go/descriptor/descriptor.go
Skarlso: Isn't this copied from OCM? If we change this, won't we break existing component versions?

jakobmoellerdev: Its not directly copied. Its mostly a simplified version of the existing descriptor packages

bindings/go/descriptor/extensions/scheme.go
jakobmoellerdev: I think thats why it was under extension. Do you have a better place for it?

Skarlso: I don't because putting it into another package would be not okay. But extension also feels wrong, because it's such a crucial part. Meh, it's okay. Let's leave it here, if you are fine with the location.

bindings/go/descriptor/descriptor.go
```diff
@@ -0,0 +1,139 @@
...
+	SourceRefs    []SourceRef       `json:"sourceRefs,omitempty"`
+	Type          string            `json:"type"`
+	Relation      string            `json:"relation"`
+	Access        runtime.Raw       `json:"access"`
+	Digest        *Digest           `json:"digest,omitempty"`
+	Size          int64             `json:"size,omitempty"`
+	CreationTime  string            `json:"creationTime,omitempty"`
+}
+
+func (r *Resource) GetIdentity() map[string]string {
```
jakobmoellerdev, line 55: Still on a pointer

bindings/go/descriptor/descriptor.go
```diff
@@ -0,0 +1,139 @@
...
+}
+
+type Source struct {
+	ObjectMeta    `json:",inline"`
+	ExtraIdentity map[string]string `json:"extraIdentity,omitempty"`
+	Type          string            `json:"type"`
+	Access        runtime.Raw       `json:"access"`
+}
+
+func (r *Source) GetIdentity() map[string]string {
```
jakobmoellerdev, line 71: Still on a pointer

bindings/go/descriptor/local_access.go
```diff
@@ -0,0 +1,33 @@
...
+	runtime.Type `json:"type"`
+	// LocalReference is the repository local identity of the blob.
+	// it is used by the repository implementation to get access
+	// to the blob and if therefore specific to a dedicated repository type.
+	LocalReference string `json:"localReference"`
+	// MediaType is the media type of the object represented by the blob
+	MediaType string `json:"mediaType"`
+	// GlobalAccess is an optional field describing a possibility
+	// for a global access. If given, it MUST describe a global access method.
+	GlobalAccess *runtime.Raw `json:"globalAccess,omitempty"`
```
jakobmoellerdev, line 24: Still on a pointer (maybe needed? unsure)

bindings/go/descriptor/scheme.go
```diff
@@ -0,0 +1,16 @@
...
+
+var Scheme = runtime.NewScheme()
+
+func init() {
+	MustAddToScheme(Scheme)
+}
+
+func MustAddToScheme(scheme *runtime.Scheme) {
+	obj := &LocalBlob{}
+	scheme.MustRegisterWithAlias(obj, runtime.NewUngroupedVersionedType(LocalBlobAccessType, LocalBlobAccessTypeV1))
```
jakobmoellerdev, line 15: Missing a group name

jakobmoellerdev, line 15: The version is the complete type, this would currently fail

jakobmoellerdev (changes requested):
The entire Descriptor structs are lacking comments and documentation . a doc.go would also be nice.

Apart from that and some smaller things, LGTM 💯

bindings/go/descriptor/local_access.go
Skarlso: It kind-a has to be for a propert omit empty.. 🤔 Otherwise the checks are going to be crazy not to mention the resulting JSON will have some empty values instead of values completely not being their. Since Raw is has an embedded type field.

jakobmoellerdev (commented):
I still think the descriptor package should have a `doc.go` and the comments should have examples attached to them so it is easier to work with in code. WDYT?

Skarlso:
Ah yeah. That makes sense.

jakobmoellerdev (approved):

PR #21, opened 2025-03-18
feat: DAG Library used by KRO adopted for OCM

jakobmoellerdev:
We want to use the KRO Graph Modelling Feature for Dependency Modelling while transferring resources. For this we are copying and adapting KROs take on Resource Graphs with a few adjustments:

1. We introduce generic capability to make the graphs work with anything in cmp.Ordered
2. We introduce InDegree/OutDegree to be able to selectively traverse the graph, by processing a graph through the indegree/outdegree.

This is in preparation for our transfer library, which will need the directed acyclic graph capabilities for proper dependency management and ordering


bindings/go/dag/dag_test.go
```diff
@@ -0,0 +1,194 @@
...
+				}
+			}
+
+			order, err := d.TopologicalSort()
+			r.NoError(err, "error sorting the graph")
+
+			expected := strings.Split(g.Want, ",")
+			r.ElementsMatch(order, expected, "unexpected result from TopologicalSort for nodes=%q edges=%q, got %q, want %q", g.Nodes, g.Edges, order, expected)
+
+			// checkValidTopologicalOrder(t, d, order)
```
frewilhelm, line 191: What about this out-commented function?

bindings/go/dag/dag.go
```diff
@@ -0,0 +1,328 @@
...
+	// In kro, this would be the children of a resource.
+	Edges map[T]map[string]any
+}
+
+// DirectedAcyclicGraph represents a directed acyclic graph.
+type DirectedAcyclicGraph[T cmp.Ordered] struct {
+	// Vertices stores the nodes in the graph
+	Vertices  map[T]*Vertex[T]
+	OutDegree map[T]int // In-degree of each vertex (number of outgoing edges)
+	InDegree  map[T]int // Out-degree of each vertex (number of incoming edges)
```
frewilhelm, line 50: Are the comments correct?

bindings/go/dag/dag.go
```diff
@@ -0,0 +1,328 @@
...
+		// Only initialize the map if the edge was added
+		fromNode.Edges[to] = map[string]any{}
+		// Only increment the out-degree and in-degree if the edge was added
+		d.OutDegree[from]++
+		d.InDegree[to]++
+
+		// Check if the graph is still a DAG
+		hasCycle, cycle := d.HasCycle()
+		if hasCycle {
+			// Ehmmm, we have a cycle, let's remove the edge we just added
```
frewilhelm, line 151: :D

bindings/go/dag/dag.go
```diff
@@ -0,0 +1,328 @@
...
+	for node, degree := range d.InDegree {
+		if degree == 0 {
+			roots = append(roots, node)
+		}
+	}
+	return roots
+}
+
+func (d *DirectedAcyclicGraph[T]) TopologicalSort() ([]T, error) {
+	if cyclic, nodes := d.HasCycle(); cyclic {
```
frewilhelm, line 180: Can this happen? Don't we check for cycles every time we add a new edge and delete the edge if it is a cycle

jakobmoellerdev: It is theoretically possible that someone adds edges without cycle protection by direct access. Then this can be cyclic. I think there is a test for that somewhere :D

bindings/go/dag/dag.go
jakobmoellerdev: Erhhhhhh no I think I mixed them up, mb :D thanks for the hint

bindings/go/dag/dag.go
```diff
@@ -0,0 +1,328 @@
...
+	"maps"
+	"slices"
+	"sort"
+	"strings"
+)
+
+var ErrSelfReference = fmt.Errorf("self-references are not allowed")
+
+// Vertex represents a node/vertex in a directed acyclic graph.
+type Vertex[T cmp.Ordered] struct {
```
Skarlso, line 34: I wonder. Shouldn't we make this concurrent safe with a mutex?

bindings/go/dag/dag.go
```diff
@@ -0,0 +1,328 @@
...
+}
+
+func (d *DirectedAcyclicGraph[T]) TopologicalSort() ([]T, error) {
+	if cyclic, nodes := d.HasCycle(); cyclic {
+		return nil, &CycleError{
+			Cycle: nodes,
+		}
+	}
+
+	visited := make(map[T]bool)
```
Skarlso, line 186: I'd prefer using `struct{}{}`. But not a hard stopper. :)

bindings/go/dag/dag.go
```diff
@@ -0,0 +1,328 @@
...
+	vertices := d.GetVertices()
+
+	var dfs func(T)
+	dfs = func(node T) {
+		visited[node] = true
+
+		// Sort the neighbors to ensure deterministic order
+		neighbors := make([]T, 0, len(d.Vertices[node].Edges))
+		for neighbor := range d.Vertices[node].Edges {
+			neighbors = append(neighbors, neighbor)
```
Skarlso, line 199: I know this goes through the neighbors in the below loop and does a recursion on !visited ones, but what's the reason for adding it in the first place if it's visited already? The order would still remain once sorted. 🤔

bindings/go/dag/dag.go
```diff
@@ -0,0 +1,328 @@
...
+// GetVertices returns the nodes in the graph in sorted alphabetical
+// order.
+func (d *DirectedAcyclicGraph[T]) GetVertices() []T {
+	nodes := make([]T, 0, len(d.Vertices))
+	for node := range d.Vertices {
+		nodes = append(nodes, node)
+	}
+
+	// Ensure deterministic order. This is important for TopologicalSort
+	// to return a deterministic result.
```
Skarlso, line 230: If we care so much about order, I wonder why not just have them as Slices and have a map as a side tracker for IDs if really need it. If we end up sorting during all GetVertices call which is quite plenty, this could be super bad for performance.

Skarlso: Though I would have to run a few benchmarks first to be sure about this statement.

bindings/go/dag/dag.go
jakobmoellerdev: I think if we want concurrency safety, it needs to be on the entire map and not just on the vertex. WDYT?

bindings/go/dag/dag.go
jakobmoellerdev: TBH the Topological Sort is something we dont really need in our traversal logic for OCM (because we have concurrent traversals and traverse based on indegree and outdegree, not topology alone). We can even kill the TopologicalSort. It is good for debugging though

bindings/go/dag/dag.go
jakobmoellerdev: map to bool in this case is just a nicer interface, and I dont think we optimize for that much (yet?)

bindings/go/dag/dag.go
jakobmoellerdev: We care about edge ordering, not necessarily slice ordering. Thats quite different

bindings/go/dag/dag.go
Skarlso: Yeah, fine. :D

bindings/go/dag/dag.go
Skarlso: So you're saying that even if the slice is ordered or rather created in some order it doesn't mean that a slice would help vs a map since you'll have to order it anyways.

Skarlso (dismissed):
Well done! Nice job! :)

bindings/go/dag/dag.go
Skarlso: Yeah, agreed.

Skarlso (approved):

PR #22, opened 2025-03-18
feat: Common Transport Format Library support

jakobmoellerdev:

```
// Package ctf provides various interfaces and types for working with Common Transport Format Archives (CTF)
//
// The Common Transport Format describes a file system structure that can be
// used for the representation of [content](https://github.com/opencontainers/image-spec)
// of an OCI repository.
//
// Therefore, it can be used to describe a subset of repositories of an OCI registry with a subset of
// artifacts, that can then be imported again into any OCI registry.
//
// # A CTF is a file or directory containing
//
//   - artifact-index.json: This JSON file describes the contained artifact (versions).
//
//   - blobs
//
//     The blobs directory contains the blobs described by the
//     artifact index as a flat file list. These are layer blobs or artifact
//     blobs for the artifact descriptors.
//
//     Every file has a filename according
//     to its digest (https://github.com/opencontainers/image-spec/blob/main/descriptor.md#digests).
//     Hereby the algorithm separator character is replaced by a dot (".").
//
//     Every file SHOULD be referenced, directly or indirectly, in the artifact
//     descriptor by a descriptor according the OCI Image Specification (https://github.com/opencontainers/image-spec/blob/main/descriptor.md).
//
//     The artifact index describes the OCI manifests (image manifests and index
//     manifests), which refer to further non-manifest blobs.
//     Files not referenced by the artifacts described by the index SHOULD be ignored.
//
// The FileFormat of a CTF can differ: as directory of an
// operating system file system or a virtual file system (FormatDirectory) or as content of
// a TAR archive (unzipped - FormatTAR or zipped - FormatTGZ).
// The descriptor SHOULD be the first file if stored in an archive.
//
// This package also offers a legacy compatibility layer access for the ArtifactSet, a now no longer recommended
// artifact format that was used in the past by OCM CLI to package local blobs. We now instead recommend packaging
// in the format of OCI Image Layouts as they are almost identical. See ArtifactSet for more details.
```

This is a partial solution to https://github.com/open-component-model/ocm-project/issues/443 and requires https://github.com/open-component-model/open-component-model/pull/18 for its blob access abstraction.

Open:

- Legacy Compatibility API to Artifact Sets (for old OCM CTFs)
- Various Tests that check the CTF (still lots of things to do there)


bindings/go/ctf/artifact_set.go
```diff
@@ -0,0 +1,333 @@
...
+// It will start owning the readcloser in the blob and needs to be closed due to this.
+// It is backed by a tar filesystem that is either read from a tar file or a gzip file.
+// The gzip format is not detected by ArtifactSetMediaType, but by the magic number in the first 512 bytes of the file.
+// The blob must be a valid ArtifactSet, otherwise an error is returned.
+//
+// The ArtifactSet mereley supports read-only and any future write operation is no longer supported.
+// All ArtifactSet's encountered in the wild MUST be converted to OCI Image Layouts.
+//
+// See ArtifactSet for more information about the format.
+func NewArtifactSetFromBlob(b blob.ReadOnlyBlob) (ArtifactSet, error) {
```
Skarlso, line 88: Do we really want to return an interface here?

bindings/go/ctf/artifact_set.go
```diff
@@ -0,0 +1,333 @@
...
+
+var _ ArtifactSet = (*artifactSet)(nil)
+
+// NewArtifactSetFromBlob creates a new ArtifactSet from a blob.
+// It will start owning the readcloser in the blob and needs to be closed due to this.
+// It is backed by a tar filesystem that is either read from a tar file or a gzip file.
+// The gzip format is not detected by ArtifactSetMediaType, but by the magic number in the first 512 bytes of the file.
+// The blob must be a valid ArtifactSet, otherwise an error is returned.
+//
+// The ArtifactSet mereley supports read-only and any future write operation is no longer supported.
```
Skarlso, line 84:
```suggestion
// The ArtifactSet merely supports read-only and any future write operation is no longer supported.
```

bindings/go/ctf/artifact_set.go
Skarlso: I see why. Any reason to have it backed by an unexported type? This will make this a bit difficult to work with from the outside.

bindings/go/ctf/cas_file_blob.go
```diff
@@ -0,0 +1,66 @@
...
+
+func (b *CASFileBlob) ReadCloser() (io.ReadCloser, error) {
+	return b.blob.ReadCloser()
+}
+
+func (b *CASFileBlob) Digest() (digest string, known bool) {
+	if b.HasPrecalculatedDigest() {
+		return b.digest, true
+	}
+	dig, known := b.blob.Digest()
```
Skarlso, line 44: Doesn't this require a lock? I see both the functions have one, then this part should also have one, I think. 🤔

bindings/go/ctf/filesystem_ctf.go
```diff
@@ -0,0 +1,203 @@
...
+
+	return c.writeFile(filepath.Join(
+		BlobsDirectoryName,
+		ToBlobFileName(dig),
+	), data)
+}
+
+// ToBlobFileName converts a digest to a blob file name by replacing the ":" with ".", which is the
+// default separator for blobs in the CTF under BlobsDirectoryName.
+func ToBlobFileName(digest string) string {
```
Skarlso, line 195: If this is something that the user could influence, I'm pretty sure we should be careful with replacing `.` otherwise we might expose ourselves to a directory traversal attack.

bindings/go/ctf/tar.go
```diff
@@ -0,0 +1,272 @@
...
+		}
+		wg.Wait()
+
+		var err error
+		for i := 0; i < len(blobs); i++ {
+			err = errors.Join(<-errs)
+		}
+		if err != nil {
+			return err
+		}
```
Skarlso, line 167: Use errgroup?

Skarlso: Further, we should definitely add cancellations here, right? Via context.

Skarlso (commented):
Well done! I have some comments and questions. :)

bindings/go/ctf/filesystem_ctf.go
jakobmoellerdev: im not sure how that can be used for exposure but im happy to be taught. I think this conversion is static and can never be modified

bindings/go/ctf/artifact_set.go
jakobmoellerdev: Can expose the type directly too since all fields are unexported. 👍 I think generally it should not make a difference right now

bindings/go/ctf/filesystem_ctf.go
Skarlso: What if the name is `.:/` or something. that would result in `..` involuntary providing a traverse once that's written to the filesystem.

jakobmoellerdev: then that wouldnt be a valid digest anymore would it?

Skarlso: Indeed. That's why I'm asking if this could be something a user would provide? Or if there is a valid digest check before calling this method or whether we should add one here? :)

jakobmoellerdev: I think adding a valid digest is a good idea, ill add it

bindings/go/ctf/tar.go
```diff
@@ -0,0 +1,272 @@
...
+		return fmt.Errorf("unable to setup file system ctf: %w", err)
+	}
+	if err := os.Mkdir(path, 0o755); os.IsExist(err) {
+	} else if err != nil {
+		return fmt.Errorf("unable to create directory: %w", err)
+	} else {
+		if fsCTF, err = OpenCTFFromOSPath(path, O_RDWR); err != nil {
+			return fmt.Errorf("unable to setup file system ctf: %w", err)
+		}
+	}
```
frewilhelm, line 137:
```suggestion
	if err := os.Mkdir(path, 0o755); err != nil {
		if !os.IsExist(err) {
			return fmt.Errorf("unable to create directory: %w", err)
		} 
	} else {
		if fsCTF, err = OpenCTFFromOSPath(path, O_RDWR); err != nil {
			return fmt.Errorf("unable to setup file system ctf: %w", err)
		}
	}
```

IMO this is a bit easier to read but should make the same. WDYT?

bindings/go/ctf/tar.go
```diff
@@ -0,0 +1,272 @@
...
+	var fsCTF CTF
+
+	blobs, err := ctf.ListBlobs()
+	if err != nil {
+		return fmt.Errorf("unable to list blobs: %w", err)
+	}
+
+	if fsCTF, err = OpenCTFFromOSPath(path, O_RDWR|O_CREATE); err != nil {
+		return fmt.Errorf("unable to setup file system ctf: %w", err)
+	}
```
frewilhelm, line 129: If I am not mistaken, then `OpenCTFFromOSPath()` creates a new directory for `path` (check the call `filesystem.NewFS(...)`). So, the following `os.Mkdir(...)` should not be necessary, right?

bindings/go/ctf/cas_file_blob.go
jakobmoellerdev: Just looked at this again. Why would the lock be needed here? The functions individually lock already (first a read lock in HasPrecalculatedDigest, and then a write lock in SetPrecalculatedDigest)

bindings/go/ctf/tar.go
jakobmoellerdev: fixed, pls double check

bindings/go/ctf/tar.go
jakobmoellerdev: fixed, pls double check

bindings/go/ctf/tar.go
jakobmoellerdev: changed to errgroups, then noticed our io has no context cancellation, then added ctx to everything in CTF :D

bindings/go/ctf/tar.go
```diff
@@ -0,0 +1,272 @@
...
+		return ErrUnsupportedFormat
+	}
+}
+
+// ArchiveDirectory archives the CTF to the specified path with FormatDirectory.
+// The blobs are copied to the directory and the index is written to the index file.
+// The CTF is not modified and only read from.
+// The directory is created if it does not exist.
+// The blobs are written to the blobs directory concurrently.
+func ArchiveDirectory(ctf CTF, path string) error {
```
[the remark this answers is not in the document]
jakobmoellerdev: Just double checked, in NewFS there is a call to fs.Abs() which would resolve this

bindings/go/ctf/filesystem_ctf.go
jakobmoellerdev: added validation of digest to every call

bindings/go/ctf/cas_file_blob.go
jakobmoellerdev: I now wrote the lock path explicitly so there can never be a lock race

bindings/go/ctf/tar.go
Skarlso: Yay. :D

bindings/go/ctf/tar.go
```diff
@@ -0,0 +1,267 @@
...
+	var fsCTF CTF
+
+	blobs, err := ctf.ListBlobs(ctx)
+	if err != nil {
+		return fmt.Errorf("unable to list blobs: %w", err)
+	}
+
+	if fsCTF, err = OpenCTFFromOSPath(path, O_RDWR|O_CREATE); err != nil {
+		return fmt.Errorf("unable to setup file system ctf: %w", err)
+	}
```
frewilhelm, line 139:
```suggestion

	blobs, err := ctf.ListBlobs(ctx)
	if err != nil {
		return fmt.Errorf("unable to list blobs: %w", err)
	}

	fsCTF, err := OpenCTFFromOSPath(path, O_RDWR|O_CREATE)
	if err != nil {
		return fmt.Errorf("unable to setup file system ctf: %w", err)
	}
```

Skarlso (approved):

PR #23, opened 2025-03-18
chore: fixup dependabot config

jakobmoellerdev:

This makes sure we receive dependabot updates in our go bindings

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -18,6 +18,7 @@ updates:
   - component/github-actions
 - package-ecosystem: "gomod"
   directories:
+    - "/cli"
     - "/bindings/go/*"
   groups:
     go:
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -17,57 +17,13 @@ updates:
   - kind/skip-release-notes
   - component/github-actions
 - package-ecosystem: "gomod"
-  directory: "/"
+  directories:
+    - "/bindings/go/*"
   groups:
     go:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -17,57 +17,13 @@ updates:
   - kind/skip-release-notes
   - component/github-actions
 - package-ecosystem: "gomod"
-  directory: "/"
+  directories:
+    - "/bindings/go/*"
```
frewilhelm, line 21: Do we have to update this field, when we add go-code to other locations?

frewilhelm:
```suggestion
      - "**/*"
```
This should do the job as well and prevent that we have to update the config each time we add go-code to another location

frewilhelm: > i dont want to update go code everywhere right?

My assumption is that `dependabot` updates code-dependencies. So, every go-code that we write and that has dependencies should be regularly checked by `dependabot`. Did I misunderstand something?

As far as I understood this repository, `binding/go` will contain the library code. But I expected a cli-tool that is also written in go and is located in `cli/`. Maybe this assumption was wrong. If my assumption was correct, then we need to adjust the `dependabot` configfile, when we introduce new go-code.

jakobmoellerdev: Looked at this again, I think youre right. However I do not really want dependabot to have to scan every single directory in our repository

frewilhelm (approved):

PR #29, opened 2025-03-20
fix: use original k8s deep copy gen for core runtime

jakobmoellerdev:

This uses the original deep copy gen instead of the controller runtime version as well as copies over the deep copy behavior from unstructured so that we use raw and unstructured in the same way as kubernetes


Skarlso (approved):
LGTM

PR #30, opened 2025-03-20
chore: add blackduck workflow

morri-son:
Add workflow for Blackduck SCA scanning, similar to all other repos.

### DIFF
--- a/.github/workflows/trigger-blackduck-scan.yaml
+++ b/.github/workflows/trigger-blackduck-scan.yaml
@@ -15,7 +15,7 @@ on:
 
 jobs:
   trigger-scan:
-    uses: open-component-model/.github/.github/workflows/blackduck-scan.yaml@main
+    uses: open-component-model/.github/.github/workflows/blackduck-scan.yaml@47e68e42a6d78125d519e6e008c2b203385a0a72
     with:
       # required to be able to differentiate between PRs and pushes in the called workflow (rapid or full scan)
       event_type: ${{ github.event_name }} 


jakobmoellerdev (commented):
Are you sure this can run on the root of the repo? It looks like this might need to run per module. WDYT?

morri-son:
> Are you sure this can run on the root of the repo? It looks like this might need to run per module. WDYT?

Ah, ok. This is a monorepo. Then we need to add more depth for the search. I did this in the PR now: DETECT_DETECTOR_SEARCH_DEPTH: 3

.github/workflows/blackduck_scan.yaml
```diff
@@ -0,0 +1,101 @@
...
+          blackducksca_scan_full: true
+
+      - name: Run Black Duck SCA Scan (Pull Requests)
+        if: ${{ github.event_name == 'pull_request_target' }}
+           # The action sets blackducksca_scan_full internally: for pushes to true and PRs to false
+        uses: blackduck-inc/black-duck-security-scan@v2.0.0
+        env:
+          DETECT_PROJECT_USER_GROUPS: opencomponentmodel
+          DETECT_PROJECT_VERSION_DISTRIBUTION: opensource
+          DETECT_EXCLUDED_DIRECTORIES: .bridge
```
jakobmoellerdev, line 49: why do we have a .bridge directory here?

.github/workflows/blackduck_scan.yaml
```diff
@@ -0,0 +1,101 @@
...
+          NODE_TLS_REJECT_UNAUTHORIZED: true
+        with:
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          blackducksca_url: ${{ secrets.BLACKDUCK_URL }}
+          blackducksca_token: ${{ secrets.BLACKDUCK_API_TOKEN }}
+          blackducksca_scan_full: true
+
+      - name: Run Black Duck SCA Scan (Pull Requests)
+        if: ${{ github.event_name == 'pull_request_target' }}
+           # The action sets blackducksca_scan_full internally: for pushes to true and PRs to false
```
frewilhelm, line 44: nit: typo and indention looks weird?

.github/workflows/blackduck_scan.yaml
```diff
@@ -0,0 +1,101 @@
...
+
+      # Check Black Duck status and upload status file as artifact.
+      # This step is required to be set as always(), so the status file is uploaded even if the Black Duck scan fails.
+      - name: Check Black Duck status
+        if: always()
+        id: check_blackduck_status
+        shell: bash
+        run: |
+          # Use find to locate status file
+          STATUS_FILE=$(find "/home/runner/work/ocm-cicd-playground/ocm-cicd-playground/.bridge/Blackduck SCA Detect Execution/detect/runs" -name "status.json" | head -n 1)
```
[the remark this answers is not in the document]
frewilhelm: I am confused by this `find` command. Basically it finds all locations of files called `status.json` in `.../runs` and its subdirectories. But then, we only return the first path of that list (`head -n 1`). That seems a bit arbitrary. 

If `status.json` is located in `.../runs`, then this location will always be returned.
If `status.json` is not located in `.../runs`, but in one or more subdirectories of `.../runs`, then we take the first path from the list, that is returned by `find` and which is sorted first numerical and then alphabetical.

Where do we expect `status.json` to be and are we aware of the things mentioned above?
What is the purpose of that call?

.github/workflows/blackduck_scan.yaml
```diff
@@ -0,0 +1,101 @@
...
+      
+          if [ -z "$STATUS_FILE" ]; then
+            echo "::warning file=status.json::No Black Duck status file found"
+            exit 1
+          else
+            ISSUE_COUNT=$(jq '.issues | length' "$STATUS_FILE")
+            
+            if [[ "$ISSUE_COUNT" -eq 0 ]]; then
+              echo "status_file_path=$STATUS_FILE" >> "$GITHUB_OUTPUT"
+              echo "Black Duck scan successfully executed. Status JSON will be uploaded as an artifact to the GitHub action.""
```
frewilhelm, line 76: Why do we upload the Status JSON if there were no issues?

.github/workflows/blackduck_scan.yaml
```diff
@@ -0,0 +1,101 @@
...
+            else
+              # Issues exist, fail step but save file path for upload
+              echo "status_file_path=$STATUS_FILE" >> "$GITHUB_OUTPUT"
```
frewilhelm, line 79: This is redundant to L75

.github/workflows/blackduck_scan.yaml
morri-son: I removed the param.

.github/workflows/blackduck_scan.yaml
```diff
@@ -0,0 +1,101 @@
...
+          blackducksca_scan_full: false
+
+      # Check Black Duck status and upload status file as artifact.
+      # This step is required to be set as always(), so the status file is uploaded even if the Black Duck scan fails.
+      - name: Check Black Duck status
+        if: always()
+        id: check_blackduck_status
+        shell: bash
+        run: |
+          # Use find to locate status file
```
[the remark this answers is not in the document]
morri-son: no, one can set an output folder for the detect agent, but that always gets a "runs" folder that contains a timestamp and this is where the status.json is located. Therefore I need to use the find command below.

.github/workflows/blackduck_scan.yaml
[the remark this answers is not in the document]
morri-son: as mentioned in the comment above, detect adds addition folders and places the results in. In the cicd playground this for example looks like:

```
2025-03-21 11:26:29.4318 UTC [Blackduck SCA Detect Execution] INFO: Results are available at "/home/runner/work/ocm-cicd-playground/ocm-cicd-playground/.bridge/Blackduck SCA Detect Execution/detect/runs/2025-03-21-11-25-15-064/scan/open_component_model_ocm_cicd_playground_main_BlackDuck_DeveloperMode_Result.json"
```

The find works as expected and has been tested in the ocm-cicd-playground: https://github.com/open-component-model/ocm-cicd-playground/actions/runs/13990680306/job/39173733533

.github/workflows/blackduck_scan.yaml
morri-son: Issues <> issues :-). The issues in the result file are solely for issues with the Blackduck detect scan itself, not the security issues found. The status.json should be uploaded in any case as it represents the results found in the scan, regardless if it contains security issues and/or technical issues during the scan.

.github/workflows/blackduck_scan.yaml
morri-son: I don't get the point. Why is this redundant? One is in if, the other in else. In both cases, technical issues / no technical issues, the json should be uploaded.

frewilhelm: because if it is executed in `if` and in `else`, then the statement will be executed either way. Thus, it can be put before the if-condition itself.

.github/workflows/blackduck_scan.yaml
```diff
@@ -0,0 +1,96 @@
...
+            # Extract and print issue details
+            jq -r '.issues[] | "\(.type): \(.title)\n  Details: \((.messages | if type == "string" then [.] else . end) | join("; "))"' "$STATUS_FILE" | \
+            while IFS= read -r line; do
+              echo "::error::$line"
+            done
+            echo
+            echo "Black Duck Overall Status:"
+            jq -r '.overallStatus[0].key + " - " + .overallStatus[0].status' "$STATUS_FILE"
+            exit 1
+          fi
```
frewilhelm, line 89:
```suggestion
          echo "status_file_path=$STATUS_FILE" >> "$GITHUB_OUTPUT"
          echo "Black Duck scan successful"
          
          if [[ "$ISSUE_COUNT" -neq 0 ]]; then
            # Technical issues with scan exist, fail step but save file path for upload
            echo "::error file=$STATUS_FILE::Black Duck scan had technical issues:"
            
            # Extract and print issue details
            jq -r '.issues[] | "\(.type): \(.title)\n  Details: \((.messages | if type == "string" then [.] else . end) | join("; "))"' "$STATUS_FILE" | \
            while IFS= read -r line; do
              echo "::error::$line"
            done
            echo
            echo "Black Duck Overall Status:"
            jq -r '.overallStatus[0].key + " - " + .overallStatus[0].status' "$STATUS_FILE"
            exit 1
          fi
```

.github/workflows/blackduck_scan.yaml
morri-son: ah, ok. I moved the export to GITHUB_OUTPUT up and removed the duplicate occupancy.

jakobmoellerdev (commented):
is this equivalent to https://github.com/open-component-model/ocm/pull/1365?

morri-son:
@jakobmoellerdev 
> is this equivalent to [open-component-model/ocm#1365](https://github.com/open-component-model/ocm/pull/1365)?

Yes, it is. All the PRs in the four repos - ocm, ocm-controller, ocm-k8s-toolkit and this new one - are identical, but the additional `DETECT_DETECTOR_SEARCH_DEPTH: 3`  in this repo.

frewilhelm:
> Yes, it is. All the PRs in the four repos - ocm, ocm-controller, ocm-k8s-toolkit and this new one - are identical, but the additional DETECT_DETECTOR_SEARCH_DEPTH: 3 in this repo.

Then it would probably make sense to define the workflow in `.github` as reusable workflow that takes an argument for `DETECT_DETECTOR_SEARCH_DEPTH`.

Would this be technically possible?

morri-son:
> Then it would probably make sense to define the workflow in `.github` as reusable workflow that takes an argument for `DETECT_DETECTOR_SEARCH_DEPTH`.
> 
> Would this be technically possible?

Technically yes, but since you anyway need a workflow that calls this workflow, you will have workflow code in each of the repositories. Since we're about to remove all but one repo, I propose to go with the current solution, not making it central.

frewilhelm:
> Since we're about to remove all but one repo

What do you mean by that?

I see the big benefit of a central reusable workflow in code-adjustments. As we had seen last week, some things did not work out correctly (e.g. the path was hardcoded to the playground repository). This must be changed in all repositories right now. Using a reusable workflow would only mean one adjustment in one repository.

jakobmoellerdev:
We are not removing the other repos. This is moot. We are introducing a monorepo, thats true but we will keep the others around until we have feature parity at least, meaning like 12-24 months

morri-son:
Hi @jakobmoellerdev and @frewilhelm , I slightly adopted the workflow to make it generic. https://github.com/open-component-model/.github/pull/51 adds it to the .github repo. This PR now just contains a workflow file that triggers the central workflow (similar to what we did with BDBA in the release workflow). Pease check. Once ok, I will remove the blackduck workflows from the other repos and replace them with the same trigger.

frewilhelm (dismissed):

.github/workflows/trigger-blackduck-scan.yaml
```diff
@@ -0,0 +1,27 @@
...
+    paths-ignore:
+    - '.github/workflows/**'
+    branches: [ "main" ]
+  schedule:
+    - cron:  '10 0 * * 0'
+  workflow_dispatch:
+
+jobs:
+  trigger-scan:
+    uses: open-component-model/.github/.github/workflows/blackduck-scan.yaml@main
```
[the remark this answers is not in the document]
morri-son: will do once https://github.com/open-component-model/.github/pull/52 is merged and I can use the commit ID in all trigger workflows.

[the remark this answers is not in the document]
frewilhelm: ah yes, you are completely right :D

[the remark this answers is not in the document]
jakobmoellerdev: this has now been merged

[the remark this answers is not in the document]
morri-son: added the commit SHA instead of main: https://github.com/open-component-model/open-component-model/pull/30/files#diff-8a7de65b0116a7b8300df77f8ef17d37aa1bc788afe5e70c743618e77835e66eR18

frewilhelm (approved):

PR #32, opened 2025-03-21
fix: fixup module path for descriptor

jakobmoellerdev:

The go.mod module name was incorrectly set to the descriptor without v2 suffix which causes go mod resolution to fail

relates to https://github.com/open-component-model/ocm-website/pull/424


Skarlso (approved):

jakobmoellerdev:
@Skarlso the API path was also not updated correctly, package should be v2 after all because deepcopy-gen does not support differing package and folder

jakobmoellerdev:
> That's why I was asking if we should have the module _outside_ the v2 folder. But this is fine as well.

I think its canonical to have one module per major version, so this should be okay

Skarlso (approved):

PR #33, opened 2025-03-21
fix: allow bidirectional raw conversion due to switch away from interface

jakobmoellerdev:

This allows for bidirectional raw conversion by adding a path in convert that specifically converts back into Raw.

This is especially useful now because we switched from using interface types in our structs to only raw, so it is important that our conversion works as intended everytime.

```
// Convert transforms one Typed object into another. Both 'from' and 'into' must be non-nil pointers.
//
// Special Cases:
//   - Raw → Raw: performs a deep copy of the underlying []byte data.
//   - Raw → Typed: unmarshals YAML from raw.Data into the typed object (if the type is registered).
//   - Typed → Raw: marshals the object to JSON, applies canonicalization, and stores the result in Raw.Data. (See Raw.UnmarshalJSON)
//   - Typed → Typed: performs a deep copy using Typed.DeepCopyTyped, with reflection-based assignment.
//
// Errors are returned if:
//   - Either argument is nil.
//   - A type is not registered in the Scheme (for Raw conversions).
//   - A reflection-based assignment fails due to type mismatch.
```

This also reintroduces the Typed constraint on the registry because we now have generators that support our methods, which means there is no need to use any and we can rely on the methods in the interfaces


bindings/go/runtime/registry.go
```diff
@@ -154,7 +128,7 @@ func (r *Scheme) NewObject(typ Type) (any, error) {
 		}
 		object = reflect.New(t).Interface()
 
-		return object, nil
+		return object.(Typed), nil
```
Skarlso, line 131: Type check here?

bindings/go/runtime/registry.go
```diff
@@ -178,45 +152,85 @@ func (r *Scheme) Decode(data io.Reader, into any) error {
...
+//   - A reflection-based assignment fails due to type mismatch.
+func (r *Scheme) Convert(from Typed, into Typed) error {
+	if from == nil || into == nil {
+		return fmt.Errorf("both 'from' and 'into' must be non-nil")
+	}
+
+	fromType := from.GetType()
+
+	// Handle Raw → Raw (cheap shortcut)
+	if rawFrom, ok := from.(*Raw); ok {
```
Skarlso, line 176: Use a switch on type instead of readability?

bindings/go/runtime/registry.go
```diff
@@ -178,45 +152,85 @@ func (r *Scheme) Decode(data io.Reader, into any) error {
...
+//   - Raw → Typed: unmarshals Raw.Data JSON via json.Unmarshal into the Typed object (if Typed.GetType is registered).
+//   - Typed → Raw: marshals the Typed with json.Marshal, applies canonicalization, and stores the result in Raw.Data.
+//     (See Raw.UnmarshalJSON for equivalent behavior)
+//   - Typed → Typed: performs a deep copy using Typed.DeepCopyTyped, with reflection-based assignment.
+//
+// Errors are returned if:
+//   - Either argument is nil.
+//   - A type is not registered in the Scheme (for Raw conversions).
+//   - A reflection-based assignment fails due to type mismatch.
+func (r *Scheme) Convert(from Typed, into Typed) error {
```
Skarlso, line 168: The scheme contains a mutex. Don't we need to use that here? 🤔

bindings/go/runtime/registry.go
jakobmoellerdev: if the object is created through reflection above and we did an assertion through the the registry during creation, how would this type check be needed?

Skarlso: Ah you're right, sorry the diff was blurring out that section of the code and I didn't check.

bindings/go/runtime/registry.go
jakobmoellerdev: Its used inside the call of IsRegistered which should be fine. WDYT?

jakobmoellerdev (commented):
To reviewer: It might show here that we want to at least have enough knowledge of type information and top level keys. The Raw type is now able to inject type information but we may need a mix of Raw and Unstructured for that, something like `RawUnstructured` that contains `map[string]json.RawMessage` so that we are able to easily default certain fields. Im happy to hear opinions though

bindings/go/generator/ocmtypegen/main.go
```diff
@@ -1,114 +1,248 @@
...
+func main() {
+	if len(os.Args) < 2 {
+		slog.Info("Usage: generator <root-folder>")
+		os.Exit(1)
+	}
+	root := os.Args[1]
+
+	packages, err := findGoPackages(root)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Error finding Go packages: %v\n", err)
```
Skarlso, line 31: You're using slog and fmt.Fprintf and slog.Error all over the place. :D Choose one. :D

bindings/go/generator/ocmtypegen/main.go
```diff
@@ -1,114 +1,248 @@
...
+		}
 	}
-	sourceFile := os.Args[1]
-	outputFile := filepath.Join(filepath.Dir(sourceFile), "zz_generated.type.ocm.software.go")
+	return false
+}
 
-	pkg, structs, err := ParseStructsWithGenerateDirective(sourceFile)
+// hasRuntimeTypeField checks if the struct has a field named Type of type runtime.Type.
+func hasRuntimeTypeField(s *ast.StructType) bool {
```
Skarlso, line 150: This got to be the most one that could break. Especially if we rename. We should definitely leave a comment on the Type saying if you rename this make sure you rename the generator or something. WDYT?

bindings/go/runtime/registry.go
Skarlso: Hmmm. But there is no defence if someone uses this directly. humm humm.

bindings/go/generator/ocmtypegen/main.go
jakobmoellerdev: added a log statement for this case specifically

bindings/go/generator/ocmtypegen/main.go
jakobmoellerdev: 👍 chose slog

bindings/go/runtime/registry.go
jakobmoellerdev: what do you mean with uses it directly? where would we need a lock?

Skarlso (approved):

PR #34, opened 2025-03-21
feat: introduce internal descriptor runtime

jakobmoellerdev:

```
// Package runtime defines an internal runtime to work
// with component descriptors in all schema versions without
// restricting the code to the public API for future major changes.
//
// This package should be preferred whenever working with non-serialization
// relevant routines (such as working with accesses or components)
//
// This package SHOULD NOT be used for serialization and any kind of
// attempt at serialization will fail on purpose, be buggy and is not
// supported!
//
// Instead, use dedicated converters to the scheme versions available.
//
// This package is not guaranteed to be stable, but will always
// support conversion from and to all advertised component descriptor
// schemes.
//
// Thus, it is much easier to migrate from/to than if one would program
// against the scheme directly.
//
// The conversions offered in this package are guaranteed to be lossless unless
// otherwise specified, so it is usually safe to convert back and forth between
// scheme versions.
package runtime
```

this can be used by other packages such as the oci bindings to work with descriptors internally


bindings/go/descriptor/runtime/descriptor.go
```diff
@@ -0,0 +1,281 @@
...
+	// References component dependencies that can be resolved in the current context.
+	References []Reference `json:"-"`
+}
+
+// ResourceRelation describes whether the component is created by a third party or internally.
+type ResourceRelation string
+
+const (
+	// LocalRelation defines a internal relation
+	// which describes a internally maintained resource in the origin's context.
```
fabianburth, line 69:
```suggestion
	// which describes an internally maintained resource in the origin's context.
```

bindings/go/descriptor/runtime/descriptor.go
```diff
@@ -0,0 +1,281 @@
...
+)
+
+// A Resource is a delivery artifact, intended for deployment into a runtime environment, or describing additional content,
+// relevant for a deployment mechanism.
+// For example, installation procedures or meta-model descriptions controlling orchestration and/or deployment mechanisms.
+// See https://github.com/open-component-model/ocm-spec/blob/main/doc/01-model/02-elements-toplevel.md#resources
+// +k8s:deepcopy-gen=true
+type Resource struct {
+	ElementMeta `json:",inline"`
+	// SourceRefs defines a list of source names.
```
fabianburth, line 83:
```suggestion
	// SourceRefs defines a list of sources used to generate the resource.
```
The "used to generate the resource" part is kind of optional. `source names` is just not really accurate for the `SourceRefs` data structure.

bindings/go/descriptor/runtime/descriptor.go
```diff
@@ -0,0 +1,281 @@
...
+	// These entries reference the sources defined in the
+	// component.sources.
+	SourceRefs []SourceRef `json:"-"`
+	// Type describes the type of the object.
```
fabianburth, line 87:
```suggestion
	// Type describes the type of the resource.
```

bindings/go/descriptor/runtime/descriptor.go
```diff
@@ -0,0 +1,281 @@
...
+// Source elements do not have specific additional formal attributes.
+// See https://github.com/open-component-model/ocm-spec/blob/main/doc/01-model/02-elements-toplevel.md#sources
+// +k8s:deepcopy-gen=true
+type Source struct {
+	ElementMeta `json:",inline"`
+	Type        string        `json:"-"`
+	Access      runtime.Typed `json:"-"`
+}
+
+// Reference describes the reference to another component in the registry.
```
fabianburth, line 118:
```suggestion
// Reference describes the reference to another component.
```
might lead to confusion

bindings/go/descriptor/runtime/descriptor.go
```diff
@@ -0,0 +1,281 @@
...
+// A component version may refer to other component versions by adding a reference to the component version.
+//
+// The Open Component Model makes no assumptions about how content described by the model is finally deployed or used.
+// This is left to external tools.
+//
+// Tool specific deployment information is formally represented by other artifacts with an appropriate type.
```
fabianburth, line 124:
```suggestion
// Tool specific deployment information is formally represented by other artifacts with an appropriate type and/or by labels.
```

bindings/go/descriptor/runtime/descriptor.go
```diff
@@ -0,0 +1,281 @@
...
+//
+// In addition to the common artifact information, a resource may optionally describe a reference to the source by specifying its artifact identity.
+//
+// See https://github.com/open-component-model/ocm-spec/blob/main/doc/02-processing/01-references.md#referencing
+// +k8s:deepcopy-gen=true
+type Reference struct {
+	ElementMeta `json:",inline"`
```
fabianburth, line 131:
```suggestion
type Reference struct {
         // The name in ElementMeta describes the name of the reference itself within this component (not the name of the referenced component). But the version in ElementMeta specifies the version of the referenced component.
	ElementMeta `json:",inline"`
```
Since `Reference` is kind of weird with regards to `ElementMeta`, we might want to add an additional comment here.

bindings/go/descriptor/runtime/descriptor.go
```diff
@@ -0,0 +1,281 @@
...
+// +k8s:deepcopy-gen=true
+type Meta struct {
+	// Version is the schema version of the component descriptor.
+	Version string `json:"-"`
+}
+
+// ObjectMeta defines an object that is uniquely identified by its name and version.
+// Additionally the object can be defined by an optional set of labels.
+// It is an implementation of the Element Identity as per
+// https://github.com/open-component-model/ocm-spec/blob/main/doc/01-model/03-elements-sub.md#element-identity
```
fabianburth, line 157: I think this comment belongs to `ElementMeta`, doesn't it?

bindings/go/descriptor/runtime/descriptor.go
```diff
@@ -0,0 +1,281 @@
...
+	}
+	return base
+}
+
+// ElementMeta defines an object with name and version containing labels.
+// +k8s:deepcopy-gen=true
+type ElementMeta struct {
+	ObjectMeta `json:",inline"`
+	// ExtraIdentity is the identity of an object.
+	// An additional label with key "name" is not allowed
```
fabianburth, line 186:
```suggestion
	// An additional identity attribute with key "name" is not allowed
```

bindings/go/descriptor/runtime/convert_v2.go
```diff
@@ -0,0 +1,430 @@
...
+	"maps"
+	"strings"
+	"time"
+
+	v2 "ocm.software/open-component-model/bindings/go/descriptor/v2"
+	"ocm.software/open-component-model/bindings/go/runtime"
+)
+
+// ConvertFromV2 converts a v2.Descriptor to the internal Descriptor format.
+func ConvertFromV2(descriptor *v2.Descriptor) (*Descriptor, error) {
```
fabianburth, line 15: I imagine, we would also implement a validation function on the internal component descriptor? Is this something we'd like to keep separate or should this be incorporated here? Currently, you could easily get an internal component descriptor with e.g. 2 resources with the same identities, which could later on mess with other function implemented upon the component.

bindings/go/descriptor/runtime/doc.go
```diff
@@ -0,0 +1,24 @@
+// Package runtime defines an internal runtime to work
+// with component descriptors in all schema versions without
+// restricting the code to the public API for future major changes.
```
fabianburth, line 3: maybe, we also add here that this makes it easier to work with the embedded types in the internal version since we do not need to constantly convert

fabianburth (approved):
Looks good!
Mostly nitpicky suggestions to change some comments to avoid potential confusion.

bindings/go/descriptor/runtime/doc.go
jakobmoellerdev: added a small section on ease of use with embedded types

fabianburth (approved):

PR #35, opened 2025-03-25
feat: migrate configuration v1

Skarlso:

Closes https://github.com/open-component-model/ocm-project/issues/445.


cli/configuration/v1/flat_map.go
```diff
@@ -0,0 +1,39 @@
...
+			if err := scheme.Convert(config.GetType(), &cfg); err != nil {
+				merged.Configurations = append(merged.Configurations, config)
+			} else {
+				flattenCandidates = append(flattenCandidates, &cfg)
+			}
+		}
+
+		cfg, err := FlatMap(flattenCandidates...)
+		if err != nil {
+			return nil, err
```
frewilhelm, line 29: `FlatMap` does not return any error, except this one. Thus, we do not need to return an error.

cli/configuration/v1/ocm_config.go
```diff
@@ -0,0 +1,235 @@
...
+	}
+	if path := getFromExecutableDir(); path != "" {
+		paths = append(paths, path)
+	}
+
+	if len(paths) > 0 {
+		return paths, nil
+	}
+
+	return nil, fmt.Errorf("OCM config not found in any known locations")
```
frewilhelm, line 153: Should we print out the known locations as a support-message?

cli/configuration/v1/config.go
```diff
@@ -0,0 +1,45 @@
...
+	scheme.MustRegisterWithAlias(&Config{}, runtime.NewUngroupedVersionedType(ConfigType, ConfigTypeV1))
+}
+
+const (
+	ConfigType   = "generic.config.ocm.software"
+	ConfigTypeV1 = "v1"
+)
+
+// Config holds configuration entities loaded through a configuration file.
+// +k8s:deepcopy-gen=true
```
Skarlso, line 21: @jakobmoellerdev How do you run the generate workflow? :D It keeps erroring with weird errors like, module not found, no main module, etc...

jakobmoellerdev: I think your problem is that youre generating types in the CLI. The code generator is currently setup to only run on go modules. Check the tools section for that

jakobmoellerdev: After you included the cli, it should just be `task generate`

Skarlso: I tried that but it gave me different errors. Let me do that again.

Taskfile.yml
```diff
@@ -25,11 +25,16 @@ includes:
...
-
+  cli/configuration/v1:
+    optional: true
+    taskfile: ./cli/configuration/v1/Taskfile.yml
+    dir: ./cli/configuration/v1
 
 vars:
   GO_MODULES:
     sh: find {{ .ROOT_DIR }}/bindings/go -name go.mod -exec dirname {} \; | sed 's|{{ .ROOT_DIR }}/bindings/go/||'
+  GO_GENERATE:
```
jakobmoellerdev, line 36: I think this requires appropriate fixes in github actions as well.

jakobmoellerdev: We need the bindings in github actions just for task distribution, imo its totally fine to have a task that prints them for integration in github actions, then we dont need the scripts there.

jakobmoellerdev: You can use a task instead of this line: https://github.com/open-component-model/open-component-model/blob/main/.github/workflows/ci.yml#L31 then you can unify it all

Skarlso: Nice

cli/configuration/v1/config.go
```diff
@@ -0,0 +1,46 @@
...
+type Config struct {
+	Type           runtime.Type    `json:"type"`
+	Configurations []Configuration `json:"configurations"`
+}
+
+// Configuration holds a single configuration entity.
+// +k8s:deepcopy-gen=true
+type Configuration struct {
```
jakobmoellerdev, line 29: Why does thing exist? Couldnt we just use runtime.Raw directly in the array?

cli/configuration/v1/flat_map.go
Skarlso: You're right.

cli/configuration/v1/config.go
Skarlso: Actually have this removed locally while refactoring runtime.Type. :D But didn't commit it yet because of our upcoming talk and because my test is failing. :)

cli/configuration/v1/ocm_config.go
Skarlso: Sure!

jakobmoellerdev (approved):

PR #38, opened 2025-03-26
fix: make sure that type generation on local blob access is fine

jakobmoellerdev:

fixes up the Type that was embedded as struct in the local blob which is invalid


fabianburth (approved):
LGTM (the pr description could have been a little more explanatory ;) )

PR #40, opened 2025-03-27
feat: setup go workspace

jakobmoellerdev:

because we can release all our modules at the same time, we are perfectly fine with committing our go workspace.

A dependency resolution will now look like this:

```
module github.com/jakobmoellerdev/goworktestconsumer

go 1.24.1

require (
	github.com/jakobmoellerdev/open-component-model/bindings/go/ctf v0.0.0-20250327102339-9e336774a59c
)

require (
	github.com/jakobmoellerdev/open-component-model/bindings/go/blob v0.0.0-20250327102339-9e336774a59c // indirect
	github.com/nlepage/go-tarfs v1.2.1 // indirect
	github.com/opencontainers/go-digest v1.0.0 // indirect
	github.com/opencontainers/image-spec v1.1.1 // indirect
	golang.org/x/sync v0.12.0 // indirect
)
```

if imported via go get. At the same time all commits can now finally touch up all modules at the same time and they are always synced.

 A simple example can be found here https://github.com/jakobmoellerdev/goworktest/tree/main.

You can simply `go get github.com/jakobmoellerdev/goworktest/mod2` and observe the dependency resolution.

This fixes the issue that we can only edit one go mod at once when we have interdependant go modules. This is very similar to how kubernetes maintains their staging repository, with the difference that we immediately release to the module path.


fabianburth (approved):

.github/workflows/ci.yml
```diff
@@ -125,24 +118,19 @@ jobs:
...
-        project: ${{ fromJSON(needs.discover_modules.outputs.modules_json) }}
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
+      - name: Setup Go
+        uses: actions/setup-go@v5
         with:
-          sparse-checkout: |
-            ${{ matrix.project }}
+          go-version: '1.24.x'
```
frewilhelm, line 133: I don't understand why `go` is required now in these steps wrt to the changes above?

PR #45, opened 2025-03-31
feat: convert compat function that can translate old and new descriptors

jakobmoellerdev:

Offers a compatibility method on runtime descriptors that can convert arbitrary access types to new types that we know in our scheme. This is useful for making old component descriptors work correctly


jakobmoellerdev:
closing in favor of using n registries and legacy types

PR #51, opened 2025-04-01
feat(docs): ocm orchestration spec adr

fabianburth:
This is a proposal on how to implement essential ocm ecosystem operations as well as extension through the formalization of a generic cel based yaml pipeline syntax.

Fixes https://github.com/open-component-model/ocm-project/issues/444


docs/adr/0005_transformation.md
```diff
@@ -0,0 +1,467 @@
+# A Specification for Orchestration of OCM Component Operations
+
+* Status: proposed
+* Deciders: Gergely Brautigam, Fabian Burth, Jakob Moeller
+* Date: 2025-02-17
+
+Technical Story: Design a specification that enables the orchestration of
+arbitrary operations on components and their resources.
```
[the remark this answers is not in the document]
jakobmoellerdev: I would like to name it OCM Transformation Specification or OCM Transform Spec in short. Is that okay?

jakobmoellerdev (commented):
Looks great already! Currently looking at the benefits/cons mostly which seem reasonable. IMO this is almost acceptable ootb.

needs a pass through for typo correction. quite a few sentences that are incoherent / incomplete scattered over the proposal.

docs/adr/0005_transformation.md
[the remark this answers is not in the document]
jakobmoellerdev: if you think it should be an orchestration spec, I think you could just call the type `OCMProcessingGraph` and then work from there

docs/adr/0005_transformation.md
```diff
@@ -0,0 +1,467 @@
...
+**Con**
+
+- **Complexity**:
+  The generic graph traversal logic introduces significant complexity that might
+  be hard to understand and debug.
+
+> Although, the clear separation between algorithm and business logic
+> introduced through this formalization is a clear benefit.
+
+- **Hard to Generate**
```
ikhandamirov, line 436: To put this specification into a context, it would be helpful to have a chapter describing, who, how and when is supposed to generate the orchestration spec, especially for bigger products, consisting of dozens of components with hundreds of resources.

docs/adr/0005_transformation.md
```diff
@@ -0,0 +1,467 @@
...
+    - Requires to really understand the dependencies between the operations
+    - Manual creation is quite difficult and error-prone
+
+## Conclusion
+
+The concept behind the **ocm orchestration specification** quite clean. The
+reusable graph orchestration logic provides significant value.
+
+The main **disadvantage** is the complexity of the manual creation of a
+specification. We assume that the manual creation is a rare task. Either, the
```
ikhandamirov, line 461: Manual creation is rather tedious. Seems only be feasible for podinfo-like small examples.

ikhandamirov (commented):
Nicely written. Easy to understand.

docs/adr/0005_transformation.md
```diff
@@ -0,0 +1,467 @@
...
+  - root component references `ocm.software/leaf-component:1.0.0`
+  - leaf component is stored at
+    `quay.io/ocm-component-model/transfer-source/ocm.software/leaf-component:1.0.0`
+
+  Both components should be transferred to
+  `ghcr.io/ocm-component-model/transfer-target`
+  and `quay.io/ocm-component-model/transfer-target` in a single transfer
+  process.
+
+- **Transfer resources between different storage systems**
```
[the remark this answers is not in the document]
fabianburth: Yeah, the requirement is meant as "resource transfer as part of component transfer". I'll clarify.

docs/adr/0005_transformation.md
fabianburth: I don't disagree. Although, I think one has to consider that:
- not the entire spec would have to be created manually, one could also use the generator we plan to provide
- as a future outlook, there could be tools that support users in creating custom transformation specifications

docs/adr/0005_transformation.md
fabianburth: Added a small usage section

fabianburth:
Thanks for the feedback! I revised the ADR again and tried to incorporate everything!
PTAL @ikhandamirov @jakobmoellerdev

frewilhelm (approved):
very cool!

PR #57, opened 2025-04-07
chore: optimize filesystem ctf implementation to use file size information everywhere.

jakobmoellerdev:

This makes it so filesystem CTFs always use file size information where available to issue a CopyN command which means we do not require a buffer. If there is no size information, the implementation falls back to buffering.

Avoids unnecessary allocs of CTF based blobs


bindings/go/ctf/filesystem_ctf_test.go
```diff
@@ -193,3 +195,76 @@ func Test_FileSystemCTF_FileSystemOperations(t *testing.T) {
...
+			},
+			wantErr: "blob does not have a digest that can be used to save it",
+		},
+		{
+			name: "digest not known",
+			blob: &mockBlob{
+				digest: "sha256:123",
+				known:  false,
+			},
+			wantErr: "blob does not have a digest that can be used to save it",
```
frewilhelm, line 243: Is such a setup even possible? It sounds counterintuitive to have the digest filled, but the `known` field `false`.

frewilhelm (approved):
lgtm - just a little nit about one test

bindings/go/ctf/filesystem_ctf_test.go
jakobmoellerdev: consider it a sanity check if you use a blob implementation that doesnt expose a digest!

PR #58, opened 2025-04-07
feat: introduce subset identity matching

jakobmoellerdev:

adds a subset matcher for identities in runtime

```
// IdentitySubset is a ChainableIdentityMatcher that checks if the identity sub is a subset of the identity base.
// It is useful to check if an identity is a subset of another identity, and thus can be considered
// an IdentityEqual matcher on a subset of the identity.
//
// Note that giving an empty subset will always return true, as an empty identity is a subset of any identity.
func IdentitySubset(sub Identity, base Identity) bool
```

useful for writing selector logic in descriptors later


bindings/go/runtime/identity_match_test.go
```diff
@@ -204,3 +204,124 @@ func TestIdentity_Match(t *testing.T) {
 		})
 	}
 }
+
+func TestIdentitySubset(t *testing.T) {
+	type args struct {
+		b runtime.Identity
+		a runtime.Identity
```
frewilhelm, line 211: It makes the test easier to read if you refactor the variable names to what they represent. `a` should be `sub` and `b` should be `base`, right?

bindings/go/runtime/identity_match.go
```diff
@@ -56,6 +56,23 @@ func IdentityEqual(a Identity, b Identity) bool {
...
+// Note that giving an empty subset will always return true, as an empty identity is a subset of any identity.
+func IdentitySubset(sub Identity, base Identity) bool {
+	if len(sub) > len(base) {
+		return false
+	}
+	for k, vsub := range sub {
+		if vm, found := base[k]; !found || vm != vsub {
+			return false
+		}
+	}
```
frewilhelm, line 72: I am having some trouble understanding this part.

`k` = key
`vsub` = value of subset?
`vm` = value of subset match?

frewilhelm (approved):
lgtm - nit: the description of variables and tests seems a bit lazy^^

jakobmoellerdev:
Double check pls @frewilhelm

frewilhelm (approved):
Thank you for adjusting the names!

PR #67, opened 2025-04-15
feat: normalise with json v3

jakobmoellerdev:

This is a WIP / Draft PR that adds json normalisation and the v3 algorithm to the OCM descriptor binding library. This still needs more fixes and validation and currently there are some fields in the v2 descriptor that are not serialized correctly. still need to decide what we do with that.


jakobmoellerdev:
@fabianburth please review

jakobmoellerdev:
To Reviewers: we should determine if we break this algorithm to v4alpha1 instead of v3

bindings/go/descriptor/normalisation/engine/jcs/doc.go
```diff
@@ -0,0 +1,91 @@
...
+//	    "ocm.software/open-component-model/bindings/go/descriptor/normalisation/engine/jcs"
+//	)
+//
+//	// Basic normalization
+//	normalized, err := jcs.Normalise(input, nil)
+//
+//	// With exclusion rules
+//	excludes := jcs.MapExcludes{
+//	    "field1": nil,  // Exclude field1
+//	    "field2": jcs.NoExcludes{},  // Include field2
```
frewilhelm, line 69: This seems weird. If I don't want that field excluded, I should not add it to the `jcs.MaxExcludes{}`. What am I missing?

bindings/go/descriptor/normalisation/engine/jcs/normalise.go
```diff
@@ -0,0 +1,431 @@
...
+
+	"github.com/cyberphone/json-canonicalization/go/src/webpki.org/jsoncanonicalizer"
+)
+
+// Normalise is a helper function that prepares and marshals a normalized value.
+// It takes an input value and exclusion rules, and returns the canonicalized JSON representation.
+//
+// Parameters:
+//   - v: The input value to normalize
+//   - ex: Exclusion rules to apply during normalization
```
frewilhelm, line 19: From the `doc.go` it is described that `Normalise` can also take a `mapper` which contains a transformation. This is not the same as the exclusion rules, is it?

frewilhelm (approved):
lgtm besides of nits

I would be fine to use v4alpha1. Although this implementation resembles the ocm v1 v3-normalisation, we (practically) cannot make sure that they behave the same way. Thus, a break should be considered.

bindings/go/descriptor/normalisation/engine/jcs/doc.go
jakobmoellerdev: You are correct but there are two ways on this. NoExcludes is an exclusion rule like any other. So semantically "field2": jcs.NoExcludes{} is even equivalent to not setting the field. NoExcludes is mostly used within the engine itself to return if no fields should be excluded as part of `Field(name string, value interface{}) (string, interface{}, ExcludeRules)` which expects to return exclusion rules. I can remove this from the example.

bindings/go/descriptor/normalisation/engine/jcs/normalise.go
jakobmoellerdev: actually MapValue is also an exclusion rule, which is... unfortunate now that I think about it. Maybe they should be called transformation rules instead?

bindings/go/descriptor/normalisation/engine/jcs/doc.go
frewilhelm: ah, I see - thanks for clarifying. Maybe this can lead to confusion in the documentation without any hint (like the comment you just gave). Thus, I would either clarify it or omit it.

bindings/go/descriptor/normalisation/engine/jcs/normalise.go
jakobmoellerdev: I changed the names to transformationRules instead of exclude rules

frewilhelm: sounds reasonable

bindings/go/descriptor/normalisation/engine/jcs/normalise.go
```diff
@@ -0,0 +1,362 @@
...
+//
+// Returns:
+//   - []byte: The canonicalized JSON representation
+//   - error: Any error that occurred during normalization
+func Normalise(v interface{}, ex TransformationRules) ([]byte, error) {
```
frewilhelm, line 24:
```suggestion
func Normalise(v interface{}, rules TransformationRules) ([]byte, error) {
```

bindings/go/descriptor/normalisation/engine/jcs/normalise.go
```diff
@@ -0,0 +1,362 @@
...
+func (n *null) Value() interface{}                     { return nil }
+func (n *null) SetField(name string, value Normalised) { panic("set field on null") }
+
+// Null represents a normalized null value.
+var Null Normalised = (*null)(nil)
+
+// PrepareNormalisation converts an input value into a normalized structure,
+// by marshaling it to JSON and then unmarshaling into a map or array.
+// This is the main entry point for the normalization process.
+func PrepareNormalisation(n Normalisation, v interface{}, excludes TransformationRules) (Normalised, error) {
```
frewilhelm, line 228:
```suggestion
func PrepareNormalisation(n Normalisation, v interface{}, rules TransformationRules) (Normalised, error) {
```

bindings/go/descriptor/normalisation/engine/jcs/normalise.go
```diff
@@ -0,0 +1,362 @@
...
+		return prepareArray(n, rawArray, excludes)
+	}
+
+	// If both fail, try as a basic value
+	return n.NewValue(v), nil
+}
+
+// Prepare recursively converts an input value into a normalized structure,
+// applying any exclusion rules along the way.
+func Prepare(n Normalisation, v interface{}, ex TransformationRules) (Normalised, error) {
```
frewilhelm, line 252:
```suggestion
func Prepare(n Normalisation, v interface{}, rules TransformationRules) (Normalised, error) {
```

bindings/go/descriptor/normalisation/engine/jcs/normalise.go
```diff
@@ -0,0 +1,362 @@
...
+	}
+	// Apply any normalisation filter if available.
+	if filter, ok := ex.(NormalisationFilter); ok {
+		return filter.Filter(result)
+	}
+	return result, nil
+}
+
+// prepareStruct normalizes a map by applying exclusion rules to each field.
+func prepareStruct(n Normalisation, v map[string]interface{}, ex TransformationRules) (Normalised, error) {
```
frewilhelm, line 293:
```suggestion
func prepareStruct(n Normalisation, v map[string]interface{}, rules TransformationRules) (Normalised, error) {
```

frewilhelm (approved):

PR #68, opened 2025-04-15
feat: add jsonschema based descriptor validation for v2 descriptors

jakobmoellerdev:

this adds json schema validation to descriptors based on the latest 2020-12 schema. It is based on the original json schema in the ocm codebase but has stricter type syntax.


bindings/go/descriptor/v2/validate.go
```diff
@@ -0,0 +1,94 @@
...
+	"sigs.k8s.io/yaml"
+)
+
+// JSONSchema contains the embedded JSON schema for validating Open Component Model descriptors.
+//
+//go:embed resources/schema-2020-12.json
+var JSONSchema []byte
+
+// getJSONSchema is a singleton that compiles the JSON schema once and caches it for reuse.
+var getJSONSchema = sync.OnceValues[*jsonschema.Schema, error](func() (*jsonschema.Schema, error) {
```
fabianburth, line 23: Would it make sense to offer this method publicly? Otherwise, if - for whatever reason - someone would like to use the JSON Schema, he might use the raw bytes (since they are public) and repeat the compilation?

fabianburth (dismissed):

fabianburth (approved):
Looks good to me!

PR #70, opened 2025-04-16
feat: setup OCIRepository

jakobmoellerdev:

Creates the OCIRepository Type which is an alias of our old OCIRegistry type.


Skarlso (approved):

Skarlso:
I was too fast. :D

frewilhelm (approved):
lgtm, but is creating the type all that this PR wants to achieve?

Skarlso:
@frewilhelm Correct. :)

PR #71, opened 2025-04-16
feat: component constructors

jakobmoellerdev:

This PR is about implementing component constructors. It is still heavily WIP and not yet ready but I use it for cross validation of the reference bindings against CTFs (since component constructors only work on local ctfs at first)

The concept is the following:

1. There is a dedicated schema with Input Specifications for the `component-constructor.yaml` file that is very closely related but not identical to a component descriptor
2. Whenever a resource is added, it is checked if we have an input or access type defined. For access types the resource is accepted as is, for input types, a input method registry is called and it is checked if the input specification is known.
3. For a found input spec, the input method is called, which can process the resource as per the specification. The contract defines that at the end, a resource MUST have an access after the input was processed. For example the utf8 input type generates local blobs that contain utf8 text from the specification.
4. After all inputs are (concurrently) processed, the descriptor is fully rendered and added to the component version store.

A later plugin system can hook new Input Methods into the Constructor library, allowing for flexibility (e.g. by adding a helm repo input method that directly pulls the chart into a local blob).

The only inbuilt input types that are available by default are file and text based inputs that can pack generic blobs. Any special content such as Helm chart input methods will be implemented separately because they will be in other go modules. This is so that while building the CLI we can decide which input methods will be supported and selectively compile OCM for different audiences.

TODO:
- [ ] How to pass Input Method specific configuration to the method (e.g. a setting from .ocmconfig needs to configure how the helm chart input method adds the chart)
- [ ] Confirm that the plugin system can really be integrated smoothly
- [x] Adding more tests and explanations.
- [x] ~In addition to Input UTF8 and File, add a Dir input that automatically archives.~ created https://github.com/open-component-model/ocm-project/issues/494 for this

Once this is done, the constructor library can be embedded by the CLI to be used in "add cv" commands so that new component versions can be created from scratch via the constructor specification.

fix https://github.com/open-component-model/ocm-project/issues/460


jakobmoellerdev:
I have made some progress here but also still have a few challenges:

1. I wanted to find a way to include full HELM support compiled in without having a plugin binary. For this I tried both an external plugin and a go mod dependency. TBH both sucked because the HELM dependency tree is bloated and ahving a plugin for something as widespread as HELM felt like a useability loss. As such I started to work on `helmlite`, a stripped down version of the HELM go module that focuses only and purely on 1) replicating HELM Packaging and Loading behavior 1:1 and 2) staying FAR away from any other dependency that is not strictly related to 1). This lead to an input method that is so stripped down that we can safely include it as a compiled in dependency.
2. I discovered that almost all of our input methods we use centrally now do not need external dependencies anymore which is great.

Now for the bad news:
I do not seem to find a smart way to decouple an input method into something that can be extracted into a plugin.

Ideally I would like a flow like this:

1. Load Component Constructor Centrally
2. Iterate over Resources / Sources
3. For every Resource / Source with an Input Specification, call Input plugin, and receive AccessSpec
4. Then update this access spec and save it in the descriptor.

This sounded fine until I noticed that for some Input plugins to work, access to the component version repository is necessary. This is because If you want to add e.g. a local blob, you would introspect the input specification, then create a blob from that, and then upload that to that same component version repository where you want to store the component version.

Here comes the problem: I dont want the input methods to KNOW how to access the component version repositories, because that would mean they would also need to understand the credential system and so on.

Instead im now thinking of a flow like this:

1. Input Method gets called with Input Specification (as before), but returns a Blob and Resource. If blob is set, then a local blob should be added by the constructor. If NOT, then the blob has been uploaded already by the input method and we should accept the resource as is
2. Constructor adds local blobs on its own thereby encapsulating repository knowledge to itself.

That means input methods only need to know the repository specification when they actually need to interact with it directly.

Now to the second problem: How to pass over configurations needed from OCM to the Input Method. This is actually related to config passing to OCM in general. IMO input plugins should be able to request certain types of configurations from .ocmconfig and then we can pass them that information. This will allow strong typing guarantees with a central config system. @Skarlso and I will need to talk through this on how we can pass this config data centrally then. 

Last but not least, input methods will be required to provide Consumer Identities for the credential system just like the cv repository so that we can ask them which credentials they might require from the credential graph. Then we can pass them these credentials whenever we call them.

jakobmoellerdev:
Im closing this one due to size concerns. I will start pushing the constructor lib separately in smaller PRs and then fixup the add cv command

PR #72, opened 2025-04-16
feat: compressed / decompressed blobs

jakobmoellerdev:

adds a library set of functions to the blob package that allows for transparently compressed or decompressed blobs. currently only gzip is implemented

useful if one wants to work with transparently compressed / non-compressed content (e.g. a file that automatically should be compressed like the file input type)

needed for https://github.com/open-component-model/ocm-project/issues/460

### DIFF
--- a/bindings/go/blob/compression/blob.go
+++ b/bindings/go/blob/compression/blob.go
@@ -1,0 +1,202 @@
+// Package compression provides functionality for handling compressed blobs in the Open Component Model.
+// It supports various compression methods and provides utilities for working with compressed data streams.
+package compression
+
+import (
+	"compress/gzip"
+	"errors"
+	"fmt"
+	"io"
+	"strings"
+
+	"ocm.software/open-component-model/bindings/go/blob"
+)
+
+const (
+	MediaTypeGzip       = "application/gzip"
+	MediaTypeGzipSuffix = "+gzip"
+)
+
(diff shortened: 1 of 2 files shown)


bindings/go/blob/compression/blob.go
```diff
@@ -0,0 +1,202 @@
...
+	default:
+		compressed = gzip.NewWriter(writer)
+	}
+
+	_, err := io.Copy(compressed, reader)
+	writer.CloseWithError(errors.Join(err, compressed.Close(), reader.Close()))
+}
+
+// Decompress creates a decompressed version of the given blob if it is compressed.
+// It detects the compression method based on the blob's media type and returns a new
```
frewilhelm, line 111: Why do we only decompress based on the media type?

We could also use something like [`DetectCompressionFormat`](https://pkg.go.dev/github.com/containers/image/v5/pkg/compression@v5.34.3#DetectCompressionFormat) to get the compression format.

jakobmoellerdev: we could, but tbh im not sure if we should do that. if the media type is unknown, would you prefer that?

frewilhelm: I think there are good reasons to rely on the mediaType and enforce that it is set. So, I am fine with the current approach.

frewilhelm (approved):

PR #78, opened 2025-04-23
chore: remove redundant Taskfile value as it is infered

Skarlso:


Skarlso:
This is currently not possible: https://github.com/go-task/task/issues/2204

PR #80, opened 2025-04-25
fix: make sure that manifests are discovered always and set global access

jakobmoellerdev:

This makes it so that the global access is set to a layer if we have a non-manifest type, and an oci image if we do. This is because single layers are not full oci images so they should be references as oci image layers.

### DIFF
--- a/bindings/go/oci/internal/pack/pack.go
+++ b/bindings/go/oci/internal/pack/pack.go
@@ -20,6 +20,7 @@ import (
 	resourceblob "ocm.software/open-component-model/bindings/go/oci/blob"
 	internaldigest "ocm.software/open-component-model/bindings/go/oci/internal/digest"
 	"ocm.software/open-component-model/bindings/go/oci/internal/identity"
+	"ocm.software/open-component-model/bindings/go/oci/internal/introspection"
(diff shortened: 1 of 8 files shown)


bindings/go/oci/internal/pack/pack.go
```diff
@@ -40,6 +41,10 @@ type Options struct {
 	// ManifestAnnotations are annotations that will be added to single layer Artifacts
 	// They are not used for OCI Layouts.
 	ManifestAnnotations map[string]string
+
+	// EnforceGlobalAccess indicates if new resources should contain a global access regardless whether the
+	// access is guaranteed to be valid or not
+	EnforceGlobalAccess bool
```
fabianburth, line 47: Could you elaborate the purpose of this option?

frewilhelm (approved):

bindings/go/oci/internal/pack/pack.go
jakobmoellerdev: it is used for testing as I cannot enforce this logic without a remote repository such as in integration tests. Since the function is internal it is fine to be exported.

PR #81, opened 2025-04-28
feat: add JSON schema generator to runtime package

Skarlso:


bindings/go/runtime/json_schema_generator.go
```diff
@@ -0,0 +1,31 @@
...
+
+	"github.com/invopop/jsonschema"
+)
+
+// GenerateJSONSchemaForType takes a Type and uses reflection to generate a JSON Schema representation for it.
+// It will also use the correct type representation as we don't marshal the type in object format.
+func GenerateJSONSchemaForType(obj Typed) ([]byte, error) {
+	r := &jsonschema.Reflector{
+		Mapper: func(i reflect.Type) *jsonschema.Schema {
+			if i == reflect.TypeOf(Type{}) {
```
jakobmoellerdev, line 15: please turn if statement around. Also is it intended that nil is returned if there is no match?

bindings/go/runtime/go.mod
```diff
@@ -4,16 +4,21 @@ go 1.24.1
 
 require (
 	github.com/cyberphone/json-canonicalization v0.0.0-20241213102144-19d51d7fe467
+	github.com/invopop/jsonschema v0.13.0
```
jakobmoellerdev, line 7: Worried about this dependency, are we sure on its stability?

bindings/go/runtime/json_schema_generator_test.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+func (c *customType) DeepCopyTyped() Typed {
+	c2 := *c
+	return &c2
+}
+
+var _ Typed = &customType{}
+
+func TestGenerateJSONSchemaForType(t *testing.T) {
```
jakobmoellerdev, line 30: Im missing a failure case

jakobmoellerdev (changes requested):
Mostly concerned about the lib on the runtime.

bindings/go/runtime/json_schema_generator.go
Skarlso: Normally, I would agree. But here it's explicit that we do something different in case of this type instead of returning `nil` in case it's NOT that type. 🤔  I think this rather explicitness, reads better.

That said, if you prefer it the other way around I'm okay with that. I'm not dying on this hill. :D 

Yes, the intended use is to return nil, if you don't care about the type:

```go
		if t := r.Mapper(t); t != nil {
			return t
		}
                ... // normal flow
```

bindings/go/runtime/go.mod
Skarlso: I'm not. Seems like a stable library, being used by quite a lot of people. There is precious else we can do other than try to implement everything ourselves. That would be rather daunting.

bindings/go/runtime/json_schema_generator_test.go
Skarlso: Plus the function requires to pass in an struct that satisfies the interface. So I can't pass garbage.

bindings/go/runtime/json_schema_generator.go
```diff
@@ -0,0 +1,46 @@
...
+// GenerateJSONSchemaForType takes a Type and uses reflection to generate a JSON Schema representation for it.
+// It will also use the correct type representation as we don't marshal the type in object format.
+func GenerateJSONSchemaForType(obj Typed) ([]byte, error) {
+	if obj == nil {
+		return nil, fmt.Errorf("cannot generate JSON schema for nil object")
+	}
+
+	switch v := obj.(type) {
+	case *Unstructured:
+		if v.Data == nil {
```
jakobmoellerdev, line 19: can we generate a json scheme for a nil object? that would be weird wouldnt it?

Skarlso: No, I don't think it can generate anything from a nil object indeed. That's why we error out here if Data is nil.

jakobmoellerdev: if we cant generate anything from a nil then why dont we just return an error no matter if we are nil or not?

Skarlso: Err, I didn't quite get that, sorry. We can generate something from something though. So if it's not nil, it will create something like :

```json
...{"Unstructured":{"properties":{"Data":{"type":"object"}}...
```

Is that not good?

jakobmoellerdev: no because theres no property such as Data :)

Skarlso: you mean no explicit property, it's the thing _inside_ data that you want in the schema

jakobmoellerdev: no i mean that Data is not a valid property of that object. its an incorrect schema.

Skarlso: Okay, so basically, just error out for now if the type is either Unstructured or Raw. :)

jakobmoellerdev (approved):
Thanks for this, great stuff :)

PR #82, opened 2025-04-28
feat:  setup get component version command

jakobmoellerdev:
this sets up the get component version command. This is a very early PR that we will need to split up into multiple sub PRs with corrections in modules but it serves as an integration point.

This implements the Get Component Version command as described in the reference documentation. Not included are things such as reference lookups and other smaller features that are present in old OCM but not new OCM. Nevertheless it will serve as a starting point for us. Depending on how we progress, we will either get this merged in with hardcoded oci support or with plugin support already

Eventually fixes https://github.com/open-component-model/ocm-project/issues/473 but not before this needs to be extremely slimmed down and broken into various subissues in https://github.com/open-component-model/ocm-project/issues/448.

fix https://github.com/open-component-model/open-component-model/pull/82


jakobmoellerdev:
This can now be started for a review but I still need to start adding tests as well as the recursive flag to allow recursive component resolution

cli/cmd/get/component-version/encode.go
```diff
@@ -0,0 +1,89 @@
...
+		return nil, 0, fmt.Errorf("encoding component version descriptor as %q failed: %w", output, err)
+	}
+	return bytes.NewReader(data), int64(len(data)), nil
+}
+
+func encodeDescriptorsAsNDJSON(descs []*descruntime.Descriptor) ([]byte, error) {
+	var buf bytes.Buffer
+	encoder := json.NewEncoder(&buf)
+	for _, desc := range descs {
+		// TODO(add formatting options for scheme version with v2 as only option)
```
Skarlso, line 40: can you add your username to the todos? :)

cli/internal/context/context.go
```diff
@@ -0,0 +1,169 @@
...
+	"sync"
+
+	"github.com/spf13/cobra"
+
+	"ocm.software/open-component-model/bindings/go/credentials"
+	"ocm.software/open-component-model/bindings/go/plugin/manager"
+	"ocm.software/open-component-model/cli/configuration/v1"
+)
+
+type key struct{}
```
Skarlso, line 14: usually it's done more meaningfully
```go
type contextKey string
const key contextKey = "key"
```

what I'm trying to say, is that `key` is meaningless. Let's give it some kind of name or something that defines this is ocmContextKey... maybe even just rename it?

cli/configuration/v1/flat_map.go
```diff
@@ -19,7 +19,7 @@ func FlatMap(configs ...*Config) *Config {
 		flattenCandidates := make([]*Config, 0)
 		for _, config := range config.Configurations {
 			var cfg Config
-			if err := scheme.Convert(config, &cfg); err != nil {
+			if err := Scheme.Convert(config, &cfg); err != nil {
```
Skarlso, line 22: I wonder. Scheme has a map inside. This all will eventually end up on `setupOCMConfig` for example. Do you think scheme is thread-safe?

Skarlso (commented):
Just first run. Generally, this is gorgeous. :) 👍 I have a few questions, then I would like to run this actually locally once I have more time.

cli/internal/context/context.go
jakobmoellerdev: An empty struct on package level is actual the defacto standard for context keys that you do not want to make accessible externally. That is something that you do not want accessible outside the package that the key is defined in. Now, someone would not be able to access this context without this package. With a string, that would be possible.

Skarlso: The type wouldn't be the same. Even if it is a key, the type is an unexported thing.

So:

package1
```go
package package1

import "context"

type contextKey string

const key contextKey = "key"

func CreateContext() context.Context {
	ctx := context.WithValue(context.Background(), key, "this-is-the-value")

	return ctx
}

```

package2:

```go
package package2

import (
	"fmt"
	"test-context/package1"
)

func GetContext() {
	ctx := package1.CreateContext()

	value := ctx.Value("key")
	fmt.Println("value: ", value)
}
```

And main:

```go
package main

import "test-context/package2"

func main() {
	package2.GetContext()
}

```

prints:

```
value:  <nil>
```

However main with package1's getContext:

```
➜ go run main.go
value:  <nil>
Value:  this-is-the-value
```

Anyways... I just wanted this key to be named so we know what it is :D

cli/internal/plugin/spec/config/v2alpha1/config.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+// Merge merges the provided configs into a single config.
+func Merge(configs ...*Config) *Config {
+	if len(configs) == 0 {
+		return nil
+	}
+
+	merged := new(Config)
+	_, _ = scheme.DefaultType(merged)
```
Skarlso, line 121: Won't this overwrite different types to a single type or do I understand this incorrectly?

cli/internal/context/context.go
jakobmoellerdev: understood now, will adjust

cli/internal/plugin/spec/config/v2alpha1/config.go
jakobmoellerdev: this is for the merged configuration only. thats why I think using the default type for the merged config is fine.

cli/configuration/v1/flat_map.go
jakobmoellerdev: I believe it is acceptable to consider that Scheme is not thread safe and thus FlatMap is also not threadsafe because the pattern to add things to a scheme is to do it once at startup

cli/cmd/get/component-version/cmd.go
```diff
@@ -0,0 +1,144 @@
...
+If no type is given, the repository path is interpreted based on introspection and heuristics.
+`,
+			compref.DefaultPrefix,
+			strings.Join([]string{ociv1.Type, ctfv1.Type}, "|"),
+			strings.Join([]string{ociv1.ShortType, ociv1.ShortType2, ctfv1.ShortType, ctfv1.ShortType2}, "|"),
+		),
+		Example: strings.TrimSpace(`
+Getting a single component version:
+
+get component-version ghcr.com/open-component-model/ocm//ocm.software/ocmcli:0.23.0
```
Skarlso, line 51: ghcr.io :)

Skarlso (dismissed):
LGTM, nice work! :nod-cry:

Skarlso (approved):

PR #83, opened 2025-04-29
feat: logging flags for configurable logging in the CLI

jakobmoellerdev:

Eventually there will be commands that will not want to output debug logging all the time. Thus the default level should change and we should have configurable logging setup via flags. (think of a generic get cv that shouldnt output all fetch logs from plugins or the OCI implementation)

fix https://github.com/open-component-model/ocm-project/issues/474


cli/cmd/root.go
```diff
@@ -1,30 +1,68 @@
 package cmd
 
 import (
+	"fmt"
+	"log/slog"
 	"os"
 
 	"github.com/spf13/cobra"
+
+	v1 "ocm.software/open-component-model/cli/configuration/v1"
```
frewilhelm, line 10:
```suggestion
	"ocm.software/open-component-model/cli/configuration/v1"
```

cli/cmd/root.go
```diff
@@ -1,30 +1,68 @@
...
   artifacts, like Component Archives, Common Transport Archive,
   Component Repositories, and Component Versions.`,
-	// Uncomment the following line if your bare application
-	// has an action associated with it:
-	RunE: func(cmd *cobra.Command, args []string) error {
-		return cmd.Help()
-	},
-	DisableAutoGenTag: true,
+			// Uncomment the following line if your bare application
+			// has an action associated with it:
```
frewilhelm, line 31:
```suggestion
```

docs/reference/cli/ocm_completion_fish.md
```diff
@@ -28,6 +28,39 @@ ocm completion fish [flags]
...
+                         3. The current working directory:
+                         - $PWD/ocm/config
+                         - $PWD/.ocmconfig
+                         4. The directory of the current executable:
+                         - $EXE_DIR/ocm/config
+                         - $EXE_DIR/.ocmconfig
+                         Using the option, this configuration file be used instead of the lookup above.
+      --logformat enum   set the log output format that is used to print individual logs
+                            json: Output logs in JSON format, suitable for machine processing
+                            text: Output logs in human-readable text format, suitable for console output (must be one of [json text]) (default json)
```
frewilhelm, line 53: This looks a bit odd. Shouldn't this `(must be one of [json text]) (default json)` be on L51?

frewilhelm: The same is true for `loglevel` and `logoutput`

jakobmoellerdev: Why should this be on this line? The enum specification refers to https://github.com/open-component-model/open-component-model/blob/867a2779a47c1bbb020d0bdbd8e9217536ce2ccb/cli/internal/enum/flag.go#L56 so the `must be one of...` string comes always after the flag description. The `default ...` comment is added by cobra for the default value of the flag.

That said, because the flag description doesn't stop in the first line I wouldnt expect it on L51

cli/log/flag.go
```diff
@@ -0,0 +1,156 @@
...
+		return nil, fmt.Errorf("failed to get the log output from the command flag: %w", err)
+	}
+
+	var outputWriter io.Writer
+	switch output {
+	case OutputStdout:
+		outputWriter = cmd.OutOrStdout()
+	case OutputStderr:
+		outputWriter = cmd.OutOrStderr()
+	}
```
Skarlso, line 115: Do you also need a default here erroring with invalid choice?

Skarlso (dismissed):
Two questions, otherwise LGTM

cli/cmd/root.go
```diff
@@ -1,30 +1,66 @@
...
+// setupRoot sets up the root command with the necessary setup for all cli commands.
+func setupRoot(cmd *cobra.Command, _ []string) error {
+	logger, err := log.GetBaseLogger(cmd)
+	if err != nil {
+		return fmt.Errorf("could not retrieve logger: %w", err)
+	}
+	slog.SetDefault(logger)
+
+	if cfg, err := v1.GetFlattenedOCMConfigForCommand(cmd); err != nil {
+		logger.Debug("could not get configuration", slog.String("error", err.Error()))
```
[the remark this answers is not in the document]
jakobmoellerdev: Hmm ideally I would like to separate between Config not found and config rendering fails. I added this logic here because otherwise if there is no ocmconfig such as in CI we error out. WDYT?

cli/log/flag.go
jakobmoellerdev: yes, good catch, thanks!

docs/reference/cli/ocm_completion_fish.md
jakobmoellerdev: I adopted the second one because the first one is not possible without changing the way that enum argument parsing works and I would like to not complicate things

frewilhelm (approved):

PR #87, opened 2025-04-30
feat: add docker config or native store backed OCI credential repository

jakobmoellerdev:

This adds a repository implementation for the credential graph that is backed by a store that comes either directly from the host (wincred on windows, pass / secretservice on linux, osxkeychain on mac, as well as default docker config locations from environment variables) or uses specified inlined or file-specified docker config JSONs as per specification.

The credential graph can only reference these as fallback repositories (see credential graph and repository concepts for more details). The Store can by principle resolve any entry backed by a DockerConfig repository config entry (see docker_config.go).

We now have the full process for credential resolution:

1. Introspect a Resource with an Access or an OCI Repository Specification
2. Generate a Consumer Identity based on it (using the provided methods in this PR)
3. Ask with that Identity in the Graph
4. Either have direct credentials in the graph, or fallback to repositories
5. If falling back, then use DockerConfig entries in graph to ask corresponding configs.
7. If docker config does not point to specific inline configurations or files, fall back to native stores or environment lookups and ask that
8. If any entry is found return it from the Credential Resolution

This ensures maximum lookup flexibility while fully integrating the graph.

Note that the docker config repository is a "special" repository in that it never needs credentials on its own for a lookup (unlike a third party server provider such as a vault repository). Thus, some of its repository methods are stubbed out on purpose so that they get ignored in the credential graph lookup.

fix https://github.com/open-component-model/ocm-project/issues/476


bindings/go/oci/credentials/docker_config.go
```diff
@@ -0,0 +1,179 @@
...
+//   - Shell expansion for file paths (e.g., ~ for home directory)
+//   - Temporary file creation for inline configurations
+//   - Integration with the native host credential store
+func getStore(ctx context.Context, dockerConfig credentialsv1.DockerConfig) (remotecredentials.Store, error) {
+	var store remotecredentials.Store
+	var err error
+	switch {
+	case dockerConfig.DockerConfigFile == "" && dockerConfig.DockerConfig == "":
+		slog.InfoContext(ctx, "attempting to load docker config from default locations or native host store")
+		store, err = remotecredentials.NewStoreFromDocker(remotecredentials.StoreOptions{
```
Skarlso, line 141: I don't particularly like overwriting error values. But it's fine unless you think that this code section here could possibly be called concurrently. If the answer is yes, that means that this is a potential exploit. ( the error value can be overwritten concurrently, and possibly could result in a nil error continuing a segment of code that you didn't wish for ).

bindings/go/oci/credentials/docker_config.go
```diff
@@ -0,0 +1,179 @@
...
+			dirname, err := os.UserHomeDir()
+			if err != nil {
+				return nil, fmt.Errorf("failed to get user home directory: %w", err)
+			}
+			path = strings.ReplaceAll(path, "~", dirname)
+		}
+		if _, err := os.Stat(path); err != nil {
+			slog.WarnContext(ctx, "failed to find docker config file, thus the config will not offer any credentials", "path", path)
+		}
+		store, err = remotecredentials.NewStore(path, remotecredentials.StoreOptions{})
```
Skarlso, line 173: Maybe add a small comment ( so others don't have to do what I did which is check if NewStore throws an error on none existing config ) that `NewStore` returns no error on `if os.IsNotExist(err) {`.

Skarlso (approved):
Really just two remarks. Otherwise this is brilliantly done. :)

bindings/go/oci/credentials/docker_config.go
jakobmoellerdev: good point, added comment accordingly

bindings/go/oci/credentials/docker_config.go
jakobmoellerdev: agreed. i split this method to make it more readable now please double check! :)

Skarlso: thanks! <3

Skarlso (approved):

PR #88, opened 2025-05-06
fix: oci logger formatting

jakobmoellerdev:

the oci Base and Operation logger were currently working incorrectly. The Base Logger didnt get updated when a new slog default logger was specified, and the Operation logger wrongly encapsulated the attributes causing double logged LEVELs and wrongly encoded messages.

I fixed this behavior and adjusted the test, which even checked for the incorrect value (my bad ...)


hilmarf:
is it possible to configure the logger, so that no timestamps are logged?

frewilhelm (approved):

Skarlso (approved):

PR #92, opened 2025-05-07
feat: support source repositories

jakobmoellerdev:

The OCI repository now supports adding and retrieving artifacts that are Sources as well. These map directly to OCM sources as per specification.

Outstanding TODOs:

- [x] ~Had one instance where in the repo, the OCM specification is used to verify its digest against the descriptor digest. sources dont have digest information (was surprised too) so I have to think how I code this~ (I think the verification of the content is already covered by the oci puller so this argument is moot)
- [x] There is still some duplicated code in the repo for local blobs as well as in pack that I would like to unify if possible.
- [x] I need to add test cases that add sources. I will say that adding sources that are not OCI Images is the norm so they will be added as local layers almost always. Think of a git repository you might want to embed in the component version. In this instance one has to upload them as a tar. We have tooling available for that already though


bindings/go/oci/blob/artifact_blob.go
```diff
@@ -42,45 +52,48 @@ func NewResourceBlobWithMediaType(resource *descriptor.Resource, b blob.ReadOnly
 		}
 	}
 
-	if resource.Digest == nil {
-		if digAware, ok := b.(blob.DigestAware); ok {
-			if dig, ok := digAware.Digest(); ok {
-				digSpec, err := digestSpec(dig)
-				if err != nil {
-					return nil, fmt.Errorf("failed to parse digest spec from blob: %w", err)
+	if resource, ok := artifact.(*descriptor.Resource); ok {
```
Skarlso, line 55: This is a bit weird... do I see it correctly that this type assert is in here twice? 🤔

bindings/go/oci/blob/artifact_blob.go
```diff
@@ -11,28 +11,38 @@ import (
...
+	size      int64
 }
 
-// NewResourceBlobWithMediaType creates a new ResourceBlob instance with the given resource,
+// NewArtifactBlobWithMediaType creates a new ArtifactBlob instance with the given artifact,
 // blob data, and media type. This constructor ensures that all necessary
-// information is properly initialized for the ResourceBlob to function correctly.
-func NewResourceBlobWithMediaType(resource *descriptor.Resource, b blob.ReadOnlyBlob, mediaType string) (*ResourceBlob, error) {
+// information is properly initialized for the ArtifactBlob to function correctly.
+func NewArtifactBlobWithMediaType(artifact descriptor.Artifact, b blob.ReadOnlyBlob, mediaType string) (*ArtifactBlob, error) {
```
Skarlso, line 29: Please find a way to not do this. Is this altering the `artifact` in the process of creating a new ArtifactBlob?

Because it's also setting the size of it. After that, `resource` is never used again anywhere.

bindings/go/oci/repository.go
```diff
@@ -380,29 +476,61 @@ func (repo *Repository) getStore(ctx context.Context, component string, version
...
+	if err := internaldigest.Apply(res.Digest, desc.Digest); err != nil {
+		return fmt.Errorf("failed to apply digest to resource: %w", err)
+	}
+	res.Access = access
+
+	return nil
+}
+
+// UploadSource uploads a [*descriptor.Source] to the repository.
+func (repo *Repository) UploadSource(ctx context.Context, target runtime.Typed, src *descriptor.Source, b blob.ReadOnlyBlob) (err error) {
```
Skarlso, line 502: This alters the passed in source. That's not apparent from the function description nor the name of the function. I don't really like side-effects that need to be discovered by usage.

I can tolerate them if they are well documented however. :) That said, I would be even happier if this just returns the altered source. Then I can use it like `src, err = UploadSource` which makes it clearer that the source will be a different source after the call.

That said, if you want to go with just a documentation I can accept that.

Skarlso (commented):
Over-all structure is okay.

bindings/go/oci/repository.go
jakobmoellerdev: I agree fundamentally, I changed the method contracts now since the impact is still bearable

bindings/go/oci/blob/artifact_blob.go
jakobmoellerdev: I agree, I changed the contract of the method to no longer modify the artifact and splitted the artifact modification into a different method.

bindings/go/oci/blob/artifact_blob.go
jakobmoellerdev: It was like that yes, but I changed the logic now so this no longer occurs.

bindings/go/oci/repository_test.go
```diff
@@ -893,3 +1077,330 @@ func setupLegacyComponentVersion(t *testing.T, store *ocictf.Store, ctx context.
...
+				content := []byte("oci layout content")
+				desc := ociImageSpecV1.Descriptor{
+					MediaType: ociImageSpecV1.MediaTypeImageLayer,
+					Digest:    digest.FromBytes(content),
+					Size:      int64(len(content)),
+				}
+
+				// Push the content
+				if err := layout.Push(t.Context(), desc, bytes.NewReader(content)); err != nil {
+					panic(fmt.Sprintf("failed to push content: %v", err))
```
Skarlso, line 1291: You could just pass in r to this and call r.NoError? Totally up to you, a panic here IS weird... :D

Skarlso (approved):

bindings/go/oci/repository_test.go
jakobmoellerdev: holy moly youre right.

Skarlso (approved):

PR #93, opened 2025-05-07
feat: artifact interface

jakobmoellerdev:

this introduces the artifact interface. A commond interface through which one can jointly program logic against both Sources and Resources. Makes it much easier to program component version repositories for joint logics.

### DIFF
--- a/bindings/go/descriptor/runtime/artifact.go
+++ b/bindings/go/descriptor/runtime/artifact.go
@@ -1,0 +1,45 @@
+package runtime
+
+import (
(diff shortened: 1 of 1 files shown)


frewilhelm (approved):
lgtm - but I ask myself if it makes sense to create the file `artifact.go`.

The `resource` and `source` are defined in `descriptor.go`.

jakobmoellerdev:
i choose this because the interface is explicitly outside of descriptors. its more of an additional convenience and not part of the official descriptor contract.

PR #94, opened 2025-05-07
fix: plugin manager test

Skarlso:

The problem was that all tests are using the same name and some didn't properly clean up the socket or the port or something and the next test wasn't able to run.

However, maybe I even need a way to overwrite the name of the socket. 🤔  Right now it's just generated from the name of the file.

### DIFF
--- a/bindings/go/plugin/client/sdk/plugin_test.go
+++ b/bindings/go/plugin/client/sdk/plugin_test.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"io"
+	"log/slog"
 	"net"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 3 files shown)


bindings/go/plugin/client/sdk/plugin_test.go
```diff
@@ -18,13 +19,14 @@ import (
 )
 
 func TestPluginSDK(t *testing.T) {
+	slog.SetLogLoggerLevel(slog.LevelDebug)
 	r := require.New(t)
 
 	output := bytes.NewBuffer(nil)
-	location := "/tmp/test-plugin-plugin.socket"
+	location := "/tmp/test-plugin-flow-plugin.socket"
```
jakobmoellerdev, line 26: suggestion here: use the t.Name() function

jakobmoellerdev (approved):

bindings/go/plugin/client/sdk/plugin_test.go
Skarlso: Will correct in an upcoming one.

PR #99, opened 2025-05-08
chore: add retaining log levels from plugins

Skarlso:

Keep log levels from plugins at the level that the plugin defined and relog them with a plugin ID on the level they defined.

Also, harden the log streamer so that if the plugin quits or the context is cancelled, the streaming attempts stop completely.

Closes https://github.com/open-component-model/ocm-project/issues/489


jakobmoellerdev (dismissed):
We should eventually add an option to the manager to make the logger whose logs we will stream to configurable and not use the default logger. Feel free to setup a separate task for that with low prio

PR #108, opened 2025-05-14
feat: allow all fs.FS implementations for file blobs

jakobmoellerdev:

this makes it so that file blobs can deal with any fs.FS implementation.

That means that we open the ecosystem for inbuilt types such as mapfs but also for other types from libraries such as tarfs.

This is a prerequisite to CTF support of fs.FS.

Note that by default, compatibility is read-only and write support is based on support of our interface methods.


Skarlso (approved):
This will be a nice extension point!

PR #110, opened 2025-05-15
docs: add component constructor support

jakobmoellerdev:

see the ADR for a concrete description or the issue for a short summary

https://github.com/open-component-model/ocm-project/issues/501


jakobmoellerdev:
Even though the PR is not fully ready it has enough content to be reviewed

docs/adr/0006_component_constructors.md
```diff
@@ -0,0 +1,334 @@
...
+## Context and Problem Statement
+
+Whenever users of OCM want to build component versions, it is important to understand that the existing binding libraries offer support to create component versions with API Calls.
+
+However, when working within the context of a CLI, the user experience is not as straightforward. There is a need for a more user-friendly approach to constructing component versions, especially for those who may not be familiar with the underlying API calls in the Go API directly.
+
+As such we want to continue supporting commands in the form of
+
+```shell
+ocm add componentversions ./ctf component-constructor.yaml
```
Skarlso, line 18: I wonder if we should rename that to `ocm create componentversions` to make it even more straight-forward.

docs/adr/0006_component_constructors.md
```diff
@@ -0,0 +1,334 @@
...
+          type: file
+          path: ./testdata/text.txt
+```
+
+In this example:
+
+- The input of type `file` tells the constructor to use the registered `file` input method.
+- This method processes `./testdata/text.txt` and generates an access specification.
+- The result is included in the final component descriptor.
+- The format of the generated access specification depends on the specific input method used.
```
Skarlso, line 86: Perhaps show the resulting descriptor?

docs/adr/0006_component_constructors.md
```diff
@@ -0,0 +1,334 @@
...
+// The method will get called with the raw specification specified in the constructor and is expected
+// to return a ResourceMethodResult or an error.
+//
+// A method can be supplied with credentials from any credentials system by requesting a consumer identity with
+// ResourceInputMethod.GetCredentialConsumerIdentity. The resulting identity MAY be used to uniquely identify the consuming
+// method and to request credentials from the credentials system. The credentials system is not part of this interface and
+// is expected to be supplied by the caller of the input method.
+//
+// The resolved credentials MAY be passed to the input method via the credentials map, but a method MAY
+// work without credentials as well.
```
Skarlso, line 186: Hmmm. I would argue that if the credentials are request then they MUST be passed in, right?

docs/adr/0006_component_constructors.md
```diff
@@ -0,0 +1,334 @@
...
+    // GetSourceInputMethod returns the input method for the given source constructor specification.
+    GetSourceInputMethod(ctx context.Context, resource *constructor.Resource) (input.ResourceInputMethod, error)
+}
+```
+
+To allow for actual construction of component versions, the constructor library provides a central entry function:
+
+```go
+// TargetRepository defines the interface for a target repository that can store component versions created 
+// by the constructor but also act as a target for the constructor to add resources to.
```
Skarlso, line 254: I'm not sure I understand this sentence. So does this literally mean that the TargetRepository can add components + resources? If yes, then let's simplify this sentence a bit?

docs/adr/0006_component_constructors.md
```diff
@@ -0,0 +1,334 @@
...
+type TargetRepository interface {
+    // AddLocalResource adds a local resource to the repository.
+    // The resource must be referenced in the component descriptor.
+    // Resources for non-existent component versions may be stored but may be removed during garbage collection.
```
Skarlso, line 258: How does it know that the component version doesn't have this resource? And how do you even add a Resource without a component version? I would think that having a component version to add a resource is sort of like a prerequisite?

docs/adr/0006_component_constructors.md
```diff
@@ -0,0 +1,334 @@
...
+* Complex Interfaces for a lot of different scenarios (input and access specs are mutually exclusive and lead to branching)
+
+## Conclusion
+
+The new component constructor library will:
+
+- Provide a simple and extensible way to construct component versions based on a `component-constructor.yaml` specification.
+- Support various input methods and access specifications, and upload the resulting component version to an OCM repository.
+- Download resources and upload them as local blobs.
+- Process component version resources and sources in parallel to improve efficiency.
```
Skarlso, line 333: You mean that it will download things in parallel right? Because for the construction I think it needs to be sequential.

Skarlso (commented):
A few questions really. Overall the design is okay. Though I will read this again.

docs/adr/0006_component_constructors.md
jakobmoellerdev: I think we could create an alias but its not super straight forward because not every add is equivalent to create. WDYT?

docs/adr/0006_component_constructors.md
jakobmoellerdev: a plugin cannot request them, it can only be provided with them. So there is no way for a plugin to say "i need credentials now" with this call. Instead they can just return an error in GetConsumerIdentity if they dont want to resolve credentials

docs/adr/0006_component_constructors.md
Skarlso: Gotcha. It's not that important. :)

docs/adr/0006_component_constructors.md
Skarlso: Ah I see. I probably misunderstood the sentence then.

docs/adr/0006_component_constructors.md
jakobmoellerdev: Resources are independent of component version lifecycles. Local Resources have to be added BEFORE a component version. If a blob gets added twice, then the call should be indempotent

docs/adr/0006_component_constructors.md
jakobmoellerdev: for construction, resources and source inputs can be processed concurrently I believe. What makes you think otherwise?

docs/adr/0006_component_constructors.md
Skarlso: Ah, yah, that makes sense now.

docs/adr/0006_component_constructors.md
jakobmoellerdev: I added a resulting descriptor example. please double check to make sure

docs/adr/0006_component_constructors.md
Skarlso: I have a distinct memory, that the construction of the component descriptor needed to follow a strict order. But maybe that memory is flawed. :)

jakobmoellerdev: Are you referring to the processing of individual resources and their order within a component version? Im not aware of such logic, but im happy to accomodate anything in the ADR that you know of or can point me to!

docs/adr/0006_component_constructors.md
```diff
@@ -0,0 +1,363 @@
...
+## Outcome
+
+TBD
+
+## A simplified `bindings/go/constructor` library
+
+The core idea of this proposal is the introduction of a new module `bindings/go/constructor` that will be responsible for
+
+* Providing centralized access and interaction with `component-constructor.yaml` style files and their schemas
+* Providing centralized extension points for adding new Input Methods for building component versions in addition
```
Skarlso, line 51: Do you mean like, adding ANY kind of input method by a user? Or something else?

docs/adr/0006_component_constructors.md
```diff
@@ -0,0 +1,363 @@
...
+  to directly referencing an access.
+
+## Input Methods and Contracts
+
+A key feature of the `component-constructor.yaml` file is the ability to define **inputs**—specifications that describe where to retrieve data and how to store it in a resource for a newly created component version.
+
+### Purpose of Input Methods
+
+An **input method** replaces the `access` field in a standard OCM resource or source. When specified, the input is interpreted at runtime during component construction and used to generate an appropriate access specification based on the processed input data.
```
Skarlso, line 60: By runtime, do you mean during running this new command that constructs the component descriptor in the end?

docs/adr/0006_component_constructors.md
```diff
@@ -0,0 +1,363 @@
...
+
+Each input method must determine whether it should produce a `localBlob` or a global access specification.
+
+---
+
+### "By Value" Construction
+
+Any input or access specification can be marked as **constructed by value**. In this case:
+
+- The constructor directly embeds the resource into the component version.
```
Skarlso, line 140: And here you don't mean that the actual byte stream is added to the component version yaml file. Just to be 100% clear on that, right?

docs/adr/0006_component_constructors.md
```diff
@@ -0,0 +1,363 @@
...
+
+2. **Parse Component Versions**  
+   The constructor library parses the specified component versions from the YAML file.
+
+3. **Process Resources and Sources**  
+   For each component version, the constructor handles all resources and sources:
+
+    - **Input Method Specified**  
+      If an input method is defined:
+        - If it returns `ResourceInputMethodResult.ProcessedBlobData`, the constructor uploads the local blob to the target OCM repository.
```
Skarlso, line 164: So this is also upload like in `by value` but the difference here is that it will not add it to the component descriptor?

docs/adr/0006_component_constructors.md
```diff
@@ -0,0 +1,363 @@
...
+}
+
+// ResourceInputMethod is the interface for processing a resource with an input method declared as per
+// [spec.Resource.Input]. Note that spec.Resource's who have their access predefined, are never processed
+// with a ResourceInputMethod, but are directly added to the component version repository.
+// any spec.Resource passed MUST have its [spec.Resource.Input] field set.
+// If the input method does not support the given input specification it MAY reject the request
+//
+// The method will get called with the raw specification specified in the constructor and is expected
+// to return a ResourceMethodResult or an error.
```
Skarlso, line 208:
```suggestion
// to return a ResourceInputMethodResult or an error.
```

right?

docs/adr/0006_component_constructors.md
```diff
@@ -0,0 +1,363 @@
...
+    // to get the source input method for the component specification when processing sources with an input method.
+    SourceInputMethodProvider
+
+    // While constructing a component version, the constructor library will use the 
+    // given function to decide whether a resource should be processed by value or not.
+    ProcessByValue func(*constructor.Resource) bool
+}
+
+// Construct constructs a component version based on the given specification and options.
+func Construct(ctx context.Context, specification *constructor.Constructor, opts Options) ([]*descriptor.Descriptor, error)
```
Skarlso, line 340: What is `*constructor.Constructor`? Did I miss the definition of that?

docs/adr/0006_component_constructors.md
```diff
@@ -0,0 +1,363 @@
...
+  less experienced engineers.
+* Extensibility: The component constructor schema must stay flexible enough to allow for the addition of new input methods
+  and the ability to extend existing ones.
+* Maintainability: The constructor logic and input methods should stay separated from existing OCI libraries and
+  binding libraries. This will allow us to maintain the component constructor schema and its input methods without
+  affecting the existing libraries.
+
+## Outcome
+
+TBD
```
Skarlso, line 44: Do you have any alternatives? If not, you could pre-fill this with the decision outcome because it will pretty much be this. :D I wouldn't mind. I don't really have anything against this approach. I'm pondering for a better interface, maybe I can come up with some better names. :P

Skarlso (commented):
A few more questions for clarification. Otherwise, LGTM honestly.

docs/adr/0006_component_constructors.md
jakobmoellerdev: Yeah, I was referring to generic input methods

docs/adr/0006_component_constructors.md
jakobmoellerdev: at runtime means in the moment the constructor is actually used, ie. within the CLIs command call

docs/adr/0006_component_constructors.md
jakobmoellerdev: The component version contains a descriptor which receives this resource through the embedding. Does that make more sense?

docs/adr/0006_component_constructors.md
jakobmoellerdev: Actually if ProcessedBlobData is provided, then the component version will also have a local blob reference like in "by value". However "by value" forces results that were not some local blob to become local blobs.

docs/adr/0006_component_constructors.md
jakobmoellerdev: The Constructor is part of the specification as part of https://github.com/open-component-model/open-component-model/pull/111. The package is quite large since it is mostly a redefinition of the component descriptors so thats why I didnt add it to the contract. Do you prefer me to add that here?

docs/adr/0006_component_constructors.md
jakobmoellerdev: Sure, I was gonna wait for initial feedback before suggesting a decision :D

docs/adr/0006_component_constructors.md
Skarlso: Hmm, hmmm. Maybe just something small, you know? I'm thinking that reading an ADR should be possible to understand things without the need to navigate away? Not sure.

frewilhelm (commented):
In general very nicely written. Thanks a lot!

Maybe what I am missing are examples why the extensibility is beneficial. Why should the model be extensible? What input types are likely to be implemented?

docs/adr/0006_component_constructors.md
Skarlso: Yep

docs/adr/0006_component_constructors.md
```diff
@@ -0,0 +1,363 @@
...
+```
+
+To allow for actual construction of component versions, the constructor library provides a central entry function:
+
+```go
+// TargetRepository defines the interface for a target repository that can store component versions and associated local resources
+type TargetRepository interface {
+    // AddLocalResource adds a local resource to the repository.
+    // The resource must be referenced in the component descriptor.
+    // Resources for non-existent component versions may be stored but may be removed during garbage collection.
```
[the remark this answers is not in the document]
jakobmoellerdev: This is a blatant copy & paste of the repository interface in our OCI implementation. Do you think we should remove the comment here and reference it? Ultimately the interface is very generic.

docs/adr/0006_component_constructors.md
Skarlso: Meh, it's fine. I was looking for it, but couldn't find anything that would say we can't do something like that. So this is fine.

docs/adr/0006_component_constructors.md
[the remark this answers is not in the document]
jakobmoellerdev: I added some context

docs/adr/0006_component_constructors.md
jakobmoellerdev: I added a reference to the specification as a separate section as it was too huge for the ADR to include the entire spec

docs/adr/0006_component_constructors.md
Skarlso: You didn't actually commit this. :D

docs/adr/0006_component_constructors.md
```diff
@@ -0,0 +1,379 @@
...
+
+
+// SourceInputMethod is the interface for processing a source with an input method declared as per
+// [spec.Source.Input]. Note that spec.Source's who have their access predefined, are never processed
+// with a ResourceInputMethod, but are directly added to the component version repository.
+// any spec.Source passed MUST have its [spec.Source.Input] field set.
+// If the input method does not support the given input specification it MAY reject the request
+//
+// The method will get called with the raw specification specified in the constructor and is expected
+// to return a ResourceMethodResult or an error.
```
Skarlso, line 256:
```suggestion
// to return a SourceMethodInputResult or an error.
```

frewilhelm (commented):
I was wondering if `by reference`-accesses would download the referenced resource as well to calculate the resource digest to put it in the CV. This is the current behaviour of ocm lib v1. However, this is not mentioned in this ADR.

There several arguments for downloading the resources (or getting the resource digest through an OCI manifest) and against it. For instance, secure delivery vs. separation of concerns vs. substantial changes in the cli usage for users...

Is the discussion or "feature" also in scope for this ADR or should we postpone it?

docs/adr/0006_component_constructors.md
jakobmoellerdev: now i did. *pinky promise*

jakobmoellerdev:
> I was wondering if by reference-accesses would download the referenced resource as well to calculate the resource digest to put it in the CV. This is the current behaviour of ocm lib v1. However, this is not mentioned in this ADR.

I agree, I ammended a section on the ADR to describe specifically how to handle digest behavior through a digest provider. The advantage of a dedicated digest provider is that we can more efficiently compute digests: while in the past OCMv1 had to download entire resources, I propose that the underlying implementation can decide how to amend the digest. For OCI for example, the implementation would just need to resolve the descriptor.

> There several arguments for downloading the resources (or getting the resource digest through an OCI manifest) and against it. For instance, secure delivery vs. separation of concerns vs. substantial changes in the cli usage for users...

For me secure delivery means resources should be able to be provided with a digest for our most supported accesses. But for extension purposes we should probably go for a flow such as above. I ammended the PR as such with a separate section. Also note that this only applies to resources, because sources do not have digests.

> Is the discussion or "feature" also in scope for this ADR or should we postpone it?

definetly IN SCOPE for ADR, but implementation may need to be done in steps.

PTAL at the changed ADR @frewilhelm and thanks for bringing it up!

Skarlso (approved):
LGTM now

PR #111, opened 2025-05-19
feat: add component constructor specification

jakobmoellerdev:

This adds component constructors with a corresponding JSON Scheme. 
see `doc.go` for more information.

This is the first step in partializing the big bad PR for component constructors.

This introduces specifications only, without including loging for handling component construction itself. It rather just describes what a construction should look like


bindings/go/constructor/spec/v1/constructor.go
```diff
@@ -0,0 +1,290 @@
...
+
+	"ocm.software/open-component-model/bindings/go/runtime"
+)
+
+type ComponentConstructor struct {
+	Components []Component `json:"components"`
+}
+
+func (c *ComponentConstructor) UnmarshalJSON(data []byte) error {
+	// Unmarshal the JSON data into the ComponentConstructor struct
```
Skarlso, line 16: Why? 🤔  Is it because during  failure it could potentially blank an existing struct? Would that be bad? 🤔

bindings/go/constructor/spec/v1/validate.go
```diff
@@ -0,0 +1,92 @@
...
+	"sigs.k8s.io/yaml"
+)
+
+// JSONSchema contains the embedded JSON schema for validating Open Component Model descriptors.
+//
+//go:embed resources/schema-2020-12.json
+var JSONSchema []byte
+
+// GetJSONSchema is a singleton that compiles the JSON schema once and caches it for reuse.
+var GetJSONSchema = sync.OnceValues[*jsonschema.Schema, error](func() (*jsonschema.Schema, error) {
```
Skarlso, line 21: Nice :)

Skarlso (approved):
Just a single question. Otherwise this is LGTM.

bindings/go/constructor/spec/v1/constructor.go
jakobmoellerdev: because there is the potential to have one component descriptor instead of many as part of the schema. so I have 2 potential unmarshals. However While im writing this I noticed to add the second case X) thanks for bringing it up again

Skarlso (approved):

PR #166, opened 2025-05-21
Add configuration to plugins

Skarlso:


bindings/go/plugin/manager/manager.go
```diff
@@ -64,7 +65,7 @@ func WithIdleTimeout(d time.Duration) RegistrationOptionFn {
 // RegisterPlugins walks through files in a folder and registers them
 // as plugins if connection points can be established. This function doesn't support
 // concurrent access.
-func (pm *PluginManager) RegisterPlugins(ctx context.Context, dir string, opts ...RegistrationOptionFn) error {
+func (pm *PluginManager) RegisterPlugins(ctx context.Context, config *v1.Config, dir string, opts ...RegistrationOptionFn) error {
```
Skarlso, line 68: @jakobmoellerdev I imagined this to be called from CLI side or wherever and then a parsed config would be given to the registration. From there, using Filter, we get the right config to the right plugin. This sounds simple enough....

bindings/go/plugin/go.mod
```diff
@@ -10,15 +10,19 @@ require (
 	ocm.software/open-component-model/bindings/go/descriptor/runtime v0.0.0-20250521120014-424308d81e7b
 	ocm.software/open-component-model/bindings/go/descriptor/v2 v2.0.0-20250520105023-82a459e39e46
 	ocm.software/open-component-model/bindings/go/runtime v0.0.1
+	ocm.software/open-component-model/cli v0.0.0-20250521134817-41f55c131044
```
jakobmoellerdev, line 13: this is definetly incorrect, CLI should not be a dependency to plugin

bindings/go/plugin/internal/testplugin/main.go
```diff
@@ -110,6 +115,11 @@ func main() {
 
 	logger.Info("registered test plugin")
 
+	capabilities.AddConfigType(runtime.Type{
```
jakobmoellerdev, line 118: this type should be derived from the config above right? now you configure it manually

bindings/go/plugin/manager/manager.go
jakobmoellerdev: I was hoping the config would be passed as a RegistrationOptionFn

bindings/go/plugin/manager/manager.go
```diff
@@ -169,7 +170,21 @@ func (pm *PluginManager) fetchPlugins(ctx context.Context, conf *mtypes.Config,
...
 
-func (pm *PluginManager) addPlugin(ctx context.Context, plugin mtypes.Plugin, output *bytes.Buffer) error {
+func (pm *PluginManager) addPlugin(ctx context.Context, ocmConfig *v1.Config, plugin mtypes.Plugin, capabilitiesCommandOutput *bytes.Buffer) error {
+	// Determine Configuration requirements.
+	types := &mtypes.Types{}
+	if err := json.Unmarshal(capabilitiesCommandOutput.Bytes(), types); err != nil {
+		return fmt.Errorf("failed to unmarshal capabilities: %w", err)
+	}
+
+	// TODO: config is a flattened config map; but is that right? Will all configs be in `Configuration`?
```
jakobmoellerdev, line 180: this looks ok :)

jakobmoellerdev: but the best way to test is to write an end to end test for this!

jakobmoellerdev (commented):
Im just missing a test where you parse a string based config (e.g. a yaml) with a custom plugin config type, and then filter that and pass it around.

Apart from that, lifecycle wise this looks good :)

bindings/go/plugin/internal/testplugin/main.go
Skarlso: >this type should be derived from the config above right?

Hm? Which one would that be?

Skarlso:
> Im just missing a test where you parse a string based config (e.g. a yaml) with a custom plugin config type, and then filter that and pass it around.

How would I do this if my configuration is an actual struct? 🤔  What scenario would that cover that this test isn't doing right now https://github.com/open-component-model/open-component-model/pull/166/files#diff-2a6832e44914f859b3da975a3e670f68cd5bea364e85fab7f29875c165953894R83 if I understand you correctly. Like, which part here would be string based config that is marshalled into a config that we then pass in as a struct anyways? Could you elaborate please? :)

bindings/go/plugin/go.mod
Skarlso: https://github.com/open-component-model/open-component-model/pull/171

bindings/go/plugin/internal/testplugin/main.go
jakobmoellerdev: I was hoping you would have a scheme that you derive the type from but for this test its fine :)

jakobmoellerdev:
I think for the above, my preference would be that you have a test that actually tests the e2e flow of the configuration being passed and then that triggering an update but I guess you kinda already have that. So consider my point moot!

jakobmoellerdev (commented):
Mostly great extension, 2 questions otherwise LGTM

bindings/go/plugin/manager/endpoints/endpoint_constructors.go
```diff
@@ -42,3 +42,9 @@ func (c *EndpointBuilder) MarshalJSON() ([]byte, error) {
 func (c *EndpointBuilder) GetHandlers() []Handler {
 	return c.Handlers
 }
+
+// AddConfigType adds a configuration type to the list of supported config types.
+// TODO(Gergely Brautigam): This could live on the config object's side.
```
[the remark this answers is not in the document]
Skarlso: I think I can delete this. :D It was some previous thing I was thinking about.

bindings/go/plugin/internal/testplugin/main.go
```diff
@@ -92,11 +92,16 @@ func (m *TestPlugin) GetIdentity(ctx context.Context, typ repov1.GetIdentityRequ
...
+type Config struct {
+	MaximumNumberOfPotatoes string `json:"maximumNumberOfPotatoes"`
+}
+
 func main() {
 	args := os.Args[1:]
 	// log messages are shared over stderr by convention established by the plugin manager.
 	logger := slog.New(slog.NewJSONHandler(os.Stderr, &slog.HandlerOptions{
-		Level: slog.LevelDebug, // debug level here is respected when sending this message.
+		Level: slog.LevelInfo, // debug level here is respected when sending this message.
```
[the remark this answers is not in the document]
Skarlso: Ah no, this can stay on debug.

jakobmoellerdev (approved):

PR #189, opened 2025-05-23
feat: add component version command

jakobmoellerdev:

see `docs/reference/cli/ocm_add_component-version.md` for more details 

fix https://github.com/open-component-model/ocm-project/issues/491

### DIFF
--- a/cli/cmd/add/cmd.go
+++ b/cli/cmd/add/cmd.go
@@ -1,0 +1,20 @@
+package add
+
+import (
+	"github.com/spf13/cobra"
+
+	componentversion "ocm.software/open-component-model/cli/cmd/add/component-version"
+)
+
+// New represents any command that is related to retrieving ( "get"ting ) objects
+func New() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "add {component-version|component-versions|cv|cvs}",
+		Short: "Add anything to OCM",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return cmd.Help()
+		},
+	}
+	cmd.AddCommand(componentversion.New())
+	return cmd
+}
--- a/cli/cmd/add/component-version/cmd.go
+++ b/cli/cmd/add/component-version/cmd.go
@@ -1,0 +1,270 @@
+package componentversion
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"log/slog"
+	"os"
+	"path/filepath"
+	"strconv"
+	"strings"
+
+	"github.com/spf13/cobra"
+	"sigs.k8s.io/yaml"
+
+	"ocm.software/open-component-model/bindings/go/blob"
+	"ocm.software/open-component-model/bindings/go/blob/filesystem"
+	"ocm.software/open-component-model/bindings/go/constructor"
+	constructorv1 "ocm.software/open-component-model/bindings/go/constructor/spec/v1"
+	"ocm.software/open-component-model/bindings/go/credentials"
+	descriptor "ocm.software/open-component-model/bindings/go/descriptor/runtime"
+	ctfv1 "ocm.software/open-component-model/bindings/go/oci/spec/repository/v1/ctf"
+	"ocm.software/open-component-model/bindings/go/plugin/manager"
+	v1 "ocm.software/open-component-model/bindings/go/plugin/manager/contracts/ocmrepository/v1"
+	"ocm.software/open-component-model/bindings/go/plugin/manager/types"
+	"ocm.software/open-component-model/bindings/go/runtime"
+	ocmctx "ocm.software/open-component-model/cli/internal/context"
+	"ocm.software/open-component-model/cli/internal/flags/file"
+)
+
+const (
+	FlagConcurrencyLimit         = "concurrency-limit"
+	FlagRepositoryRef            = "repository"
+	FlagComponentConstructorPath = "constructor"
+	FlagCopyResources            = "copy-resources"
+	FlagBlobCacheDirectory       = "blob-cache-directory"
+
+	DefaultComponentConstructorBaseName = "component-constructor"
+	LegacyDefaultArchiveName            = "transport-archive"
+)
+
+func New() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:        fmt.Sprintf("component-version"),
+		Aliases:    []string{"cv", "component-versions", "cvs"},
+		SuggestFor: []string{"component", "components", "version", "versions"},
+		Short:      fmt.Sprintf("Add component version(s) to an OCM Repository stored as Common Transport Format Archive (CTF) based on a %[1]q file", DefaultComponentConstructorBaseName),
+		Args:       cobra.NoArgs,
+		Long: fmt.Sprintf(`Add component version(s) to an OCM Common Transport Format Archive (CTF) that can be reused
+for transfers.
+
+A %[1]q file is used to specify the component version(s) to be added. It can contain both a single component or many
+components. The component reference is used to determine the repository to add the components to.
+
+By default, the command will look for a file named "%[1]q.yaml" or "%[1]q.yml" in the current directory.
+If given a path to a directory, the command will look for a file named "%[1]s.yaml" or "%[1]s.yml" in that directory.
+If given a path to a file, the command will attempt to use that file as the %[1]q file.
+
+In case the component archive does not exist, it will be created by default.
+`,
+			DefaultComponentConstructorBaseName,
+		),
+		Example: strings.TrimSpace(fmt.Sprintf(`
+Adding component versions to a non-default CTF named %[1]q based on a non-default default %[2]q file:
+
+add component-version ./path/to/%[1]s ./path/to/%[2]s.yaml
+`, LegacyDefaultArchiveName, DefaultComponentConstructorBaseName)),
+		RunE:              AddComponentVersion,
+		DisableAutoGenTag: true,
+	}
+
+	cmd.Flags().Int(FlagConcurrencyLimit, 4, "maximum amount of parallel requests to the repository for resolving component versions")
+	file.VarP(cmd.Flags(), FlagRepositoryRef, string(FlagRepositoryRef[0]), LegacyDefaultArchiveName, "path to the repository")
+	file.VarP(cmd.Flags(), FlagComponentConstructorPath, string(FlagComponentConstructorPath[0]), DefaultComponentConstructorBaseName+".yaml", "path to the repository")
+	cmd.Flags().Bool(FlagCopyResources, false, "copy external resources by-value to the archive")
+	cmd.Flags().String(FlagBlobCacheDirectory, "blobs", "path to the blob cache directory")
+
+	return cmd
+}
+
+func AddComponentVersion(cmd *cobra.Command, _ []string) error {
+	pluginManager := ocmctx.FromContext(cmd.Context()).PluginManager()
+	if pluginManager == nil {
+		return fmt.Errorf("could not retrieve plugin manager from context")
+	}
+
+	credentialGraph := ocmctx.FromContext(cmd.Context()).CredentialGraph()
+	if credentialGraph == nil {
+		return fmt.Errorf("could not retrieve credential graph from context")
+	}
+
+	concurrencyLimit, err := cmd.Flags().GetInt(FlagConcurrencyLimit)
+	if err != nil {
+		return fmt.Errorf("getting concurrency-limit flag failed: %w", err)
+	}
+
+	copyResources, err := cmd.Flags().GetBool(FlagCopyResources)
+	if err != nil {
+		return fmt.Errorf("getting copy-resources flag failed: %w", err)
+	}
+
+	repoSpec, err := GetRepositorySpec(cmd)
+	if err != nil {
+		return fmt.Errorf("getting repository spec failed: %w", err)
+	}
+
+	cacheDir, err := cmd.Flags().GetString(FlagBlobCacheDirectory)
+
+	constructorSpec, err := GetComponentConstructor(cmd)
+	if err != nil {
+		return fmt.Errorf("getting component constructor failed: %w", err)
+	}
+
+	instance := constructorProvider{
+		cache:          cacheDir,
+		targetRepoSpec: repoSpec,
+		PluginManager:  pluginManager,
+	}
+
+	_, err = constructor.Construct(cmd.Context(), constructorSpec, constructor.Options{
+		TargetRepositoryProvider:        instance,
+		ResourceRepositoryProvider:      instance,
+		ResourceInputMethodProvider:     instance,
+		ResourceDigestProcessorProvider: instance,
+		ProcessByValue: func(resource *constructorv1.Resource) bool {
+			if resource.Relation == constructorv1.ExternalRelation {
+				return copyResources
+			}
+			return true
+		},
+		ConcurrencyLimit: concurrencyLimit,
+	})
+
+	return err
+}
+
+func GetRepositorySpec(cmd *cobra.Command) (runtime.Typed, error) {
+	repoRef, err := file.Get(cmd.Flags(), FlagRepositoryRef)
+	if err != nil {
+		return nil, fmt.Errorf("getting repository reference flag failed: %w", err)
+	}
+	var accessMode ctfv1.AccessMode = ctfv1.AccessModeReadWrite
+	if !repoRef.Exists() {
+		accessMode += "|" + ctfv1.AccessModeCreate
+	}
+	repoSpec := ctfv1.Repository{
+		Path:       repoRef.String(),
+		AccessMode: accessMode,
+	}
+	return &repoSpec, nil
+}
+
+func GetComponentConstructor(cmd *cobra.Command) (*constructorv1.ComponentConstructor, error) {
+	constructorFlag, err := file.Get(cmd.Flags(), FlagComponentConstructorPath)
+	if err != nil {
+		return nil, fmt.Errorf("getting component constructor path flag failed: %w", err)
+	}
+	if constructorFlag.IsDir() {
+		return nil, fmt.Errorf("path %q is a directory but must point to a component constructor", constructorFlag.String())
+	} else if !constructorFlag.Exists() {
+		return nil, fmt.Errorf("component constructor %q does not exist", constructorFlag.String())
+	}
+	constructorStream, err := constructorFlag.Open()
+	if err != nil {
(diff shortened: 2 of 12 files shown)


cli/cmd/add/cmd.go
```diff
@@ -0,0 +1,20 @@
+package add
+
+import (
+	"github.com/spf13/cobra"
+
+	componentversion "ocm.software/open-component-model/cli/cmd/add/component-version"
+)
+
+// New represents any command that is related to retrieving ( "get"ting ) objects
```
Skarlso, line 9: This is an add command. :)

cli/cmd/add/component-version/cmd.go
```diff
@@ -0,0 +1,359 @@
...
+add component-version  --%[1]s ./path/to/%[2]s --%[3]s ./path/to/%[4]s.yaml
+`, FlagRepositoryRef, LegacyDefaultArchiveName, FlagComponentConstructorPath, DefaultComponentConstructorBaseName)),
+		RunE:              AddComponentVersion,
+		DisableAutoGenTag: true,
+	}
+
+	cmd.Flags().Int(FlagConcurrencyLimit, 4, "maximum amount of parallel requests to the repository for resolving component versions")
+	file.VarP(cmd.Flags(), FlagRepositoryRef, string(FlagRepositoryRef[0]), LegacyDefaultArchiveName, "path to the repository")
+	file.VarP(cmd.Flags(), FlagComponentConstructorPath, string(FlagComponentConstructorPath[0]), DefaultComponentConstructorBaseName+".yaml", "path to the repository")
+	cmd.Flags().String(FlagBlobCacheDirectory, filepath.Join(".ocm", "cache"), "path to the blob cache directory")
```
Skarlso, line 103: Should this be on the path of the running command?

cli/cmd/add/component-version/cmd.go
```diff
@@ -0,0 +1,359 @@
...
+			tracker := trackers[key]
+			tracker.UpdateMessage(tracker.Message + " constructed")
+			tracker.Increment(1)
+			tracker.MarkAsDone()
+			return nil
+		}
+		// TODO Add Resource and Source tracking in more detail
+		go func() {
+			for _, tracker := range trackers {
+				tracker.Start()
```
Skarlso, line 324: Does this need some kind of stop mechanism / timeout?

cli/cmd/add/component-version/cmd.go
```diff
@@ -0,0 +1,359 @@
...
+	if stop != nil {
+		stop()
+	}
+
+	time.Sleep(10 * time.Millisecond)
+	return err
+}
+
+func GetRepositorySpec(cmd *cobra.Command) (runtime.Typed, error) {
+	repoRef, err := file.Get(cmd.Flags(), FlagRepositoryRef)
```
Skarlso, line 186: I'm pretty sure we need a filepath.Clean here. WDYT?

cli/cmd/add/component-version/cmd.go
```diff
@@ -0,0 +1,359 @@
...
+		return nil, fmt.Errorf("reading component constructor %q failed: %w", constructorFlag.String(), err)
+	}
+	data := constructorv1.ComponentConstructor{}
+	if err := yaml.Unmarshal(constructorData, &data); err != nil {
+		return nil, fmt.Errorf("unmarshalling component constructor %q failed: %w", constructorFlag.String(), err)
+	}
+
+	converted := constructorruntime.ConvertToRuntimeConstructor(&data)
+
+	return converted, nil
```
Skarlso, line 231:
```suggestion
	return constructorruntime.ConvertToRuntimeConstructor(&data), nil
```

cli/internal/plugin/builtin/oci/resource_repository.go
```diff
@@ -0,0 +1,163 @@
...
+		return nil, fmt.Errorf("unsupported type %s for downloading the resource", t)
+	}
+}
+
+// TODO Repeated calls with separate credentials will always use the first credentials set.
+//
+//	we need to be able to dynamically inject credentials to an existing repository instance.
+func (p *ResourceRepositoryPlugin) getRepository(spec *ociv1.Repository, creds map[string]string) (Repository, error) {
+	key := spec.BaseUrl
+	if repo, ok := p.repoCache.Get(key); ok {
```
Skarlso, line 142: Wait. Is this right? If I have multiple things from different things that happened to have the same baseURL they could potentially use the same cred even though they might NOT have that cred? I feel like that's a security issue isn't it?

Skarlso (commented):
I have some initial comments and questions with a first run-through. :)

cli/cmd/add/component-version/cmd.go
```diff
@@ -0,0 +1,359 @@
...
+	if err != nil {
+		return fmt.Errorf("registering constructor progress tracker failed: %w", err)
+	}
+
+	_, err = constructor.ConstructDefault(cmd.Context(), constructorSpec, opts)
+	if stop != nil {
+		stop()
+	}
+
+	time.Sleep(10 * time.Millisecond)
```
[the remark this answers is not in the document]
jakobmoellerdev: youre right, will attempt to remove this

cli/cmd/add/component-version/cmd.go
jakobmoellerdev: Currently we dont have a central cache directory that we can configure via OCM config. I think the next best thing would be to introduce the old OCM v1 caching configuration. WDYT?

cli/cmd/add/component-version/cmd.go
jakobmoellerdev: the starting itself does not block, so imo it doesnt need a timeout. Will double check

cli/cmd/add/component-version/cmd.go
jakobmoellerdev: file.Get fetches a filepath. What would be the reason for a Clean here?

cli/cmd/add/component-version/cmd.go
```diff
@@ -0,0 +1,359 @@
...
+
+	opts, stop, err := registerConstructorProgressTracker(cmd, opts)
+	if err != nil {
+		return fmt.Errorf("registering constructor progress tracker failed: %w", err)
+	}
+
+	_, err = constructor.ConstructDefault(cmd.Context(), constructorSpec, opts)
+	if stop != nil {
+		stop()
+	}
```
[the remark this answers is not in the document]
Skarlso: Thanks. Maybe it's fine. :)

cli/cmd/add/component-version/cmd.go
Skarlso: The filepath is user defined isn't it? Unless I'm wrong. In that case, it's prone to injections.

cli/cmd/add/component-version/cmd.go
Skarlso: Yah, works for me.

cli/cmd/add/component-version/cmd.go
jakobmoellerdev: its a flag so how would that injection attack work? Im just wondering how a possible attack vector might look like. I think we can add clean in the file package of course

Skarlso: Nah, you're fine, I think I misunderstood where the value is coming from.

cli/internal/plugin/builtin/oci/resource_repository.go
jakobmoellerdev: removed the cache for now

cli/cmd/add/cmd.go
jakobmoellerdev: thanks! adjusted

cli/cmd/add/component-version/cmd.go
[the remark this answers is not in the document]
jakobmoellerdev: I did adjust this towards a proper defer handling now

cli/cmd/add/component-version/cmd.go
```diff
@@ -0,0 +1,359 @@
...
+}
+
+func (prov *constructorProvider) GetTargetRepository(ctx context.Context, _ *constructorruntime.Component) (constructor.TargetRepository, error) {
+	plugin, err := prov.pluginManager.ComponentVersionRepositoryRegistry.GetPlugin(ctx, prov.targetRepoSpec)
+	if err != nil {
+		return nil, fmt.Errorf("getting plugin for repository %q failed: %w", prov.targetRepoSpec, err)
+	}
+	var creds map[string]string
+	identity, err := plugin.GetComponentVersionRepositoryCredentialConsumerIdentity(ctx, prov.targetRepoSpec)
+	if err == nil {
```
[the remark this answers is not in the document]
jakobmoellerdev: added an else branch with a warning log!

cli/cmd/add/component-version/cmd.go
```diff
@@ -0,0 +1,359 @@
...
+			key := descriptor.Component.Name + "/" + descriptor.Component.Version
+			tracker := trackers[key]
+			tracker.UpdateMessage(tracker.Message + " constructed")
+			tracker.Increment(1)
+			tracker.MarkAsDone()
+			return nil
+		}
+		// TODO Add Resource and Source tracking in more detail
+		go func() {
+			for _, tracker := range trackers {
```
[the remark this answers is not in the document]
jakobmoellerdev: changed to sync map implementation (typed, we will probably need this anyways a few more times)

cli/internal/sync/map.go
```diff
@@ -0,0 +1,63 @@
+package sync
```
Skarlso, line 1: Why have this instead of sync.Map? 🤔

jakobmoellerdev: because I want to be type safe. I hate that sync.Map always needs type assertion

Skarlso: Fair enough.

cli/cmd/add/component-version/cmd.go
```diff
@@ -0,0 +1,381 @@
...
+			}
+			tracker.UpdateMessage(tracker.Message + " constructed")
+			tracker.Increment(1)
+			tracker.MarkAsDone()
+			return nil
+		}
+		// TODO Add Resource and Source tracking in more detail
+		go func() {
+			trackers.Range(func(_ string, tracker *progress.Tracker) bool {
+				tracker.Start()
```
Skarlso, line 325: Okay, so... You're filling up the tracker in `OnStartComponentConstruct`. But you're immediately starting this go routine. How is not immediately returning? Since the tracker is empty at that point yet until wither OnStartComponentConstruct happens, right?

cli/internal/sync/map.go
```diff
@@ -0,0 +1,63 @@
...
+
+// Load returns the value stored in the map for a key, or nil if no
+// value is present.
+// The ok result indicates whether value was found in the map.
+func (m *Map[K, V]) Load(key K) (value V, ok bool) {
+	v, ok := m.Untyped.Load(key)
+	if !ok {
+		return value, ok
+	}
+	return v.(V), ok
```
Skarlso, line 26: I think you need type assertions on these returns, right?

jakobmoellerdev: I made the map unexported now so it is impossible to write into it in a case where the return type can not be cast.

cli/cmd/add/component-version/cmd.go
Skarlso: Sure, but the Range is never ranging over anything because this is immediately started and you're only filling the tracker in that OnStartCOmponentConstruct which happens later, right? So this doesn't really do anything other than immediately just call Render with no started trackers.

jakobmoellerdev: now i got you. youre absolutely right, good catch! thanks!

cli/cmd/add/component-version/cmd.go
```diff
@@ -0,0 +1,378 @@
...
+			tracker.Increment(1)
+			tracker.MarkAsDone()
+			return nil
+		}
+		// TODO(jakobmoellerdev): Add Resource and Source tracking in more detail so we can track those as well.
+		go func() {
+			// this is the actual blocking loop
+			pw.Render()
+		}()
+
```
Skarlso, line 327:
```suggestion
		go pw.Render()
```

cli/internal/plugin/builtin/oci/resource_repository.go
```diff
@@ -0,0 +1,157 @@
...
+	ociv1 "ocm.software/open-component-model/bindings/go/oci/spec/repository/v1/oci"
+	"ocm.software/open-component-model/bindings/go/runtime"
+)
+
+type ResourceRepositoryPlugin struct {
+	scheme            *runtime.Scheme
+	manifests, layers cache.OCIDescriptorCache
+}
+
+func (p *ResourceRepositoryPlugin) GetResourceDigestProcessorCredentialConsumerIdentity(ctx context.Context, resource *descriptor.Resource) (identity runtime.Identity, err error) {
```
Skarlso, line 23:
```suggestion
func (p *ResourceRepositoryPlugin) GetResourceDigestProcessorCredentialConsumerIdentity(ctx context.Context, resource *descriptor.Resource) (runtime.Identity, error) {
```

cli/internal/plugin/builtin/oci/resource_repository.go
```diff
@@ -0,0 +1,157 @@
...
+			return nil, fmt.Errorf("error downloading resource: %w", err)
+		}
+
+		return b, nil
+	default:
+		return nil, fmt.Errorf("unsupported type %s for downloading the resource", t)
+	}
+}
+
+// TODO Repeated calls with separate credentials will always use the first credentials set.
```
Skarlso, line 136: Is still still true? 🤔

cli/internal/plugin/builtin/oci/resource_repository.go
```diff
@@ -0,0 +1,157 @@
...
+func ociImageAccessToBaseURL(access *v1.OCIImage) (string, error) {
+	ref, err := registry.ParseReference(access.ImageReference)
+	if err != nil {
+		return "", fmt.Errorf("error parsing image reference %q: %w", access.ImageReference, err)
+	}
+	baseURL := ref.Host()
+	if ref.Repository != "" {
+		baseURL += "/" + ref.Repository
+	}
+	return baseURL, nil
```
Skarlso, line 156: Errr, what is happening here? :D 

So if an image ref is like example.com/registry/repo:4444@sha... this would be `ghcr.io/registry/repo` I guess? Why are we adding the repo to the base url?

Skarlso (commented):
Just a couple of questions really, nothing major. I think this is looking awesome now.

cli/cmd/add/component-version/cmd.go
```diff
@@ -0,0 +1,378 @@
...
+	}
+}
+
+func ComponentVersionOverridePolicies() []string {
+	return []string{
+		string(ComponentVersionConflictPolicyAbortAndFail),
+		string(ComponentVersionConflictPolicySkip),
+		string(ComponentVersionConflictPolicyReplace),
+	}
+}
```
fabianburth, line 69: Is there a particular reason we call it `OverridePolicies` here instead of `ConflictPolicies`?

cli/cmd/add/component-version/cmd.go
```diff
@@ -0,0 +1,378 @@
...
+
+add component-version  --%[1]s ./path/to/%[2]s --%[3]s ./path/to/%[4]s.yaml
+`, FlagRepositoryRef, LegacyDefaultArchiveName, FlagComponentConstructorPath, DefaultComponentConstructorBaseName)),
+		RunE:              AddComponentVersion,
+		DisableAutoGenTag: true,
+	}
+
+	cmd.Flags().Int(FlagConcurrencyLimit, 4, "maximum amount of parallel requests to the repository for resolving component versions")
+	file.VarP(cmd.Flags(), FlagRepositoryRef, string(FlagRepositoryRef[0]), LegacyDefaultArchiveName, "path to the repository")
+	file.VarP(cmd.Flags(), FlagComponentConstructorPath, string(FlagComponentConstructorPath[0]), DefaultComponentConstructorBaseName+".yaml", "path to the repository")
```
fabianburth, line 103:
```suggestion
	file.VarP(cmd.Flags(), FlagComponentConstructorPath, string(FlagComponentConstructorPath[0]), DefaultComponentConstructorBaseName+".yaml", "path to the component constructor")
```

cli/cmd/add/component-version/cmd.go
```diff
@@ -0,0 +1,378 @@
...
+		Example: strings.TrimSpace(fmt.Sprintf(`
+Adding component versions to a non-default CTF named %[2]q based on a non-default default %[4]q file:
+
+add component-version  --%[1]s ./path/to/%[2]s --%[3]s ./path/to/%[4]s.yaml
+`, FlagRepositoryRef, LegacyDefaultArchiveName, FlagComponentConstructorPath, DefaultComponentConstructorBaseName)),
+		RunE:              AddComponentVersion,
+		DisableAutoGenTag: true,
+	}
+
+	cmd.Flags().Int(FlagConcurrencyLimit, 4, "maximum amount of parallel requests to the repository for resolving component versions")
```
fabianburth, line 101: is this really for "resolving component versions"?

cli/internal/plugin/builtin/oci/resource_repository.go
jakobmoellerdev: Errrhh I wish I knew but frankly I cant remember.. I will take a look at this again. Thanks for bringing it up again

Skarlso: If you figure it out leave a comment on the function. :D

cli/internal/plugin/builtin/oci/resource_repository.go
jakobmoellerdev: Will have to double check this. I think this might still be the case. We dont dynamically override credentials I believe

cli/cmd/add/component-version/cmd.go
jakobmoellerdev: Errrhh probably leftover, will normalize

cli/cmd/add/component-version/cmd.go
jakobmoellerdev: corrected

cli/cmd/add/component-version/cmd.go
jakobmoellerdev: corrected description

cli/internal/plugin/builtin/oci/resource_repository.go
Skarlso: Noice!

PR #199, opened 2025-05-26
feat: component constructor library

jakobmoellerdev:

This sets up the default constructor library

It is the implementation that is used whenever constructing components with our go bindings.

This is a barebones implementation without any input method. This means that by default, it cannot process anything.

Only by adding input methods will it become more powerful.

This extension is made possible through the plugin manager / integration with the various provider interfaces (see options.go)

part of https://github.com/open-component-model/ocm-project/issues/460

implements the first part of component construction => the construction library itself.

once we have that the next step is integrating it into the plugin system and then 
using it in https://github.com/open-component-model/open-component-model/pull/189


bindings/go/constructor/input_method_registry.go
```diff
@@ -0,0 +1,93 @@
...
+		return runtime.Type{}, err
+	}
+
+	return typ, nil
+}
+
+// GetResourceInputMethod retrieves the resource input method for a given typed object.
+// Returns the method and a boolean indicating if the method was found.
+func (r *InputMethodRegistry) GetResourceInputMethod(_ context.Context, res *constructorv1.Resource) (ResourceInputMethod, error) {
+	if res == nil || !res.HasInput() {
```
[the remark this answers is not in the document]
jakobmoellerdev: considering this a nit and resolving because I dont think it matters

bindings/go/constructor/default.go
```diff
@@ -0,0 +1,40 @@
...
+	// Scheme is the default runtime scheme for the registry
+	Scheme = runtime.NewScheme()
+	// DefaultInputMethodRegistry is the default registry instance using the default scheme
+	DefaultInputMethodRegistry = New(Scheme)
+)
+
+// MustAddToScheme registers the file and UTF8 types with the given scheme.
+// It registers both versioned and unversioned types.
+func MustAddToScheme(scheme *runtime.Scheme) {
+	// scheme.MustRegisterWithAlias(&filev1.File{},
```
Skarlso, line 17: Errr.. I assume this is for testing?

jakobmoellerdev: it was a left over from my extraction to make the PRs smaller. I can remove it

frewilhelm (commented):
Missing a `doc.go` (especially because usually they are very well written by you!)

bindings/go/constructor/construct.go
```diff
@@ -0,0 +1,499 @@
...
+	for _, source := range component.Sources {
+		if err := source.Validate(); err != nil {
+			errs = append(errs, err)
+		}
+	}
+	return errors.Join(errs...)
+}
+
+// addColocatedResourceLocalBlob adds a local blob to the component version repository and defaults fields relevant
+// to declare the spec.LocalRelation to the component version as well as default the resource version, media type and size.
```
frewilhelm, line 411: I tried to understand this description 3 times but still failed. Can you clarify this what is meant by 

> defaults fields relevant to declare the spec.LocalRelation to the component version as well as default the resource version, media type and size.

 I see a default for the resource.Version but not for the media type and size. The media type is only set if it is mediaTypeAware. What is the default if not?

bindings/go/constructor/construct.go
```diff
@@ -0,0 +1,499 @@
...
+
+	if result.ProcessedBlobData != nil {
+		processedResource, err = addColocatedResourceLocalBlob(ctx, targetRepo, component, version, resource, result.ProcessedBlobData)
+	} else if result.ProcessedResource != nil {
+		processedResource = result.ProcessedResource
+	}
+
+	if err != nil {
+		return nil, fmt.Errorf("error adding resource %q to target repository: %w", resource.ToIdentity(), err)
+	}
```
frewilhelm, line 384: Why is this error checked down here and not where the error is set/created?

bindings/go/constructor/construct_resource_test.go
```diff
@@ -0,0 +1,669 @@
...
+
+	// Create the constructor with our mocks
+	opts := Options{
+		ResourceInputMethodProvider: mockProvider,
+		TargetRepositoryProvider:    &mockTargetRepositoryProvider{repo: mockRepo},
+		ProcessResourceByValue: func(resource *constructorv1.Resource) bool {
+			return true
+		},
+	}
+	ctor := NewDefaultConstructor(opts)
```
frewilhelm, line 210: is `ctor` the short version for `constructor`?

bindings/go/constructor/construct_resource_test.go
```diff
@@ -0,0 +1,669 @@
+package constructor
```
frewilhelm, line 1: I am missing a test that contains several resources of the same or different types to make sure that not only one resource per constructor can be processed

bindings/go/constructor/construct.go
```diff
@@ -0,0 +1,499 @@
...
+		res, err = c.processSourceWithInput(ctx, targetRepo, src, component, version)
+	} else {
+		logger.Debug("processing source with existing access")
+		converted := constructorv1.ConvertToRuntimeSource(*src)
+		res = &converted
+	}
+
+	if err != nil {
+		return nil, fmt.Errorf("error processing source %q: %w", src.ToIdentity(), err)
+	}
```
frewilhelm, line 294:
```suggestion
	if src.HasInput() {
		logger.Debug("processing source with input method")
		res, err = c.processSourceWithInput(ctx, targetRepo, src, component, version)
		if err != nil {
			return nil, fmt.Errorf("error processing source %q: %w", src.ToIdentity(), err)
		}
	} else {
		logger.Debug("processing source with existing access")
		converted := constructorv1.ConvertToRuntimeSource(*src)
		res = &converted
	}


```

bindings/go/constructor/construct_source_test.go
```diff
@@ -0,0 +1,455 @@
+package constructor
```
frewilhelm, line 1: Also missing a test that includes more than one source

bindings/go/constructor/construct_test.go
```diff
@@ -0,0 +1,81 @@
+package constructor
```
frewilhelm, line 1: Do we need any test that contains resources and sources?

bindings/go/constructor/input_method_registry.go
```diff
@@ -0,0 +1,93 @@
...
+	typ, err := r.scheme.TypeForPrototype(typed)
+	if err != nil {
+		return runtime.Type{}, err
+	}
+
+	return typ, nil
+}
+
+// GetResourceInputMethod retrieves the resource input method for a given typed object.
+// Returns the method and a boolean indicating if the method was found.
```
frewilhelm, line 55: The method does not return a boolean indicating if the method was found

bindings/go/constructor/options.go
```diff
@@ -0,0 +1,49 @@
...
+
+	// While constructing a component version, the constructor library will use the
+	// given function to decide whether a resource should be processed by value or not.
+	// The ProcessResourceByValue function is OPTIONAL, if not provided, the constructor library will never process resources by value.
+	ProcessResourceByValue func(*constructor.Resource) bool
+
+	// While constructing a component version, the constructor library will use the
+	// given ConcurrencyLimit to limit the number of concurrent operations on resources and sources.
+	// The ConcurrencyLimit is OPTIONAL, if not provided, the constructor library will limit concurrency to the number of available CPU cores.
+	ConcurrencyLimit int
```
frewilhelm, line 48: This differs from the [ADR](https://github.com/open-component-model/open-component-model/blob/main/docs/adr/0006_component_constructors.md?plain=1#L395). In theory, the ADR is not accepted yet. Should we adjust the ADR, when this PR is merged?

frewilhelm (commented):
Very nice implementation so far!

bindings/go/constructor/default.go
Skarlso: It's fine to leave it in. It is useful guide later. :)

bindings/go/constructor/construct.go
jakobmoellerdev: This is moot imo. I would argue that keeping nesting level low is preferrable always. But I also understand that not having the error present immediately hinders readability. WDYT?

bindings/go/constructor/construct.go
```diff
@@ -0,0 +1,499 @@
...
+
+	result, err := method.ProcessSource(ctx, src, creds)
+	if err != nil {
+		return nil, fmt.Errorf("error getting blob from input method: %w", err)
+	}
+
+	var processedSource *descriptor.Source
+
+	if result.ProcessedBlobData != nil {
+		processedSource, err = addColocatedSourceLocalBlob(ctx, targetRepo, component, version, src, result.ProcessedBlobData)
```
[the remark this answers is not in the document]
jakobmoellerdev: it does the following:
1. Add the processed blob data as local blob (via AddLocalResource)
2. Create an access for that uploaded local blob,
3. Then add that to the source and return it

It is "colocated" because it is located alongside the component version later on, unlike arbitrary source accesses which can lie anywhere based on the access specification. Colocated == "localBlob"

bindings/go/constructor/construct.go
jakobmoellerdev: To avoid nesting. IMO its always preferrable to have codepaths the least nested way possible. In this instance I could avoid having to write an if inside another if, because down the line this is an antipattern for large code flows. (similar as the other code path where you commented). Happy to hear your thoughts on this

bindings/go/constructor/construct.go
jakobmoellerdev: lets break it down

> defaults fields relevant to declare the spec.LocalRelation to the component version

if no relation is set, default to local relation because the component version starts "owning" the resource and its locally included.


>  as well as default the resource version

if no resource version is set, set it to the component version.

> media type

a local blob may contain a media type. However if the blob doesn't declare a media type (because blobs MAY have a media type but it can also be omitted), then this defaulting cannot happen.

> and size.

This is actually a left over from when I was setting the size on resources. I think I should remove this.

bindings/go/constructor/construct_test.go
jakobmoellerdev: sure, although I didnt find them necessary. I can add one for good measure 👍

bindings/go/constructor/options.go
frewilhelm: IMHO I do not really mind. It can just lead to confusion. On the other side, we will not adjust the ADR everytime we alter the contract, i guess.

(It is not only about ConcurrencyLimit, but also ProcessResourceByValue vs. ProcessByValue)

bindings/go/constructor/construct.go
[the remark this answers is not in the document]
Skarlso: Thanks, I see.

bindings/go/constructor/spec/v1/resources/schema-2020-12.json
```diff
@@ -120,44 +120,88 @@
...
-            "$ref": "#/$defs/label"
+      "oneOf": [
+        {
+          "type": "object",
+          "description": "a source that has a predefined access specification",
+          "required": [
+            "name",
+            "version",
+            "type",
+            "access"
```
Skarlso, line 131: Is this true? Isn't it either access or input?

Skarlso: Wait, you can use anyOf in the required field though right? Something like 
```json
  "anyOf": [
    { "required": ["input"] },
    { "required": ["access"] }
  ]
```

No?

Skarlso (commented):
Generally this is super good. I just have two minor questions really and remarks. Once Frederic is okay, it's LGTM.

bindings/go/constructor/construct_test.go
jakobmoellerdev: added

bindings/go/constructor/construct_resource_test.go
jakobmoellerdev: added

bindings/go/constructor/construct.go
jakobmoellerdev: adjusted code comment

bindings/go/constructor/construct.go
```diff
@@ -0,0 +1,504 @@
...
+			errs = append(errs, err)
+		}
+	}
+	return errors.Join(errs...)
+}
+
+// addColocatedResourceLocalBlob adds a local blob to the component version repository and defaults fields relevant
+// to declare the spec.LocalRelation to the component version as well as default the resource version and media type:
+//
+//  1. If not resource relation is set, it defaults to constructorv1.LocalRelation because the resource is then located
```
frewilhelm, line 413:
```suggestion
//  1. If no resource relation is set, it defaults to constructorv1.LocalRelation because the resource is then located
```

PR #220, opened 2025-06-10
feat: allow access to all registered types in the scheme

jakobmoellerdev:

this adds functions for

- copying all types or specific types from one scheme to another
- getting all existing types from the registry

This is useful if one works with multiple schemes and one scheme works as an aggregation of 

This also removes a leftover where there could be versioned and non versioned defaults. since this was removed, we can remove the code path for the preference of the versioned default.

### DIFF
--- a/bindings/go/runtime/registry.go
+++ b/bindings/go/runtime/registry.go
@@ -6,6 +6,8 @@ import (
 	"io"
 	"maps"
 	"reflect"
+	"slices"
+	"strings"
 	"sync"
 
 	"github.com/cyberphone/json-canonicalization/go/src/webpki.org/jsoncanonicalizer"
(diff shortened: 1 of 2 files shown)


bindings/go/runtime/registry.go
```diff
@@ -54,6 +82,68 @@ func (r *Scheme) Clone() *Scheme {
 	return clone
 }
 
+// RegisterSchemes calls RegisterScheme for each scheme in the list, registering all types from each scheme.
+// Conflicts between Scheme's passed will result in an error on the first conflict found.
+// Registration might still have occurred for some types before the error is returned.
```
frewilhelm, line 87: Is this a problem or why are you describing this behaviour?

frewilhelm (approved):

PR #224, opened 2025-06-11
feat: introduce not found error for oci bindings

jakobmoellerdev:

- introduces a binding not found error that can be coded against when using GetComponentVersion. This allows to check for existence of a component version (useful for CLI and constructor)
- enforce provider after changes to runtime (makes sure provider is always present in descriptors)
- separates the interfaces for the repository out to make it easier to read / debug / change independent from the oci implementation

part of https://github.com/open-component-model/ocm-project/issues/460 as we need this to properly code a "not found" case as a check for component construction (check if component version exists before we try to push a new one from the constructor)


bindings/go/oci/interface.go
```diff
@@ -0,0 +1,182 @@
...
+type ComponentVersionRepository interface {
+	// AddComponentVersion adds a new component version to the repository.
+	// If a component version already exists, it will be updated with the new descriptor.
+	// The descriptor internally will be serialized via the runtime package.
+	// The descriptor MUST have its target Name and Version already set as they are used to identify the target
+	// Location in the Store.
+	AddComponentVersion(ctx context.Context, descriptor *descriptor.Descriptor) error
+
+	// GetComponentVersion retrieves a component version from the repository.
+	// Returns the descriptor from the most recent AddComponentVersion call for that component and version.
```
Skarlso, line 37:
```suggestion
	// Returns the descriptor for the given component name and version.
```

bindings/go/oci/interface.go
```diff
@@ -0,0 +1,182 @@
...
+	// If the component version does not exist, it returns ErrNotFound.
+	GetComponentVersion(ctx context.Context, component, version string) (*descriptor.Descriptor, error)
+
+	// ListComponentVersions lists all component versions for a given component.
```
Skarlso, line 41:
```suggestion
	// ListComponentVersions lists all versions for a given component.
```

PR #226, opened 2025-06-11
chore!: refactor memory blob implementation

jakobmoellerdev:

This reimagines the current blob in memory implementation to be easier to understand and debug.

It reduces one abstraction layer from the memory blob implementation so that it is easier to debug and use. The PR also separates the memory blob implementation from the actual blob interface

we deprecate the original method used in the lib to push people to adopt the new constructor method in the inmemory package.

Now the inmemory blob is parallel to the existing filesystem blob.

This is important because right now, we have an implicit behavior in which the Read Call will Seek back to the start of the data implicitly. This is usually not wanted. Instead we should offer a separate bytes Reader that can do independent seeks.


bindings/go/blob/inmemory/blob.go
```diff
@@ -0,0 +1,167 @@
...
+	"bytes"
+	"fmt"
+	"io"
+	"sync"
+
+	"github.com/opencontainers/go-digest"
+)
+
+// New forwards a given [io.Reader] to be able to be used as a ReadOnlyBlob.
+// It does this by wrapping the [io.Reader] in a Blob, to allow for catching information such as
```
fabianburth, line 13:
```suggestion
// It does this by wrapping the [io.Reader] in a Blob, to allow for caching information such as
```

bindings/go/blob/inmemory/blob.go
```diff
@@ -0,0 +1,167 @@
...
+	source io.Reader // source is loaded via Blob.Load on the first ReadCloser call.
+	loaded bool      // indicates if the data has been loaded from source.
+	err    error     // error encountered during Blob.Load, if any
+}
+
+// ReadCloser returns a reader to incrementally access byte stream content
+// It is the caller's responsibility to close the reader.
+//
+// ReadCloser MUST be safe for concurrent use, serializing access as necessary.
+// ReadCloser MUST be able to be called multiple times, which each invocation
```
fabianburth, line 43:
```suggestion
// ReadCloser MUST be able to be called multiple times, with each invocation
```

jakobmoellerdev:
Added some missing legacy deprecated typings that would otherwise break people. Also added concurrent read tests

fabianburth (approved):

PR #228, opened 2025-06-11
feat: component version repository provider

jakobmoellerdev:

In the current OCI lib there is no implementation that can switch based on the Type on which repository to create / resolve efficiently, leaving it up to users / the CLI.

This implementation provides an efficient and easy way to create component version repositories based on repository specifications.

This ultimately solves the problem of allowing an implementation that can be used for the plugin manager through the provider interface.

### DIFF
--- a/bindings/go/oci/repository/provider/provider_test.go
+++ b/bindings/go/oci/repository/provider/provider_test.go
@@ -1,0 +1,89 @@
+package provider_test
+
+import (
+	"os"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"golang.org/x/sync/errgroup"
+
+	"ocm.software/open-component-model/bindings/go/blob/filesystem"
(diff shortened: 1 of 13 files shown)


bindings/go/oci/repository/provider/provider_test.go
```diff
@@ -0,0 +1,89 @@
...
+		// Now verify that the same component version can be retrieved from the second repository
+		retrievedDesc2, err := repo2.GetComponentVersion(ctx, name, version)
+		r.NoError(err)
+		r.Equal(name, retrievedDesc2.Component.Name)
+
+		t.Run("concurrent access to the same repository", func(t *testing.T) {
+			r := require.New(t)
+			// Attempt to add the same component version concurrently
+			eg, ctx := errgroup.WithContext(t.Context())
+			for i := 0; i < 10; i++ {
```
Skarlso, line 74: Fun fact... you can now do `for range 10 {`. :)

PR #242, opened 2025-06-13
feat: introduce file / path flag

jakobmoellerdev:

this refactors the flag implementations and adds a flag type for file paths so that it is easy to pass file / directory references via flag. this flag automatically stats the path and collects underlying filesystem information.

This prepares for various commands such as `add cv` which need to work with files / filesystems on a regular basis and could reuse a flag implementation that does the heavy lifting for them

### DIFF
--- a/cli/internal/flags/file/flag.go
+++ b/cli/internal/flags/file/flag.go
@@ -21,6 +21,9 @@ type Flag struct {
 }
 
 func (f *Flag) String() string {
+	if f.path == nil {
+		return ""
+	}
 	return *f.path
 }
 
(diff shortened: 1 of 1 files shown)


cli/internal/flags/file/flag.go
```diff
@@ -0,0 +1,84 @@
...
+const Type = "path"
+
+// Flag defines a path flag that checks if the value is an existing file.
+type Flag struct {
+	path *string
+	fs.FileInfo
+}
+
+func (f *Flag) String() string {
+	return *f.path
```
Skarlso, line 24: Requires a nil check before dereferencing.

cli/internal/flags/file/flag.go
```diff
@@ -0,0 +1,84 @@
...
+	return nil
+}
+
+func (f *Flag) Type() string {
+	return Type
+}
+
+func Var(f *pflag.FlagSet, name string, value string, usage string) {
+	actual := strings.Clone(value)
+	flag := Flag{path: &actual}
```
Skarlso, line 58: Shouldn't this also call Set like VarP does? 🤔

cli/internal/flags/file/flag.go
```diff
@@ -0,0 +1,84 @@
...
+	f.Var(&flag, name, usage)
+}
+
+func VarP(f *pflag.FlagSet, name, shorthand string, value string, usage string) {
+	actual := strings.Clone(value)
+	flag := &Flag{}
```
Skarlso, line 64: I'm confused here. Sorry. This is calling set while VAr is calling Flag{path: &actual}. Why? :D

Skarlso (commented):
Looks okay, just questions.. :)

cli/internal/flags/file/flag.go
jakobmoellerdev: youre right :)

PR #248, opened 2025-06-16
feat!: make referrer indexes and lookup optional and disable by default

jakobmoellerdev:

this makes the referrer index, the index creation, subject reference on the index, as well as the listing via referrer API optional and hidden behind a configurable policy that is by default disabled. (this is the expected behavior if moving from a legacy OCM repository, as we didnt have referrer support before)

### DIFF
--- a/bindings/go/oci/integration/integration_test.go
+++ b/bindings/go/oci/integration/integration_test.go
@@ -185,7 +185,7 @@ func Test_Integration_OCIRepository(t *testing.T) {
 			uploadDownloadBarebonesComponentVersion(t, repo, "test-component", "v1.0.0")
 		})
 
-		t.Run("basic upload and download of a component version", func(t *testing.T) {
+		t.Run("basic upload and download of a component version (with index based referrer tracking)", func(t *testing.T) {
 			repo, err := oci.NewRepository(oci.WithResolver(resolver), oci.WithReferrerTrackingPolicy(oci.ReferrerTrackingPolicyByIndexAndSubject))
 			r.NoError(err)
 			uploadDownloadBarebonesComponentVersion(t, repo, "test-component", "v1.0.0")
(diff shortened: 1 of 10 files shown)


bindings/go/oci/integration/integration_test.go
```diff
@@ -181,6 +185,12 @@ func Test_Integration_OCIRepository(t *testing.T) {
 			uploadDownloadBarebonesComponentVersion(t, repo, "test-component", "v1.0.0")
 		})
 
+		t.Run("basic upload and download of a component version", func(t *testing.T) {
+			repo, err := oci.NewRepository(oci.WithResolver(resolver), oci.WithReferrerTrackingPolicy(oci.ReferrerTrackingPolicyByIndexAndSubject))
+			r.NoError(err)
+			uploadDownloadBarebonesComponentVersion(t, repo, "test-component", "v1.0.0")
+		})
```
Skarlso, line 192: Basically testing that it's still the same thing?  Should this be asserting something else? The name of it is definitely the same.

bindings/go/oci/repository.go
```diff
@@ -134,7 +137,16 @@ func (repo *Repository) ListComponentVersions(ctx context.Context, component str
...
 		},
-	})
+	}
+
+	switch repo.referrerTrackingPolicy {
+	case ReferrerTrackingPolicyByIndexAndSubject:
+		opts.LookupPolicy = lister.LookupPolicyReferrerWithTagFallback
+	case ReferrerTrackingPolicyNone:
+		opts.LookupPolicy = lister.LookupPolicyTagOnly
+	}
```
Skarlso, line 147: Default case in which the setting is uninitialized?

bindings/go/oci/integration/integration_test.go
jakobmoellerdev: I agree, will add the referrer policy to the test

bindings/go/oci/repository.go
jakobmoellerdev: default case would be zero-value which is ReferrerTrackingPolicyNone

fabianburth (approved):

PR #251, opened 2025-06-16
feat: constructor callbacks

jakobmoellerdev:

This adds optional lifecycle callbacks to the constructor 

```
type ComponentConstructionCallbacks struct {
	// OnStartComponentConstruct is called before the construction of a component version starts.
	OnStartComponentConstruct func(ctx context.Context, component *constructor.Component) error
	// OnEndComponentConstruct is called after the construction of a component version ends.
	// If an error occurs during the construction, the error is passed as a parameter.
	OnEndComponentConstruct func(ctx context.Context, descriptor *descriptor.Descriptor, err error) error

	// OnStartResourceConstruct is called before the construction of a resource starts.
	OnStartResourceConstruct func(ctx context.Context, resource *constructor.Resource) error
	// OnEndResourceConstruct is called after the construction of a resource ends.
	// If an error occurs during the construction, the error is passed as a parameter.
	OnEndResourceConstruct func(ctx context.Context, resource *descriptor.Resource, err error) error

	// OnStartSourceConstruct is called before the construction of a source starts.
	OnStartSourceConstruct func(ctx context.Context, source *constructor.Source) error
	// OnEndSourceConstruct is called after the construction of a source ends.
	// If an error occurs during the construction, the error is passed as a parameter.
	OnEndSourceConstruct func(ctx context.Context, source *descriptor.Source, err error) error
}
```

this allows plugging in progress trackers for tools like the CLI

### DIFF
--- a/bindings/go/constructor/construct.go
+++ b/bindings/go/constructor/construct.go
@@ -56,7 +56,7 @@ func (c *DefaultConstructor) Construct(ctx context.Context, constructor *constru
 
 	for i, component := range constructor.Components {
 		componentLogger := logger.With("component", component.Name, "version", component.Version)
-		componentLogger.Info("constructing component")
+		componentLogger.Debug("constructing component")
 
 		eg.Go(func() error {
 			if c.opts.OnStartComponentConstruct != nil {
@@ -87,7 +87,7 @@ func (c *DefaultConstructor) Construct(ctx context.Context, constructor *constru
 		return nil, fmt.Errorf("error constructing components: %w", err)
 	}
 
-	logger.Info("component construction completed successfully", "num_components", len(descriptors))
+	logger.Debug("component construction completed successfully", "num_components", len(descriptors))
 	return descriptors, nil
 }
 
@@ -187,7 +187,7 @@ func (c *DefaultConstructor) processDescriptor(
 
 	for i, resource := range component.Resources {
 		resourceLogger := logger.With("resource", resource.ToIdentity())
-		resourceLogger.Info("processing resource")
+		resourceLogger.Debug("processing resource")
 
 		eg.Go(func() error {
 			if c.opts.OnStartResourceConstruct != nil {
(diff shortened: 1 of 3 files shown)


bindings/go/constructor/construct.go
```diff
@@ -56,10 +56,20 @@ func (c *DefaultConstructor) Construct(ctx context.Context, constructor *constru
 
 	for i, component := range constructor.Components {
 		componentLogger := logger.With("component", component.Name, "version", component.Version)
-		componentLogger.Info("constructing component")
+		componentLogger.Debug("constructing component")
 
 		eg.Go(func() error {
+			if c.opts.OnStartComponentConstruct != nil {
```
Skarlso, line 62: Do you think we could have multiple of these ala chained observers?

bindings/go/constructor/construct.go
```diff
@@ -56,10 +56,20 @@ func (c *DefaultConstructor) Construct(ctx context.Context, constructor *constru
...
+				if err := c.opts.OnStartComponentConstruct(egctx, &component); err != nil {
+					return fmt.Errorf("error starting component construction for %q: %w", component.ToIdentity(), err)
+				}
+			}
 			desc, err := c.construct(egctx, &component)
+			if c.opts.OnEndComponentConstruct != nil {
+				if err := c.opts.OnEndComponentConstruct(egctx, desc, err); err != nil {
+					return fmt.Errorf("error ending component construction for %q: %w", component.ToIdentity(), err)
+				}
+			}
```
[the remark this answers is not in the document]
jakobmoellerdev: what would be the problem here? if we fail out and should just log than thats the behavior of the callback right?

bindings/go/constructor/construct.go
jakobmoellerdev: then you could have one OnStartComponentConstruct that delegates to multiple others right?

PR #256, opened 2025-06-17
feat: introduce oci backed resource digest processor

jakobmoellerdev:

add a method / interface that allows us to process and append / pin digests to resources backed by OCI Images efficiently via resolve calls.

This allows the component constructor to efficiently construct OCI based blob digests.

### DIFF
--- a/bindings/go/oci/ctf/store.go
+++ b/bindings/go/oci/ctf/store.go
@@ -8,7 +8,6 @@ import (
 	"log/slog"
 	"os"
 	"slices"
-	"strings"
 	"sync"
 
 	"github.com/opencontainers/go-digest"
@@ -157,12 +156,14 @@ func (s *Repository) Resolve(ctx context.Context, reference string) (ociImageSpe
 	// if we do not have a pure digest, we need to parse the reference
 	// loosely because it could be that registry/repository information is prefixed to the actual reference.
 	if _, err := digest.Parse(reference); err != nil {
-		// if we have an encoded digest, prefer to use that one only,
-		// else if we have a tag, use that one.
-		if idx := strings.Index(reference, "@"); idx != -1 {
-			reference = reference[idx+1:]
-		} else if idx := strings.Index(reference, ":"); idx != -1 {
-			reference = reference[idx+1:]
+		ref, err := looseref.ParseReference(reference)
+		if err != nil {
+			return ociImageSpecV1.Descriptor{}, fmt.Errorf("invalid reference %q: %w", reference, err)
+		}
+		if ref.ValidateReferenceAsDigest() == nil {
+			reference = ref.Reference.Reference
+		} else if ref.ValidateReferenceAsTag() == nil {
+			reference = ref.Tag
 		}
 	}
 
--- a/bindings/go/oci/ctf/store.go
+++ b/bindings/go/oci/ctf/store.go
@@ -137,6 +137,8 @@ func (s *Repository) Push(ctx context.Context, expected ociImageSpecV1.Descripto
 
 // Resolve resolves a reference string to its corresponding descriptor in the CTF archive.
 // The reference should be in the format "repository:tag" so it will be resolved against the index.
+// The reference can also be just a tag or a digest, in which case the repository is based on the base repository.
+// Alternatively, it is also possible to provide a digest directly, e.g., "sha256:abc123...".
 // If a full reference is given, it will be resolved against the blob Repository immediately.
 // Returns the descriptor if found, or an error if the reference is invalid or not found.
 func (s *Repository) Resolve(ctx context.Context, reference string) (ociImageSpecV1.Descriptor, error) {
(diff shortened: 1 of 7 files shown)


bindings/go/oci/ctf/store.go
```diff
@@ -152,11 +154,15 @@ func (s *Repository) Resolve(ctx context.Context, reference string) (ociImageSpe
...
-		reference = strings.TrimPrefix(reference, prefix)
+	// if we do not have a pure digest, we need to parse the reference
+	// loosely because it could be that registry/repository information is prefixed to the actual reference.
+	if _, err := digest.Parse(reference); err != nil {
+		if idx := strings.Index(reference, ":"); idx != -1 {
+			reference = reference[idx+1:]
+		}
+		if idx := strings.Index(reference, "@"); idx != -1 {
+			reference = reference[:idx]
+		}
```
Skarlso, line 165: Doesn't this _remove_ the digest part? Is that what you want?

jakobmoellerdev: hmm youre right im currently not discovering a failed resolve because the tag is still resolvable, i want it to be the other way around. good catch!

bindings/go/oci/repository.go
```diff
@@ -197,6 +197,94 @@ func (repo *Repository) AddLocalSource(ctx context.Context, component, version s
...
+		}
+		globalAccess, err := repo.scheme.NewObject(typed.GlobalAccess.GetType())
+		if err != nil {
+			return nil, fmt.Errorf("error creating typed global blob access with help of scheme: %w", err)
+		}
+		if err := repo.scheme.Convert(typed.GlobalAccess, globalAccess); err != nil {
+			return nil, fmt.Errorf("error converting global blob access: %w", err)
+		}
+		res.Access = globalAccess
+		return repo.ProcessResourceDigest(ctx, res)
```
Skarlso, line 232: I'm 70% sure there is an infinite loop here. I'm trying to write a test for it.

Skarlso: Oh okay, so it will keep getting the type, until basically there is this error:

```
local blob access does not have a global access and cann...
```

So shouldn't it just stop by then? I don't know. :D This is the type I used:
```go
				selfRef := &v2.LocalBlob{
					LocalReference: "self-ref",
					MediaType:      "application/octet-stream",
				}
				selfRef.GlobalAccess = &runtime.Raw{
					Type: runtime.Type{
						Name:    "localBlob",
						Version: "v1",
					},
					Data: []byte(`{"type":"localBlob/v1","localReference":"self-ref","mediaType":"application/octet-stream", "globalAccess":{"type":"localBlob/v1","localReference":"self-ref","mediaType":"application/octet-stream"}}`),
				}
```

Result in three time error:

```
2025/06/17 13:42:54 INFO operation starting realm=oci operation="process resource digest" resource.name=recursive-resource resource.version=1.0.0
2025/06/17 13:42:54 INFO operation starting realm=oci operation="process resource digest" resource.name=recursive-resource resource.version=1.0.0
2025/06/17 13:42:54 INFO operation starting realm=oci operation="process resource digest" resource.name=recursive-resource resource.version=1.0.0
2025/06/17 13:42:54 ERROR operation failed realm=oci operation="process resource digest" duration=2.625µs resource.name=recursive-resource resource.version=1.0.0 error="local blob access does not have a global access and cannot be used"
2025/06/17 13:42:54 ERROR operation failed realm=oci operation="process resource digest" duration=11.542µs resource.name=recursive-resource resource.version=1.0.0 error="local blob access does not have a global access and cannot be used"
2025/06/17 13:42:54 ERROR operation failed realm=oci operation="process resource digest" duration=180.042µs resource.name=recursive-resource resource.version=1.0.0 error="local blob access does not have a global access and cannot be used"
```

Is that okay?

bindings/go/oci/ctf/store.go
Skarlso: 👁️ 👄 👁️

Skarlso: This will now fail anything that has a port in. For example `registry:5000/repo:tag`, right?

bindings/go/oci/repository.go
jakobmoellerdev: for me that is acceptable because it is weird as hell

Skarlso: Okay, works for me. :)

bindings/go/oci/ctf/store.go
Skarlso: Not particularly fail I guess, so `registry:5000/repo:tag` will just become `tag`.

Skarlso: No, actually, Index is the first occurrence. So this would become `5000/repo:tag`.

Skarlso: Nice, now it looks good, I think.

PR #258, opened 2025-06-17
fix: allow arbitrary raw json/yaml structures in label values and accept a version field

jakobmoellerdev:

as per OCM specification it is possible to define arbitrary label values as JSON / YAML structures. see https://github.com/open-component-model/ocm-spec/blob/main/doc/01-model/03-elements-sub.md#labels


bindings/go/descriptor/v2/descriptor.go
```diff
@@ -258,10 +262,28 @@ type SignatureInfo struct {
...
-	Name  string `json:"name"`
-	Value string `json:"value"`
+	// Name is the unique name of the label.
+	Name string `json:"name"`
+	// Value is the json/yaml data of the label
+	Value json.RawMessage `json:"value"`
 	// Signing describes whether the label should be included into the signature
 	Signing bool `json:"signing,omitempty"`
+	// Version is the optional specification version of the attribute value
+	Version string `json:"version,omitempty"`
```
frewilhelm, line 272: what is the `Version` for? I don't see any usage or tests for it.

frewilhelm: The spec even says

> The specification version for the label content. If no version is specified, implicitly the version v1 is assumed.

So, we must set the version to `v1` if empty.

jakobmoellerdev: no, we must not. we must assume v1, but it must not be serialized

PR #271, opened 2025-06-24
chore: apply latest versions to the CLI module

Skarlso:


cli/internal/plugin/builtin/oci/plugin.go
```diff
@@ -151,7 +155,7 @@ func (p *Plugin) createRepository(spec *ociv1.Repository, credentials map[string
 		oci.WithResolver(urlResolver),
 		oci.WithScheme(p.scheme),
 		oci.WithCreator(Creator),
-		oci.WithOCIDescriptorCache(p.memory),
+		oci.WithManifestCache(p.memory),
```
jakobmoellerdev, line 158: I think we need a layer cache, I probably missed this the first time (and 2 memory structs)

Skarlso:
>https://github.com/open-component-model/open-component-model/actions/runs/15845688199/job/44667219596?pr=271#step:5:49 the test failures need a correction in the tests by setting a mocked provider name, then were good to go

Yah, I know. I'm fixing it.

Skarlso:
Dang, there are some more problems here. I also need to update constructor because I can't run the tests.

```
GOROOT=/usr/local/go #gosetup
GOPATH=/Users/skarlso/go #gosetup
/usr/local/go/bin/go test -c -o /Users/skarlso/Library/Caches/JetBrains/GoLand2025.1/tmp/GoLand/___go_test_ocm_software_open_component_model_cli_cmd.test ocm.software/open-component-model/cli/cmd #gosetup
# ocm.software/open-component-model/bindings/go/constructor/spec/v1
/Users/skarlso/go/pkg/mod/ocm.software/open-component-model/bindings/go/constructor@v0.0.0-20250611123713-bf94006044ed/spec/v1/convert.go:126:16: cannot use labels[i].Value (variable of type string) as json.RawMessage value in assignment

```

Skarlso:
For reference, constructor updated because I updated the plugin manager because it was failing because it didn't have Ilya's fix for the missing return of the specific error. :D So now I have to deal with that.

ikhandamirov (approved):

jakobmoellerdev (changes requested):
we need the layer cache in https://github.com/open-component-model/open-component-model/pull/271/files#diff-0113166a2da6a768a51b7d68a2a524b44cd35ced37398c08d6c83f8a2e74afc0R158

PR #288, opened 2025-06-26
feat: add handlers and implementation for outside credentials

Skarlso:


jakobmoellerdev (commented):
shouldnt this be switched to use the real repository contracts as well? were still on v1 after all?

Skarlso:
@jakobmoellerdev Ah, you're right. I should add the wrapper here as well. I thought it's enough because it's using Resolve or something like that. I'll do the update.

Skarlso:
@jakobmoellerdev I _think_ that's pretty much it, I'm trying to integrate it into the CLI and see how it works.

Skarlso:
I just spotted I missed RegisterInternalCredentialRepositoryPlugin register type.

PR #289, opened 2025-06-26
feat: introduce file input method

jakobmoellerdev:

see `bindings/go/constructor/input/file/doc.go`

fix https://github.com/open-component-model/ocm-project/issues/503


bindings/go/constructor/input/file/blob.go
```diff
@@ -0,0 +1,71 @@
...
+
+	b, err := filesystem.GetBlobFromOSPath(file.Path)
+	if err != nil {
+		return nil, err
+	}
+
+	mediaType := file.MediaType
+	if mediaType == "" {
+		// see https://github.com/gabriel-vasile/mimetype/blob/master/supported_mimes.md for supported types
+		mime, _ := mimetype.DetectFile(file.Path)
```
Skarlso, line 60: Don't we care about this error? 🤔

bindings/go/constructor/input/file/method.go
```diff
@@ -0,0 +1,108 @@
...
+	return nil, ErrFilesDoNotRequireCredentials
+}
+
+// ProcessResource processes a file-based resource input by converting the input specification
+// to a v1.File format, reading the file from the filesystem, and returning the processed
+// blob data. This method handles automatic media type detection and optional compression
+// as specified in the file configuration.
+//
+// The method performs the following steps:
+//  1. Converts the resource input to v1.File specification
```
Skarlso, line 59: Here too. I'm not sure these are valuable. And the other one. Maybe make it broader, more conceptual. Then it makes sense.

bindings/go/constructor/input/file/blob.go
jakobmoellerdev: we dont, from the go doc you can see that if we fail the detection, we default to application/octet-stream

Skarlso: True. 👍

fabianburth (approved):
LGTM

bindings/go/constructor/input/file/method.go
jakobmoellerdev: I dont really have a better way to make it conceptual but theres always room for improvement!

PR #291, opened 2025-06-27
fix: resolve to correct value in case of empty mediatype

Skarlso:

Closes https://github.com/open-component-model/ocm-project/issues/528


bindings/go/oci/ctf/store_test.go
```diff
@@ -378,3 +378,34 @@ func TestResolveWithRegistry(t *testing.T) {
 		assert.Equal(t, digest.Digest(digestStr), desc.Digest)
 	})
 }
+
+func TestResolveWithEmptyMediaType(t *testing.T) {
```
jakobmoellerdev, line 382: You should also test integration with an old CTF that you open

Skarlso: What would that change? Wouldn't it be the same in regards to the mediatype just being empty? 🤔

bindings/go/oci/ctf/store_test.go
```diff
@@ -378,3 +378,60 @@ func TestResolveWithRegistry(t *testing.T) {
...
+
+	t.Run("resolve with empty media type defaults to image manifest", func(t *testing.T) {
+		desc, err := store.Resolve(ctx, "test-tag")
+		assert.NoError(t, err)
+		assert.Equal(t, ociImageSpecV1.MediaTypeImageManifest, desc.MediaType)
+		assert.Equal(t, digest.Digest(digestStr), desc.Digest)
+	})
+}
+
+func TestResolveWithCompatibilityCTFTestdata(t *testing.T) {
```
Skarlso, line 413: @jakobmoellerdev Is this.. right? 🤔 I mean I tested the artifact that it's empty. That's correct, right?

jakobmoellerdev: I think its important to check if we can actually access the CTF by reading a component version

jakobmoellerdev (commented):
This is actually mostly LGTM. Only thing im thinking now is that we should have an already pre existing transport archive sitting somewhere in the actual CTF binding lib. Not a huge deal, but I was wondering if thats reusable? (I just want to avoid too many archives / files in the repo) WDYT?

Skarlso:
Are you saying that instead of copying over, I should open it from under `bindings/go/ctf/testdata`? I would strongly advice against that. Especially because it's a separate go module. I know you could technically load it but that's a boundary violation.

We could _maybe_ move it top level and both of those modules could use it, but I don't see a huge benefit in that honestly.

fabianburth (approved):

jakobmoellerdev (approved):

PR #292, opened 2025-06-27
feat: support for `+` in cv versions

ikhandamirov:

If a component version contains a "`+`", in the derived OCI tag it will be replaced by "`.build-`".

Fixes open-component-model/ocm-project#530


bindings/go/oci/ctf/store.go
```diff
@@ -74,7 +74,8 @@ func (s *Store) Reference(reference string) (fmt.Stringer, error) {
 
 // ComponentVersionReference creates a reference string for a component version in the format "component-descriptors/component:version".
 func (s *Store) ComponentVersionReference(component, version string) string {
-	return fmt.Sprintf("%s/component-descriptors/%s:%s", wellKnownRegistryCTF, component, version)
+	tag := oci.ToTag(version)
```
jakobmoellerdev, line 77: This is missing a comment on the adjustment

bindings/go/oci/repository.go
```diff
@@ -38,6 +38,15 @@ import (
 	"ocm.software/open-component-model/bindings/go/runtime"
 )
 
+// plusSubstitute is used to substitute the plus character ('+') in OCI tags.
+// An OCM version is allowed to contain the plus character, but OCI tags do not allow it.
+// Because the OCI tag of an artifact representing an OCM component is derived from the respective component
+// version, this replacement is required.
```
jakobmoellerdev, line 44: I think it would be good if you could reference the relevant part of the OCM spec here, but otherwise this is great 💯

bindings/go/oci/repository.go
```diff
@@ -162,6 +175,9 @@ func (repo *Repository) GetComponentVersion(ctx context.Context, component, vers
 	if err != nil {
 		return nil, err
 	}
+	if !strings.HasSuffix(reference, version) {
```
jakobmoellerdev, line 178: I think it makes sense to log these warnings directly in the call within getStore / ComponentVersionReference and not the repository

bindings/go/oci/repository.go
```diff
@@ -712,3 +728,8 @@ func getDescriptorOCIImageManifest(ctx context.Context, store spec.Store, refere
 	}
 	return manifest, index, nil
 }
+
+// ToTag converts an OCM version to a valid OCI tag by replacing possible '+' characters.
```
jakobmoellerdev, line 732: missing a reference to the OCM spec

jakobmoellerdev:
```suggestion
// LooseSemverToOCITag converts an OCM version to a valid OCI tag by replacing possible '+' characters.
```

bindings/go/oci/repository.go
ikhandamirov: I suggest logging directly in the `LooseSemverToOCITag`. There is also a map caching problematic versions, to ensure that there is only one log entry per "problematic" version. PTAL at the new `semver_tag.go`

bindings/go/oci/semver_tag.go
```diff
@@ -0,0 +1,40 @@
...
+// version, this replacement is required. See also:
+// - https://github.com/open-component-model/ocm-spec/blob/main/doc/04-extensions/03-storage-backends/oci.md#version-mapping
+const (
+	plusSubstitute = ".build-"
+	plus           = "+"
+)
+
+// versionsWithPlus cache is used to keep track of version numbers, a warning has already been logged for.
+// The intention is to avoid logging the same warning multiple times for the same version number.
+var versionsWithPlus = make(map[string]string)
```
jakobmoellerdev, line 21: this is not concurrency safe. if you really want to do this you need to make this safe for loading.

jakobmoellerdev: I think when we talked about it last time I mentioned this is overkill for now, so id prefer to remove the caching again.

bindings/go/oci/semver_tag.go
```diff
@@ -0,0 +1,40 @@
...
+
+// LooseSemverToOCITag converts an OCM version to a valid OCI tag by replacing 1 possible occurrence of the '+' character.
+// If there is more than one occurrence of the '+' character, the expectation is that this is caught later by the
+// OCI tag validation.
+// See also:
+// - https://github.com/open-component-model/ocm-spec/blob/main/doc/04-extensions/03-storage-backends/oci.md#version-mapping
+// - https://semver.org/#spec-item-10
+func LooseSemverToOCITag(version string) string {
+	idx := strings.LastIndex(version, plus)
```
jakobmoellerdev, line 30: I dont really understand why you dont use `strings.Replace(version, plus, plusSubstitute, 1)` here, is there any behavior that you wanted to avoid?

bindings/go/oci/repository.go
```diff
@@ -99,8 +99,9 @@ func (repo *Repository) AddComponentVersion(ctx context.Context, descriptor *des
 		return fmt.Errorf("failed to add descriptor to store: %w", err)
 	}
 
-	// Tag the manifest with the reference
-	if err := store.Tag(ctx, *manifest, version); err != nil {
+	// Tag the manifest with the version, but first remove prohibited characters.
+	tag := LooseSemverToOCITag(version)
+	if err := store.Tag(ctx, *manifest, tag); err != nil {
```
jakobmoellerdev, line 104: why dont you just use the reference here?
```suggestion
	if err := store.Tag(ctx, *manifest, reference); err != nil {
```

jakobmoellerdev (changes requested):

bindings/go/oci/semver_tag.go
ikhandamirov: Initially I wanted to replace the last occurrence, similar to https://github.com/open-component-model/ocm/blob/main/api/ocm/extensions/repositories/genericocireg/component.go#L80. Never mind. You are right.

bindings/go/oci/repository.go
ikhandamirov: I don't use `reference` here, because you were not using it in the first place. And it look like you did it for a reason, because if replaced with `reference`, the CTF tests start to fail.

```
        	Error:      	Received unexpected error:
        	            	failed to get manifest: failed to resolve reference "ctf.ocm.software/component-descriptors/test-component:v1.0.0": not found
        	Test:       	Test_Integration_CTF/direct/basic_upload_and_download_of_a_component_version
--- FAIL: Test_Integration_CTF/direct/basic_upload_and_download_of_a_component_version (0.01s)
```

So, do we want to keep the tag/version here?

bindings/go/oci/repository.go
ikhandamirov: As discussed, we don't do caching for now.

bindings/go/oci/repository.go
jakobmoellerdev: This seems to be buggy behavior. you should probably check out why this is failing because the reference should work.

ikhandamirov: Seems to be fixed. Ptal.

PR #295, opened 2025-06-27
feat: add healthcheck to the component version repository interface

Skarlso:

### DIFF
--- a/bindings/go/oci/ctf/store.go
+++ b/bindings/go/oci/ctf/store.go
@@ -54,6 +54,8 @@ type Store struct {
 	archive ctf.CTF
 }
 
+// Ping for CTF return always true. This is because if it doesn't exist it will be created. If it does exist
+// it's all good. Which means it doesn't make any sense to check it.
 func (s *Store) Ping(ctx context.Context) error {
 	return nil
 }
--- a/bindings/go/oci/ctf/store.go
+++ b/bindings/go/oci/ctf/store.go
@@ -54,6 +54,10 @@ type Store struct {
 	archive ctf.CTF
 }
 
+func (s *Store) Ping(ctx context.Context) error {
+	return nil
+}
+
 // StoreForReference returns a new Store instance for a specific repository within the CTF archive.
 func (s *Store) StoreForReference(_ context.Context, reference string) (spec.Store, error) {
 	rawRef, err := s.Reference(reference)
--- a/bindings/go/oci/integration/integration_test.go
+++ b/bindings/go/oci/integration/integration_test.go
@@ -181,7 +181,8 @@ func Test_Integration_HealthCheck_Authentication(t *testing.T) {
 
 		resolver, err := urlresolver.New(urlresolver.WithBaseURL(reg))
 		r.NoError(err)
-		// No authentication client set
+		// explicitly set default client to avoid token fetch round
+		resolver.SetClient(http.DefaultClient)
 
 		repo, err := oci.NewRepository(oci.WithResolver(resolver))
 		r.NoError(err)
@@ -189,9 +190,7 @@ func Test_Integration_HealthCheck_Authentication(t *testing.T) {
 		// Health check should fail without credentials for ghcr.io
 		// GHCR returns 403 during token exchange when no credentials provided
 		err = repo.CheckHealth(t.Context())
-		r.Error(err)
-		r.True(strings.Contains(err.Error(), "403") || strings.Contains(err.Error(), "denied"),
-			"Expected 403 or denied error, got: %v", err)
+		r.ErrorContains(err, "401")
 	})
 
 	t.Run("resolver ping with valid authentication succeeds", func(t *testing.T) {
@@ -216,9 +215,7 @@ func Test_Integration_HealthCheck_Authentication(t *testing.T) {
 		// Direct resolver ping should fail for ghcr.io with invalid credentials
 		// GHCR returns 403 during token exchange when credentials are invalid
 		err = resolver.Ping(t.Context())
-		r.Error(err)
-		r.True(strings.Contains(err.Error(), "403") || strings.Contains(err.Error(), "denied"),
-			"Expected 403 or denied error, got: %v", err)
+		r.ErrorContains(err, "403")
 	})
 
 	t.Run("resolver ping without authentication fails", func(t *testing.T) {
@@ -226,13 +223,12 @@ func Test_Integration_HealthCheck_Authentication(t *testing.T) {
 
 		resolver, err := urlresolver.New(urlresolver.WithBaseURL(reg))
 		r.NoError(err)
+		resolver.SetClient(http.DefaultClient)
 
 		// Direct resolver ping should fail for ghcr.io without credentials
 		// GHCR returns 403 during token exchange when no credentials provided
 		err = resolver.Ping(t.Context())
-		r.Error(err)
-		r.True(strings.Contains(err.Error(), "403") || strings.Contains(err.Error(), "denied"),
-			"Expected 403 or denied error, got: %v", err)
+		r.ErrorContains(err, "401")
 	})
 }
 
--- a/bindings/go/oci/integration/integration_test.go
+++ b/bindings/go/oci/integration/integration_test.go
@@ -129,6 +129,113 @@ func Test_Integration_OCIRepository_BackwardsCompatibility(t *testing.T) {
 	})
 }
 
+func Test_Integration_HealthCheck_Authentication(t *testing.T) {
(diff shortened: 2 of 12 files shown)


Skarlso:
Humpf, I realise that this is gonna be problematic. I will have to do some separate releases for the interfaces and their implementations.

bindings/go/oci/repository.go
```diff
@@ -149,6 +149,33 @@ func (repo *Repository) ListComponentVersions(ctx context.Context, component str
...
+// Validate checks if the OCI repository is accessible and properly configured.
+func (repo *Repository) Validate(ctx context.Context) (err error) {
+	done := log.Operation(ctx, "validate repository")
+	defer func() {
+		done(err)
+	}()
+
+	// Use a test component reference to validate repository access
+	// Use a valid component name format for the test
+	testReference := repo.resolver.ComponentVersionReference("example.com/validation-test", "1.0.0")
```
Skarlso, line 161: ah actually we might need a component reference with the validation.... ? I'm wondering how that will work with CTF.

bindings/go/oci/repository.go
```diff
@@ -148,6 +148,23 @@ func (repo *Repository) ListComponentVersions(ctx context.Context, component str
...
+	store, err := repo.resolver.StoreForReference(ctx, testReference)
+	if err != nil {
+		return fmt.Errorf("failed to resolve test reference for validation: %w", err)
+	}
+
+	// Perform a lightweight operation to test connectivity and authentication
+	// We use Resolve() with a non-existent reference to test basic connectivity
+	// A successful connection will return ErrNotFound, while connection failures
+	// will return network/authentication errors
+	_, err = store.Resolve(ctx, "validation-test")
```
jakobmoellerdev, line 163: I just checked the specification again and there is in fact something that we could use for remote registries:

https://github.com/opencontainers/distribution-spec/blob/main/spec.md#determining-support as per this endpoint we can use the root URL with a get request to determine availability of the OCI specification with inbuilt authentication support. WDYT?

bindings/go/oci/resolver/url/resolver.go
```diff
@@ -57,6 +57,14 @@ func (resolver *CachingResolver) Reference(reference string) (fmt.Stringer, erro
 	return registry.ParseReference(reference)
 }
 
+func (resolver *CachingResolver) Ping(ctx context.Context) error {
```
Skarlso, line 60: TODO: Add tests

bindings/go/oci/ctf/store.go
```diff
@@ -54,6 +54,10 @@ type Store struct {
 	archive ctf.CTF
 }
 
+func (s *Store) Ping(ctx context.Context) error {
+	return nil
```
jakobmoellerdev, line 58: needs a quick explanation

bindings/go/oci/interface.go
```diff
@@ -155,4 +160,9 @@ type Resolver interface {
 	// format represents a valid reference that can be used for a given store returned
 	// by StoreForReference.
 	Reference(reference string) (fmt.Stringer, error)
+
+	// Ping does a healthcheck for the underlying registry. The implementation varies based on the implementing
```
jakobmoellerdev, line 164:
```suggestion
	// Ping does a healthcheck for the underlying Store. The implementation varies based on the implementing
```

bindings/go/oci/interface.go
```diff
@@ -155,4 +160,9 @@ type Resolver interface {
...
+	// technology. For OCI this means a spec check on the registry, which also does an authenticated call if credentials
+	// are configured. For CTF we return true.
```
jakobmoellerdev, line 166: I think we should explain why we do what we do in the implementations and not here and reference that

bindings/go/oci/interface.go
```diff
@@ -47,6 +47,11 @@ type ComponentVersionRepository interface {
 	// https://github.com/opencontainers/distribution-spec/blob/v1.1.0/spec.md#listing-referrers
 	ListComponentVersions(ctx context.Context, component string) ([]string, error)
 
+	// HealthCheck checks if the repository is accessible and properly configured.
+	// This method verifies that the underlying OCI registry is reachable and that authentication
+	// is properly configured. It performs a lightweight check without modifying the repository.
+	HealthCheck(ctx context.Context) error
```
jakobmoellerdev, line 53:
```suggestion
	CheckHealth(ctx context.Context) error
```

to stay consistent with our current VerbSubject

jakobmoellerdev: We should probably add that this is repository specific right?

bindings/go/oci/resolver/url/resolver.go
```diff
@@ -57,6 +57,14 @@ func (resolver *CachingResolver) Reference(reference string) (fmt.Stringer, erro
...
+	r, err := remote.NewRegistry(resolver.baseURL)
```
jakobmoellerdev, line 61: isnt this registry something we can create when we create the resolver from the base url? Also im wondering if you tested this against a real registry, because from what I can see you never pass the credentials / the http Client anywhere.

I was expecting an integration test to verify this against ghcr and something like this:

```
remote.Registry{
		RepositoryOptions: remote.RepositoryOptions{
			PlainHTTP: resolver.plainHTTP,
			Client:    repo.Client,
		},
	}
```

instead of default construction

jakobmoellerdev (changes requested):
I have one concern regarding auth, but pattern wise this looks great!

bindings/go/oci/resolver/url/resolver.go
Skarlso: I think it's actually just this

```go
	
	r.PlainHTTP = resolver.plainHTTP
	if resolver.baseClient != nil {
		r.Client = resolver.baseClient
	}
```

The client is there. This might be nice. :)

bindings/go/oci/interface.go
jakobmoellerdev: best speak to @fabianburth because he needs this as well for extraction

jakobmoellerdev (changes requested):
Missing an Integration Test case for OCI working against ghcr. Other than that this looks goof for me!

Skarlso:
>Missing an Integration Test case for OCI working against ghcr. Other than that this looks good for me!

I know I know. :D I'm writing it. :DDDD

bindings/go/oci/integration/integration_test.go
```diff
@@ -129,6 +129,113 @@ func Test_Integration_OCIRepository_BackwardsCompatibility(t *testing.T) {
...
+		resolver.SetClient(createAuthClient(reg, "invalid-user", "invalid-password"))
+
+		repo, err := oci.NewRepository(oci.WithResolver(resolver))
+		r.NoError(err)
+
+		// Health check should fail for ghcr.io with invalid credentials
+		// GHCR returns 403 during token exchange when credentials are invalid
+		err = repo.CheckHealth(t.Context())
+		r.Error(err)
+		r.True(strings.Contains(err.Error(), "403") || strings.Contains(err.Error(), "denied"),
```
Skarlso, line 175: So this is a bit shit. Apparently, I can't just ping ghcr... It immediately redirects to a token exchange that results in a a 403.

Skarlso: Technically, it can return a 401 too.
```
➜ curl -I -X GET https://ghcr.io/v2/
HTTP/2 401
content-type: application/json
docker-distribution-api-version: registry/2.0
strict-transport-security: max-age=63072000; includeSubDomains; preload
www-authenticate: Bearer realm="https://ghcr.io/token",service="ghcr.io",scope="repository:user/image:pull"
date: Wed, 02 Jul 2025 14:17:29 GMT
content-length: 73
x-github-request-id: E086:25DF22:CB3F9:CD484:68653F79

```

But apparently, oras isn't getting that.

jakobmoellerdev: the get request you are receiving is correct because your curl doesnt have a token.
the ping needs to have the token set

jakobmoellerdev: Basically what im saying is that the check health failure you receive here is completely fine because the invalid credentials make it unhealthy. for me this is okay. WDYTß

Skarlso: Meh, maybe it's fine. I can't do anything about it. It's inside ORAS.

jakobmoellerdev: hmm thats interesting. you might want to check if the authentication client is setup correctly.

PR #351, opened 2025-07-03
feat: add configurable temp folder location

Skarlso:


Skarlso:
The problem is passing it to the CTF (WorkWithinCTF). For the plugin manager this would actually be totally fine. Since, I can just pass in whatever is from the CLI in `setupPluginManager`. And NewManager could take that as a parameter and then pass it down the chain to the relevant location. 🤔 

The CTF is a completely different beast.

Skarlso:
Hm, the context thing is internal. I can't use that to pass this around unfortunately.

PR #352, opened 2025-07-03
feat: `dir` input method

ikhandamirov:

Introduces an input method of type `dir`

Fixes open-component-model/ocm-project#494


ikhandamirov:
Status: the "meat" is still to be implemented - `GetV1DirBlob()`

bindings/go/input/dir/blob.go
```diff
@@ -0,0 +1,210 @@
...
+// contents from the filesystem.
+//
+// The function performs the following steps:
+//  1. Validates that the directory path is not empty
+//  2. Reads the directory contents using an instance of the virtual FileSystem
+//  3. Packs the directory contents into a tar archive
+//  4. Applies different configuration options of the v1.Dir specification
+func GetV1DirBlob(dir v1.Dir) (blob.ReadOnlyBlob, error) {
+	if dir.Path == "" {
+		return nil, fmt.Errorf("dir path must not be empty")
```
jakobmoellerdev, line 30: Nit: shouldnt the packDir fail here?

bindings/go/input/dir/blob.go
```diff
@@ -0,0 +1,210 @@
...
+	fs, err := filesystem.NewFS(baseDir, os.O_RDONLY)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create filesystem while trying to access %v: %w", path, err)
+	}
+
+	// Create a new tar writer.
+	var buf bytes.Buffer
+	tw := tar.NewWriter(&buf)
+	defer func() {
+		closeErr := tw.Close()
```
jakobmoellerdev, line 74: you can use errors.Join here because the function already excludes nil errors passed.
```suggestion
		err = errors.Join(err, tw.Close())
```

bindings/go/input/dir/blob.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+
+	// Create a new virtual FileSystem instance based on the provided directory path.
+	fs, err := filesystem.NewFS(baseDir, os.O_RDONLY)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create filesystem while trying to access %v: %w", path, err)
+	}
+
+	// Create a new tar writer.
+	var buf bytes.Buffer
```
jakobmoellerdev, line 71: this will buffer the entire length of the data in memory. is this what you want here? if not you maybe want to open a pipe.

bindings/go/input/dir/blob.go
```diff
@@ -0,0 +1,210 @@
...
+		if !include {
+			continue
+		}
+
+		// Create tar header.
+		header, err := tar.FileInfoHeader(info, "")
+		if err != nil {
+			return fmt.Errorf("failed to create tar header: %w", err)
+		}
+		header.Name = relPath
```
jakobmoellerdev, line 141: this needs a ocmment on why youre doing that I believe

bindings/go/input/dir/blob.go
```diff
@@ -0,0 +1,210 @@
...
+
+		// Write the header to the tar archive.
+		if err := tw.WriteHeader(header); err != nil {
+			return fmt.Errorf("failed to write tar header to tar archive: %w", err)
+		}
+
+		// If the entry is a file, copy its content to the tar archive.
+		if entry.Type().IsRegular() {
```
jakobmoellerdev, line 149: what if the entrz is neither regular nor a directory? maybe we should error or at least warn the user?

bindings/go/input/dir/blob.go
```diff
@@ -0,0 +1,210 @@
...
+			file, err := fs.OpenFile(relPath, os.O_RDONLY, 0o644)
+			if err != nil {
+				return fmt.Errorf("failed to open file: %w", err)
+			}
+			defer func() {
+				closeErr := file.Close()
```
jakobmoellerdev, line 155: this has the same errors.Join option as above

bindings/go/input/dir/blob.go
```diff
@@ -0,0 +1,210 @@
...
+				if closeErr != nil && err == nil {
+					err = closeErr
+				}
+			}()
+
+			if _, err := io.Copy(tw, file); err != nil {
```
jakobmoellerdev, line 161: you should use the native Write functionality or CopyN and make use of the file size to limit the reader

bindings/go/input/dir/blob.go
```diff
@@ -0,0 +1,210 @@
...
+				return fmt.Errorf("failed to write file contents to tar archive: %w", err)
+			}
+		}
+
+		// If the entry is a directory, recursively process its contents.
+		if entry.IsDir() {
```
jakobmoellerdev, line 167: nit: we might need to protect from infinite recursion here (think of a symlink of a directory causing a cycle)

bindings/go/input/dir/blob.go
ikhandamirov: Moved.

bindings/go/input/dir/blob.go
```diff
@@ -0,0 +1,210 @@
...
+		// Get FileInfo for the entry.
+		info, err := entry.Info()
+		if err != nil {
+			return fmt.Errorf("failed to get file information: %w", err)
+		}
+
+		// Construct the full path of the entry.
+		entryPath := filepath.Join(currentDir, entry.Name())
+		// Get the relative path of the entry with respect to the base directory.
+		relPath, err := filepath.Rel(baseDir, entryPath)
```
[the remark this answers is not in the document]
ikhandamirov: entry.Name() is just the last element of path. But this code could indeed be simplified, which I already did.

bindings/go/input/dir/blob.go
jakobmoellerdev: I meant an explanation on why you even need to rewrite the header path in the first place instead of using the one that is already in the header

ikhandamirov: Because I want to preserve the subfolder structure. Otherwise, the header name only contains the last element on the path. I.e. the tar would contain a flat collection of files. Which is not what we want, do we?

bindings/go/input/dir/blob.go
ikhandamirov: Yes, I was kind of expecting this one. I'll take a look. In case you have links to documentation or examples at hand, please let me know. Would be greatly appreciated.

bindings/go/input/dir/blob.go
ikhandamirov: Maybe I finally know, where you come from. If not, let's discuss, so I don't have to guess.
In OCM V1 there are two methods: `tarutils.PackFsIntoTar()`and `tarutils.TarFlatFs()`. I oriented myself on the former, because the task says that the new implementation should be "based on the dir input in old OCM".

bindings/go/input/dir/blob.go
```diff
@@ -0,0 +1,210 @@
...
+		include, err := isPathIncluded(relPath, excludePatterns, includePatterns)
+		if err != nil {
+			return fmt.Errorf("failed to check if an entry should be included in the tar archive: %w", err)
+		}
+		if !include {
+			continue
+		}
+
+		// Create tar header.
+		header, err := tar.FileInfoHeader(info, "")
```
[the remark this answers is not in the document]
ikhandamirov: Does OCM V1 `tarutils.RegularFileInfoHeader()` work as you expect it? I'll take a closer look at it.

bindings/go/input/dir/blob.go
ikhandamirov: OK, sorry for the missing explanation. I've updated two comments, please see below. I wouldn't refer to OCMv1 behavior though.

```go
// walkDirContents does recursive packaging of the directory contents, while keeping the subfolder structure.
```

```go
		// Set header name to the relative path of the entry with respect to the base directory,
		// to preserve the subfolder structure in the tar archive.
		header.Name = entryPath
```

bindings/go/input/dir/blob.go
```diff
@@ -0,0 +1,262 @@
...
+	// - Handle FollowSymlinks option.
+
+	// Pack directory contents as a tar archive.
+	reader, err := packDirToTar(dir.Path, &dir)
+	if err != nil {
+		return nil, fmt.Errorf("error producing blob for a dir input: %w", err)
+	}
+
+	// Wrap the tar archive in a ReadOnlyBlob.
+	var dirBlob blob.ReadOnlyBlob = inmemory.New(reader, inmemory.WithMediaType(dir.MediaType))
```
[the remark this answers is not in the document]
ikhandamirov: Or it is defined by the user in the input spec. Could be "application/vnd.gardener.landscaper.blueprint.v1+tar" or whatever specified, from my understanding.

bindings/go/input/dir/blob.go
ikhandamirov: As discussed in the call, a TODO is added for the future. Not a prerequisite for the merge of this PR.

bindings/go/input/dir/blob.go
```diff
@@ -0,0 +1,291 @@
...
+			if err := tw.WriteHeader(header); err != nil {
+				return fmt.Errorf("failed to write tar header to tar archive: %w", err)
+			}
+
+			// Copy file content to the tar archive.
+			file, err := fileSystem.OpenFile(entryPath, os.O_RDONLY, 0o644)
+			if err != nil {
+				return fmt.Errorf("failed to open file %q: %w", entryPath, err)
+			}
+			if size := fi.Size(); size > blob.SizeUnknown {
```
jakobmoellerdev, line 180: imo you don need to check for fi, because the header should have the size available. then you don't need to run against SizeUnknown

bindings/go/input/dir/blob.go
```diff
@@ -0,0 +1,291 @@
...
+
+	// Create a pipe for streaming the tar data
+	pr, pw := io.Pipe()
+
+	// Start a goroutine to create the tar and write the data to the pipe.
+	go func() {
+		defer func() {
+			err = pw.Close()
+			if err != nil {
+				log.Base().Log(ctx, slog.LevelError, fmt.Sprintf("failed to close PipeWriter: %s", err.Error()))
```
jakobmoellerdev, line 92: why do you log here if you already return the close error?

ikhandamirov: This is how we discussed it last time. It is a `defere` in a `goroutine`. The error is not returned. I now removed the named return value to avoid confusion.

bindings/go/input/dir/blob.go
jakobmoellerdev: this is defensive programming because the value you modify it to is always >= 0. imo the check should be removed for simplicity sake

ikhandamirov: Done: https://github.com/open-component-model/open-component-model/pull/352/commits/e51deb27dbb47be5a1127b1422d7880dc14dcb1d

jakobmoellerdev:
I will move this issue into ready for review so that others are signalled to take a look here. This looks really good already

bindings/go/input/dir/blob.go
```diff
@@ -0,0 +1,279 @@
...
+			}
+
+			// Process subdirectory contents.
+			if err := walkDirContents(ctx, entryPath, baseDir, opt, fileSystem, tw); err != nil {
+				return err
+			}
+
+		case header.Typeflag == tar.TypeSymlink:
+			/*// The entry is a symlink.
+			if !opt.FollowSymlinks {
```
jakobmoellerdev, line 200: Would be cool if you could add a TODO here with your name on the impact of this change

jakobmoellerdev (commented):
Just one last comment from my side, apart from that LGTM

bindings/go/input/dir/blob.go
ikhandamirov: Done! Thanks!

jakobmoellerdev:
now you need to register a new module here https://github.com/open-component-model/ocm-website/tree/main/static/open-component-model/bindings/go/input and then you can add it to the CLI!

PR #356, opened 2025-07-07
chore(docs): update meeting notes for next TSC call

jakobmoellerdev:

updates the TSC notes for the upcoming TSC meeting


frewilhelm:
The title of the document needs to be adjusted to `06-07`

PR #364, opened 2025-07-07
chore(docs): ocm resource download ADR

jakobmoellerdev:

Enable easy and simple interpretation of downloaded resources fetched from component versions based on a new set of specifications that integrates seamlessly with the new OCM Plugin System, while providing a rearchitected foundation from the OCM v1 download handlers.


docs/adr/0007_resource_download.md
```diff
@@ -0,0 +1,194 @@
...
+```yaml
+type: transformer.blob.ocm.software/v1alpha1
+transformers:
+- id: extract-oci-artifact
+  spec:
+    type: extract.oci.artifact.ocm.software/v1alpha1
+    # additional configuration for the transformer
+```
+
+This allows transformer registration via configuration, enabling portable setups and declarative ocm configuration structures.
```
Skarlso, line 112: I'm struggling with this part:
> This allows transformer registration via configuration

How? I mean this is just passed to the transformer, right? That means it already "exists" technically. This allows to configure that for sure. But I don't see how this allows "registration". Like, for me, registration would mean that for example, if it's a binary, it would point to the binary to load during startup or something.

I guess my question is, what part of this is allowing registration? The fact that the config type `extract.oci.artifact.ocm.software/v1alpha1` exists in the config? Is that enough?

jakobmoellerdev: For me the registration is coming from the `id`. The id defines an instance of the config that you can directly access (see the config for that)

Skarlso: Ahhha. Okay. I thought so. That makes sense. Can we mention that please to make it explicit? 🙇

docs/adr/0007_resource_download.md
```diff
@@ -0,0 +1,194 @@
...
+
+By default, this transformer will:
+
+1. Introspect the tarball stream to identify the OCI Layout, and access its index.
+2. Identify the top-level artifact within the OCI Layout, by introspecting the index.
+3. Extract the artifact's content into a specified destination directory, preserving the OCI Layout structure.
+
+By default, the transformer will extract all layers from the top-level artifact and unpack it into the specified destination directory, while trying to respect the original media type of the layer.
+
+#### Example transformation command
```
Skarlso, line 141: I really appreciate this example. ❤️

Skarlso (commented):
Done with two comments left. :) 👍

ikhandamirov (commented):
Looks good to me. If my comments regarding the name of the flag don't resonate, feel free to just resolve.

docs/adr/0007_resource_download.md
```diff
@@ -0,0 +1,195 @@
...
+This requires the support of a new set of handlers that can reuse the existing OCM binding libraries to interpret the resource data and provide a more granular download infrastructure.
+
+To specify the exact download that is required, we introduce a new concept called a **BlobTransformer**, that is simpler than the existing "download handler" concept, but still allows for a flexible and extensible way to handle resource data.
+
+This will allow us to define how the resource data should be transformed before it is returned to the user — e.g., extracting a tarball, converting a JSON file, etc.
+
+In the OCM CLI, users will be able to transform their downloaded resource by specifying a `BlobTransformer` on the command line:
+
+```shell
+ocm download resources <component-reference> <resource-reference> --download-transformer extract-oci-artifact
```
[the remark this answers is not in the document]
jakobmoellerdev: I agree, naming is always hard. If you have another suggestion, please let me know :D

docs/adr/0007_resource_download.md
```diff
@@ -0,0 +1,195 @@
...
+```shell
+ocm download resources ghcr.io/open-component-model/ocm//ocm.software/ocmcli:0.26.0 -o ./ --identity name=ocmcli,os=linux,arch=amd64
+```
+
+*This will download the `ocmcli` binary as a binary stream, which can be run as is.*
+
+With Transformation
+
+```shell
+ocm download resources ghcr.io/open-component-model/samples//sample-chart -o ./ --identity name=podinfo --transform extract-oci-artifact
```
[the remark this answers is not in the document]
jakobmoellerdev: The flag definetly needs unification, i did a mistake during documentation here. Lets agree on one term and then use that?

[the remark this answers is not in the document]
jakobmoellerdev: unified the name to transform

PR #368, opened 2025-07-08
feat: make the base logger settable

Skarlso:

### DIFF
--- a/bindings/go/oci/internal/lister/lister.go
+++ b/bindings/go/oci/internal/lister/lister.go
@@ -114,12 +114,12 @@ func (lister *Lister) List(ctx context.Context, opts Options) ([]string, error) 
 	if err != nil {
 		return nil, err
 	}
-	log.Base().Log(ctx, slog.LevelDebug, "listed version candidates", slog.Int("count", len(candidates)))
+	log.Base(ctx).Log(ctx, slog.LevelDebug, "listed version candidates", slog.Int("count", len(candidates)))
 	sorted, err := lister.sort(ctx, opts, candidates)
 	if err != nil {
 		return nil, err
 	}
-	log.Base().Log(ctx, slog.LevelDebug, "sorted version candidates")
+	log.Base(ctx).Log(ctx, slog.LevelDebug, "sorted version candidates")
 
 	return sorted, nil
 }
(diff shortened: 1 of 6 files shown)


Skarlso:
I didn't want to break the Base contract that's why I thought I do it like this. Also I can just call.

```go
	repo := &Repository{
		scheme:                     options.Scheme,
		localArtifactManifestCache: options.LocalManifestCache,
		localArtifactLayerCache:    options.LocalLayerCache,
		resolver:                   options.Resolver,
		creatorAnnotation:          options.Creator,
		resourceCopyOptions:        *options.ResourceCopyOptions,
		referrerTrackingPolicy:     options.ReferrerTrackingPolicy,
		logger:                     options.Logger,
	}

	repo.setBaseLogger()
```

And it would work fine from there. :D

But your suggestion is cleaner.

jakobmoellerdev:
Yah you have a point but I think it was a mistake that I added it as a package level variable. its always an anti pattern long term...

bindings/go/oci/internal/lister/lister.go
```diff
@@ -114,12 +114,12 @@ func (lister *Lister) List(ctx context.Context, opts Options) ([]string, error)
 	if err != nil {
 		return nil, err
 	}
-	log.Base().Log(ctx, slog.LevelDebug, "listed version candidates", slog.Int("count", len(candidates)))
+	log.Base(ctx).Log(ctx, slog.LevelDebug, "listed version candidates", slog.Int("count", len(candidates)))
```
jakobmoellerdev, line 117: that looks completely off. youre passing the context twice even though the log interface already has a context with it. whats the purpose of the context then?

Skarlso: huh, where am I passing it twice?

jakobmoellerdev: See for example https://github.com/veqryn/slog-context to see how this can be used

Skarlso: Ah interesting. I take a look at that package.

Skarlso:
Superseeded by https://github.com/open-component-model/open-component-model/pull/374

PR #375, opened 2025-07-09
feat(oci)!: introspect OCI Layouts with graph

jakobmoellerdev:

This makes it so that the OCI library can now dynamically discover top level artifacts in OCI layouts passed to upload scenarios. This is achieved by building a successor graph dynamically based on the index. The call itself is implemented to be extremely efficient for 90% of our cases. (1-10 main artifacts with <100 references).

This effectively allows us to use layouts without passing in a source access and we can simplify our OCI repository interfaces signficantly. This means we are now much closer to the old OCMv1 Main Artifact Fetching in Artifact Sets.

This also allows library users to read OCI Layouts without knowing a source reference, making it extremely easy to write laoyut transformations (i.e. a single artifact layout with a helmchart can now be easily extracted). This is part of https://github.com/open-component-model/ocm-project/issues/523 efforts so that we can build easy layout blob transformations

preparation for https://github.com/open-component-model/ocm-project/issues/553


bindings/go/oci/tar/top_level_artifacts.go
```diff
@@ -0,0 +1,60 @@
...
+// It uses content.Successors to find successors of each artifact.
+// The function returns a slice of top-level artifacts.
+func TopLevelArtifacts(ctx context.Context, fetcher content.Fetcher, candidates []ociImageSpecV1.Descriptor) []ociImageSpecV1.Descriptor {
+	// If there's only one artifact, it's automatically a top-level artifact
+	if len(candidates) <= 1 {
+		return candidates
+	}
+
+	// Build a set of all referenced digests
+	var mu sync.Mutex
```
Skarlso, line 24: Hmm, I don't think you need another mutex if your calling origin is inside a WaitGroup that already does synced access. 🤔

Skarlso: Eh, I think you're right. Sync WaitGroup is only about coordination. It doesn't handle syncing. My bad.

jakobmoellerdev:
While the code I wrote here was clean I noticed a race in the OCI writer (not reader) that I fixed up

PR #389, opened 2025-07-11
feat: add central configuration types for logging and filesystem

Skarlso:


bindings/go/configuration/filesystem/v1alpha1/config.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+// Config represents the top-level configuration for the plugin manager.
+//
+// +k8s:deepcopy-gen:interfaces=ocm.software/open-component-model/bindings/go/runtime.Typed
+// +k8s:deepcopy-gen=true
+// +ocm:typegen=true
+type Config struct {
+	Type runtime.Type `json:"-"`
+	// TempFolder defines places where plugins and other functionalities can put ephemeral files under.
```
jakobmoellerdev, line 31: just missing a defaulting to os temp dir in the comment

bindings/go/configuration/filesystem/v1alpha1/config.go
```diff
@@ -0,0 +1,105 @@
...
+		}
+		merged = Merge(cfgs...)
+		if merged == nil {
+			merged = &Config{}
+		}
+	} else {
+		merged = new(Config)
+	}
+
+	// TODO: Question: Should we do this, or leave it empty signaling that the using code can create their own location?
```
jakobmoellerdev, line 76: I think we want plugins to be able to choose os.TempDir() right?

Skarlso: Yes. So I think we shouldn't default this then. Otherwise, it will always have a value that will point to os.TempDir if not defined explicitly.

jakobmoellerdev (approved):

PR #390, opened 2025-07-14
Configure internal plugins 4

Skarlso:


cli/go.mod
```diff
@@ -11,7 +11,7 @@ require (
 	golang.org/x/sync v0.16.0
 	golang.org/x/sys v0.34.0
 	ocm.software/open-component-model/bindings/go/blob v0.0.3
-	ocm.software/open-component-model/bindings/go/configuration v0.0.1
+	ocm.software/open-component-model/bindings/go/configuration v0.0.2-0.20250711153804-7cbbb713b74e
```
jakobmoellerdev, line 14: Please release a new version of configuration first

cli/cmd/cmd.go
```diff
@@ -16,6 +16,10 @@ import (
 	"ocm.software/open-component-model/cli/internal/flags/log"
 )
 
+const (
+	tempFolderFlag = "temp-folder"
```
jakobmoellerdev, line 20: what are you doing with this flag?

cli/cmd/setup.go
```diff
@@ -29,6 +31,88 @@ func setupOCMConfig(cmd *cobra.Command) {
...
+	// Set the type for the filesystem config
+	fsCfg.Type = runtime.NewVersionedType(v1alpha1.ConfigType, v1alpha1.Version)
+
+	// Create a runtime.Raw directly with the filesystem config
+	raw := &runtime.Raw{
+		Type: fsCfg.Type,
+	}
+
+	// Use the v1 scheme to encode the filesystem config
+	if err := v1.Scheme.Convert(fsCfg, raw); err != nil {
```
jakobmoellerdev, line 102: why do you not just initialize the config directly?

cli/cmd/setup.go
```diff
@@ -29,6 +31,88 @@ func setupOCMConfig(cmd *cobra.Command) {
...
+	var fsCfg *v1alpha1.Config
+	var err error
+
+	if cfg == nil {
+		slog.WarnContext(cmd.Context(), "could not get configuration to initialize filesystem config")
+		fsCfg = &v1alpha1.Config{}
+	} else {
+		fsCfg, err = v1alpha1.LookupConfig(cfg)
+		if err != nil {
+			slog.WarnContext(cmd.Context(), "could not get filesystem configuration", slog.String("error", err.Error()))
```
jakobmoellerdev, line 49: imo this is at most a debug log

bindings/go/ctf/ctf.go
```diff
@@ -96,7 +97,8 @@ type ReadOnlyBlobStore interface {
 // OpenCTF opens a CTF at the specified path with the specified format.
 // the CTF may be backed by a temporary directory if the format is FormatTAR or FormatTGZ.
 // In this case, the temporary directory is used to extract the archive before returning access on that path.
-func OpenCTF(ctx context.Context, path string, format FileFormat, flag int) (CTF, error) {
+// The fsConfig parameter allows customization of filesystem behavior, such as temporary directory location.
+func OpenCTF(ctx context.Context, path string, format FileFormat, flag int, fsConfig *v1alpha1.Config) (CTF, error) {
```
jakobmoellerdev, line 101:
```suggestion
// The optional config parameter allows customization of filesystem behavior, such as temporary directory location.
func OpenCTF(ctx context.Context, path string, format FileFormat, flag int, config *v1alpha1.Config) (CTF, error) {
```

bindings/go/ctf/ctf.go
```diff
@@ -110,7 +112,13 @@ func OpenCTF(ctx context.Context, path string, format FileFormat, flag int) (CTF
 			return nil, fmt.Errorf("unable to hash path to determine temporary ctf: %w", err)
 		}
 
-		tmp, err := os.MkdirTemp("", fmt.Sprintf("ctf-%x-*", hash.Sum(nil)))
+		// Use filesystem configuration for temporary directory
+		tempDir := os.TempDir()
```
jakobmoellerdev, line 116: super tiny nit: imo we should not initialize this variable by default as it may do an env lookup.

bindings/go/ctf/ctf_compatibility_test.go
```diff
@@ -45,7 +45,7 @@ func Test_CTF_Basic_ReadOnly_Compatibility(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			ctx := t.Context()
 			r := require.New(t)
-			archive, discovered, err := ctf.OpenCTFByFileExtension(ctx, tc.path, ctf.O_RDONLY)
+			archive, discovered, err := ctf.OpenCTFByFileExtension(ctx, tc.path, ctf.O_RDONLY, nil)
```
jakobmoellerdev, line 48: am I the only one that hates nil parameters? maybe :D

jakobmoellerdev (commented):
The general design of the changes is nice! I am wondering about some nits and if you can really change all of these modules at once. I.e. youre changing the CLI but are also changing the OCI binding. The changes in OCI won't propagate to CLI if you do not separate these PRs out or create follow-ups. Whats your plan here?

bindings/go/ctf/ctf_compatibility_test.go
Skarlso: Nah, for sure. But this is the default I guess. That's why I added an extra test to test the config explicitly. What would you prefer? An options list?

bindings/go/ctf/ctf.go
Skarlso: Fair 👍

jakobmoellerdev:
As i explained above, the modules are not separated. The only reason you compile is because you do not upgrade the modules yet. But for example, in OCI you have usages of OpenCTF that would be affected by your changes and that would anyhow require a second PR

cli/cmd/setup.go
Skarlso: Ah, something like this would work, right?
```go
	raw := &runtime.Raw{}
	if err := v1.Scheme.Convert(fsCfg, raw); err != nil {
		return fmt.Errorf("failed to convert filesystem config to raw: %w", err)
	}
	cfg.Configurations = append(cfg.Configurations, raw)
```

Since config already should have type set through the scheme?

Skarlso: Yah, it appears to be working... Tests are passing at least. :D

docs/reference/cli/ocm_add.md
```diff
@@ -24,37 +24,38 @@ ocm add {component-version|component-versions|cv|cvs} [flags]
...
+                                debug: Show all logs including detailed debugging information
+                                info:  Show informational messages and above
+                                warn:  Show warnings and errors only (default)
+                                error: Show errors only
+                             (must be one of [debug error info warn]) (default info)
+      --logoutput enum       set the log output destination
+                                stdout: Write logs to standard output (default)
+                                stderr: Write logs to standard error, useful for separating logs from normal output
+                             (must be one of [stderr stdout]) (default stdout)
+      --temp-folder string   Specify a custom temporary folder path for filesystem operations.
```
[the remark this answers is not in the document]
jakobmoellerdev: Documentation is preferred for me.
If the CLI flag takes precedence and there is a value in the config, I believe we should log a warning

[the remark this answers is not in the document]
Skarlso: But the CLI overwriting _is_ the actual feature. 🤔  Like, when I define it I know what I'm doing I want to override the existing config. If we do that, and you define mutliple flags, would you expect multiple warnings?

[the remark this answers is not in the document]
Skarlso: Okay.

cli/cmd/setup_test.go
```diff
@@ -0,0 +1,346 @@
...
+		TempFolder: tempFolder,
+	}
+
+	// Serialize the config to JSON
+	data, err := json.Marshal(cfg)
+	if err != nil {
+		panic(err)
+	}
+
+	return &runtime.Raw{
```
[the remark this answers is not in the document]
jakobmoellerdev: because it would be consistent with the rest of the repository, we never manually build Raws. ultimately its a nit.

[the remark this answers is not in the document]
Skarlso: Ah okay. Let me fix that up then. I didn't particularly knew since setup had no tests. So I just winged it. :) And this was easier. :D

docs/reference/cli/ocm_add.md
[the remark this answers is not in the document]
jakobmoellerdev: that sounds off. isnt that something that only should get logged as part of CLI?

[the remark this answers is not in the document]
Skarlso: No wait, I'm an idiot. Disregard.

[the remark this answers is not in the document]
Skarlso: Done.

cli/cmd/setup_filesystem_config.go
```diff
@@ -0,0 +1,88 @@
...
+			}
+		}
+	}
+
+	ctx := ocmctx.WithFilesystemConfig(cmd.Context(), fsCfg)
+	cmd.SetContext(ctx)
+}
+
+// hasFilesystemConfig checks if the central configuration already contains filesystem configuration
+func hasFilesystemConfig(cfg *v1.Config) bool {
```
jakobmoellerdev, line 57: this is not a safe way to check for this config because it could be versioned or unversioned.

cli/cmd/setup_filesystem_config.go
```diff
@@ -0,0 +1,88 @@
...
+
+	"ocm.software/open-component-model/bindings/go/configuration/filesystem/v1alpha1"
+	v1 "ocm.software/open-component-model/bindings/go/configuration/v1"
+	"ocm.software/open-component-model/bindings/go/runtime"
+	ocmctx "ocm.software/open-component-model/cli/internal/context"
+)
+
+// setupFilesystemConfig sets up file system configuration entity.
+func setupFilesystemConfig(cmd *cobra.Command) {
+	value, _ := cmd.PersistentFlags().GetString(tempFolderFlag)
```
jakobmoellerdev, line 17: why do you ignore the error here?

cli/internal/plugin/builtin/oci/repository.go
```diff
@@ -44,12 +46,15 @@ func createRepository(
...
 		return nil, fmt.Errorf("failed to create URL resolver: %w", err)
 	}
-	repo, err := oci.NewRepository(
+	options := []oci.RepositoryOption{
 		oci.WithResolver(urlResolver),
 		oci.WithCreator(Creator),
 		oci.WithManifestCache(manifests),
 		oci.WithLayerCache(layers),
-	)
+		oci.WithFilesystemConfig(filesystemConfig), // the filesystem config being empty is a valid config
```
jakobmoellerdev, line 54: i guess with empty you mean nil?

jakobmoellerdev (commented):
just two things I noticed during the final run, other than that great job 🥳

cli/internal/plugin/builtin/oci/repository.go
Skarlso: Yes.

cli/cmd/setup_filesystem_config.go
Skarlso: Thanks. I wanted to fix that.

cli/cmd/setup_filesystem_config.go
```diff
@@ -14,13 +14,17 @@ import (
 
 // setupFilesystemConfig sets up file system configuration entity.
 func setupFilesystemConfig(cmd *cobra.Command) {
-	value, _ := cmd.PersistentFlags().GetString(tempFolderFlag)
+	value, err := cmd.PersistentFlags().GetString(tempFolderFlag)
+	if err != nil {
+		slog.ErrorContext(cmd.Context(), "could not read temp folder flag value", slog.String("error", err.Error()))
+
+		// no point to continue if we don't have a value
+		return
```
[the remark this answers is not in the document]
Skarlso: Fixed.

jakobmoellerdev (approved):

PR #403, opened 2025-07-15
feat(resolvers): adapt ocm v1 resolver behaviour

fabianburth:
This PR introduces the ocm config type defined by ocm allowing to specify resolver, as well as the corresponding resolver behaviour.

This PR adresses: https://github.com/open-component-model/ocm-project/issues/513


bindings/go/repositories/artifactrepository/interface.go
```diff
@@ -0,0 +1,63 @@
...
+	// Note that UploadResource is special in that it considers both
+	// - the Access from [descriptor.Resource.Access]
+	// - the Target Access from the given target specification
+	// It might be that during the upload, the source pointer may be updated with information gathered during upload
+	// (e.g. digest, size, etc).
+	//
+	// The content of form blob.ReadOnlyBlob is expected to be a (optionally gzipped) tar archive that can be read with
+	// tar.ReadOCILayout, which interprets the blob as an OCILayout.
+	//
+	// The given OCI Layout MUST contain the resource described in source with an v1.OCIImage specification,
```
jakobmoellerdev, line 26: These descriptions are invalid for non-oci repositories

bindings/go/repositories/artifactrepository/interface.go
```diff
@@ -0,0 +1,63 @@
...
+
+// ResourceRepository defines the interface for storing and retrieving OCM resources
+// independently of component versions from a Store Implementation
+type ResourceRepository interface {
+	// UploadResource uploads a [descriptor.Resource] to the repository.
+	// Returns the updated resource with repository-specific information.
+	// The resource must be referenced in the component descriptor.
+	// Note that UploadResource is special in that it considers both
+	// - the Access from [descriptor.Resource.Access]
+	// - the Target Access from the given target specification
```
jakobmoellerdev, line 19: this is an old description that needs to be adjusted from main

bindings/go/repositories/doc.go
```diff
@@ -0,0 +1,25 @@
+// Package repositories provides an abstraction layer for working
+// with different OCM (Open Component Model) repository technologies.
+// It defines a set of interfaces and types that allow developers to interact
+// with component version repositories in a technology-agnostic way.
```
jakobmoellerdev, line 4:
```suggestion
// Package repositories provides an abstraction layer for working
// with different OCM (Open Component Model) repository technologies.
```

bindings/go/repositories/doc.go
```diff
@@ -0,0 +1,25 @@
...
+//
+// This module enables technology-specific repository implementations (such as
+// OCI, CTF) to conform to a common set of interfaces. By programming against
+// these interfaces, consumers of the module can seamlessly switch between
+// repository backends without changing their application logic.
```
jakobmoellerdev, line 9:
```suggestion
// By programming against these technology-specific repository interfaces, consumers of the module can seamlessly switch between
// repository backends without changing their application logic.
```

jakobmoellerdev (commented):
mostly ok!

bindings/go/repositories/componentrepository/fallback/repository.go
```diff
@@ -0,0 +1,344 @@
...
+	"golang.org/x/sync/errgroup"
+
+	"ocm.software/open-component-model/bindings/go/blob"
+	descriptor "ocm.software/open-component-model/bindings/go/descriptor/runtime"
+	"ocm.software/open-component-model/bindings/go/repositories/componentrepository"
+	resolverruntime "ocm.software/open-component-model/bindings/go/repositories/componentrepository/resolver/config/runtime"
+	"ocm.software/open-component-model/bindings/go/runtime"
+)
+
+// Deprecated: FallbackRepository is an implementation for the deprecated config
```
jakobmoellerdev, line 25: deprecated notices usually are at the end of the code comment no?

jakobmoellerdev: I think fallback repositories in general are not deprecated, but ones from v1 configs are right? if you plan a separate implementation i suggest to introduce a v1 package and deprecate the entire package

bindings/go/repositories/componentrepository/fallback/repository.go
```diff
@@ -0,0 +1,344 @@
...
+// retrieve the component version from each repository.
+func (f *FallbackRepository) GetComponentVersion(ctx context.Context, component, version string) (*descriptor.Descriptor, error) {
+	repos := f.RepositoriesForComponentIterator(ctx, component)
+	for repo, err := range repos {
+		if err != nil {
+			return nil, fmt.Errorf("getting repository for component %s failed: %w", component, err)
+		}
+		desc, err := repo.GetComponentVersion(ctx, component, version)
+		if err != nil {
+			if errors.As(err, new(*componentrepository.NotFoundError)) {
```
jakobmoellerdev, line 112: does this not already cover a nil check for err != nil??

bindings/go/repositories/componentrepository/fallback/repository.go
```diff
@@ -0,0 +1,344 @@
...
+	repos := f.RepositoriesForComponentIterator(ctx, component)
+
+	var errGroup errgroup.Group
+	var versionsMu sync.Mutex
+	accumulatedVersions := make(map[string]struct{})
+
+	errGroup.SetLimit(goruntime.NumCPU())
+
+	for repo, err := range repos {
+		errGroup.Go(func() error {
```
jakobmoellerdev, line 141: initialize the errgroup with errgroup.WithContext then you dont need to manually check for context cancellation

bindings/go/repositories/componentrepository/fallback/repository.go
```diff
@@ -0,0 +1,344 @@
...
+
+// Deprecated: FallbackRepository is an implementation for the deprecated config
+// type "ocm.config.ocm.software/v1". This concept of fallback resolvers is deprecated
+// and only added for backwards compatibility.
+// New concepts will likely be introduced in the future (contributions welcome!).
+//
+// ListComponentVersions accumulates a deduplicated list of the versions of the
+// given component from all repositories specified by resolvers with a matching
+// component prefix in the order of their priority (higher priorities first).
+func (f *FallbackRepository) ListComponentVersions(ctx context.Context, component string) ([]string, error) {
```
jakobmoellerdev, line 131: QQ: does List really always go to the Repo and not reuse its cached entries for the component? this looks a bit off

bindings/go/repositories/componentrepository/fallback/repository.go
```diff
@@ -0,0 +1,344 @@
...
+		for index, resolver := range f.resolvers {
+			if resolver.Prefix != "" && !strings.HasPrefix(component, resolver.Prefix) {
+				continue
+			}
+			repo, err := f.getRepositoryFromCache(ctx, index, resolver)
+			if err != nil {
+				yield(nil, fmt.Errorf("getting repository for resolver %v failed: %w", resolver, err))
+				return
+			}
+			if repo == nil {
```
jakobmoellerdev, line 285: do we really need this check here?

bindings/go/repositories/componentrepository/fallback/repository.go
```diff
@@ -0,0 +1,344 @@
...
+// Compared to RepositoriesForComponent, using the iterator allows for lazy
+// evaluation and can be more efficient when only a few repositories are
+// needed (e.g., when leveraged by the CLI code to do a simple GetComponentVersion).
+func (f *FallbackRepository) RepositoriesForComponentIterator(ctx context.Context, component string) iter.Seq2[componentrepository.ComponentVersionRepository, error] {
+	return func(yield func(componentrepository.ComponentVersionRepository, error) bool) {
+		for index, resolver := range f.resolvers {
+			if resolver.Prefix != "" && !strings.HasPrefix(component, resolver.Prefix) {
+				continue
+			}
+			repo, err := f.getRepositoryFromCache(ctx, index, resolver)
```
jakobmoellerdev, line 280: why do you not call yield(repo, err) here?

bindings/go/repositories/componentrepository/fallback/repository.go
```diff
@@ -0,0 +1,344 @@
...
+	}
+}
+
+// Deprecated: FallbackRepository is an implementation for the deprecated config
+// type "ocm.config.ocm.software/v1". This concept of fallback resolvers is deprecated
+// and only added for backwards compatibility.
+// New concepts will likely be introduced in the future (contributions welcome!).
+//
+// GetResolvers returns a copy of the resolvers used by this repository.
+func (f *FallbackRepository) GetResolvers() []*resolverruntime.Resolver {
```
jakobmoellerdev, line 303: why do you need this deep copy?

bindings/go/repositories/componentrepository/fallback/repository.go
```diff
@@ -0,0 +1,344 @@
...
+	}
+	repo, err := f.repositoryProvider.GetComponentVersionRepository(ctx, specification, credentials)
+	if err != nil {
+		return nil, fmt.Errorf("getting component version repository for %q failed: %w", specification, err)
+	}
+	return repo, nil
+}
+
+func (f *FallbackRepository) getRepositoryFromCache(ctx context.Context, index int, resolver *resolverruntime.Resolver) (componentrepository.ComponentVersionRepository, error) {
+	f.repositoriesForResolverCacheMu.Lock()
```
jakobmoellerdev, line 331: to optimize, mack this cache mu a RWMutex

bindings/go/repositories/componentrepository/resolver/config/runtime/config.go
```diff
@@ -0,0 +1,50 @@
...
+	// If resolvers are defined, it is possible to use component version names on the
+	// command line without a repository. The names are resolved with the specified
+	// resolution rule.
+	//
+	// They are also used as default lookup repositories to lookup component references
+	// for recursive operations on component versions («--lookup» option).
+	Resolvers []Resolver `json:"-"`
+}
+
+// Resolver assigns a priority and a prefix to a single OCM repository specification
```
jakobmoellerdev, line 33: This is missing a deprecation notice

bindings/go/repositories/componentrepository/resolver/config/runtime/config.go
```diff
@@ -0,0 +1,50 @@
...
+	// Resolvers define a list of OCM repository specifications to be used to resolve
+	// dedicated component versions.
+	// All matching entries are tried to lookup a component version in the following
+	//    order:
+	//    - highest priority first
+	// If resolvers are defined, it is possible to use component version names on the
+	// command line without a repository. The names are resolved with the specified
+	// resolution rule.
+	//
+	// They are also used as default lookup repositories to lookup component references
```
jakobmoellerdev, line 28: There is no mention of a default priority here. Is this intended?

bindings/go/repositories/componentrepository/resolver/config/spec/v1/config.go
```diff
@@ -0,0 +1,132 @@
...
+var scheme = runtime.NewScheme()
+
+func init() {
+	scheme.MustRegisterWithAlias(&Config{},
+		runtime.NewVersionedType(ConfigType, Version),
+		runtime.NewUnversionedType(ConfigType),
+	)
+}
+
+// Deprecated: Resolvers are deprecated and are only added for backwards
```
jakobmoellerdev, line 28: I am missing a yaml example here

bindings/go/repositories/componentrepository/fallback/repository.go
```diff
@@ -0,0 +1,344 @@
...
+// given component from all repositories specified by resolvers with a matching
+// component prefix in the order of their priority (higher priorities first).
+func (f *FallbackRepository) ListComponentVersions(ctx context.Context, component string) ([]string, error) {
+	repos := f.RepositoriesForComponentIterator(ctx, component)
+
+	var errGroup errgroup.Group
+	var versionsMu sync.Mutex
+	accumulatedVersions := make(map[string]struct{})
+
+	errGroup.SetLimit(goruntime.NumCPU())
```
jakobmoellerdev, line 138: should be configurable imo

bindings/go/repositories/componentrepository/fallback/repository.go
```diff
@@ -0,0 +1,344 @@
...
+// Deprecated: FallbackRepository is an implementation for the deprecated config
+// type "ocm.config.ocm.software/v1". This concept of fallback resolvers is deprecated
+// and only added for backwards compatibility.
+// New concepts will likely be introduced in the future (contributions welcome!).
+//
+// NewFallbackRepository creates a new FallbackRepository instance.
+func NewFallbackRepository(_ context.Context, repositoryProvider componentrepository.ComponentVersionRepositoryProvider, credentialProvider componentrepository.CredentialProvider, res ...*resolverruntime.Resolver) (*FallbackRepository, error) {
+	resolvers := make([]*resolverruntime.Resolver, len(res))
+
+	// copy the resolvers to ensure immutability
```
jakobmoellerdev, line 61: why do we need immutability here?

bindings/go/repositories/componentrepository/fallback/repository.go
fabianburth: I thought they are. In previous discussions, discussed alternative ideas (such as mapping component names to repositories based on a glob or regex matching) are repository providers. Besides, I'm not sure whether I'd call such a behaviour "fallback". But those are semantic nitpics tbh.

Overall, I think introducing a v1 package and deprecating it sounds like a good idea.

bindings/go/repositories/componentrepository/fallback/repository.go
fabianburth: we met up and checked the old code together - it actually does go to the repo.

bindings/go/repositories/componentrepository/fallback/repository.go
fabianburth: Moved it to `getRepositoryForSpecification()`. As we do not know the `provider` in there, I still think it's a good idea to perform the check.

bindings/go/repositories/componentrepository/fallback/repository.go
fabianburth: I want to prevent weird side effects that might happen if someone changes the resolver that is pointed to within the repository.

bindings/go/repositories/componentrepository/fallback/repository.go
fabianburth: to prevent possible side effects caused by resolvers being changed outside of the fallback code

bindings/go/repositories/componentrepository/fallback/repository.go
jakobmoellerdev: imo the value, err := pattern is indicative that if err == nil, value must be not nil, so I consider it defensive programming. The contract clearly should return a valid repo imo. WDYT?

bindings/go/repositories/componentrepository/fallback/repository.go
fabianburth: wasn't sure if we currently have a consistent way to pass such a config and this was consistent with the rest of the current code base, do you want me to change it?

jakobmoellerdev: You will need to expose a setting in the library either way. we will introduce centrally available settings through the config with inbuilt and external plugins separately. for more info reach out to @Skarlso

fabianburth: So, do you want me to do something for now?

jakobmoellerdev: Ideally you talk to @Skarlso and ask him how to best integrate such a config later so we can be sure it is working in the future

bindings/go/repositories/componentrepository/resolver/config/v1/spec/config.go
```diff
@@ -0,0 +1,161 @@
...
+	merged := new(Config)
+	merged.Type = configs[0].Type
+	merged.Resolvers = make([]*Resolver, 0)
+	merged.Aliases = map[string]*runtime.Raw{}
+
+	for _, cfg := range configs {
+		// Merge aliases
+		for alias, raw := range cfg.Aliases {
+			if existing, exists := merged.Aliases[alias]; exists {
+				if existing != nil && raw != nil {
```
Skarlso, line 150: Shouldn't this also check if the values are equal and not just not nil?

fabianburth: So, you'd like to throw an error or log a warning if there are two identical aliases mapping to different repositories?

PR #407, opened 2025-07-15
feat: blob transformer plugins

jakobmoellerdev:

implements blob transformers as per https://github.com/open-component-model/open-component-model/blob/main/docs/adr/0007_resource_download.md#blobtransformers-and-contracts with extension into the credential system if needed

fix https://github.com/open-component-model/ocm-project/issues/552

### DIFF
--- a/bindings/go/plugin/manager/registries/blobtransformer/converter.go
+++ b/bindings/go/plugin/manager/registries/blobtransformer/converter.go
@@ -1,0 +1,59 @@
+package blobtransformer
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"os"
+
+	"ocm.software/open-component-model/bindings/go/blob"
+	"ocm.software/open-component-model/bindings/go/blob/filesystem"
+	blobtransformerv1 "ocm.software/open-component-model/bindings/go/plugin/manager/contracts/blobtransformer/v1"
+	"ocm.software/open-component-model/bindings/go/plugin/manager/registries/blobs"
+	"ocm.software/open-component-model/bindings/go/plugin/manager/types"
+	"ocm.software/open-component-model/bindings/go/runtime"
+)
+
+type converter struct {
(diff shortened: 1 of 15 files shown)


bindings/go/plugin/manager/registries/blobtransformer/converter.go
```diff
@@ -0,0 +1,59 @@
...
+}
+
+func (c *converter) TransformBlob(ctx context.Context, blob blob.ReadOnlyBlob, spec runtime.Typed, credentials map[string]string) (blob.ReadOnlyBlob, error) {
+	tmp, err := os.CreateTemp("", "blob")
+	if err != nil {
+		return nil, fmt.Errorf("failed to create temp file: %w", err)
+	}
+
+	defer func() {
+		err = errors.Join(err, tmp.Close())
```
Skarlso, line 29: don't we want to remove this file?

bindings/go/plugin/manager/registries/plugins/wait_for_plugin.go
```diff
@@ -83,8 +84,18 @@ func getPluginLocation(ctx context.Context, plugin *types.Plugin) (string, error
...
 
+	var reader io.Reader
+	switch {
+	case plugin.Stdout != nil && plugin.Stderr != nil:
+		reader = io.MultiReader(plugin.Stdout, plugin.Stderr)
+	case plugin.Stdout != nil:
+		reader = plugin.Stdout
+	case plugin.Stderr != nil:
+		reader = plugin.Stderr
+	}
```
Skarlso, line 95: I'm unsure about this. The reason I didn't add scanning stderr here, is that it could potentially pollute the output and jeopardize the reading of a single line which should be there normally.

That said, maybe this is fine, since we read all lines and we just skip it if we couldn't parse it.

Skarlso: It's fine, let's leave it and see. I _think_ the parsing will save this if anything pollutes stderr.

PR #421, opened 2025-07-16
feat(oci): use interfaces from repositories module

fabianburth:
Related to: https://github.com/open-component-model/open-component-model/pull/403

This PR added the ocm repository interfaces to a module that is intended to only work on the abstraction level of ocm repositories (and therefore, does not have storage technology specific code) and can be imported by oci.


bindings/go/oci/repository.go
```diff
@@ -366,15 +374,18 @@ func (repo *Repository) GetLocalSource(ctx context.Context, component, version s
 		done(err)
 	}()
 
-	var b LocalBlob
+	var b fetch.LocalBlob
 	var artifact descriptor.Artifact
 	if b, artifact, err = repo.localArtifact(ctx, component, version, identity, annotations.ArtifactKindSource); err != nil {
+		if errors.Is(err, errdef.ErrNotFound) {
+			return nil, nil, componentrepository.NewNotFoundError(fmt.Sprintf("component version %q/%q not found", component, version), err)
```
jakobmoellerdev, line 381: why does this error look so strange? Why couldn't the error here just be a static string?

fabianburth: can you show me an example of what you'd have in mind?

bindings/go/oci/integration/integration_test.go
```diff
@@ -345,10 +344,12 @@ func Test_Integration_OCIRepository(t *testing.T) {
 				"username": testUsername,
 				"password": password,
 			})
+			ocirepo, ok := repo.(oci.ComponentVersionRepository)
```
jakobmoellerdev, line 347: why do you need this assert here now?

bindings/go/oci/repository.go
```diff
@@ -178,6 +180,9 @@ func (repo *Repository) GetComponentVersion(ctx context.Context, component, vers
 	}
 
 	desc, _, _, err = getDescriptorFromStore(ctx, store, reference)
+	if errors.Is(err, errdef.ErrNotFound) {
+		return desc, errors.Join(fmt.Errorf("component version %s/%s not found: %w", component, version, componentrepository.ErrNotFound), err)
```
jakobmoellerdev, line 184:
```suggestion
		return desc, errors.Join(componentrepository.ErrNotFound, fmt.Errorf("component version %s/%s not found: %w", component, version, err))
```
i think you should switch this around as the first error will be the first one shown in the join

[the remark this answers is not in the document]
jakobmoellerdev: https://github.com/open-component-model/open-component-model/pull/421#discussion_r2212512200

bindings/go/oci/integration/integration_test.go
fabianburth: because the `RepositoryProvider` interface returns a repository of type `componentrepository.ComponentVersionRepository` but everything within oci works with `oci.ComponentVersionRepository` which is also a `ResourceDigestProcessor`.

fabianburth: I'd argue (or rather you argued) that typically when using a provider, one wouldn't really expect to be able to use technology specific functionality.

fabianburth: Damn, very good catch!
I missed pulling that change in.

bindings/go/oci/repository/provider/provider.go
```diff
@@ -105,9 +108,7 @@ func (b *CachingComponentVersionRepositoryProvider) GetComponentVersionRepositor
 // to its corresponding object type in the scheme. It ensures that the type is set correctly
 func getConvertedTypedSpec(scheme *runtime.Scheme, repositorySpecification runtime.Typed) (runtime.Typed, error) {
 	repositorySpecification = repositorySpecification.DeepCopyTyped()
-	if _, err := scheme.DefaultType(repositorySpecification); err != nil {
-		return nil, fmt.Errorf("failed to ensure type for repository specification: %w", err)
-	}
+	_, _ = scheme.DefaultType(repositorySpecification)
```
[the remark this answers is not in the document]
fabianburth: Pretty sure I had a state at one point where this error was the only reason why something wasn't working, removed it for testing and forgot to add it back in

PR #428, opened 2025-07-17
feat(constructor): update interfaces and omit empty name and/or version in constructor `ToIdentity`

fabianburth:
This PR updates the constructor module to use the interfaces from the central repositories module instead of from the oci module. Besides, it also aligns the constructors `ToIdentity` handling with the descriptors `ToIdentity` handling, where name and/or. version are omitted if empty.


jakobmoellerdev (commented):
needs an update to the latest module version

jakobmoellerdev (approved):

PR #433, opened 2025-07-17
feat: oci blob transformers for Single Layer Artifacts

Skarlso:


Skarlso:
So I re-read 0007 couple times now and realised that my implementation of the interface is sitting in there and that is not correct. Because that needs to live in the plugin package basically. `BlobTransformer` needs to be an interface declared here, but implemented by a plugin.

So I need a registry and all that jazz.

bindings/go/blob/transformer/testdata/oci-layout.tar.gz
jakobmoellerdev, line 1: please dont add blobs into the repo, gotta keep it small! :D

Skarlso:
Okay, I think this is now ready for review.

In CLI this is how I would register it: 

```go
	if err := ocitransformerplugin.Register(manager.BlobTransformerRegistry); err != nil {
		return fmt.Errorf("could not register OCI transformer plugin: %w", err)
	}

```

with: 

```go
// ociTransformerWrapper adapts the OCI transformer to the blob transformer interface
type ociTransformerWrapper struct {
	transformer *ocitransformer.Transformer
}

// TransformBlob implements the blob transformer interface with credentials parameter
func (w *ociTransformerWrapper) TransformBlob(ctx context.Context, input blob.ReadOnlyBlob, config runtime.Typed, credentials map[string]string) (blob.ReadOnlyBlob, error) {
	// OCI transformer doesn't use credentials, so we ignore that parameter
	return w.transformer.TransformBlob(ctx, input, config)
}

// Register registers the OCI transformer as an internal plugin.
func Register(registry *blobtransformer.Registry) error {
	ociTransformer := &ociTransformerWrapper{
		transformer: ocitransformer.New(),
	}
	
	// Create a prototype config instance for registration
	prototype := &spec.Config{}
	
	if err := blobtransformer.RegisterInternalBlobTransformerPlugin(
		registry.Scheme(),
		registry,
		ociTransformer,
		prototype,
	); err != nil {
		return fmt.Errorf("failed to register OCI blob transformer plugin: %w", err)
	}

	return nil
}
```

And the todo from Jakob would be something like selecting the a transformer if it exists and the result is the blob... something like: 

```go
	transformerPlugin, err := transformerRegistry.GetPlugin(cmd.Context(), transformerConfig)
	if err != nil {
		return fmt.Errorf("failed to lookup transformer %q: %w", transformer, err)
	}

	// Transform the blob
	transformedBlob, err := transformerPlugin.TransformBlob(cmd.Context(), data, transformerConfig, nil)
	if err != nil {
		return fmt.Errorf("failed to transform blob: %w", err)
	}
```

but I don't have the rest of the logic down yet. :)

jakobmoellerdev (changes requested):
As discussed we wanna get rid of this hard coding and replace it with proper config lookups based on the latest changes. after that we should be good to go here!

bindings/go/oci/go.mod
```diff
@@ -11,6 +11,7 @@ require (
 	github.com/veqryn/slog-context v0.8.0
 	golang.org/x/sync v0.16.0
 	ocm.software/open-component-model/bindings/go/blob v0.0.6
+	ocm.software/open-component-model/bindings/go/configuration v0.0.5
```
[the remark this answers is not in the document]
Skarlso: Errr... configuration is at 0.0.5.

bindings/go/oci/transformer/extract.go
```diff
@@ -0,0 +1,183 @@
...
+		// Check if this layer matches any of the rule's selectors
+		matched := false
+		for _, selector := range rule.LayerSelectors {
+			if selector.Matches(layerInfo) {
+				matched = true
+				break
+			}
+		}
+
+		if matched {
```
jakobmoellerdev, line 125: nit: switch this if and continue and then move the code out of the if block to make nesting smaller

jakobmoellerdev (commented):
I think we need a test for the default file names / without selectors

jakobmoellerdev (approved):

PR #438, opened 2025-07-18
chore: make sure that OCI binding can be fed directly with a tempfolder and doesnt need a file system config

jakobmoellerdev:

we want to get rid of the filesystem config completely from our default libraries so that we can bind against the configs only in CLI and co


bindings/go/oci/repository_options.go
```diff
@@ -53,9 +52,8 @@ type RepositoryOptions struct {
 	// If not provided, slog.Default() will be used.
 	Logger *slog.Logger
 
-	// FilesystemConfig is the configuration for filesystem operations.
-	// If not provided, default filesystem configuration will be used.
-	FilesystemConfig *v1alpha1.Config
+	// TempFolder is the default temporary filesystem folder for any temporary cached data
+	TempFolder string
```
fabianburth, line 56: absolute nit, but maybe sticking with the common terminology `TempDir` would be preferable

fabianburth (dismissed):

fabianburth (approved):

Issue #424, opened 2025-07-17
Check if unix domain socket is still in-use by a process if not, remove it


Skarlso:
The problem we are facing during start-up of a plugin is if the socket that it is about to create is still there it will error and never start. This results in some buggy behaviour and some problems that we don't catch or we just time-out.

What we should do is create a lock file for the socket, put the PID into the lock file for the socket and then during startup check if the lock file exists, if yes, we check if the PID is still alive, if yes, we error saying an instance of the plugin is already running. Nothing to do about that.

But if the PID is gone, we can safely delete the lock file and the socket and create a new one. This would prevent accidental deletion of the socket file by another process but still allow orphaned sockets to be removed and the plugin to start up clean.

PR #442, opened 2025-07-18
feat: add `extract.oci.artifact.ocm.software` config

Skarlso:

This is part of the work for https://github.com/open-component-model/open-component-model/issues/424.


bindings/go/configuration/extract/v1alpha1/config.go
```diff
@@ -0,0 +1,91 @@
...
+	"fmt"
+	"time"
+
+	v1 "ocm.software/open-component-model/bindings/go/configuration/v1"
+	"ocm.software/open-component-model/bindings/go/runtime"
+)
+
+const (
+	// ConfigType defines the type identifier for transformation configurations
+	ConfigType = "extract.oci.artifact.ocm.software"
```
jakobmoellerdev, line 14: im guessing this is a todo? what is this config type? :D

bindings/go/configuration/extract/v1alpha1/config.go
```diff
@@ -0,0 +1,91 @@
...
+)
+
+var scheme = runtime.NewScheme()
+
+func init() {
+	scheme.MustRegisterWithAlias(&Config{}, runtime.NewVersionedType(ConfigType, Version))
+}
+
+// Config represents the top-level configuration for the transformation.
```
jakobmoellerdev, line 23: I dont understand this pr, now this is for transformation, the config type doesnt fit and we have the Temp Folder in here which is already in another config

jakobmoellerdev (commented):
Hey there, struggling a bit with understanding the PR. would you mind filling out the description a little?

Skarlso:
@jakobmoellerdev Was trying to create the configuration entity descsribed in 0007 https://github.com/open-component-model/open-component-model/blob/main/docs/adr/0007_resource_download.md#example-configuration

jakobmoellerdev:
Got it thanks for the context! Sometimes I have amnesia :D then its all good!

bindings/go/configuration/extract/v1alpha1/config.go
```diff
@@ -0,0 +1,91 @@
...
+//
+// +k8s:deepcopy-gen:interfaces=ocm.software/open-component-model/bindings/go/runtime.Typed
+// +k8s:deepcopy-gen=true
+// +ocm:typegen=true
+type Config struct {
+	Type runtime.Type `json:"-"`
+	// TempFolder defines places where plugins and other functionalities can put ephemeral files under.
+	// If not defined, os.TempDir is used as a default.
+	TempFolder string `json:"tempFolder,omitempty"`
```
jakobmoellerdev, line 32: This currently shares configuration settings with the filesystem config so this looks off to me.

Skarlso: Ah, so maybe, this just contains the type?

bindings/go/configuration/extract/v1alpha1/config.go
```diff
@@ -0,0 +1,91 @@
...
+
+const (
+	// ConfigType defines the type identifier for transformation configurations
+	ConfigType = "extract.oci.artifact.ocm.software"
+)
+
+var scheme = runtime.NewScheme()
+
+func init() {
+	scheme.MustRegisterWithAlias(&Config{}, runtime.NewVersionedType(ConfigType, Version))
```
fabianburth, line 20: Why don't we register the unversioned config type here too? We do filter for it in the `LookupConfig()` function.

bindings/go/configuration/extract/v1alpha1/config.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+// Config represents the top-level configuration for the transformation.
+//
+// +k8s:deepcopy-gen:interfaces=ocm.software/open-component-model/bindings/go/runtime.Typed
+// +k8s:deepcopy-gen=true
+// +ocm:typegen=true
+type Config struct {
+	Type runtime.Type `json:"-"`
```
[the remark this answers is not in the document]
Skarlso: I'm honestly not sure, because some of these omit the type and some of them don't. :D I'm not sure when to do witch.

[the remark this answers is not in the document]
jakobmoellerdev: if you have a runtime package youre not supposed to use that for serialization, however here you work with a spec package, this one needs the serialization!

bindings/go/configuration/extract/v1alpha1/config.go
```diff
@@ -0,0 +1,91 @@
...
+
+func (d Duration) MarshalJSON() ([]byte, error) {
+	return json.Marshal(time.Duration(d).String())
+}
+
+func (d *Duration) UnmarshalJSON(b []byte) error {
+	var v interface{}
+
+	return json.Unmarshal(b, &v)
+}
```
fabianburth, line 45: What is the duration needed for?

bindings/go/configuration/extract/v1alpha1/doc.go
```diff
@@ -0,0 +1,2 @@
+// Package v1alpha1 any config that can influence the transformator's behavior.
```
fabianburth, line 1:
```suggestion
// Package v1alpha1 contains any config that can influence the transformator's behavior.
```

bindings/go/configuration/extract/v1alpha1/config.go
Skarlso: I thought it looks good. :D ( nothing I'll remove it. )

fabianburth:
I think due to the updates I did for the central configuration package, you'll have to adjust the directory setup (e.g. `configuration/extract/v1alpha1/spec).

bindings/go/configuration/extract/v1alpha1/config.go
fabianburth: 😂

bindings/go/configuration/extract/v1alpha1/config.go
Skarlso: Added layer configuration. Please see if that makes sense. 🤔

bindings/go/configuration/extract/v1alpha1/spec/config.go
```diff
@@ -0,0 +1,82 @@
...
+
+// Config represents the top-level configuration for the transformation.
+//
+// +k8s:deepcopy-gen:interfaces=ocm.software/open-component-model/bindings/go/runtime.Typed
+// +k8s:deepcopy-gen=true
+// +ocm:typegen=true
+type Config struct {
+	Type runtime.Type `json:"type"`
+	// Layer defines configuration options on selecting the right Layer.
+	Layer Layer `json:"layer,omitempty"`
```
jakobmoellerdev, line 35: Generally this looks much better. Im just wondering why the media type is needed? is this a selector? How is this supposed to work. I was missing the logic for the behavior on the structs

bindings/go/configuration/extract/v1alpha1/spec/config.go
```diff
@@ -0,0 +1,82 @@
...
+
+var scheme = runtime.NewScheme()
+
+func init() {
+	scheme.MustRegisterWithAlias(&Config{}, runtime.NewVersionedType(ConfigType, Version))
+}
+
+// Layer has configuration information for layers to process.
+type Layer struct {
+	Index     string `json:"index,omitempty"`
```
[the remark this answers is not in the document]
Skarlso: it should have been. I made it an `int`. 🤔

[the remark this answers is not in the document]
Skarlso: Damn, it's a string. WTH :D

bindings/go/configuration/extract/v1alpha1/spec/config.go
jakobmoellerdev: yes! as per specification you need to identify the correct layer by media type. but if we want a selector the spec should be a proper selector imo

jakobmoellerdev: I think we should invest in a proper specification here. That means for me that we need to find out what we want.

I think we want to be able to somehow tell the transformer to select _multiple_ layers by either _index_ or _mediaType_ or in the future even _annotations_ or other fields. That means we should probably invest in a proper selector here.

Additionally we need to identify how to map this layer to the filesystem. I am open for suggestions here

Skarlso: Okay, I added a basic selection logic. I feel like I might have went to far with it though...

Skarlso: I felt like, if we _do_ provide a selector, we actually also have to provide a nice way of matching. So I added LayerInfo and a properties type of matcher.

bindings/go/configuration/extract/v1alpha1/spec/config.go
```diff
@@ -0,0 +1,214 @@
...
+	Operator LayerSelectorOperator `json:"operator"`
+	// Values is an array of string values. If the operator is In or NotIn,
+	// the value array must be non-empty. If the operator is Exists or DoesNotExist,
+	// the value array must be empty.
+	Values []string `json:"values,omitempty"`
+}
+
+// LayerSelector allows selecting layers based on index, mediatype, and annotations.
+// +k8s:deepcopy-gen=true
+type LayerSelector struct {
```
jakobmoellerdev, line 60: selector logic needs some tests

bindings/go/configuration/extract/v1alpha1/spec/config.go
```diff
@@ -0,0 +1,214 @@
...
+	// MatchLabels is a map of {key,value} pairs. A single {key,value} in the matchLabels
+	// map is equivalent to an element of matchExpressions, whose key field is "key", the
+	// operator is "In", and the value array contains only "value". *Note* this is called
+	// match_Labels_ to be in-line with other, more common, matching selector operations.
+	// Though technically, no labels are matched.
```
jakobmoellerdev, line 65: Then lets not call it MatchLabels. I think in OCI, we call manifest addins annotations. https://github.com/opencontainers/image-spec/blob/main/descriptor.md#properties

so that would mean the match happens based on that.

jakobmoellerdev (commented):
apart from tests LGTM. maybe move the selector into a separate file for readability. nice job!

bindings/go/configuration/extract/v1alpha1/spec/config.go
Skarlso: Cool! Will do.

bindings/go/configuration/extract/v1alpha1/spec/config.go
Skarlso: Also, why the hell are you reviewing shit at 1AM dude? :D

bindings/go/configuration/extract/v1alpha1/spec/config.go
Skarlso: I also added annotations into the mix and added a BUNCH of tests. Like, after the skeleton was done it was just copy&paste. :D

PR #462, opened 2025-07-25
chore: update CLI documentation directory path in Taskfile

morri-son:

To be able to mount the reference doc as Hugo module on the website, the docs need to reside in the ocm.software/open-component-model/cli module.


jakobmoellerdev (commented):
you actually have to move the docs as well, its not enough to change the target!

PR #477, opened 2025-08-01
feat: add interface to blobs for plugin implementation

Skarlso:

Part of https://github.com/open-component-model/ocm-project/issues/553.

By adding this interface, we will allow for the blobtransformer plugin registry to implement this interface instead of the one in the plugin manager right now.


bindings/go/blob/go.mod
```diff
@@ -5,15 +5,24 @@ go 1.24.1
 require (
 	github.com/opencontainers/go-digest v1.0.0
 	github.com/stretchr/testify v1.10.0
+	ocm.software/open-component-model/bindings/go/runtime v0.0.2
```
jakobmoellerdev, line 8: Hmm i really dont know if we want the runtime to be part of the dependency chain for blobs

bindings/go/blob/go.mod
```diff
@@ -5,15 +5,24 @@ go 1.24.1
...
+)
+
+require (
```
jakobmoellerdev, line 11: two indirect blocks, maybe you used go.work?

Skarlso: Ah yes. Damn.

Skarlso: Actually no. That's the plain result of a go mod tidy. ¯\\_(ツ)\_\/¯

bindings/go/blob/go.mod
jakobmoellerdev: resolved. we try a separate module and if that doesnt work we go for package inclusion

bindings/go/blob/go.mod
Skarlso: So I tried to checkout main go.mod and sum file, and then run go mod tidy and simply this is what it comes up with. If I try to manually merge the groups it just reverts my change. ☹️

jakobmoellerdev: I just checked it out myself moved it together and ran go mod tidy, this didnt result in any change :) Pushed the commit to your branch

PR #480, opened 2025-08-01
chore: move ocm-k8s-toolkit to open-component-model monorepo

jakobmoellerdev:

This moves the OCM K8s toolkit to the new OCM monorepo

This is a large item so it will take some time until it is fully migrated.

I migrated to taskfiles as well

fix https://github.com/open-component-model/ocm-k8s-toolkit/issues/235
fix https://github.com/open-component-model/ocm-k8s-toolkit/issues/200


Skarlso (dismissed):

Skarlso (approved):

Skarlso:
Indeed. :D

PR #500, opened 2025-08-05
feat(dag)!: convert dag maps to sync maps for concurrent use

fabianburth:
During development of https://github.com/open-component-model/ocm-project/issues/563, we decided to implement a DAG renderer. This renderer is implemented based on a render loop.

We discussed multiple different approaches to allow to render loop to render the graph concurrently to the discovery without data races:
- locking with sync.RWMutex(quite hard to read and maintain, pollutes the algorithm code)
- observer pattern (works and is a clean and efficient solution, but quite hard to maintain and debug)
- sync.Maps (rather easy to implement and maintain)

Thus, we decided to go with sync.Maps. To allow all our DAG to theoretically be rendered with that DAG renderer, we want to change the general used dag package DAG to such a sync.Map graph.

This will require follow up changes in other packages (such as the credential package). 

Contributes to https://github.com/open-component-model/ocm-project/issues/563


jakobmoellerdev (changes requested):
Not entirely convinced of the representation implementation and replacing the package entirely. maybe we offer this side by side? not sure.

also, what happens if the DAG is called with Discover concurrently? I feel like things will go crazy because of conflicting attributes.

jakobmoellerdev (approved):

PR #511, opened 2025-08-06
feat(cli): add tree renderer

fabianburth:
This PR introduces:
- a render loop for live display on command line
- a tree renderer for rendering a DAG with a single root as shown below

```
── A (completed)
   ├─ B (completed)
   │  ╰─ C (completed)
   ╰─ D (completed)
```

Contributes to https://github.com/open-component-model/ocm-project/issues/564

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ env:
   # Check Only Changed Modules
   # If true, only changed modules will be analyzed.
   # If false, all discovered modules will be analyzed.
-  check_only_changed: ${{ github.event_name == 'pull_request' }}
+  check_only_changed: ${{ github.event_name == 'pull_request_target' }}
   REF: ${{ github.event_name == 'pull_request_target' && github.event.pull_request.head.ref || github.ref }}
   REPO: ${{ github.event_name == 'pull_request_target' && github.event.pull_request.head.repo.full_name || github.repository }}
 


cli/internal/renderer/graph/tree/renderer.go
```diff
@@ -0,0 +1,124 @@
...
+	"fmt"
+	"io"
+
+	"github.com/jedib0t/go-pretty/v6/list"
+
+	syncdag "ocm.software/open-component-model/bindings/go/dag/sync"
+	"ocm.software/open-component-model/cli/internal/renderer/graph"
+)
+
+type RendererOptions[T cmp.Ordered] struct {
```
jakobmoellerdev, line 15: nit: please move options to separate `renderer_options.go` and add documentation on when and how to use them

cli/internal/renderer/graph/tree/renderer.go
```diff
@@ -0,0 +1,124 @@
...
+// Renderer renders a tree structure from a DirectedAcyclicGraph.
+type Renderer[T cmp.Ordered] struct {
+	// The listWriter is used to write the tree structure. It holds manages
+	// the indentation and style of the output.
+	listWriter list.Writer
+	// The vertexSerializer is a function that serializes a vertex to a string.
+	// It MUST perform READ-ONLY access to the vertex and its attributes.
+	vertexSerializer VertexSerializer[T]
+	// The root ID of the tree to render.
+	root T
```
jakobmoellerdev, line 53: is the root really static? if one can repeatedly call render with different roots, I think it would be better to move root out and into Render

cli/internal/renderer/graph/utils.go
```diff
@@ -0,0 +1,55 @@
...
+	})
+
+	slices.SortFunc(neighbors, func(a, b T) int {
+		return compareByOrderIndex(ctx, vertex, a, b)
+	})
+
+	return neighbors
+}
+
+func compareByOrderIndex[T cmp.Ordered](ctx context.Context, vertex *syncdag.Vertex[T], a, b T) int {
```
jakobmoellerdev, line 32: nit: this probably needs a method doc on what it actually achieves

cli/internal/renderer/graph/utils.go
```diff
@@ -0,0 +1,55 @@
...
+	orderA := getOrderIndex(ctx, vertex, a)
+	orderB := getOrderIndex(ctx, vertex, b)
+
+	// If both edges have order indices, compare them.
+	if orderA != nil && orderB != nil {
+		return *orderA - *orderB
```
jakobmoellerdev, line 38: if both orderA and orderB are set, isn't it wrong to subtract? instead of calling

```suggestion
            return cmp.Compare(*orderA, *orderB)
```

or even

```suggestion
            a, b = *orderA, *orderB
```

PR #512, opened 2025-08-06
fix: support `compress` option for `file` and `dir` inputs

ikhandamirov:

This PR makes it possible to create resource from `file`and `dir` with `Compress` option set to `true`. In such cases, when creating respective OCI layer, the size and the digest of the blob are now calculated dynamically.

The use of the fix is not limited to `file`and `dir` inputs, but applicable to all cases, when the size or the digest of the blob are not known in advance.

Fixes open-component-model/ocm-project#566


bindings/go/oci/blob/artifact_blob.go
```diff
@@ -193,6 +195,60 @@ func (r *ArtifactBlob) OCIDescriptor() ociImageSpecV1.Descriptor {
 	}
 }
 
+// HasSizeAndDigest checks if the blob has both a pre-calculated size and a pre-calculated  digest.
+func (r *ArtifactBlob) HasSizeAndDigest() bool {
```
jakobmoellerdev, line 199: I think this is not reliable enough. Just because the digest and size are not available, doesn't mean that the artifact blob doesn't expose a digest and/or size, just if it is available as a precomputed variable.

so I think it may be needed to attempt to fetch the digest and or size and if that fails, to compute them with the buffer.

jakobmoellerdev: I think actually you want something like `ExposesSize` and that can be checked by asserting that the blob implements an interface called [`DigestAware` for digests](https://github.com/open-component-model/open-component-model/blob/main/bindings/go/blob/blob.go#L67-L73) and [`SizeAware`](https://github.com/open-component-model/open-component-model/blob/main/bindings/go/blob/blob.go#L42-L49) for sizes.

bindings/go/oci/blob/artifact_blob.go
```diff
@@ -193,6 +195,60 @@ func (r *ArtifactBlob) OCIDescriptor() ociImageSpecV1.Descriptor {
...
+	return r.HasPrecalculatedSize() && r.HasPrecalculatedDigest()
+}
+
+// NewArtifactWithBufferedBlob creates a new ArtifactBlob with an in-memory buffered blob.
+// It can be used to covert ArtifactBlob with unknown size or digest to a new instance where those fields a set.
+func (r *ArtifactBlob) NewArtifactWithBufferedBlob() (result *ArtifactBlob, err error) {
```
jakobmoellerdev, line 205:
```suggestion
// BufferEagerly creates a new ArtifactBlob backed by an in-memory buffered blob.
// It can be used to covert ArtifactBlob's with unknown size or digest to a new instance where those fields are available for use with OCI Artifact Descriptors
func (r *ArtifactBlob) BufferEagerly() (result *ArtifactBlob, err error) {
```

bindings/go/oci/blob/artifact_blob.go
```diff
@@ -193,6 +195,60 @@ func (r *ArtifactBlob) OCIDescriptor() ociImageSpecV1.Descriptor {
...
+
+	// Create buffered in-memory blob and load it.
+	inMemoryBlob := inmemory.New(reader, inmemory.WithMediaType(mediaType))
+	err = inMemoryBlob.Load()
+	if err != nil {
+		return nil, fmt.Errorf("failed to load in-memory blob: %w", err)
+	}
+
+	// Create a new instance of the descriptor.
+	var newArtifact descriptor.Artifact
```
jakobmoellerdev, line 226: I really don't like this, can we make this method available on Artifact instead?

bindings/go/oci/internal/pack/pack.go
```diff
@@ -124,10 +124,20 @@ type OCILayerConvertableBlob interface {
 }
 
 // NewBlobOCILayer creates a new OCI layer descriptor for a OCILayerConvertableBlob.
-func NewBlobOCILayer(b *ociblob.ArtifactBlob, opts ResourceBlobOCILayerOptions) (ociImageSpecV1.Descriptor, error) {
+// It ensures that the involved ArtifactBlob has the size and the digest set. If the incoming ArtifactBlob does not have a size or a digest,
+// a new instance of ArtifactBlob having both fields set is created and returned. Otherwise, the initial ArtifactBlob is returned.
+func NewBlobOCILayer(b *ociblob.ArtifactBlob, opts ResourceBlobOCILayerOptions) (*ociblob.ArtifactBlob, ociImageSpecV1.Descriptor, error) {
```
jakobmoellerdev, line 129:
```suggestion
// It ensures that the involved ArtifactBlob has the size and the digest set so it can be uploaded to OCI registries which need these fields for API interactions. If the incoming ArtifactBlob does not have a size or a digest,
// a new instance of ArtifactBlob having both fields set is created and returned. Otherwise, the initial ArtifactBlob is returned.
func PrepareArtifactBlobForOCI(b *ociblob.ArtifactBlob, opts ResourceBlobOCILayerOptions) (*ociblob.ArtifactBlob, ociImageSpecV1.Descriptor, error) {
```

bindings/go/oci/blob/artifact_blob.go
ikhandamirov: Reworked in pair programming. New function name is `Buffer()`.

bindings/go/oci/internal/pack/pack.go
ikhandamirov: Done.

PR #519, opened 2025-08-07
chore: removes deferrlint

jakobmoellerdev:

deferrlint is too early stage to be in our repo right now, lets remove it


Skarlso (approved):
👍  I'm sorry 😿

PR #535, opened 2025-08-11
feat: adjust filename based on oci artifact media type

Skarlso:

This is part of https://github.com/open-component-model/ocm-project/issues/553 effort. As documented in the code, unfortunately, packaging and creating files for HELM MUST follow a certain requirement detailed here: https://helm.sh/docs/topics/charts/#charts-and-versioning.

We could do several other things like we initially planned with CEL integration, but this is an easier path for now to go.

### DIFF
--- a/bindings/go/oci/transformer/extract.go
+++ b/bindings/go/oci/transformer/extract.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"log/slog"
 	"slices"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


jakobmoellerdev (commented):
generally looks good, just some tweaks to make it easier to understand!

jakobmoellerdev (approved):
oh how I hate that we have to do this. Would be cool if you could link this PR to the appropriate issue and add some description

Skarlso:
Now that this is done, https://github.com/open-component-model/open-component-model/pull/530 can proceed.

PR #541, opened 2025-08-12
feat: support stateful, serialized io.Reader's in direct blobs

jakobmoellerdev:

when we don't have io.ReaderAt backed implementations of blobs (for example when working with io.Pipe which is not resettable), then we need to base our implementation solely on io.Reader. However the blobs ReadCloser() methods can in theory be concurrently called, so we need a serializing blob that is able to protect from bad reads.

This fixes not being able to use pure io.Readers and allows direct blobs to receive an io.Reader natively, switching to serialized access under the hood if needed.

Needed to actually migrate input methods like dir (https://github.com/open-component-model/ocm-project/issues/580)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -265,15 +265,15 @@ jobs:
             .github/workflows/ci.yml
             ${{ matrix.project }}
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@76621b61decf072c1cee8dd1ce2d2a82d33c17ed # v3
+        uses: github/codeql-action/init@df559355d593797519d70b90fc8edd5db049e7a2 # v3
         with:
           languages: go
           queries: security-extended
       - name: Autobuild
-        uses: github/codeql-action/autobuild@76621b61decf072c1cee8dd1ce2d2a82d33c17ed # v3
+        uses: github/codeql-action/autobuild@df559355d593797519d70b90fc8edd5db049e7a2 # v3
 
       - name: Perform CodeQL Analysis
-        uses: github/codeql-action/analyze@76621b61decf072c1cee8dd1ce2d2a82d33c17ed # v3
+        uses: github/codeql-action/analyze@df559355d593797519d70b90fc8edd5db049e7a2 # v3
         with:
           category: "/language:go"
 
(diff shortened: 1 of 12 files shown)


bindings/go/blob/direct/serial/blob.go
```diff
@@ -0,0 +1,102 @@
...
+			_, err := seeker.Seek(0, io.SeekStart)
+			return err
+		}
+		return nil
+	})
+	l.close = func() error {
+		if !acquired.Load() {
+			// Release the token if we did acquire it before.
+			l.parent.token <- struct{}{}
+		}
```
Skarlso, line 72: Am I stupid or is this comment wrong? :D

bindings/go/blob/direct/blob_test.go
```diff
@@ -183,6 +184,49 @@ func TestCompatibility(t *testing.T) {
...
+			waitForParallel := make(chan struct{}, 1)
+			go func() {
+				defer func() {
+					waitForParallel <- struct{}{}
+				}()
+				// Wait for the reader to be created
+				rc, err := b.ReadCloser()
+				require.NoError(t, err)
+				require.NotNil(t, rc)
+				result, err := io.ReadAll(br)
```
Skarlso, line 218: Shouldn't this be using `rc` instead of `br`?

bindings/go/blob/direct/serial/blob.go
Skarlso: Yeah, but aren't you supposed to `if acquired.Load()` instead of `if !acquired.Load()`? Since `// acquired is true if the token was acquired.`

jakobmoellerdev: you are completely right >.<. thanks for catching this!

bindings/go/blob/direct/blob_test.go
jakobmoellerdev: good catch!

fabianburth (dismissed):

Skarlso (dismissed):
This _looks_ okay now. :D

Skarlso (approved):

fabianburth (approved):

PR #542, opened 2025-08-12
fix: add another conformance for legacy JSON normalization (v3)

jakobmoellerdev:

Implemented a conformance testing suite for validating legacy JSON normalization (v3). Included test datasets and adjusted handling of provider parsing logic to handle both string and byte inputs to not break conformance

fixed a broken case in which the wrong normalisation was computed. If the provider only had a single string value, it was incorrectly normalised to the string instead of a map with the "name" key

this paves the way to get exactly the same output as with old OCM CLI for the newest normalisation algorithm.

relates to https://github.com/open-component-model/ocm-project/issues/579


bindings/go/descriptor/normalisation/json/v4alpha1/normalisation.go
```diff
@@ -151,13 +151,23 @@ func DefaultComponent(d *v2.Descriptor) {
...
+		}
+	case string:
+		var provider map[string]any
+		if err := json.Unmarshal([]byte(v), &provider); err == nil {
+			return provider
+		}
+	}
+	return map[string]any{
+		"name": v,
 	}
```
fabianburth, line 170:
```suggestion
func ProviderAsMap(v any) any {
    var provider map[string]any
	switch v := v.(type) {
	case []byte:
		if err := json.Unmarshal(v, &provider); err == nil {
			return provider
		}
	case string:
		if err := json.Unmarshal([]byte(v), &provider); err == nil {
			return provider
		}
	}
	return map[string]any{
		"name": v,
	}
```

Does the caller expect the return type to be `any` here?

PR #549, opened 2025-08-13
feat(565): update constructor to support working dir

matthiasbruns:
Cherry picking constructor changes from https://github.com/open-component-model/open-component-model/pull/540

Changes (high level):

- Added WorkingDirectory handling to the constructor bindings


bindings/go/constructor/construct_test.go
```diff
@@ -177,6 +177,7 @@ components:
 	mockRepo := newMockTargetRepository()
 
 	// Create the constructor with our mocks
+	// TODO: # 565 Add WorkingDirectory tests
```
matthiasbruns, line 180: on it

bindings/go/constructor/construct_resource_test.go
```diff
@@ -789,3 +789,5 @@ components:
 	assert.Len(t, mockRepo.addedLocalResources, 0)
 	assert.Len(t, mockRepo.addedVersions, 1)
 }
+
+// TODO: # 565 Add WorkingDirectory tests
```
matthiasbruns, line 793: on it

matthiasbruns:
> While true, the constructor itself does not actually control the context of the input methods because it doesnt own the plugins. That responsibility lies within the CLI. At least thats what my little brain is trying to make sense of right now, could be wrong >.>

I'll trust your little brain on that - pr is closed

PR #557, opened 2025-08-13
feat(565): update input to support working dir

matthiasbruns:
Cherry picking constructor changes from https://github.com/open-component-model/open-component-model/pull/540

Changes (high level):

- Added WorkingDirectory handling to the input bindings


bindings/go/input/dir/method.go
```diff
@@ -106,3 +116,31 @@ func (i *InputMethod) ProcessSource(ctx context.Context, src *constructorruntime
 		ProcessedBlobData: fileBlob,
 	}, nil
 }
+
+// ensureWdOrPath ensures that the dirPath is absolute.
+// If the path is relative, it prepends the working directory to it.
+// If the working directory is empty, it does nothing.
+func ensureWdOrPath(dirPath, workingDirectory string) string {
```
jakobmoellerdev, line 123: Im wondering about this. Can there be a case in which the dirPath is relative but outside of the workingDirectory that is set?

Example:

workingDirectory: "my/working/dir"
dirPath: "../../../path"

how would you resolve this now? shouldn't you create a root from "my/working/dir" and then ensure that you can open dirPath from within that root? That way you can avoid path escalation

matthiasbruns: > Im wondering about this. Can there be a case in which the dirPath is relative but outside of the workingDirectory that is set?

Is this a problem? Why would we want to sandbox the users directories?

jakobmoellerdev: because this can lead to attack vectors if not explicitly enabled. IMO we should not default to a potentially insecure "escape-allow". If we do allow this imo it should be opt-in, not opt-out

bindings/go/input/dir/method.go
```diff
@@ -106,3 +125,49 @@ func (i *InputMethod) ProcessSource(ctx context.Context, src *constructorruntime
...
 }
+
+// ensureWdOrPath ensures that the dirPath is absolute.
+// If the path is relative, it prepends the working directory to it.
+// If the working directory is empty, it does nothing.
+// It also ensures that the resulting path does not escape the working directory.
+func ensureWdOrPath(path, workingDirectory string) (string, error) {
+	if filepath.IsAbs(path) {
+		return path, nil
+	}
```
fabianburth, line 136: Even if the filepath is absolute, wouldn't we want to make sure, it's not outside the working directory?

matthiasbruns:
> Would it make sense to put the `ensureWdOrPath` function as a public function somewhere, so it can be reused?

I askes @jakobmoellerdev  where this place would be - maybe you can give me a pointer :D

matthiasbruns:
> for reuse of your function consider something like
> 
> ```
> func GetBlobInWorkingDirectory(path, workingDir string) (*Blob, error) {
> 	path, err := ensureWdOrPath(path, workingDir)
> 	if err != nil {
> 		return nil, err
> 	}
> 	return GetBlobFromOSPath(path)
> }
> ```
> 
> in the blob/filesystem bindings

https://github.com/open-component-model/open-component-model/pull/557

After pr #572 557 is merged, I will update this one

bindings/go/input/dir/method.go
```diff
@@ -65,6 +87,7 @@ func (i *InputMethod) ProcessResource(ctx context.Context, resource *constructor
 		return nil, fmt.Errorf("error converting resource input spec: %w", err)
 	}
 
+	// TODO: update call from https://github.com/open-component-model/open-component-model/pull/572
```
matthiasbruns, line 90: while https://github.com/open-component-model/open-component-model/pull/572/files#diff-2d1251bfe35486a20c64d0f11feafb6da27931958411c17bf86215f222e0662fR85 works for files, it might not be applicable for the dir handling, since this happens in the input module

bindings/go/input/dir/method.go
```diff
@@ -97,6 +120,7 @@ func (i *InputMethod) ProcessSource(ctx context.Context, src *constructorruntime
...
```
matthiasbruns, line 123: while https://github.com/open-component-model/open-component-model/pull/572/files#diff-2d1251bfe35486a20c64d0f11feafb6da27931958411c17bf86215f222e0662fR85 works for files, it might not be applicable for the dir handling, since this happens in the input module

bindings/go/input/dir/blob.go
```diff
@@ -31,11 +32,24 @@ var ErrEmptyPath = errors.New("dir path must not be empty")
...
+	if _, err := filesystem.EnsurePathInWorkingDirectory(dir.Path, workingDirectory); err != nil {
+		if strings.Contains(err.Error(), "empty path") {
+			return nil, ErrEmptyPath
+		}
+		if strings.Contains(err.Error(), "no such file or directory") {
+			return nil, fmt.Errorf("path does not exist in working directory %q: %w", workingDirectory, err)
+		}
+		return nil, fmt.Errorf("error ensuring path %q in working directory %q: %w", dir.Path, workingDirectory, err)
+	}
+
```
fabianburth, line 52:
```suggestion
	if _, err := filesystem.EnsurePathInWorkingDirectory(dir.Path, workingDirectory); err != nil {
		return nil, fmt.Errorf("error ensuring path %q in working directory %q: %w", dir.Path, workingDirectory, err)
	}

```

We shouldn't parse the error string. Instead, we should use `errors.Is`.
Although, I don't think this is necessary here. I think you can just remove those error checks and rely on the error message from `EnsurePathInWorkingDirectory` to be descriptive enough.

bindings/go/input/dir/blob_test.go
```diff
@@ -127,14 +129,16 @@ func TestGetV1DirBlob_Reproducibility(t *testing.T) {
...
 			require.NoError(t, err)
 			require.NotNil(t, b)
 
 			// Read the tar data after modification.
 			readerAfter, err := b.ReadCloser()
 			require.NoError(t, err)
-			defer readerAfter.Close()
+			defer func(readerBefore io.ReadCloser) {
+				require.NoError(t, readerBefore.Close())
+			}(readerAfter)
```
fabianburth, line 141:
```suggestion
			defer func() {
				require.NoError(t, readerAfter.Close())
			}()
```

Parameter name seems confusing. But generally, I think you can just use a closure here.

bindings/go/input/dir/blob.go
```diff
@@ -31,11 +32,24 @@ var ErrEmptyPath = errors.New("dir path must not be empty")
...
-//  2. Reads the directory contents using an instance of the virtual FileSystem
-//  3. Packs the directory contents into a tar archive
-//  4. Applies different configuration options of the v1.Dir specification
-func GetV1DirBlob(ctx context.Context, dir v1.Dir) (blob.ReadOnlyBlob, error) {
+//  2. Ensures that the directory path is within the working directory
+//     (this is to prevent directory traversal attacks and ensure security)
+//  3. Reads the directory contents using an instance of the virtual FileSystem
+//  4. Packs the directory contents into a tar archive
+//  5. Applies different configuration options of the v1.Dir specification
+func GetV1DirBlob(ctx context.Context, dir v1.Dir, workingDirectory string) (blob.ReadOnlyBlob, error) {
```
fabianburth, line 40: Would it make sense to make `workingDirectory` optional and default working directory to the working directory of the process?

bindings/go/input/dir/blob_test.go
```diff
@@ -329,7 +333,9 @@ func TestGetV1DirBlob_Standard_Cases(t *testing.T) {
 			// Test reading data.
 			reader, err := b.ReadCloser()
 			require.NoError(t, err)
-			defer reader.Close()
+			defer func(readerBefore io.ReadCloser) {
+				require.NoError(t, readerBefore.Close())
+			}(reader)
```
fabianburth, line 338: See comment above.

bindings/go/input/dir/blob_test.go
```diff
@@ -338,7 +344,9 @@ func TestGetV1DirBlob_Standard_Cases(t *testing.T) {
 				// Decompress gzipped data.
 				gzReader, err := gzip.NewReader(bytes.NewReader(data))
 				require.NoError(t, err)
-				defer gzReader.Close()
+				defer func(readerBefore io.ReadCloser) {
+					require.NoError(t, readerBefore.Close())
+				}(gzReader)
```
fabianburth, line 349: See comment above.

fabianburth (changes requested):
I know this was like this before you touched it already, but since the tests make a lot of asserts (`require.NoError(t,...)`), we might want to think about introducing an `r := require.New(t)` and to be able to subsequently call `r.NoError(...)` without having to pass the testing environment.
You can check other tests for reference.

bindings/go/input/dir/blob.go
matthiasbruns: We can either handle this here or in the caller (cli). I would do this in the caller to prevent confusion later on

bindings/go/input/dir/blob.go
fabianburth: I think we should move it then.

PR #572, opened 2025-08-15
feat: only allow paths to be in working directory

matthiasbruns:
Based on the comment https://github.com/open-component-model/open-component-model/pull/557#issuecomment-3189225662

Changes (high level):

- Added WorkingDirectory handling to the input bindings

### DIFF
--- a/bindings/go/blob/filesystem/file.go
+++ b/bindings/go/blob/filesystem/file.go
@@ -91,7 +91,7 @@ func GetBlobFromOSPath(path string) (*Blob, error) {
 func GetBlobInWorkingDirectory(path, workingDir string) (*Blob, error) {
 	path, err := EnsurePathInWorkingDirectory(path, workingDir)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("failed to ensure path in working directory: %w", err)
 	}
 	return GetBlobFromOSPath(path)
 }
@@ -103,17 +103,14 @@ func GetBlobInWorkingDirectory(path, workingDir string) (*Blob, error) {
 func EnsurePathInWorkingDirectory(path, workingDirectory string) (_ string, err error) {
 	if filepath.IsAbs(path) {
 		if path, err = filepath.Rel(workingDirectory, path); err != nil {
-			return "", err
+			return "", fmt.Errorf("failed to create relative path for %q based on working directory %q: %w", path, workingDirectory, err)
 		}
 	}
 
-	fd, err := os.OpenInRoot(workingDirectory, path)
+	_, err = os.OpenInRoot(workingDirectory, path)
 	if err != nil {
-		return "", err
+		return "", fmt.Errorf("failed to open path %q in root %q: %w", path, workingDirectory, err)
 	}
-	defer func() {
-		err = errors.Join(err, fd.Close())
-	}()
 
 	return filepath.Join(workingDirectory, path), nil
 }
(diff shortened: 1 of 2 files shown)


bindings/go/blob/filesystem/file_test.go
```diff
@@ -45,7 +45,9 @@ func TestGetLocalBlobFromLocation(t *testing.T) {
 
 	reader, err := b.ReadCloser()
 	r.NoError(err)
-	defer reader.Close()
+	defer func(reader io.ReadCloser) {
+		_ = reader.Close()
+	}(reader)
```
fabianburth, line 50:
```suggestion
	defer func() {
		r.NoError(reader.Close())
	}()
```

bindings/go/blob/filesystem/file_unix_test.go
```diff
@@ -62,3 +62,76 @@ func TestCopyBlobToOSPath_NamedPipe_Blocking(t *testing.T) {
 		r.Equal(testData, data)
 	}
 }
+
+func Test_GetBlobInWorkingDirectory(t *testing.T) {
```
fabianburth, line 66: I think we should rather unit test the `EnsurePathInWorkingDirectory` function here.

fabianburth (changes requested):
Looks good now! 
I'd like to propose two more tiny changes (which are arguably kind of nits), then we are ready to merge.

bindings/go/blob/filesystem/file.go
```diff
@@ -81,3 +81,39 @@ func GetBlobFromOSPath(path string) (*Blob, error) {
 
 	return data, nil
 }
+
+// GetBlobInWorkingDirectory returns a blob that reads from the operating system file system,
+// ensuring that the path is resolved against the specified working directory.
+// It uses EnsurePathInWorkingDirectory to ensure the path is relative to the working directory.
+// The blob is created using the GetBlobFromOSPath function.
+// If the path is not absolute, it will be resolved against the working directory.
+// If the path is invalid or cannot be resolved, it returns an error.
```
fabianburth, line 90:
```suggestion
// GetBlobInWorkingDirectory returns a blob that reads from the operating system file system,
// ensuring that the path is resolved against the specified working directory.
// It uses EnsurePathInWorkingDirectory to ensure the path is within the working directory.
// The blob is created using the GetBlobFromOSPath function.
// If the path is not absolute, it will be resolved against the working directory.
// If the path is invalid or cannot be resolved, it returns an error.
```

bindings/go/blob/filesystem/file_unix_test.go
matthiasbruns: we are testing this indirectly, but sure, I can also write tests for this

fabianburth: I know, but we are testing the `EnsurePathInWorkingDirectory` behaviour here. If we added 5 other functions leveraging `EnsurePathInWorkingDirectory`, we wouldn't want to test this behaviour for each of those.

matthiasbruns: already done - tests are added :)

matthiasbruns: I'll keep it - it's another public method, so why not keep them - we might also change the impl there which would have the same result if we had multiple callers

fabianburth (approved):

Issue #564, opened 2025-08-14
Blob and location needs to be media type aware


Skarlso:
**What happened**:

This is a missing feature on the blob and the Location type in the Plugin.

**What you expected to happen**:

Right now, if upload with helm is done, it's an octet-stream because it's not setting the MediaType on the blob during upload.

**How to reproduce it (as minimally and precisely as possible)**:

Upload a Component Version with `ocm add cv` or via Library to the Remote Repository with the `helm` plugin. Then, do `ocm get cv -oyaml` and observe the `localBlob` type instead of being uploaded as oci artifact

**Anything else we need to know**:

This needs to be adjusted and pull through the environment. If media type is set, use .(MediaTypeAware) and use the media type to pass in to the Location in the Helm input method. Once it's done, the Blob upload needs to apply it correctly so we have a helm type. The right MediaType can be fetched from the blob then.

**Environment**:

jakobmoellerdev:
Once we have the Blob Wrapper in place from #573 location needs to be adjusted to respect the media type in the plugin manager.

PR #573, opened 2025-08-15
feat: add mediaType to the blob and location

Skarlso:

I know, I know. I will release the modules separately, but instead of a gazillion prs and discussions on separate pages, I wanted to discuss the changes together because they are small enough.

Fixed https://github.com/open-component-model/open-component-model/issues/564

Part Of https://github.com/open-component-model/ocm-project/issues/553 and needed for https://github.com/open-component-model/open-component-model/pull/530


bindings/go/helm/cmd/main.go
```diff
@@ -173,10 +173,19 @@ func processHelmResource(ctx context.Context, request *v1.ProcessResourceInputRe
 		return nil, fmt.Errorf("error copying blob data: %w", err)
 	}
 
+	// Demonstrate MediaTypeAware usage
+	mediaType := "application/x-gzip" // Default media type for Helm charts
```
Skarlso, line 177: Not entirely sure what's the default? Octet stream?

bindings/go/blob/filesystem/file_blob.go
```diff
@@ -20,12 +20,18 @@ type Blob struct {
 	fileSystem fs.FS
 	// path is the original path to the blob.
 	path string
+	// mediaType is the media type of the blob.
+	mediaType string
```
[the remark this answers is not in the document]
Skarlso: Added.

bindings/go/plugin/manager/registries/blobs/blobs.go
```diff
@@ -10,10 +10,25 @@ import (
...
+		return nil, err
+	}
+
+	if location.MediaType != "" {
+		if mtOverrideable, ok := b.(blob.MediaTypeOverrideable); ok {
+			mtOverrideable.SetMediaType(location.MediaType)
+		}
+	}
+
+	return b, nil
```
Skarlso, line 33: This will be replaced with 
```go
// CreateBlobData creates a blob based on the location.
func CreateBlobData(location types.Location) (blob.ReadOnlyBlob, error) {
	var b blob.Blob
	var err error

	switch location.LocationType {
	case types.LocationTypeLocalFile:
		b, err = filesystem.GetBlobFromOSPath(location.Value)
	default:
		return nil, fmt.Errorf("unsupported location type: %s", location.LocationType)
	}
	if err != nil {
		return nil, err
	}

	w := direct.New(direct.NewReadCloserWrapper(b), direct.WithMediaType(location.MediaType))
	return w, nil
}
```

Once https://github.com/open-component-model/open-component-model/pull/593 is merged.

Skarlso:
This is now done as all interim parts have been merged. 🎉

PR #574, opened 2025-08-15
feat(cli): json/yaml graph renderer

fabianburth:

Contributes to https://github.com/open-component-model/ocm-project/issues/563


cli/internal/render/cmd.go
```diff
@@ -0,0 +1,19 @@
+package render
+
+import "strings"
+
+const (
+	CursorUp  = "\x1b[A" // Move cursor up one line
+	EraseLine = "\x1b[K" // Erase the current line
+)
+
+func EraseNLines(n int) string {
```
jakobmoellerdev, line 10: nice utility, but needs a comment that this doesnt work everywhere if reused

cli/internal/render/graph/list/renderer.go
```diff
@@ -0,0 +1,212 @@
...
+//	── A
+//	   ├─ B
+//	   │  ╰─ C
+//	   ╰─ D
+//
+// Each letter corresponds to a vertex in the DirectedAcyclicGraph. The concrete
+// representation of the vertex is defined by the VertexMarshalizer.
+type Renderer[T cmp.Ordered] struct {
+	// The objects is a slice of objects that will be rendered.
+	objects []any
```
jakobmoellerdev, line 40: wouldn't that be typed as well to a different generic parameter?

cli/internal/render/graph/list/renderer.go
```diff
@@ -0,0 +1,212 @@
...
+// serializable object from a vertex.
+type VertexMarshalizer[T cmp.Ordered] interface {
+	Marshalize(*syncdag.Vertex[T]) (any, error)
+}
+
+// VertexMarshalizerFunc is a function type that implements the VertexMarshalizer
+// interface.
+type VertexMarshalizerFunc[T cmp.Ordered] func(*syncdag.Vertex[T]) (any, error)
+
+// Marshalize implements the VertexMarshalizer interface for VertexMarshalizerFunc.
```
jakobmoellerdev, line 68: why is this called Marshalizer and not Marshaller like everything else in go?

cli/internal/render/graph/list/renderer.go
```diff
@@ -0,0 +1,212 @@
...
+
+// renderObjects renders the objects based on the specified output format.
+func (t *Renderer[T]) renderObjects(writer io.Writer) error {
+	var (
+		err  error
+		data []byte
+	)
+	switch t.outputFormat {
+	case render.OutputFormatJSON:
+		data, err = t.encodeObjectsAsJSON()
```
jakobmoellerdev, line 158: nit: how about passing the writer directly to these methods to write into them? that way you avoid passing the byte slice everywhere

cli/internal/render/graph/list/renderer_options.go
```diff
@@ -0,0 +1,38 @@
...
+	"cmp"
+
+	syncdag "ocm.software/open-component-model/bindings/go/dag/sync"
+	"ocm.software/open-component-model/cli/internal/render"
+)
+
+type RendererOptions[T cmp.Ordered] struct {
+	// VertexMarshalizer converts a vertex to an object of type U. U is expected
+	// to be a serializable type (e.g., a struct or map).
+	// The marshalizer MUST perform READ-ONLY access to the vertex and its
```
jakobmoellerdev, line 13: can't we assure that somehow?

jakobmoellerdev (commented):
Looks neat!

cli/internal/render/graph/list/renderer.go
fabianburth: I did have that initially. But it makes the code significantly more cumbersome and exhausting to read. Besides, I didn't really see a downside to this flexibility, especially since the user fills this himself through the `VertexMarshalizer` and may do arbitrary type checks in there.

But feel free to correct me, if I have an oversight here.

cli/internal/render/graph/list/renderer.go
fabianburth: Yeah, I'm not 100% satisfied with the name either. 
- Didn't call it `Marshaller` because it does not actually marshal but only provide a "marshallable" object. 
- Didn't call it serializer because thought it might be confusing because it has a different interface than in the tree renderer.

You think we should go with `Marshaller` anyway? (Or alternative suggestions?)

cli/internal/render/graph/list/renderer.go
fabianburth: Sure, can do that :)

cli/internal/render/graph/list/renderer.go
jakobmoellerdev: IMO you should still go with `Marshaller` because in mind mind, you "marshal" from a Vertex into an object, that you then Marshal again (i.e. to JSON or text). WDYT?

cli/internal/render/graph/list/renderer_json_test.go
```diff
@@ -0,0 +1,323 @@
...
+			}
+			return map[string]any{
+				"id":    v.ID,
+				"state": traversalState.String(),
+			}, nil
+		})
+		renderer := New(d, "A", WithOutputFormat[string](render.OutputFormatJSON), WithVertexMarshalizer(marshalizer))
+		waitFunc := render.RunRenderLoop(ctx, renderer, render.WithRefreshRate(10*time.Millisecond), render.WithRenderOptions(render.WithWriter(writer)))
+
+		time.Sleep(30 * time.Millisecond)
```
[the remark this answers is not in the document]
fabianburth: I think, here - we would be rather safe without the `synctest.Wait()`, because:
- we do not do any external API calls (so the time the render loop runs is rather constant)
- the render loop is idempotent

But based on my current understanding, this is the correct approach. But please correct me if I'm wrong.
The `time.Sleep()` ensures that the render loop has had to run an iteration. The `sync.Wait()` waits for the render loop to block again. 
So, in this code, to complete its current run and start blocking as it is waiting to receive values on either `ticker.C` or `ctx.Done` channel:
```
	for {
		select {
		case <-ctx.Done():
			return ctx.Err()
		case <-ticker.C:
			if err := refreshOutput(ctx, renderer, renderState); err != nil {
				return err
			}
		}
	}
```

So, this fails - or is at least racy. But if we comment in the `synctest.Wait()`, this test will deterministically succeed every time. 
```
		renderer := New(d, "A", WithOutputFormat[string](render.OutputFormatNDJSON), WithVertexMarshalizer(marshalizer))
		waitFunc := render.RunRenderLoop(ctx, renderer, render.WithRefreshRate(10*time.Millisecond), render.WithRenderOptions(render.WithWriter(writer)))

		time.Sleep(10 * time.Millisecond)
		// synctest.Wait()
		output := buf.String()
		expected := `{"id":"A","state":"discovering"}
`
		r.Equal(expected, output)
		buf.Reset()
```

cli/internal/render/graph/list/renderer_options.go
fabianburth: We could introduce a Vertex interface that only give read only access. Since its an internal API, I thought this might just overcomplicate things.

jakobmoellerdev: maybe add a todo and call it a day. fine for now for me :)

jakobmoellerdev (approved):

PR #575, opened 2025-08-16
Feat/565 relative file pointers blob err

matthiasbruns:
Based on this comment https://github.com/open-component-model/open-component-model/pull/557#discussion_r2280203324

Add error types as output to check for in the callers in blob/file 

**Changes (high level):**

Adds the following error types and tests in `EnsurePathInWorkingDirectory`

- var ErrFileNotFound = errors.New("file not found")
- var ErrPermissionDenied = errors.New("permission denied")
- var ErrInvalidPath = errors.New("invalid path")


bindings/go/blob/filesystem/file.go
```diff
@@ -13,6 +14,12 @@ import (
 
 const DefaultFileIOBufferSize = 1 << 20 // 1 MiB
 
+var (
```
jakobmoellerdev, line 17: This looks like reinventing the wheel of already existing FS errors:
https://pkg.go.dev/io/fs#pkg-variables

Is there a specific reason you chose to go for your own variables?
If you want to specify Error Types for the Package, then I suggest to alias existing errors instead.

bindings/go/blob/filesystem/errors.go
```diff
@@ -0,0 +1,3 @@
+package filesystem
+
+import "errors"
```
jakobmoellerdev, line 3: empty file should be removed

bindings/go/blob/filesystem/file.go
```diff
@@ -109,6 +116,19 @@ func EnsurePathInWorkingDirectory(path, workingDirectory string) (_ string, err
...
+			return "", fmt.Errorf("permission denied for path %q in directory %q: %v %w", path, workingDirectory, ErrPermissionDenied, err)
+		}
+		if errors.Is(err, os.ErrInvalid) {
+			return "", fmt.Errorf("invalid path %q in directory %q: %w %w", path, workingDirectory, ErrInvalidPath, err)
+		}
+		var pathError *fs.PathError
+		if errors.As(err, &pathError) {
+			return "", fmt.Errorf("path error for %q in directory %q: %w %w", path, workingDirectory, ErrInvalidPath, err)
+		}
 		return "", fmt.Errorf("failed to open path %q in root %q: %w", path, workingDirectory, err)
```
jakobmoellerdev, line 132: Why do we need to wrap errors here? it feels like people could just check with Errors is for existing errors?

PR #547, opened 2025-08-13
chore(docs): add ADR proposal for signing and verification of component descriptors

jakobmoellerdev:
#### What this PR does / why we need it

Enable easy interpretation of component descriptors in normalized form and provide a consistent experience for signing and verifying component descriptors.

#### Which issue(s) this PR fixes

fix https://github.com/open-component-model/ocm-project/issues/579

PR #599, opened 2025-08-21
chore(docs): add ADR proposal for signing and verification of component descriptors

matthiasbruns:
Continuation of https://github.com/open-component-model/open-component-model/pull/547

Enable easy interpretation of component descriptors in normalized form and provide a consistent experience for signing and verifying component descriptors.

fix https://github.com/open-component-model/ocm-project/issues/579


docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,358 @@
...
+- **Date:** 2025-08-27
+- **Related:** OCM signing specification and examples
+
+---
+
+## Context
+
+Component Descriptors (CDs) are signed without altering digest fields during signing. Digests for resources, component references, and the component version are produced in a dedicated step and embedded into the descriptor. Signing uses fixed canonicalization (JCS, RFC 8785) over the descriptor to compute the component-version digest and create a signature envelope. When a component version references another, the digest of the referenced descriptor is embedded in the parent’s `references[]` and contributes to the parent’s component-version digest.
+
+Some workflows require an additional **OCI Cosign** signature over the same canonical payload, optionally uploaded to an OCI registry and recorded in a transparency log (Rekor).
```
jakobmoellerdev, line 14: this is not correct. I dont know where you go the OCI registry upload from but the signature is unrelated to the storage

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,358 @@
...
+
+---
+
+## Decision
+
+### Two-step flow
+
+1. **Digest calculation** mutates the descriptor to embed required digests (resources, component references, component-version).
+2. **Signing** appends a signature envelope without computing or embedding digests. Optionally, the same payload may be **cosigned**.
```
jakobmoellerdev, line 23: cosign/sigstore is not "optional" in addition to a normal signature. rather its a completely separate signing mechanism

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,358 @@
...
+
+### Commands (verb–noun, long options only, no stdin)
+
+```bash
+# Step 1 — Digest calculation (mutates descriptor: digests only)
+ocm add digests cv <ref> --recurse --force
```
jakobmoellerdev, line 29: --recurse does what?

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,358 @@
...
+
+# Step 2 — Signing (no digest mutation)
+# Direct signing via a single --cert (+ optional password).
+# Optional OCI Cosign signing for the same canonical payload.
+ocm sign cv <ref> \
```
jakobmoellerdev, line 34: this flag set is completely against our hard coding of signing plugins. we explicitly do not wanted to introduce hard coded flags for signing implementations on the commandline which is why https://github.com/open-component-model/ocm-project/issues/544 exists. Does this here suggest dropping this again?

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,358 @@
...
+  --cosign \
+  --cosign-upload \
+  --cosign-skip-tlog \
+  --cosign-annotation <key=value> \
+  --cosign-annotation <key=value> \
+  --cosign-identity-token <token-or-path>
+```
+
+- `<ref>`: descriptor reference (file path or repository reference).
+- `--recurse`: calculate digests for referenced component versions.
```
jakobmoellerdev, line 49: I dont understand why we would need this if we already have an add digests command

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,358 @@
...
+## Behavior
+
+### Digest calculation (`ocm add digests cv`)
+
+- Computes and **embeds** digests:
+  - **Resources**: content digests.
+  - **Component references**: digest of the referenced component descriptor.
+  - **Component-version**: digest over the canonicalized descriptor.
+- May download artifacts and referenced descriptors as needed.
+- Writes the updated descriptor back to `<ref>`.
```
jakobmoellerdev, line 77: So its not just a digest calculation, but rather a complete update to the component version right?

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,358 @@
...
+
+### Signing (`ocm sign cv`)
+
+- Loads the descriptor from `<ref>`.
+- Canonicalizes the descriptor using JCS.
+- Computes the component-version digest from canonical bytes.
```
jakobmoellerdev, line 83: How? If you want to include this in the ADR this is missing a description

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,358 @@
...
+- If `--pin` is provided, compares and fails on mismatch.
```
jakobmoellerdev, line 84: how can it pin if it doesn't know what to pin against?

matthiasbruns:
fyi: texts are improved by ai 🤖

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,358 @@
...
+- Performs direct signing using `--cert` (or a default signer if available).
```
jakobmoellerdev, line 85: What is a "signer"? that is not defined

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,358 @@
...
+- If `--cosign` is set:
+  - Obtain an OIDC ID token in this order:
+    1. Use `--cosign-identity-token` if provided (token string or `@path` to file).
+    2. Else use `SIGSTORE_ID_TOKEN` environment variable if present.
+    3. Else, if interactive TTY is available, run an **interactive loopback** browser flow to retrieve the token.
```
jakobmoellerdev, line 90: how would this work across plugin / binary boundary? is TTL forwarded correctly?

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,358 @@
...
+    4. Else, run the **device flow** to retrieve the token.
+  - Produce a Cosign signature over the **same canonical bytes**.
+  - When `--cosign-upload` is present and `<ref>` maps to an OCI subject (`<name>@<digest>`), attach the Cosign signature to the registry using Cosign conventions.
```
jakobmoellerdev, line 93: if using --cosign but not --cosign-upload, how is there ever a valid signature?

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,358 @@
...
+  - Record the **Cosign bundle** (signature, certs, and optional Rekor inclusion) in the envelope for offline verification.
+- Append the signature envelope under `.signatures[]`.
```
jakobmoellerdev, line 95: the signature envelope is not arbitrary and you have to adhere to the OCM specification

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,358 @@
...
+- Do not compute or embed any digests during signing.
+
+---
+
+## CLI Examples
+
+```bash
+# Digests
+ocm add digests cv ghcr.io/org/app:1.2.3 --recurse --force
```
jakobmoellerdev, line 104: there can be N hashes for a descriptor. how do you determine the identity of the digest you are generating here? similarly for the recursion references?

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,358 @@
...
+
+# Direct sign
+ocm sign cv ghcr.io/org/app:1.2.3 \
+  --sig release@2025-08-27 \
```
jakobmoellerdev, line 108: where does this signature name suddenly come from?

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,358 @@
...
+  --cert ~/.keys/release.p12 \
+  --password-prompt
+
+# Direct sign + embed Cosign bundle (no upload)
+ocm sign cv ghcr.io/org/app:1.2.3 \
+  --sig release@2025-08-27 \
+  --cosign
```
jakobmoellerdev, line 115: this also requires OIDC no?

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,358 @@
...
+
+# Direct sign + Cosign upload to OCI and Rekor (keyless OIDC via browser/device)
+ocm sign cv ghcr.io/org/app:1.2.3 \
+  --sig release@2025-08-27 \
+  --cosign \
+  --cosign-upload \
```
jakobmoellerdev, line 121: I think the bundle needs to be embedded either way to allow offline verification

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,358 @@
...
+  --cosign-annotation git.sha=deadbeef
+
+# Cosign with pre-fetched ID token (no browser)
```
jakobmoellerdev, line 124: I think this is what you were referring to with device flow earlier?

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,358 @@
...
+```
+
+### Component-version digest
+
+```go
+sum := sha256.Sum256(canon)
+componentDigest := "sha256:" + hex.EncodeToString(sum[:])
+```
+
+### Signature envelope
```
jakobmoellerdev, line 151: this is not how digests are defined in OCM right now, see https://github.com/open-component-model/ocm-spec/blob/main/doc/02-processing/03-signing-process.md

TBH I dont think we should break the existing OCM signature Specification, so you would need to map this back

jakobmoellerdev (changes requested):

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,356 @@
...
+---
+
+## Behavior
+
+### Digest calculation (`ocm add digests cv`)
+
+- Computes and **embeds** digests:
+  - **Resources**: content digests.
+  - **Component references**: digest of the referenced component descriptor.
+  - **Component-version**: digest over the canonicalized descriptor.
```
fabianburth, line 75: We should specify what we will do with this descriptor digest. We wouldn't store it anywhere during `add digest` (would we?), but only return it, right?

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,356 @@
...
+  --cosign \
+  --cosign-upload \
+  --cosign-skip-tlog \
+  --cosign-annotation <key=value> \
+  --cosign-annotation <key=value> \
+  --cosign-identity-token <token-or-path>
+```
+
+- `<ref>`: descriptor reference (file path or repository reference).
+- `--recurse`: calculate digests for referenced component versions.
```
fabianburth, line 49: I suppose `--recurse` relates to the `add digests` command.
Not sure, if we actually want to consider the `add digests` command in this ADR. If we do so, I think this needs more detail, how you'd intend to design this recursive digest calculation.

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,356 @@
...
+- Computes the component-version digest from canonical bytes.
+- If `--pin` is provided, compares and fails on mismatch.
+- Performs direct signing using `--cert` (or a default signer if available).
+- If `--cosign` is set:
+  - Obtain an OIDC ID token in this order:
+    1. Use `--cosign-identity-token` if provided (token string or `@path` to file).
+    2. Else use `SIGSTORE_ID_TOKEN` environment variable if present.
+    3. Else, if interactive TTY is available, run an **interactive loopback** browser flow to retrieve the token.
+    4. Else, run the **device flow** to retrieve the token.
+  - Produce a Cosign signature over the **same canonical bytes**.
```
fabianburth, line 92: I think this is what is meant here - but we'd produce the cosign signature for the digest calculated based on the normalized descriptor, right?

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,356 @@
...
+  - When `--cosign-upload` is present and `<ref>` maps to an OCI subject (`<name>@<digest>`), attach the Cosign signature to the registry using Cosign conventions.
```
fabianburth, line 93: Could you elaborate how the signing and verification workflows would compare for:
- component stored in oci (so, we can upload the signature)
- component not stored in oci (so, we cannot upload the signature)

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,356 @@
...
+  - Record the **Cosign bundle** (signature, certs, and optional Rekor inclusion) in the envelope for offline verification.
+- Append the signature envelope under `.signatures[]`.
+- Do not compute or embed any digests during signing.
```
fabianburth, line 96: > - Computes the component-version digest from canonical bytes.

Those statements kind of contradict each other.

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,356 @@
...
+  - **Component-version**: digest over the canonicalized descriptor.
+- May download artifacts and referenced descriptors as needed.
+- Writes the updated descriptor back to `<ref>`.
+
+### Signing (`ocm sign cv`)
+
+- Loads the descriptor from `<ref>`.
+- Canonicalizes the descriptor using JCS.
+- Computes the component-version digest from canonical bytes.
+- If `--pin` is provided, compares and fails on mismatch.
```
fabianburth, line 84: I think currently, the ADR does mention the actual original purpose of `--pin` which was to avoid an additional roundtrip of downloading resources and descriptors after already having done that in `add digests`. In general, the relationship between those commands is not really apparent here.

docs/adr/0008_signing_verification.md
matthiasbruns: documented below the commands

docs/adr/0008_signing_verification.md
matthiasbruns: but we don't have it right? I thought I add this to the adr for completeness

docs/adr/0008_signing_verification.md
matthiasbruns: We talked about this and I thought I understood that signatures have names - but they dont as I have seen again in the spec - removed

matthiasbruns: https://github.com/open-component-model/ocm-spec/blob/main/doc/02-processing/04-signing-examples.md
as in this spec, signatures have names right? e.g. `mysig `

docs/adr/0008_signing_verification.md
matthiasbruns: removed digest calculation from this ADR and mentioned this is a requirement

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,358 @@
...
+- `--password <pw>` / `--password-prompt`: decrypt key material if protected.
+- `--cosign`: also produce an OCI Cosign signature for the same canonical payload.
+- `--cosign-upload`: if `<ref>` resolves to an OCI subject, attach/upload the Cosign signature to the registry; otherwise, store the Cosign bundle only.
+- `--cosign-skip-tlog`: disable Rekor upload.
+- `--cosign-annotation`: repeatable `key=value` annotations for the Cosign signature.
+- `--cosign-identity-token <token-or-path>`: pass an OIDC ID token directly (alternatively, use `SIGSTORE_ID_TOKEN` env).
+
+### Normalization
+
+- Canonicalization for component-version digest and signature payload is **JCS (RFC 8785)** via the existing `jcs` package.
```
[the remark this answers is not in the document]
matthiasbruns: wrong wording . updated the wording and referenced our implementation

docs/adr/0008_signing_verification.md
matthiasbruns: removing digest for now and only focus on the signing part

docs/adr/0008_signing_verification.md
matthiasbruns: removing digest for now and only focus on the signing part

docs/adr/0008_signing_verification.md
matthiasbruns: removing digest for now and only focus on the signing part

docs/adr/0008_signing_verification.md
matthiasbruns: there is some demo code attached in this adr for the cases

docs/adr/0008_signing_verification.md
matthiasbruns: removed from adr

docs/adr/0008_signing_verification.md
jakobmoellerdev: I checked this demo code and I dont think it demoes what I meant. you open a HTTP server thats true but the server handler is contained within the same binary as the calling code. As such forwarding TTY is trivial. I think the real test would be a true separate binary that you fork off the original one and see if you still receive the TTY

jakobmoellerdev:
I think this PR needs to be rewritten for several reasons:
- It needs to accomodate plugin interfaces (those were removed)
- It needs to use less AI generated flows and have actually tested proven flow examples
- It should not use flags hardcoded on the CLI, but defer to the argument interpreter EPIC and use specs as we do everywhere else
- It should use the existing credential graph and credential plugins for configuration of any credential information

docs/adr/0008_signing_verification.md
matthiasbruns: applied plugin stuff as far as I understood it

docs/adr/0008_signing_verification.md
matthiasbruns: reworked

jakobmoellerdev:
closing as discussed

PR #602, opened 2025-08-21
fix: improve credential graph error handling and resolution logic

jakobmoellerdev:

- Refined error messages for better clarity when credential resolution fails.
- Introduced `ErrNoIndirectCredentials` for clear distinction between direct and indirect resolution failures.
- Enhanced `ToGraph` to handle `nil` configurations and added fallback logic for indirect resolution.
- Updated method signatures to support the new `GraphResolver` interface.
- Adjusted indirect resolution logic to signal failure without excessive verbosity.
- Incorporated tests to validate updated error handling and resolution workflows.

Improves the readability and maintainability of credential graph resolution errors and ensures consistency across direct and indirect resolution paths. (noticed during testing of CLI interaction with credential graph)


bindings/go/credentials/graph.go
```diff
@@ -29,10 +34,17 @@ type Options struct {
 	CredentialRepositoryTypeScheme *runtime.Scheme
 }
 
+type GraphResolver interface {
+	Resolve(ctx context.Context, identity runtime.Identity) (map[string]string, error)
+}
```
fabianburth, line 39: Is there a particular reason we introduced the interface?

bindings/go/credentials/graph.go
```diff
@@ -29,10 +34,17 @@ type Options struct {
...
+
 // ToGraph creates a new credential graph from the provided configuration and options.
 // It initializes the graph structure and ingests the configuration into the graph.
 // Returns an error if the configuration cannot be properly ingested.
-func ToGraph(ctx context.Context, config *cfgRuntime.Config, opts Options) (*Graph, error) {
+func ToGraph(ctx context.Context, config *cfgRuntime.Config, opts Options) (GraphResolver, error) {
+	if config == nil {
+		config = &cfgRuntime.Config{}
+	}
```
fabianburth, line 47: I think so far, our "guideline" was to fix this by not passing nil at all (and thus, adjusting the calling functions), as we considered this defensive programming. Is there a reason we deviate from this here?

bindings/go/credentials/graph.go
jakobmoellerdev: I did not want to keep exposing the Struct to avoid potential initializations with `Struct{}` style. not strictly necessary though

fabianburth: Alright! We can definitely keep it. Was just wondering whether there was a particular reason, as you also didn't leverage a mock in the tests.

bindings/go/credentials/graph.go
```diff
@@ -72,14 +81,17 @@ func (g *Graph) Resolve(ctx context.Context, identity runtime.Identity) (map[str
...
-		// fall back to indirect resolution
-		return g.resolveFromRepository(ctx, identity)
+		// fall back to indirect resolution if we have a repository plugin provider
+		// otherwise leave error as is
+		if g.repositoryPluginProvider != nil {
+			creds, err = g.resolveFromRepository(ctx, identity)
+		}
 	case err != nil:
 		return nil, err
 	}
```
fabianburth, line 91: Is this still needed?

fabianburth (approved):

fabianburth (approved):

PR #603, opened 2025-08-21
feat: start creating a draft release for the CLI on main

jakobmoellerdev:

Sets Up Github Draft Releases that always publish the latest main build of the CLI to a draft release for testing. Take a look at https://github.com/jakobmoellerdev/open-component-model/releases/tag/untagged-f09e86dae1d88f19a9a5 where I ran the job right now on my fork!

Now we can finally start shipping the CLI to users who are not having go installed and just want to test our workflows

You might have noticed that a lot of changes in the Changelog are under fix. That is because renovate puts them up as `fix(deps)` thats why I now force renovate to always use `chore(deps)` instead for any future release

### DIFF
--- a/.github/workflows/kubernetes-controller.yml
+++ b/.github/workflows/kubernetes-controller.yml
@@ -65,37 +65,31 @@ jobs:
   publish_latest:
     name: "Publish Latest Image"
     if: github.ref == 'refs/heads/main'
-    runs-on: large_runner
-    needs: E2E
+    runs-on: ubuntu-latest
+    #    needs: E2E
     steps:
       - name: Self Hosted Runner Post Job Cleanup Action
         uses: TooMuch4U/actions-clean@9b358e33df99574ac0bdf2e92fa3db1ae1415563 # v2.2
-      - name: Generate token
-        id: generate_token
-        uses: tibdex/github-app-token@3beb63f4bd073e61482598c45c71c1019b59b73a # v2.1.0
-        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/cli.yml
```diff
@@ -2,33 +2,154 @@ name: CLI
...
+        if: github.event_name == 'push' && github.ref == 'refs/heads/main'
+        uses: actions/upload-artifact@v4
+        with:
+          name: cli-build-${{ github.run_id }}
+          if-no-files-found: error
+          path: |
+            cli/tmp/**
+
+  release:
+    name: Draft Release
```
Skarlso, line 52: Why a draft release instead of a full release?

Skarlso: ah okay. I see what you mean.

.github/workflows/kubernetes-controller.yml
```diff
@@ -65,37 +65,31 @@ jobs:
...
         uses: docker/login-action@184bdaa0721073962dff0199f1fb9940f07167d1 # v3.5.0
         with:
           registry: ghcr.io
           username: ocmbot[ocm]
-          password: ${{ steps.generate_token.outputs.token }}
+          password: ${{ github.token}}
       - name: Build & Push
-        run: task kubernetes/controller:docker-build/multi-arch PUSH=true
\ No newline at end of file
+        run: task kubernetes/controller:docker-build/multi-arch PUSH=false
```
fabianburth, line 95: wouldn't it make sense to always push to a latest tag?

PR #607, opened 2025-08-21
fix(credentials): return empty config instead of nil

fabianburth:
Can't really remember where, but pretty sure we ran into a nil pointer somewhere during implementation efforts of `get cv --recursive` and agreed that this would be the best way to fix it. (: 

Contributes to: https://github.com/open-component-model/ocm-project/issues/563

### DIFF
--- a/bindings/go/credentials/spec/config/runtime/merge_test.go
+++ b/bindings/go/credentials/spec/config/runtime/merge_test.go
@@ -15,7 +15,7 @@ func TestMerge(t *testing.T) {
 		{
 			name:     "empty configs",
 			configs:  []*Config{},
-			expected: nil,
+			expected: &Config{},
 		},
 		{
 			name: "single config",


Skarlso (approved):

bindings/go/credentials/spec/config/runtime/merge.go
```diff
@@ -3,7 +3,7 @@ package runtime
 // Merge merges the provided configs into a single config.
 func Merge(configs ...*Config) *Config {
 	if len(configs) == 0 {
-		return nil
+		return &Config{}
```
Skarlso, line 6: Just saying that this means that we will have to check in a different way that things aren't set.

Skarlso: >what are you referring to?

That now you can't do `== nil` you have to figure out what the zero value is for the struct.

PR #610, opened 2025-08-22
fix(ci): properly generate and add repo checkout to generate

jakobmoellerdev:

we had a break in our generation workflow when we changed CI to pull request target. this should fix this


jakobmoellerdev:
merging this because of CI deadlock due to pull_request_target

PR #618, opened 2025-08-22
feat(cli): support short-syntax referencing for archives

jakobmoellerdev:

- Added logic to recognize and exclude archive-like file paths (e.g., `.tar`, `.tgz`) from being incorrectly identified as domains.
- Introduced a utility function `looksLikeArchive` to validate archive paths against a set of whitelisted extensions (`.tar`, `.tgz`, `.tar.gz`).
- Extended test coverage to include cases involving archive paths.

This enhancement ensures that archive paths are appropriately handled in short-syntax references, improving the accuracy of path resolution.

### DIFF
--- a/cli/cmd/cmd_test.go
+++ b/cli/cmd/cmd_test.go
@@ -469,6 +469,42 @@ resources:
 			r.NoError(err, "could not construct component version with working directory")
 		})
 	})
+
+	t.Run("archives", func(t *testing.T) {
+		tcs := []string{
+			".tar",
+			".tar.gz",
+		}
+		for _, tc := range tcs {
+			t.Run(tc, func(t *testing.T) {
+				tmp := t.TempDir()
+				constructorYAML = fmt.Sprintf(`
+name: ocm.software/examples-01
+version: 1.0.0
+provider:
+  name: ocm.software
+resources:
+  - name: my-file-replaced
+    type: blob
+    input:
+      type: utf8/v1
+      text: foobar
+`)
+
+				// Create a replacement test file to be added to the component version
+				constructorYAMLFilePath := filepath.Join(tmp, "component-constructor.yaml")
+				r.NoError(os.WriteFile(constructorYAMLFilePath, []byte(constructorYAML), 0o600))
+
+				_, err := test.OCM(t, test.WithArgs("add", "cv",
+					"--constructor", constructorYAMLFilePath,
+					"--repository", filepath.Join("transport-archive", tc),
+					"--working-directory", tmp,
+				), test.WithOutput(logs))
+
+				r.NoError(err, "could not construct component version with working directory")
+			})
+		}
+	})
 }
 
 func Test_Version(t *testing.T) {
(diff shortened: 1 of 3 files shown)


cli/internal/reference/compref/compref.go
```diff
@@ -290,3 +294,18 @@ func looksLikeDomain(s string) bool {
...
 }
+
+var allowListArchiveFilePathExtensions = []string{".tar", ".tgz", ".tar.gz"}
+
+// looksLikeArchive checks if the string ends with an allow-listed file extension
+// this makes it so that a path like "my.path.tar" is always considered an archive, and if it should
+// be interpreted as path, it needs to be passed explicitly
+func looksLikeArchive(s string) bool {
+	for _, ext := range allowListArchiveFilePathExtensions {
+		if strings.HasSuffix(s, ext) {
```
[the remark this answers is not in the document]
jakobmoellerdev: I guess I was hoping to not have to open a file descriptor on here. I could do that of course in theory, but that would make the operation significantly more expensive. WDYT?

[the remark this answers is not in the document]
fabianburth: I'd say let's roll with the `looksLikeArchive` logic - it is very simple and maintainable, and therefore, a very practical solution for the current problem.
Besides, I think relying on file extensions is a quite common thing to do.

cli/cmd/cmd_test.go
```diff
@@ -469,6 +469,42 @@ resources:
...
+  name: ocm.software
+resources:
+  - name: my-file-replaced
+    type: blob
+    input:
+      type: utf8/v1
+      text: foobar
+`)
+
+				// Create a replacement test file to be added to the component version
```
fabianburth, line 494: I don't understand that comment

cli/cmd/cmd_test.go
```diff
@@ -469,6 +469,42 @@ resources:
...
+				constructorYAMLFilePath := filepath.Join(tmp, "component-constructor.yaml")
+				r.NoError(os.WriteFile(constructorYAMLFilePath, []byte(constructorYAML), 0o600))
+
+				_, err := test.OCM(t, test.WithArgs("add", "cv",
+					"--constructor", constructorYAMLFilePath,
+					"--repository", filepath.Join("transport-archive", tc),
+					"--working-directory", tmp,
+				), test.WithOutput(logs))
```
fabianburth, line 502: We aren't using the logs, are we?

cli/cmd/cmd_test.go
```diff
@@ -469,6 +469,42 @@ resources:
 			r.NoError(err, "could not construct component version with working directory")
 		})
 	})
+
+	t.Run("archives", func(t *testing.T) {
+		tcs := []string{
+			".tar",
+			".tar.gz",
+		}
```
fabianburth, line 477: I feel like this test should not be a table test. The table test should happen based on the allow list in the compref packages unit tests.

fabianburth (changes requested):

jakobmoellerdev:
Closing in favor of a properly tested implementation as a learning opportunity

PR #619, opened 2025-08-25
chore: add logging when a version is skipped to improve ux

Skarlso:

Fixes https://github.com/open-component-model/ocm-project/issues/593 because we decided that podinfo is an outlier and the issue will be closed regardless. However, improving debuggability helps a lot in figuring out _why_ a version was skipped from listing.


jakobmoellerdev (commented):
do you mind writing a test case for this?

Skarlso:
Sure

jakobmoellerdev:
nah i think this is okay :)

PR #703, opened 2025-08-27
feat: start creating a draft release for the CLI on main

jakobmoellerdev:

Sets Up Github Draft Releases that always publish the latest main build of the CLI to a draft release for testing. Take a look at https://github.com/jakobmoellerdev/open-component-model/releases/tag/untagged-f09e86dae1d88f19a9a5 where I ran the job right now on my fork!

Now we can finally start shipping the CLI to users who are not having go installed and just want to test our workflows

You might have noticed that a lot of changes in the Changelog are under fix. That is because renovate puts them up as `fix(deps)` thats why I now force renovate to always use `chore(deps)` instead for any future release


fabianburth (approved):

jakobmoellerdev:
checked regarding the link, draft releases are not visible.

PR #767, opened 2025-09-01
chore(docs): add ADR proposal for signing and verification of component descriptors

jakobmoellerdev:
Continuation of https://github.com/open-component-model/open-component-model/pull/547 and #599 

Enable easy interpretation of component descriptors in normalized form and provide a consistent experience for signing and verifying component descriptors.

fix https://github.com/open-component-model/ocm-project/issues/579


docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,623 @@
...
+    Note over U,CLI: Step 1: Add Digest
+    U->>CLI: ocm add digest --signature default --upload
+    CLI->>Repo: Download Component Descriptor
+    CLI->>CLI: Normalize + Compute Digest
+    CLI->>Repo: Upload Descriptor with Digest only (no signature)
+    CLI-->>U: Digest field added to descriptor
+
+    Note over U,CLI: Step 2: Sign Pinned Digest
+    U->>CLI: ocm sign componentversion --signature default@<digest> --signer rsapss
+    CLI->>Repo: Download Descriptor (with digest field)
```
Skarlso, line 101: If download is a request to the `Repo` shouldn't there be a response from the Repo here depicting the returned unsigned descriptor?

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,623 @@
...
+// - Use a well-known registered default configuration and be modifiable in their behavior, assuming sane defaults.
+// - offer versioned, stable signature implementations differentiated by the config type.
+// - Reject signing specifications if there is no credential available that is required for the handler.
+//
+// The returned signature SHOULD be attached to the descriptor `signatures` field after a successful call to Sign.
+type ComponentSignatureSigner interface {
+    // GetSigningCredentialConsumerIdentity resolves the credential consumer identity of the given configuration to use for credential resolution
+    // when signing new signatures with the given configuration.
+    // If successful, the returned identity SHOULD be used for credential resolution. (i.e. against the OCM credential graph)
+    // If unsuccessful, an error MUST be returned, and Sign CAN be called without credentials.
```
Skarlso, line 206: It can be called without credentials, but how useful is that? Also, the identity isn't always just credentials, right? So if an identity IS returned successfully then doesn't it MUST use it instead of SHOULD? How could it ignore it?

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,623 @@
...
+    GetSigningCredentialConsumerIdentity(ctx context.Context, config runtime.Typed) (identity runtime.Identity, err error)
+
+    // Sign signs the descriptor using the provided config.
+    // An extensible config SHOULD support media type and algorithm selection, if multiple are availalbe.
+    //
+    // Configurations MUST NOT contain any private key or otherwise sensitive material. This is a security risk.
```
Skarlso, line 212: Aren't we allowed static creds to be in the config? 🤔

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,623 @@
...
+    CLI->>Plugin: GetSigningCredentialConsumerIdentity(config)
+    Plugin-->>CLI: Credential Identity (OIDCIdentityToken/v1alpha1)
+    CLI->>Creds: Resolve credentials for identity
+    Creds->>Plugin: Trigger SigstoreOIDC flow
+    Plugin->>Browser: Open auth URL (with PKCE)
+    U->>Browser: Authenticate + Consent
+    Browser->>Plugin: Redirect to localhost callback with auth code. Forward auth code to plugin callback server
+    Plugin->>OIDC: Exchange auth code for ID token
+    OIDC-->>Plugin: ID token
+    Plugin-->>Plugin: Store token for reuse (local cache)
```
Skarlso, line 410: Would this be just for this flow and you'd have to reauthenticate on every consequite call? Or would we store it in under something like .config/ocm/cache with some key that lets us find it later on again?

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,623 @@
...
+
+Returned Credentials from `SigstoreOIDC/v1alpha1` after a successful normal / implicit OIDC flow:
+
+```yaml
+token: <base64 id token that contains email and openid scope => id token>
+```
+
+These credentials are retrieved via the plugin by using the [interactive OIDC flow from Sigstores OIDC implementation](https://github.com/sigstore/sigstore/blob/main/pkg/oauthflow/interactive.go#L48) via plugin.
+
+Because OIDC allows to open a browser with a redirect, the plugin can dynamically open a browser window to authenticate the user and redirect back to its own small webserver retrieving the auth code. This works across binary boundaries, because the webserver is running in the same process as the plugin. Potentially, even if the server is not running on the same machine, the browser can still open a window when the client_credentials flow is used (not realized in this ADR, but possible).
```
Skarlso, line 546: Hah! So neat that we already have a server in the plugins. :)

Skarlso (commented):
Well done, amazing job with this write-up. I don't have any major concerns but I do have a couple of questions as always. :)

docs/adr/0008_signing_verification.md
fabianburth: I guess a particular signing handler could decide to adjust and therefore mutate the returned identity for some reason? (i admit that's kind of a stretch)

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,623 @@
...
+    // If successful, the returned identity SHOULD be used for credential resolution. (i.e. against the OCM credential graph)
+    // If unsuccessful, an error MUST be returned, and Sign CAN be called without credentials.
+    GetVerifyingCredentialConsumerIdentity(ctx context.Context, config runtime.Typed) (identity runtime.Identity, err error)
+    
+    // Verify performs signature and digest checks using the provided config.
+    //
+    // An extensible config SHOULD support timeout / limit configurations for signature validation
+    // Configurations MUST NOT contain any key or otherwise sensitive material. This is a security risk.
+    // Instead, the verifier MUST use the provided credentials and well-known attributes to verify the signature.
+    // If the media type cannot be verified, the signature verification MUST fail.
```
fabianburth, line 241: What media type? (I guess that's my lack of understanding here!)

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,623 @@
...
+
+    Note over U,CLI: Signing
+    U->>CLI: ocm sign componentversion --signer rsapss
+    CLI->>Repo: Download Component Descriptor
+    CLI->>CLI: Normalize + Compute Digest
+    CLI->>Handler: GetSigningCredentialConsumerIdentity(config)
+    Handler-->>CLI: Credential Identity
+    CLI->>Creds: Resolve credentials for identity
+    Creds-->>CLI: private_key_pem_file
+    CLI->>Handler: Sign(digest, config, credentials)
```
fabianburth, line 274: I guess the `Sign(digest, config, credentials)` does not align with the handlers interface described above. As I noted, there we pass the signature instead of the digest.

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,623 @@
...
+    Handler-->>CLI: Signature (RSASSA-PKCS1-V1_5)
+    CLI->>Repo: Upload Descriptor with Signature
```
fabianburth, line 276: For `ocm add digest` we returned the resulting descriptor here, I think. Would we want to do that here too?
(I asked this above already, so depending on what we decide, feel free to just close this)

docs/adr/0008_signing_verification.md
jakobmoellerdev: you are right as always! Will add response ways

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,623 @@
...
+```
+
+### Config
+
+```yaml .ocmconfig
+- type: signing.configuration.ocm.software/v1alpha1
+  signers:
+  - name: rsapss
+    config:
+      type: RSASSA-PKCS1-V1_5/v1alpha1
```
fabianburth, line 299: follow the existing patterns in ocm config (not saying that we strictly should), this would rather look like:
```
- type: signing.configuration.ocm.software/v1alpha1
  signers:
  - signer:
       type: RSASSA-PKCS1-V1_5/v1alpha1
       name: rsapss
```
wouldn't it? Is there a particular reason you decided to propose this structure?
I also don't really understand what the name maps to. Are the signature handlers registered under a particular name? Is this mentioned anywhere?

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,623 @@
...
+
+```shell
+ocm sign componentversion --signature mysig --signer rsapss ghcr.io/open-component-model/ocm//ocm.software/ocm:0.17.0
+```
+
+#### Via `--signer-spec`
+
+```yaml ./rsapss.yaml
+type: RSASSA-PKCS1-V1_5/v1alpha1
+```
```
fabianburth, line 328: wouldn't this have to include the name now?

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,623 @@
...
+
+```shell
+ocm verify componentversion --signature mysig --verifier rsapss ghcr.io/open-component-model/ocm//ocm.software/ocm:0.17.0
+```
+
+#### Via `--verifier-spec`
+
+```yaml ./rsapss.yaml
+type: RSASSA-PKCS1-V1_5/v1alpha1
+```
```
fabianburth, line 346: same as above, wouldn't this have to include the name now?

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,623 @@
...
+verifying.
+
+## Example Configuration (Sigstore)
+
+Sigstore requires a special external plugin that works as:
+
+- Signing Handler with `sign.sigstore.dev/v1alpha1`
+  - This handler can request credentials of type `OIDCIdentityToken/v1alpha1` to use for signing.
+- Verification Handler with `verify.sigstore.dev/v1alpha1`
+  - This handler can request credentials of type `PEM/v1alpha1` to use for offline verification.
```
fabianburth, line 379: Would we only support offline verification or why is nothing else mentioned here?

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,623 @@
...
+    * (If 2 step signing is used, this is done in a second command)
+
+5. **Verification** *(verify mode)*
+
+    * Filter candidate signatures by name if provided, otherwise verify all signatures present.
+    * Verify signature(s) against recomputed digest and trust material.
+
+6. **Output**
+
+    * Sign: success with updated descriptor as long as `--update` is set. Always returns the signature specification, or an error.
```
fabianburth, line 598: I think above it was called `--upload`

fabianburth (changes requested):

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,623 @@
...
+    U->>CLI: ocm add digest --signature default --upload
+    CLI->>Repo: Download Component Descriptor
+    CLI->>CLI: Normalize + Compute Digest
+    CLI->>Repo: Upload Descriptor with Digest only (no signature)
+    CLI-->>U: Digest field added to descriptor
+
+    Note over U,CLI: Step 2: Sign Pinned Digest
+    U->>CLI: ocm sign componentversion --signature default@<digest> --signer rsapss
+    CLI->>Repo: Download Descriptor (with digest field)
+    CLI->>CLI: Locate matching digest by hash algorithm + value
```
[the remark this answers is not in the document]
jakobmoellerdev: youre right this is locating by name. thanks for the correction!

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,623 @@
...
+    // If successful, the returned identity SHOULD be used for credential resolution. (i.e. against the OCM credential graph)
+    // If unsuccessful, an error MUST be returned, and Sign CAN be called without credentials.
+    GetSigningCredentialConsumerIdentity(ctx context.Context, config runtime.Typed) (identity runtime.Identity, err error)
+
+    // Sign signs the descriptor using the provided config.
+    // An extensible config SHOULD support media type and algorithm selection, if multiple are availalbe.
+    //
+    // Configurations MUST NOT contain any private key or otherwise sensitive material. This is a security risk.
+    // Instead, the signer MUST use the provided credentials and well-known attributes to sign the digest specification.
+    Sign(ctx context.Context, unsigned descruntime.Signature, config runtime.Typed, credentials map[string]string) (signed descruntime.Signature, err error)
```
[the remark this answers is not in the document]
jakobmoellerdev: good question, I checked this struct and technically no. I thought there were labels on the signatures which apparently there are not... So im doubtful now. WDYT?

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,623 @@
...
+// - offer versioned, stable verification implementations differentiated by the config type.
+// - Reject verification specifications if there is no credential available that is required for the handler to verify the signature.
+//
+// See: https://ocm.software/docs/reference/ocm-cli/verify/componentversions/
+type ComponentSignatureVerifier interface {
+    // GetVerifyingCredentialConsumerIdentity resolves the credential consumer identity of the given configuration to use for credential resolution
+    // when verifying signatures with the given configuration.
+    // If successful, the returned identity SHOULD be used for credential resolution. (i.e. against the OCM credential graph)
+    // If unsuccessful, an error MUST be returned, and Sign CAN be called without credentials.
+    GetVerifyingCredentialConsumerIdentity(ctx context.Context, config runtime.Typed) (identity runtime.Identity, err error)
```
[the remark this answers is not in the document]
jakobmoellerdev: See my comment on handling behavior without credentials. i.e. automatic file mounts or env variables.

docs/adr/0008_signing_verification.md
jakobmoellerdev: Each signature has a specific media type:

`application/vnd.ocm.signature.sigstore` for sigstore

and

`application/vnd.ocm.signature.rsa.pss` for rsapss (see https://github.com/open-component-model/ocm/blob/4a1aa4fa4668b2a0758a0d0a2d2c0e7c5d180d7e/api/tech/signing/handlers/rsa-pss/handler.go#L17)

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,623 @@
...
+
+> name: default
+> digest:
+>   hashAlgorithm: sha256
+>   normalisationAlgorithm: jsonNormalisation/v1
+>   value: cf08abae08bb874597630bc0573d941b1becc92b4916cbe3bef9aa0e89aec3f6
+```
+
+After this, as long as `--upload` is set, the descriptor will be updated with this signature field, containing a digest,
+but no actual signature value.
```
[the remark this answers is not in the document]
fabianburth: Yeah, if the default would be `--upload=true`, I think we should go with the `--dry-run` option

docs/adr/0008_signing_verification.md
jakobmoellerdev: agreed, lets discuss on the other conversation to resolve.

docs/adr/0008_signing_verification.md
jakobmoellerdev: IMHO The existing patterns imo dont make sense. Why would we have a `signer` key on `signers`. The name maps to the `--signer` attribute that you can use to shortcut to the spec

docs/adr/0008_signing_verification.md
jakobmoellerdev: why?

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,623 @@
...
+    CLI->>Repo: Download Descriptor (with digest field)
+    CLI->>CLI: Locate matching digest by hash algorithm + value
+    CLI->>Handler: GetSigningCredentialConsumerIdentity(config)
+    Handler-->>CLI: Credential Identity
+    CLI->>Creds: Resolve credentials for identity
+    Creds-->>CLI: private_key_pem_file
+    CLI->>Handler: Sign(digest, config, credentials)
+    Handler-->>CLI: Signed Signature
+    CLI->>Repo: Upload Descriptor with Signed Digest
+    CLI-->>U: Descriptor updated, ready for verification
```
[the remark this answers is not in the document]
fabianburth: I just wanted to align `ocm sign` and `ocm add digest` but after thinking about it, for `ocm add digest` we have an actual reason to return it due to the pinning. So, I think we're good without returning it!

docs/adr/0008_signing_verification.md
jakobmoellerdev: no we wouldn't I think. PTAL at the default policy (which falls back to TUF)

docs/adr/0008_signing_verification.md
jakobmoellerdev: thanks for the catch

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,623 @@
...
+         value: 390157b7...75ab2705d6
+   ```
+3. Verify the signature using the configured verifier from `.ocmconfig`.
+
+### Signing
+
+Signing uses the analogous command:
+
+```shell
+ocm sign componentversion --signature mysig --signer rsapss ghcr.io/open-component-model/ocm//ocm.software/ocm:0.17.0
```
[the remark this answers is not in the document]
fabianburth: Yeah, you are right - also, after having read through the whole adr, we have `ocm sign` and `ocm verify`, so this is the better fit! Just forget I ever suggested that!

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,623 @@
...
+**Module**
+
+```text
+bindings/go/<technology>/go.mod
+```
+
+**Package**
+
+```text
+bindings/go/<technology>/signing/method
```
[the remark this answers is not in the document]
fabianburth: thanks for explaining!

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,623 @@
...
+// See: https://ocm.software/docs/getting-started/sign-component-versions/
+// - Reject signature specifications without a precalculated digest specification
+// - Not modify the given signature digest specification in any way when signing
+// 
+// Implementations SHOULD:
+// - Use a well-known registered default configuration and be modifiable in their behavior, assuming sane defaults.
+// - offer versioned, stable signature implementations differentiated by the config type.
+// - Reject signing specifications if there is no credential available that is required for the handler.
+//
+// The returned signature SHOULD be attached to the descriptor `signatures` field after a successful call to Sign.
```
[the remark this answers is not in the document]
fabianburth: Got it - makes sense!

docs/adr/0008_signing_verification.md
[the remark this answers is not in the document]
fabianburth: Had 2 things in mind with this comment:
- We discussed whether it might make sense to separate digest and signature in a future ocm spec. In that case, digest would be a much more natural input here.
- Whether one might want to reuse this signing functionality to sign a digest other than a component digest.

So, based on my current understanding, digest would just be the more flexibel future proof option, especially because signature does not contain additional information that might be useful - as you pointed out with the labels.

docs/adr/0008_signing_verification.md
```diff
@@ -0,0 +1,623 @@
...
+
+### Signing
+
+Signing uses the analogous command:
+
+```shell
+ocm sign componentversion --signature mysig --signer rsapss ghcr.io/open-component-model/ocm//ocm.software/ocm:0.17.0
+```
+
+This downloads, normalizes, signs, and re-uploads the descriptor. Signing and verification configs live in `.ocmconfig`.
```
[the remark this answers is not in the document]
jakobmoellerdev: I was thinking that this isnt really required but an additional feature we can scope for someone. But I can add it here as well if strictly wanted.

docs/adr/0008_signing_verification.md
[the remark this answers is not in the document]
jakobmoellerdev: agreed on this. will adjust the contract

jakobmoellerdev:
Im guessing you might have a `dev-rsapss` and `prod-rsapss` for example if you are on a local machine, or you want to differentiate signing sources

fabianburth (dismissed):
Great ADR!

PR #770, opened 2025-09-01
feat(dag)!: introduce dag processing logic

fabianburth:
This PR introduces logic to perform concurrent topological traversal of a dag, with child nodes always being processed before its parent node.

- It also fixes a "bug" in the `Reverse` methods where attributes have not been copied (it was simply not implemented).
- It renames `Traverse` to `Discover` since this caused a lot of confusion.
- It fixes a flaw in the `Discover`/`DiscoverNeighbors` logic that allowed the discover neighbors function to add attributes to the neighbors which had the potential to cause races. Now, the discover neighbors logic is only passed the vertex ID and can only return the neighbor vertices IDs. 
- This is sufficient for our component cases for now. If we need attributes during discovery in the future, we'd have to add them to the edges.

Contributes to: https://github.com/open-component-model/ocm-project/issues/564

### DIFF
--- a/bindings/go/dag/sync/dag.go
+++ b/bindings/go/dag/sync/dag.go
@@ -432,17 +432,32 @@ func (d *DirectedAcyclicGraph[T]) Reverse() (*DirectedAcyclicGraph[T], error) {
 
 	// Ensure all vertices exist in the new graph
 	d.Vertices.Range(func(key, value any) bool {
-		if err := reverse.AddVertex(key.(T)); err != nil {
+		origVertex := value.(*Vertex[T])
+		attrs := make(map[string]any)
+		origVertex.Attributes.Range(func(attrKey, attrValue any) bool {
+			attrs[attrKey.(string)] = attrValue
+			return true
+		})
+		if err := reverse.AddVertex(key.(T), attrs); err != nil {
 			return false
 		}
 		return true
 	})
 
-	// Reverse the edges: Child -> Parent instead of Parent -> Child
 	d.Vertices.Range(func(key, value any) bool {
 		parent := value.(*Vertex[T])
-		parent.Edges.Range(func(child any, _ any) bool {
-			if err := reverse.AddEdge(child.(T), parent.ID); err != nil {
+		parent.Edges.Range(func(child any, edgeAttrs any) bool {
+			// Copy edge attributes
+			attrMap := make(map[string]any)
+			if edgeAttrs != nil {
+				if smap, ok := edgeAttrs.(*sync.Map); ok {
+					smap.Range(func(k, v any) bool {
+						attrMap[k.(string)] = v
+						return true
+					})
+				}
+			}
+			if err := reverse.AddEdge(child.(T), parent.ID, attrMap); err != nil {
 				return false
 			}
 			return true
--- a/bindings/go/dag/sync/dag.go
+++ b/bindings/go/dag/sync/dag.go
@@ -125,17 +125,13 @@ func (d *DirectedAcyclicGraph[T]) Clone() *DirectedAcyclicGraph[T] {
 
 // AddVertex adds a new node to the graph.
 func (d *DirectedAcyclicGraph[T]) AddVertex(id T, attributes ...map[string]any) error {
+	if _, exists := d.Vertices.Load(id); exists {
+		return fmt.Errorf("node %v already exists: %w", id, ErrAlreadyExists)
+	}
 	vertex := &Vertex[T]{
 		ID:         id,
 		Attributes: &sync.Map{},
 		Edges:      &sync.Map{},
-	}
-	return d.addRawVertex(vertex, attributes...)
-}
-
-func (d *DirectedAcyclicGraph[T]) addRawVertex(vertex *Vertex[T], attributes ...map[string]any) error {
-	if _, exists := d.Vertices.Load(vertex.ID); exists {
-		return fmt.Errorf("node %v already exists: %w", vertex.ID, ErrAlreadyExists)
 	}
 	for _, attrs := range attributes {
 		for k, v := range attrs {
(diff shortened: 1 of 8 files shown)


bindings/go/dag/sync/dag.go
```diff
@@ -432,17 +432,32 @@ func (d *DirectedAcyclicGraph[T]) Reverse() (*DirectedAcyclicGraph[T], error) {
...
 	// Ensure all vertices exist in the new graph
 	d.Vertices.Range(func(key, value any) bool {
-		if err := reverse.AddVertex(key.(T)); err != nil {
+		origVertex := value.(*Vertex[T])
+		attrs := make(map[string]any)
+		origVertex.Attributes.Range(func(attrKey, attrValue any) bool {
+			attrs[attrKey.(string)] = attrValue
+			return true
+		})
+		if err := reverse.AddVertex(key.(T), attrs); err != nil {
```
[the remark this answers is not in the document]
fabianburth: We actually cannot reuse the `vertex.Clone()` method here. `vertex.Clone()` also copies the edges. But for reversing, we have to invert the edges.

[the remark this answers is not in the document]
jakobmoellerdev: (this should be a comment in the code)

bindings/go/dag/sync/process.go
```diff
@@ -0,0 +1,261 @@
...
+				d.InDegree.Store(parent, inDegree-1)
+				parentVertex := d.MustGetVertex(parent)
+				parentVertex.Edges.Delete(id)
+				// If all prerequisites (children) of the parent have been
+				// processed and the parent has not been enqueued yet, add it to
+				// the next level.
+				if d.MustGetInDegree(parent) == 0 {
+					dag.MustGetVertex(parent).Attributes.Store(AttributeProcessingState, ProcessingStateQueued)
+					nextQueueCh <- parent
+				}
```
Skarlso, line 234: This is super hard to follow. This is in degree of `d` but then MustGetVertex is called on `dag`. Should this be `d.MustGetVertex`?

bindings/go/dag/sync/discover.go
```diff
@@ -0,0 +1,245 @@
...
+type DiscoverNeighborsFunc[T cmp.Ordered] func(ctx context.Context, v T) (neighbors []T, err error)
+
+func (f DiscoverNeighborsFunc[T]) DiscoverNeighbors(ctx context.Context, v T) (neighbors []T, err error) {
+	return f(ctx, v)
+}
+
+// Discover performs a concurrent depth-first discovery from the given root vertex.
+// For each vertex v, it calls discoverer.DiscoverNeighbors(v), which
+// return its neighbors in order or an error.
+// DiscoverNeighbors is guaranteed to be called for each vertex only once.
```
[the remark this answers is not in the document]
fabianburth: good point, so its actually not guaranteed.

bindings/go/dag/sync/discover.go
```diff
@@ -0,0 +1,245 @@
...
+		if len(options.Roots) == 0 {
+			return fmt.Errorf("no roots provided and no roots found in the dag, cannot discover")
+		}
+	}
+	doneMap := &sync.Map{}
+	errGroup := errgroup.Group{}
+
+	for _, root := range options.Roots {
+		// We ensured that the rootID vertex exists in the graph
+		v, _ := d.GetVertex(root)
```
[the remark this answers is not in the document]
fabianburth: `MustGetVertex` didn't exist when I wrote the function, I think.

bindings/go/dag/sync/process.go
```diff
@@ -0,0 +1,260 @@
...
+					vertex.Edges.Delete(child)
+					// If all parents of the child have been processed and the
+					// child has not been enqueued yet, add it to the next layer.
+					if topology.MustGetInDegree(child) == 0 {
+						d.MustGetVertex(child).Attributes.Store(AttributeProcessingState, ProcessingStateQueued)
+						nextQueueCh <- child
+					}
+				}
+				mapgraph = ToMapBasedDAG(topology)
+				_ = mapgraph
```
jakobmoellerdev, line 237: why is this here?

bindings/go/dag/sync/process.go
```diff
@@ -0,0 +1,260 @@
...
+				d.MustGetVertex(id).Attributes.Store(AttributeProcessingState, ProcessingStateProcessing)
+				if err := processor.ProcessVertex(ctx, id); err != nil {
+					d.MustGetVertex(id).Attributes.Store(AttributeProcessingState, ProcessingStateError)
+					return fmt.Errorf("failed to process vertex with id %v: %w", id, err)
+				}
+				d.MustGetVertex(id).Attributes.Store(AttributeProcessingState, ProcessingStateCompleted)
+
+				vertex := topology.MustGetVertex(id)
+				mapgraph := ToMapBasedDAG(topology)
+				_ = mapgraph
```
jakobmoellerdev, line 224: leftover?

bindings/go/dag/sync/process.go
```diff
@@ -0,0 +1,260 @@
...
+
+	// wrap this logic into a function to ensure the nextQueueCh is closed
+	f := func(nextQueueCh chan T) error {
+		defer close(nextQueueCh)
+
+		for _, id := range ids {
+			errGroup.Go(func() error {
+				// Mark the id as processed or return if already processed.
+				_, loaded := doneMap.LoadOrStore(id, true)
+				if loaded {
```
jakobmoellerdev, line 208: you never need loaded outside of this if, so inline

bindings/go/dag/sync/process.go
```diff
@@ -0,0 +1,260 @@
...
+	// Calculate the upper bound for the next queue channel
+	// this determines how many ids can be processed concurrently in the next
+	// phase
+	upperBound := 0
+	for _, id := range ids {
+		upperBound += topology.MustGetOutDegree(id)
+	}
+
+	// wrap this logic into a function to ensure the nextQueueCh is closed
+	f := func(nextQueueCh chan T) error {
```
jakobmoellerdev, line 201: instead of doing all this inline, why cant we just refactor this into separate functions?

fabianburth: Already extracted the process layer logic into a function. I didn't want to introduce another one and kind of didn't like the function interface that returns the channel.

PR #780, opened 2025-09-03
feat(575): regex glob resolver

matthiasbruns:
Provide a replacement for fallbackresolvers to reduce resolving time based on glob pattern matching.
This PR focusses on the spec/binding side. The repository will follow afterwards.

Fixes: https://github.com/open-component-model/ocm-project/issues/575


bindings/go/configuration/resolvers/v1/matcher/matcher.go
```diff
@@ -0,0 +1,156 @@
...
+
+	return m.regex.MatchString(componentName)
+}
+
+func (m *ComponentMatcher) Pattern() string {
+	return m.pattern
+}
+
+// globToRegex converts a glob pattern to a regex pattern.
+func globToRegex(glob string) string {
```
matthiasbruns, line 54: hope that is enough. I wasn't to keen on bringing glob packages into ocm

bindings/go/configuration/resolvers/v1/matcher/matcher.go
```diff
@@ -0,0 +1,156 @@
...
+type ComponentMatcher struct {
+	pattern string
+	regex   *regexp.Regexp
+}
+
+// NewComponentMatcher creates a new ComponentMatcher for the given glob pattern.
+// All patterns are treated as glob patterns and converted to regex internally.
+func NewComponentMatcher(pattern string) (*ComponentMatcher, error) {
+	if pattern == "" {
+		return &ComponentMatcher{pattern: pattern}, nil
```
jakobmoellerdev, line 21: why do we need a matcher that matches everything if we no longer have priority?

bindings/go/configuration/resolvers/v1/matcher/matcher.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+// Constraint returns the original constraint string.
+func (m *VersionMatcher) Constraint() string {
+	return m.constraint
+}
+
+// ResolverMatcher combines component name and version matching for a resolver.
+type ResolverMatcher struct {
+	componentMatcher *ComponentMatcher
```
jakobmoellerdev, line 116: make Matcher an interface and store them as such to allow switching implementations later

bindings/go/configuration/resolvers/v1/matcher/matcher.go
jakobmoellerdev: My opinion: Mostly safe against ReDoS and regex injection. Not correct for glob semantics, and should be reworked

Problems:

* `[]` handling is broken. `QuoteMeta` escapes `-`, so `[a-z]` becomes `[\-]` after your unescape, losing ranges.
* No negation. Glob `[!abc]` should map to `[^abc]`; yours matches `!` literally.
* No escape rules. `\]` or `\-` inside classes are mishandled.
* No validation. Unmatched `[` should error, not compile a bad regex.
* Semantics differ from common globs (`**`, dotfiles, path separators).

If you need real globbing in Go, use the stdlib or a library:

* `path/filepath.Match` for simple `*` and `?`.
* `github.com/bmatcuk/doublestar/v4` for `**` and more or a similar lib.

If you must keep regex conversion, drop `[]` support:

```go
func globToRegex(g string) string {
	r := regexp.QuoteMeta(g)
	r = strings.ReplaceAll(r, `\*`, ".*")
	r = strings.ReplaceAll(r, `\?`, ".")
	return "^" + r + "$"
}
```

generally, it is advisable to write a parser that does either or, not and matching.
I dont really want us to write a parser manually.

bindings/go/configuration/resolvers/v1/matcher/matcher_test.go
```diff
@@ -0,0 +1,278 @@
+package matcher
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestComponentMatcher(t *testing.T) {
```
jakobmoellerdev, line 10: this is missing quite a few common blob cases such as {} or ** syntax

bindings/go/configuration/resolvers/v1/runtime/config.go
```diff
@@ -0,0 +1,76 @@
...
+//     type: OCIRegistry
+//     baseUrl: ghcr.io
+//     subPath: open-component-model/components
+//     componentName: ocm.software/core/*
+//     semver: >1.0.0
+//
+// +k8s:deepcopy-gen:interfaces=ocm.software/open-component-model/bindings/go/runtime.Typed
+// +k8s:deepcopy-gen=true
+// +ocm:typegen=true
+type Config struct {
```
jakobmoellerdev, line 22: why do we need a runtime config for the resolvers? maybe @fabianburth has some background

bindings/go/configuration/resolvers/v1/matcher/matcher.go
jakobmoellerdev: On a second note, why do you not want a glob package in the resolver?

bindings/go/configuration/resolvers/v1/runtime/config.go
matthiasbruns: Maybe I misunderstood how to create new spec/configs in OCM

bindings/go/configuration/resolvers/v1/matcher/matcher_test.go
matthiasbruns: I will look for a glob package that isn't 6yrs old

bindings/go/configuration/resolvers/v1/runtime/config.go
fabianburth: Generally, we have a runtime config for resolvers because resolvers contain repository specifications. When using this functionality as a library, we wouldn't want to convert the repositories to `Raw`.

bindings/go/configuration/resolvers/v1/matcher/matcher.go
matthiasbruns: I want to - but all I found are pretty old

bindings/go/configuration/resolvers/v1/matcher/matcher.go
```diff
@@ -0,0 +1,156 @@
...
+	regex = strings.ReplaceAll(regex, `\]`, "]")
+
+	// Anchor the regex to match the entire string
+	return "^" + regex + "$"
+}
+
+type VersionMatcher struct {
+	constraint       string
+	semverConstraint *semver.Constraints
+}
```
fabianburth, line 73: > Out of Scope:
> 
> We discussed the versioning constraints and decided that this is explicitly out of scope for the initial implementation.

Theoretically, the issue specifically states that the version constraints are out of scope for the initial implementation. If I recall correctly, we decided this not because we thought it's too much efforts BUT because we didn't immediately want to maintain that feature.

matthiasbruns:
> We could also discuss whether we think the capabilities of https://pkg.go.dev/path?utm_source=chatgpt.com#Match are sufficient for our use case here.

Yeah I have seen it - but wasn't sure since it's only for pathing

matthiasbruns:
@jakobmoellerdev  regenerated

bindings/go/configuration/resolvers/v1/matcher/matcher_test.go
```diff
@@ -0,0 +1,109 @@
...
+	tests := []struct {
```
jakobmoellerdev, line 11: which cases in this tests are not solvable with path.Match?

matthiasbruns:
Decided in the Warroom to stick with sdtlib path instead of an external lib

matthiasbruns:
@jakobmoellerdev  @fabianburth  done

PR #781, opened 2025-09-03
fix(credentials): refactor cache identity parsing logic

jakobmoellerdev:

- Replaced manual host and port parsing with `runtime.ParseURLToIdentity`.
- Removed redundant logic for handling missing ports and default values.
- Simplified `get` and `put` methods for better clarity and maintainability.

This update streamlines credential handling by delegating identity resolution to the `runtime` library, reducing code complexity and improving reliability. Also allows for proper matching no matter the actual scheme


bindings/go/oci/repository/provider/credential_cache.go
```diff
@@ -76,18 +54,11 @@ func (cache *credentialCache) add(spec *ocirepospecv1.Repository, credentials ma
 	cache.mu.Lock()
 	defer cache.mu.Unlock()
 
-	parsedBaseURL, err := url.Parse(spec.BaseUrl)
+	identity, err := runtime.ParseURLToIdentity(spec.BaseUrl)
```
matthiasbruns, line 57: in the ghrc.io example, we would have "dummy" instead of "https" in the scheme which would be cleared afterwards - is that desired?

I guess that `ghrc.io`  should be cached under the `https` scheme in case we have an entry with https and one without.

```go
func ParseURLAndAllowNoScheme(urlToParse string) (*url.URL, error) {
	const dummyScheme = "dummy"
	if !strings.Contains(urlToParse, "://") {
		urlToParse = dummyScheme + "://" + urlToParse
	}
	parsedURL, err := url.Parse(urlToParse)
	if err != nil {
		return nil, err
	}
	if parsedURL.Scheme == dummyScheme {
		parsedURL.Scheme = ""
	}
	return parsedURL, nil
}
```

matthiasbruns (commented):
https vs empty scheme should be discussed maybe?

bindings/go/oci/repository/provider/credential_cache.go
matthiasbruns: ghrc.io is just an example - my assumption is that if not scheme or port is provided, we would fallback to https - that's at least what feels most natural for me from a view of a user

jakobmoellerdev: in the example `ghcr.io` we would land with an identity without scheme because `dummy` is removed in the method. The "fallback" you describe is the default for the OCI distribution spec. It is then the clients responsibility to use the correct scheme (ORAS does this already)

matthiasbruns: gotcha . if this it the goal, I guess the change here would work just fine

jakobmoellerdev:
@fabianburth I think you had some concerns on testing, if you give me a pointer what / where to add one ill be happy to

otherwise merge is gonna be just fine I think

jakobmoellerdev:
Thanks for even going so far as writing the test ❤️

PR #799, opened 2025-09-05
chore: group docker package updates into a single pull request

Skarlso:

Turns out neither of the defaults or extends are dealing with groupping go modules related to docker repositories. I looked at extends, group defaults, recommended and default grouppings. Looked for extendable docker configs and presets. They mostly deal with Dockerfile image versions actually. And not go modules.

### DIFF
--- a/.github/config/renovate.json5
+++ b/.github/config/renovate.json5
@@ -1,3 +1,5 @@
+// Note: This is the top-level config for Renovate.
+// For repository-specific configuration please use the renovate.json5 file at the root of the repository.
 {
   "branchPrefix": "renovate/",
   "username": "ocmbot[bot]",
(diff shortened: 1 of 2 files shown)


.github/config/renovate.json5
```diff
@@ -17,6 +17,12 @@
     'config:recommended',
     'config:best-practices',
   ],
+  "packageRules": [
+    {
+      "matchPackageNames": ["github.com/docker/cli", "github.com/docker/docker"],
```
jakobmoellerdev, line 22: this is the wrong location for this! please use the top level one in https://github.com/open-component-model/open-component-model/blob/main/renovate.json5

Also probably worth writing a glob expression for github.com/docker

jakobmoellerdev (changes requested):

PR #844, opened 2025-09-09
feat: glob resolver repo impl

matthiasbruns:
In https://github.com/open-component-model/open-component-model/pull/780 we implemented a new configuration resolver implementation that resolves components with a glob matcher. This PR implement the repository that uses the new matcher

- the cli integration will be implemented in a follow up pr

Contributes https://github.com/open-component-model/ocm-project/issues/575


matthiasbruns:
todo: add util to add non-raw 
```go
	Repository *runtime.Raw `json:"repository"`
```

bindings/go/repository/component/resolver/v1/repository.go
```diff
@@ -0,0 +1,306 @@
...
+	}
+
+	return &ResolverRepository{
+		goRoutineLimit: options.GoRoutineLimit,
+
+		repositoryProvider: repositoryProvider,
+		credentialProvider: credentialProvider,
+
+		resolvers:                    resolvers,
+		repositoriesForResolverCache: make([]repository.ComponentVersionRepository, len(resolvers)),
```
jakobmoellerdev, line 81:
```suggestion
		repositoriesForResolverCache: make([]repository.ComponentVersionRepository, 0, len(resolvers)),
```

bindings/go/repository/component/resolver/v1/repository.go
```diff
@@ -0,0 +1,306 @@
...
+		matcherCache:                 matchers,
+	}, nil
+}
+
+func (r *ResolverRepository) GetComponentVersion(ctx context.Context, component, version string) (*descriptor.Descriptor, error) {
+	repos := r.RepositoriesForComponentIterator(ctx, component)
```
jakobmoellerdev, line 87: why do we need to iterate through repositories if this resolver can be deterministic at load time?

jakobmoellerdev: I think it would be much simpler to load the order once, not cache and then use that order statically. I dont see a lot of value in lazy loading the repos. just my opinion though, happy to hear your thoughts

bindings/go/repository/component/resolver/v1/repository.go
```diff
@@ -0,0 +1,306 @@
...
+
+func (r *ResolverRepository) GetResolvers() []*resolverspec.Resolver {
+	return deepCopyResolvers(r.resolvers)
+}
+
+func (r *ResolverRepository) matchesAnyResolver(component string) bool {
+	r.matcherCacheMu.RLock()
+	defer r.matcherCacheMu.RUnlock()
+	for _, m := range r.matcherCache {
+		if m.Match(component, "") {
```
jakobmoellerdev, line 181: is this matching safe? wondering if empty string can cause side effects

matthiasbruns: version is not implemented in the binding but prepared in the signature - its `version`

```go
func (m *ResolverMatcher) Match(componentName, version string) bool {
	return m.componentMatcher.Match(componentName)
}
```

matthiasbruns:
@jakobmoellerdev  ping

bindings/go/repository/component/resolver/v1alpha1/repository.go
```diff
@@ -0,0 +1,325 @@
...
+			versionsMu.Lock()
+			defer versionsMu.Unlock()
+			for _, version := range versions {
+				if r.matchesAnyResolver(component) {
+					accumulatedVersions[version] = struct{}{}
+				}
+			}
+			return nil
+		})
+	}
```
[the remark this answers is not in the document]
matthiasbruns: https://github.com/open-component-model/open-component-model/pull/898

PR #859, opened 2025-09-09
Rsa/handler

jakobmoellerdev:

- Added a new Go module in `bindings/go/rsa` to support RSA signing configurations.
- Implemented configurations for both `RSASSA-PSS` (probabilistic) and `RSASSA-PKCS1 v1.5` (deterministic) signature schemes.
- Introduced support for signature encoding policies (`Plain` and `PEM`) to manage signature serialization and storage with both supported formats from OCMv1
- Added logic for media types, algorithms, and integration with runtime schemas.
- Autogenerated necessary deepcopy and runtime type management code for reusability.
- Wrote an extensive test suite for all possible signing/verification combinations available

fix https://github.com/open-component-model/ocm-project/issues/648


bindings/go/rsa/signing/handler/handler.go
```diff
@@ -0,0 +1,292 @@
...
+		}
+		if len(chain) == 0 {
+			return errors.New("pem signature missing certificate chain")
+		}
+		leaf := chain[0]
+		rsaPub, ok := leaf.PublicKey.(*rsa.PublicKey)
+		if !ok {
+			return errors.New("leaf cert public key is not RSA")
+		}
+		if err := verifyChainWithOptionalAnchor(leaf, chain[1:], underlying, h.roots, h.now); err != nil {
```
matthiasbruns, line 169: The pubslice could panic

bindings/go/rsa/signing/handler/handler.go
```diff
@@ -0,0 +1,292 @@
...
+	switch supported.GetSignatureEncodingPolicy() {
+	case v1alpha1.SignatureEncodingPolicyPlain:
+		return descruntime.SignatureInfo{
+			Algorithm: algorithm,
+			MediaType: supported.GetDefaultMediaType(),
+			Value:     hex.EncodeToString(rawSig),
+		}, nil
+
+	case v1alpha1.SignatureEncodingPolicyPEM:
+		fallthrough
```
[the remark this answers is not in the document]
jakobmoellerdev: this was more for explicit defaulting, so more of a style choice. I agree it can be removed though

bindings/go/rsa/signing/handler/handler.go
```diff
@@ -0,0 +1,292 @@
...
+
+// Sign produces a signature for the given digest, using RSA and the configured
+// algorithm and encoding policy. For PEM encoding, the certificate chain is
+// read from credentials and embedded into the SIGNATURE block.
+func (*Handler) Sign(
+	_ context.Context,
+	unsigned descruntime.Digest,
+	rawCfg runtime.Typed,
+	creds map[string]string,
+) (descruntime.SignatureInfo, error) {
```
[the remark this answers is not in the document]
jakobmoellerdev: in general I think there is no guidance here yet. I will say that most likely this should change to pointers because we will have to potentially deal with hundreds of signatures in which case we dont want them to all be on the stack imo. Also when dealing with JSON serialization we anyhow land on heap.

bindings/go/rsa/signing/handler/handler.go
jakobmoellerdev: actually, not sure what could panic, mind explaining?

```go
sig, algFromPEM, chain, err := rsasignature.GetSignatureFromPem([]byte(signed.Signature.Value))
if err != nil {
    return fmt.Errorf("parse pem signature: %w", err)
}
if len(chain) == 0 {
    return errors.New("pem signature missing certificate chain")
}
leaf := chain[0]
...
if err := verifyChainWithOptionalAnchor(leaf, chain[1:], underlying, h.roots, h.now); err != nil {
    ...
}
``` 

- check len(chain) == 0 before accessing chain[0].
- If len(chain) == 1, then chain[1:] is a valid empty slice, not a panic.
- If len(chain) > 1, then chain[1:] is the remainder of the slice.

matthiasbruns: `chain[1:]` I wasn't too sure about this part, but yeah after testing it you are right.

bindings/go/rsa/signing/handler/internal/pem/pem.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+// ParseRSAPublicKeyPEM scans concatenated PEM data and returns the first RSA
+// public key found plus the original parsed object it came from:
+//   - For "PUBLIC KEY": returns (*rsa.PublicKey, crypto.PublicKey)
+//   - For "RSA PUBLIC KEY": returns (*rsa.PublicKey, *rsa.PublicKey)
+//   - For "CERTIFICATE": returns (*rsa.PublicKey, *x509.Certificate)
+//
+// If none can be parsed it returns (nil, nil).
+func ParseRSAPublicKeyPEM(pemBytes []byte) (*rsa.PublicKey, any) {
```
matthiasbruns, line 54: While I get any here, an alternative would be to return a struct like this


```go
type ParsesRSAPublicKeyPEM struct {
  KeyType KeyType // const of PublicKey, RSA and Cert
  PublicKey *crypto.PublicKey
  RSA *rsa.PublicKey
  Cert *x509.Certificate


```

bindings/go/rsa/signing/handler/internal/pem/pem_signature.go
```diff
@@ -0,0 +1,93 @@
...
+// blocks as a chain.
+//
+// Returns:
+//   - signature: the bytes from the first SIGNATURE block if present, otherwise nil
+//   - algo: the value of SignaturePEMBlockAlgorithmHeader if present
+//   - chain: parsed certificates that follow (or are present in the input)
+//   - err: parsing errors (including malformed PEM or certificates)
+//
+// Empty pemData returns all-zero values and no error.
+func GetSignatureFromPem(pemData []byte) ([]byte, string, []*x509.Certificate, error) {
```
matthiasbruns, line 63: Could named returns make sense here?

matthiasbruns (commented):
beautiful

bindings/go/rsa/go.mod
```diff
@@ -0,0 +1,27 @@
+module ocm.software/open-component-model/bindings/go/rsa
+
+go 1.25.0
```
Skarlso, line 3: New modules should probably use 1.25.1. WDYT?

bindings/go/rsa/signing/handler/handler.go
```diff
@@ -0,0 +1,292 @@
...
+		err   error
+	)
+	if useSystemRoots {
+		roots, err = x509.SystemCertPool()
+		if err != nil {
+			return nil, fmt.Errorf("load system roots: %w", err)
+		}
+	} else {
+		roots = x509.NewCertPool()
+	}
```
Skarlso, line 61:
```suggestion
	var (
		roots = x509.NewCertPool()
		err   error
	)
	if useSystemRoots {
		roots, err = x509.SystemCertPool()
		if err != nil {
			return nil, fmt.Errorf("load system roots: %w", err)
		}
	}
```

Tiny nit.

bindings/go/rsa/signing/handler/internal/dn/dn.go
```diff
@@ -0,0 +1,117 @@
+// Package dn provides utilities for parsing and matching X.509 Distinguished Names (DNs).
```
Skarlso, line 1: Do... we... need all this? https://pkg.go.dev/crypto/x509 surely has most of this covered?

bindings/go/rsa/signing/handler/handler.go
```diff
@@ -0,0 +1,292 @@
...
+	ErrMissingPublicKey  = errors.New("missing public key, required for plain RSA signatures")
+)
+
+// Handler holds trust anchors and time source for X.509 validation.
+type Handler struct {
+	roots *x509.CertPool
+	now   func() time.Time
+}
+
+// New returns a Handler. If useSystemRoots is true, system trust roots are loaded.
```
Skarlso, line 48: Clarify that otherwise empty certpool is used?

bindings/go/rsa/signing/handler/handler.go
```diff
@@ -0,0 +1,292 @@
+// Package handler implements RSA signing and verification for OCM.
+// It supports both RSASSA-PSS and RSASSA-PKCS1-v1_5, and two encodings:
+//  1. Plain: hex signature bytes without certificates.
```
Skarlso, line 3: Nit: technically, the private key can use a certificate it's just not embedded in the output. Right?

bindings/go/rsa/signing/handler/internal/dn/dn.go
```diff
@@ -0,0 +1,117 @@
...
+	"STREET":     func(n *pkix.Name, v string) { n.StreetAddress = append(n.StreetAddress, v) },
+	"POSTALCODE": func(n *pkix.Name, v string) { n.PostalCode = append(n.PostalCode, v) },
+	"SN":         func(n *pkix.Name, v string) { n.SerialNumber = v },
+	"CN":         func(n *pkix.Name, v string) { n.CommonName = v },
+}
+
+// dnSplitRunes defines characters used to split DN components.
+var dnSplitRunes = []rune{'/', ';', ',', '+'}
+
+// Parse converts a string representation of a distinguished name into a pkix.Name.
```
Skarlso, line 27: Nit: Also add the comment about the fallback behaviour at the end of this function.

Skarlso (commented):
Couple of first pass comments.

PR #881, opened 2025-09-12
feat: ADR versioning strategy for monorepo components

morri-son:

Having multiple components own the same repository requires a proper versioning strategy. This ADR lists multiple options.


docs/adr/0009_versioning_strategy.md
```diff
@@ -0,0 +1,299 @@
...
+* Support straightforward hotfix and backport workflows for older minor releases.
+* Provide a machine-readable snapshot for documentation and website consumption.
+* Keep CI complexity manageable and deterministic.
+
+## Decision
+
+We adopt a combined, opinionated strategy:
+
+* Versioning: Hybrid approach — each component maintains a local `VERSION` file; the repository also maintains a generated `/VERSIONS.yaml` that mirrors the last published versions of all components (aka snapshot)
+* Branching: Component-scoped persistent release branches — branches follow the pattern `releases/<component>/v<major>.<minor>` and act as maintenance lines for each active minor release. Patches and hotfixes are applied via PRs to these branches.
```
jakobmoellerdev, line 28: give an example here

docs/adr/0009_versioning_strategy.md
```diff
@@ -0,0 +1,299 @@
...
+* Publishing: An annotated tag `<component>/v<major>.<minor>.<patch>` is the authoritative record of a published artifact.
```
jakobmoellerdev, line 29: you mean of the source of an artifact right?

docs/adr/0009_versioning_strategy.md
```diff
@@ -0,0 +1,299 @@
...
+
+### Normal Case (Automatic)
+
+The `ocm` root component version follows these automated rules for independent sub-component releases:
+
+* **Patch bump (x.y.z → z++)**: triggered by any sub-component patch release
+* **Minor bump (x.y.z → y++)**: triggered by any sub-component feature addition  
+* **Major bump (x.y.z → x++)**: triggered by any sub-component new major release
+
+Every OCM root release represents a reproducible snapshot of all sub-component versions and maintains strict SemVer compliance. The root component version is automatically computed and updated in `/VERSIONS.yaml` when any sub-component publishes a release.
```
jakobmoellerdev, line 41: I think it could be better to always create RCs for the root component. that way, we never need to have the `supress-root-release` case you were defining in the edge case below. Instead everything would be a build candidate and we can choose any build candidate to be promoted.

docs/adr/0009_versioning_strategy.md
```diff
@@ -0,0 +1,299 @@
...
+
+   Execute the GitHub Action release workflow for each affected sub-component with the "suppress-root-release" flag enabled (flag name to be finalized).
+
+   * Run release workflow for `ocm-cli` with flag: `suppress-root-release = true`
+   * Run release workflow for `ocm-controller` with flag: `suppress-root-release = true`
+   * Each workflow completes its normal release process (tag creation, artifact publishing) but skips the automatic OCM root release trigger.
+
+2. **Manual root release consolidation:**
+
+   After all coordinated sub-component releases are completed, manually trigger the "Create OCM Root Release" workflow:
```
jakobmoellerdev, line 61: I think this should be a promotion of a release candidate set, and not the creation of a new component set

docs/adr/0009_versioning_strategy.md
```diff
@@ -0,0 +1,299 @@
...
+
+   * Reads current `/VERSIONS.yaml` state to identify all unreleased changes
```
jakobmoellerdev, line 63: where is this VERSIONS.yaml and how does it look/where is it maintained?

docs/adr/0009_versioning_strategy.md
```diff
@@ -0,0 +1,299 @@
...
+   * Computes appropriate bump type (highest sub-component bump type across all coordinated releases)
```
jakobmoellerdev, line 64: Agreed, but this should have an example

docs/adr/0009_versioning_strategy.md
```diff
@@ -0,0 +1,299 @@
...
+   * Creates single OCM release tag containing all coordinated changes
+
+**Bump calculation for coordinated releases:** The root component bump type is determined by the highest impact change across all coordinated sub-components (e.g., if one patch + one minor → minor bump for root).
+
+## Considered Options
+
+I explicitly considered the following options before making the decision above. Recording them here improves traceability.
```
jakobmoellerdev, line 71:
```suggestion
We explicitly considered the following options before making the decision above. Recording them here improves traceability.
```

Speak in TSC Authors person

docs/adr/0009_versioning_strategy.md
```diff
@@ -0,0 +1,299 @@
...
+   * Run release workflow for `ocm-controller` with flag: `suppress-root-release = true`
+   * Each workflow completes its normal release process (tag creation, artifact publishing) but skips the automatic OCM root release trigger.
+
+2. **Manual root release consolidation:**
+
+   After all coordinated sub-component releases are completed, manually trigger the "Create OCM Root Release" workflow:
+
+   * Reads current `/VERSIONS.yaml` state to identify all unreleased changes
+   * Computes appropriate bump type (highest sub-component bump type across all coordinated releases)
+   * Creates single OCM release tag containing all coordinated changes
```
jakobmoellerdev, line 65: where does this tag live? the example above was only for subcomponents as far as I can see

docs/adr/0009_versioning_strategy.md
```diff
@@ -0,0 +1,299 @@
...
+
+## Considered Options
+
+I explicitly considered the following options before making the decision above. Recording them here improves traceability.
+
+### Versioning options
+
+* Per-component `VERSION` file: each component contains a `VERSION` file in its directory that is the working state for that component.
+* Central `VERSIONS.yaml`: a single root `VERSIONS.yaml` that is the canonical snapshot of component versions (human-edited or bot-maintained).
+* Hybrid: per-component `VERSION` files for working state, plus a generated `/VERSIONS.yaml` for published snapshots.
```
jakobmoellerdev, line 77: I dont think this makes sense. the generated VERSIONS.yaml would be unnecessary if it would always contain the latest versions. The only time where a dedicated file would make sense is if you cannot generate it

docs/adr/0009_versioning_strategy.md
```diff
@@ -0,0 +1,299 @@
...
+### Rejection rationale (short)
+
+* Central-only `VERSIONS.yaml` + tag-only branching: rejected because hotfix/backport workflows become cumbersome and the central file becomes a high-conflict surface.
+* Per-component `VERSION` without central snapshot: rejected because consumers (website, docs) need a single reproducible snapshot; lacking that reduces discoverability and reproducibility of root releases.
+
+## Operational Rules and Rationale
+
+### Authoritative sources
+
+* Working state (pre-publication): `ocm-<component>/VERSION` located on the appropriate branch (either `main` for ongoing development, or a `releases/...` branch for maintenance lines).
```
jakobmoellerdev, line 101: why would it be prefixed with `ocm-` ?

docs/adr/0009_versioning_strategy.md
```diff
@@ -0,0 +1,299 @@
...
+* Published state: annotated Git tag `<component>/v<major>.<minor>.<patch>` — tags are authoritative for published artifacts and their provenance.
+* Global snapshot: `/VERSIONS.yaml` — generated by CI after successful publishes and updated via OCM bot PRs; **this file is not intended for manual edits**.
+
+Rationale: Having both per-component `VERSION` files (working state) and a generated `/VERSIONS.yaml` provides autonomy and a single snapshot for consumers.
```
jakobmoellerdev, line 105: how does this provide autonomy?

docs/adr/0009_versioning_strategy.md
```diff
@@ -0,0 +1,299 @@
...
+
+### Branching model
+
+* Create and maintain persistent branches for each active minor per component using the naming pattern: `releases/<component>/v<major>.<minor>`.
+* `main` remains the integration branch for new features and non-release work.
+* Release preparation: prepare a release branch from `main` when a new minor version should be released.
+* Hotfixes: open PRs targeting the applicable `releases/<component>/vX.Y` branch; merge triggers a patch release from that branch.
+
+### Special Case: OCM Root Component (No Release Branches)
```
jakobmoellerdev, line 114: I disagree with a few statements here:

1. Why would we not have OCM-specific hotfixes? If someone wants to ship OCM, we would ship both controllers and CLI, and not just one right? that means hotfixes have to be backported just the same
2. What use do we have for a snapshot-only component? We need to do conformance testing on this version, so IMO its important to also release this version and make it our single source of truth for availability.

morri-son: 1. Having the submodules released should bump the ocm version automatically. Having the edge-case of requiring multiple submodules together, would be handled manually, by flagging the releases of the submodules as without bumping the OCM root component. The bump will happen via another workflow. I'm not sure if such - hopefully rare - use cases can be coordinated automatically, therefore kept them with flags/manual workflows. So, we're able to create hotfixes for OCM, but by hot fixing any of the submodules

2. I agree that testing is required, but why does this make an own release of ocm necessary and what is the benefit of introducing releases and related branches for ocm?

docs/adr/0009_versioning_strategy.md
```diff
@@ -0,0 +1,299 @@
...
+* **Simplified workflow** - OCM releases are triggered by sub-component releases, not by independent development
+* **Maintenance overhead reduction** - no additional branches to manage for a component that has no independent lifecycle
+
+OCM release process: `main` + `/VERSIONS.yaml` update → `ocm/vX.Y.Z` tag (no intermediate branch required).
+
+Rationale: Persistent release branches simplify backports and maintenance for components with independent code; they reduce the operational friction when maintaining older minor versions. For snapshot-only components like OCM, tags from main are sufficient.
+
+### CI and automation responsibilities
+
+#### Workflow: `create-release-branch`
```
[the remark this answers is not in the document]
morri-son: yes. I see this action as initator, as the first action to be done when new releases should be created.

docs/adr/0009_versioning_strategy.md
morri-son: correct

docs/adr/0009_versioning_strategy.md
morri-son: as mentioned above, I buy the idea of using the root component, its release branch and RC tags as a possible solution, but we need to sort out how to deal with edge-cases, both for documentation as well as having the possibility to automate this.

docs/adr/0009_versioning_strategy.md
jakobmoellerdev: then its time for rock-solid CI imho. This will be the heart of our delivery flow and it must "just work" because we will use it weekly, if not daily

docs/adr/0009_versioning_strategy.md
jakobmoellerdev: +1 for this, but I think you could state this explicitly (for me it was not clear right away)

docs/adr/0009_versioning_strategy.md
jakobmoellerdev: For me separate testing is necessary for conformance and so that not every part of ocm has to integrate conformance for everything. Instead we can choose to have signals on top level that only push down alarms for subcomponents if they fail. See the kubernetes test matrix as an example

docs/adr/0009_versioning_strategy.md
morri-son: as mentioned above, when changing the proposal and implementing using tags, we can use code to calculate all that on-the-fly.

docs/adr/0009_versioning_strategy.md
morri-son: sounds resonsonable. Let me think this through and come up with a better proposal :-)

docs/adr/0009_versioning_strategy.md
morri-son: and then simply don't care about the new release that has been created in releases/ocm/v1.4 branch? By the way, we don't create PRs for Hotfixes, patches, do we? Then there would anyway be no RC for ocm for a submodule that triggered the patch. I'm confused :-)

docs/adr/0009_versioning_strategy.md
```diff
@@ -0,0 +1,305 @@
...
+  * **Override:** Skip root release if `--suppress-root-release` flag is set (for coordinated releases)
+* Update `/VERSIONS.yaml` via OCM bot PR with new sub-component version and computed root version (if applicable)
+* Failure handling: if VERSIONS.yaml update conflicts, stop pipeline and alert maintainers
+
+**Manual Parameters:**
+
+* `suppress-root-release`: Boolean flag to prevent automatic OCM root release (for coordinated multi-component releases as described [above](#edge-case-coordinated-multi-component-releases))
+* `release_candidate`: Boolean flag to create a release candidate tag (e.g., `v0.30.0-rc.1`) instead of a final release
+* `release_candidate_name`: Release candidate suffix (e.g., `rc.1`, `rc.2`) when `release_candidate` is true
+
```
Skarlso, line 198: The less parameters the less failure prone this whole thing will be. I don't want a typo to kill the entire release process.

docs/adr/0009_versioning_strategy.md
```diff
@@ -0,0 +1,305 @@
...
+* Run component-specific tests and verification
+* Read `cli/VERSION` or `kubernetes/controller/VERSION` (fall back to latest component tag if absent)
+* Compute new patch number using release tooling
+* Create annotated tag `<component>/vX.Y.Z` using OCM bot
+* Build and publish artifacts (binaries, container images, component descriptors)
+* **Root component handling (configurable):**
+  * **Default:** Calculate root component SemVer bump and trigger automatic `ocm` release
+  * **Override:** Skip root release if `--suppress-root-release` flag is set (for coordinated releases)
+* Update `/VERSIONS.yaml` via OCM bot PR with new sub-component version and computed root version (if applicable)
+* Failure handling: if VERSIONS.yaml update conflicts, stop pipeline and alert maintainers
```
Skarlso, line 191: I hate the entire "push back an updated version file into the repository after a release" flow with a passion. Let's not replicate it.

docs/adr/0009_versioning_strategy.md
```diff
@@ -0,0 +1,305 @@
...
+**Solution:** Manual override workflow for coordinated releases:
+
+1. **Sub-component releases with root suppression:**
+
+   Execute the GitHub Action release workflow for each affected sub-component with the "suppress-root-release" flag enabled (flag name to be finalized).
+
+   * Run release workflow for `ocm-cli` with flag: `suppress-root-release = true`
+   * Run release workflow for `ocm-controller` with flag: `suppress-root-release = true`
+   * Each workflow completes its normal release process (tag creation, artifact publishing) but skips the automatic OCM root release trigger.
+
```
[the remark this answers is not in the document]
morri-son: I reworked this now completely.

docs/adr/0009_versioning_branching_releasing.md
```diff
@@ -0,0 +1,465 @@
...
+* **Simplicity**: Git tags are Git-native and require no additional file maintenance
+* **Auditability**: Complete version history is preserved in Git repository
+* **Go Module Compatibility**: Tag naming follows Go module conventions exactly
+* **Reduced Merge Conflicts**: No version files to conflict during parallel releases
+* **Single Source of Truth**: Git tags eliminate version drift between files and actual releases
+* **OCM Native**: Component Constructor used as version matrix follows OCM specification patterns
+
+Cons:
+
+* **Tag Dependency**: CI/CD pipelines must be robust in Git tag parsing and validation
```
matthiasbruns, line 458: I would add the branching overhead as a con

docs/adr/0009_versioning_branching_releasing.md
```diff
@@ -0,0 +1,465 @@
...
+- cli/v0.30.1                    # CLI component release
+- kubernetes/controller/v0.29.2  # Controller component release
+- ocm/v0.11.0                    # OCM root release (aggregates sub-components)
+```
+
+### Release Branch Strategy
+
+Sub-components use persistent release branches for maintenance:
+
+* `releases/cli/v0.30` - CLI v0.30.x maintenance line
```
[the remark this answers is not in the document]
morri-son: the concept has just been taken over from the current release process for the OCM CLI: releases are delivered from maintenance branches and act as line for delivering patches. What would you propose instead?

docs/adr/0009_versioning_branching_releasing.md
```diff
@@ -0,0 +1,465 @@
...
+`ocm` root component releases directly from `main` branch with no release branches.
+
+## Component Relationship Model
+
+### Component Independence
+
+* **No Hard Dependencies**: CLI and Controller components are developed and released independently with no direct runtime dependencies between them
+* **Bundle Validation**: `ocm` root component validates that specific combinations of sub-components work together through integration testing
+* **Breaking Changes Policy**: Following Kubernetes model, breaking changes are permitted in minor releases across all components
+* **Compatibility Scope**: Version compatibility exists only within individual sub-components (e.g., cli/v0.30.x patch compatibility), not across different components
```
Skarlso, line 80: I don't understand this sentence. :D What other different components are there other than the individual sub-components?

docs/adr/0009_versioning_branching_releasing.md
```diff
@@ -0,0 +1,465 @@
...
+        version: ${OCM_CLI_VERSION}  # will be resolved at release time
+
+      - name: controller
+        componentName: ocm.software/kubernetes/controller
+        version: ${OCM_CONTROLLER_VERSION}  # will be resolved at release time
+```
+
+### OCM Root Component Release Process
+
+**Key Principle**: Any sub-component release of type **patch** and **minor** triggers an OCM **minor** version bump, as each represents a new combination of tested components. A **major** release of a sub-component triggers an OCM **major** version bump.
```
Skarlso, line 173: That kinda seems excessive on a patch version? 🤔 Even if it is a patch version we release a new minor version? That means that every release of the root ocm component is potentially a breaking change. What's the point of a patch release for a component then?

docs/adr/0009_versioning_branching_releasing.md
```diff
@@ -0,0 +1,465 @@
...
+
+**Purpose**: Initialize a new persistent release branch for a component's minor version line
+
+**Required Steps**:
+
+1. **Branch validation**: Verify that `releases/<component>/v<major>.<minor>` doesn't already exist
+2. **Source determination**: Create branch from `main` for new minor version release
+3. **Branch creation**: Create `releases/<component>/v<major>.<minor>` branch using OCM bot
+4. **Protection rules**: Apply branch protection rules identical to `main` branch (require PR reviews, status checks, admin override settings)
+5. **Notification**: Alert component maintainers of new release branch creation
```
Skarlso, line 235: Will there be a maintainers file per component or some other structure? This is the first time I'm hearing about individual maintainers / components and notifications based on that.

docs/adr/0009_versioning_branching_releasing.md
```diff
@@ -0,0 +1,465 @@
...
+  * `ocm/v0.11.0` (OCM root release)
+
+## Operational Rules and Implementation Details
+
+### Authoritative Sources
+
+* **Published state**: Annotated Git tag `<component>/v<major>.<minor>.<patch>` — tags are authoritative for published artifacts and their provenance
+* **OCM version matrix**: `/ocm/component-constructor.yaml` — OCM Component Constructor defining sub-component versions and relationships for OCM releases
+
+*As described in the strategy above, no VERSION files are maintained - all versions are derived from Git tags.*
```
Skarlso, line 327: Do we set the tags to immutable / not deletable?

docs/adr/0009_versioning_branching_releasing.md
```diff
@@ -0,0 +1,465 @@
...
+* `controller` -> `ocm.software/kubernetes/controller`
+
+We require a single, operational specification that defines how components are versioned, how release branches are created and maintained, and how CI publishes releases.
+
+## Decision Drivers
+
+* Reproducible, auditable published releases.
+* Allow a release per component without forcing simultaneous releases of all components.
+* Support simple and robust hotfix and backport workflows for older minor releases.
+* Provide a machine-readable snapshot for documentation and website consumption.
```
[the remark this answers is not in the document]
morri-son: The website refers to the ocm root component version, but the reference docs are component-specific. We use Hugo modules for the reference docs, so we need to know what version of the submodule we need to relate to the website. Right now that action (e.g. hugo mod get ocm.software/open-component-model/cli@v0.30.0) on website updates are done manually. Having this information about the version matrix in a machine-readable format, offers to automate this later.

docs/adr/0009_versioning_branching_releasing.md
```diff
@@ -0,0 +1,465 @@
...
+The strategy includes dedicated conformance testing at the `ocm` root component level to ensure component interoperability and validate integrated component combinations before final releases of the `ocm` root component. The tests will be created over time and are not mandatory for the initial rollout of this strategy.
+
+## Description
+
+We adopt a Git-native, tag-based versioning strategy that eliminates version files in favor of authoritative Git tags:
+
+* **Versioning**: Git tags are the single source of truth for component versions - no VERSION files are maintained in the repository
+* **Branching**: Component-scoped persistent release branches following the pattern `releases/<component>/v<major>.<minor>` for maintenance lines
+* **Publishing**: Annotated Git tags with Go module-compatible naming (`<component>/v<major>.<minor>.<patch>`) serve as the authoritative record of published artifacts
+* **OCM Root Component**: Uses OCM Component Constructor YAML as version matrix and releases directly from `main` branch without release branches
```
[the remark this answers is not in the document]
morri-son: If you check a bit below, the idea is to always create new minor versions and RCs for the ocm root compopnent, whenever an increase in minor/patch happens on one of its sub-components.

docs/adr/0009_versioning_branching_releasing.md
```diff
@@ -0,0 +1,465 @@
...
+* `releases/cli/v0.30` - CLI v0.30.x maintenance line
+* `releases/kubernetes/controller/v0.29` - Controller v0.29.x maintenance line
+
+`ocm` root component releases directly from `main` branch with no release branches.
+
+## Component Relationship Model
+
+### Component Independence
+
+* **No Hard Dependencies**: CLI and Controller components are developed and released independently with no direct runtime dependencies between them
```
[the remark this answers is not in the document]
morri-son: I meant that controller and cli don't depend on each other.

docs/adr/0009_versioning_branching_releasing.md
```diff
@@ -0,0 +1,465 @@
...
+* **Bundle Validation**: `ocm` root component validates that specific combinations of sub-components work together through integration testing
+* **Breaking Changes Policy**: Following Kubernetes model, breaking changes are permitted in minor releases across all components
+* **Compatibility Scope**: Version compatibility exists only within individual sub-components (e.g., cli/v0.30.x patch compatibility), not across different components
+
+### OCM Root as Integration Bundle
+
+* **Purpose**: `ocm` root component serves as a tested bundle of working sub-component releases, containing no independent business logic
+* **Validation**: Conformance tests ensure the specific combination of sub-component versions operates correctly together
+* **Version Matrix**: Component Constructor YAML defines the tested combination without imposing runtime dependencies
```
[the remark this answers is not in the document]
morri-son: you're right. I removed the complete line.

docs/adr/0009_versioning_branching_releasing.md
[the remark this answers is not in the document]
Skarlso: If we always increase minor version what's the point of patch versions for the sub-components?

docs/adr/0009_versioning_branching_releasing.md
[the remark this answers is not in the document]
Skarlso: Please clarify that then 🙏 🙇

docs/adr/0009_versioning_branching_releasing.md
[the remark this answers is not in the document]
morri-son: @jakobmoellerdev said, let's keep it simple at the beginning 🙂. A former proposal also contained patches for the root component and only increases the minor of ocm if at least one of the sub-components also increased its minor.

docs/adr/0009_versioning_branching_releasing.md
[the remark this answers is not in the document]
morri-son: done

docs/adr/0009_versioning_branching_releasing.md
morri-son: See my other comment. @jakobmoellerdev asked to keep it the setup for the root component simple and always increase the minor version

docs/adr/0009_versioning_branching_releasing.md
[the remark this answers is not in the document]
morri-son: done

docs/adr/0009_versioning_branching_releasing.md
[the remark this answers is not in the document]
Skarlso: Okay I can accept that.

docs/adr/0009_versioning_branching_releasing.md
```diff
@@ -0,0 +1,413 @@
...
+
+#### Release Workflow for Sub-Components
+
+Sub-components follow a two-phase release process:
+
+##### Phase 1: Release Candidate Creation
+
+* Create and test release candidates from release branches
+* Generate RC tags for validation: `<component>/v<major>.<minor>.<patch>-rc.N`
+* Extended testing and validation period before final release
```
ikhandamirov, line 112:
```suggestion
* Extended period of testing and validation by external stakeholders before final release
```

docs/adr/0009_versioning_branching_releasing.md
```diff
@@ -0,0 +1,413 @@
...
+
+**Release Workflow**:
+
+1. Sub-component releases (e.g., `cli/v0.30.1`)
+2. **RC Creation Logic**:
+   * Check for existing OCM RCs for target version
+   * If no final release exists: Create new RC with incremented number
+   * If final release exists: Create RC for next version
+3. Update component constructor with new sub-component versions
+4. Run integration conformance tests (if available)
```
ikhandamirov, line 183: If there are those integration tests, why don't we run them when releasing subcomponent candidates? Why waiting for the release of the root?

docs/adr/0009_versioning_branching_releasing.md
```diff
@@ -0,0 +1,413 @@
...
+- 09:00 - Create cli/v0.31.0-rc.1
+- 10:00 - Promote to cli/v0.31.0 (final)
+- 10:01 - Auto-create ocm/v0.11.0-rc.1 with new CLI version
+- 10:05 - Run conformance tests
+
+Day 3: OCM Release
+- 09:00 - Promote ocm/v0.11.0-rc.1 → ocm/v0.11.0 (final)
+```
+
+### Example 2: Coordinated Multi-Component Release
```
ikhandamirov, line 354: Do we want to stick to biweekly releases? Is the regular duration of the grace period for release candidates also two weeks? In case so, can you please update the example accordingly?

ikhandamirov (commented):
Very comprehensive description!
I'm a bit concerned about having two levels of release candidates. Do we really need them on subcomponent level?

docs/adr/0009_versioning_branching_releasing.md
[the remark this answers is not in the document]
matthiasbruns: I wasn't there when you talked about this process. I am just asking if there might be a tag based on not branch based way of doing this?

docs/adr/0009_versioning_branching_releasing.md
morri-son: I didn't cover this aspect of release cadence so far. My assumption is that we stick to bi-weekly releases and the  opt-out for a release in case there's nothing worth or ready to be released. The [current release process doc](https://github.com/open-component-model/ocm/blob/main/RELEASE_PROCESS.md) already contains "Agreeing with the team, if a new release has to be produced. Sometimes, e.g. if there were no significant contributions, or the branch is not ready, the release can be skipped".

docs/adr/0009_versioning_branching_releasing.md
morri-son: if we're only talking about a smaller number of conformance tests that also don't take long to run, we can execute them already with the tests for sub-components. Failing early in a sub-component even before trying to create a new release for the ocm root component makes absolute sense. Integration and conformance tests are normally more complex and take longer. 

With just two components, no existing conformance tests right now and therefore no data about conformance test runs, I think it is a viable option to plan the execution of such tests with the sub-components and decide at a later point in time if the test execution should be moved completely to the ocm root component.

docs/adr/0009_versioning_branching_releasing.md
morri-son: I took over your proposal and also added the timeframe of one sprint (two weeks).

docs/adr/0009_versioning_branching_releasing.md
```diff
@@ -0,0 +1,413 @@
...
+```
+
+## Pros and Cons of the Proposal
+
+Pros:
+
+* **Simplicity**: Git tags are Git-native and require no additional file maintenance
+* **Auditability**: Complete version history is preserved in Git repository
+* **Go Module Compatibility**: Tag naming follows Go module conventions exactly
+* **Reduced Merge Conflicts**: No version files to conflict during parallel releases
```
fabianburth, line 399: I'm not sure I understand this scenario. Since you propose to release from dedicated release branches for each component anyway, how could there be merge conflicts on parallel releases?  Besides, I suppose each component would have its own VERSION file? 
Since you emphasize so much on getting rid of the version file, do you have insights into why the legacy ocm repo chose to use them over tags?

docs/adr/0009_versioning_branching_releasing.md
```diff
@@ -0,0 +1,413 @@
...
+
+Key elements:
+
+* Git tags replace VERSION files as the authoritative source for component versions
+* Component-scoped release branches for maintenance (`releases/<component>/v<major>.<minor>`)
+* OCM Component Constructor YAML manages version matrix and releases from `main` branch
+* Conformance tests validate sub-component combinations
+* Immutable GitHub releases with protected tags and signed attestations ensure supply chain security
+
+All components will run conformance tests to ensure compatibility. Tests will be created over time and are not mandatory for initial rollout.
```
fabianburth, line 42: How are the cli and the controllers even dependent on each other? What would the compatibility test look like?

docs/adr/0009_versioning_branching_releasing.md
```diff
@@ -0,0 +1,413 @@
...
+## High-level Architecture
+
+All component versions are determined exclusively from Git tags:
+
+```text
+Repository Structure:
+├── cli/                     # OCM CLI component
+├── kubernetes/controller/   # OCM Controller component  
+├── ocm/                     # `ocm` root component
+│   ├── component-constructor.yaml  # Version matrix
```
fabianburth, line 53: As far as I understand, this would be cli version + controller version. Wouldn't it be much more interesting for most users to know library module composition of cli / controller release (so, ocm.software/binding/go/oci v0.13.0, ocm.software/binding/go/repository v0.14.0, ...)?

docs/adr/0009_versioning_branching_releasing.md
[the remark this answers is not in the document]
fabianburth: I don't think there is, I believe you will always end up creating branches to deploy patches for older releases (although they could theoretically be temporary). 
What's the huge overhead in having release branches? I feel like the overhead is in having to maintain old releases, but unfortunately I don't see a way around this.

Gerald, I think you should mention the patches as rationale for the release branches in the ADR.

docs/adr/0009_versioning_branching_releasing.md
```diff
@@ -0,0 +1,413 @@
...
+* `releases/cli/v0.30` - CLI v0.30.x maintenance line
+* `releases/kubernetes/controller/v0.29` - Controller v0.29.x maintenance line
+
+`ocm` root component releases directly from `main` branch with no release branches.
+
+## Component Relationship Model
+
+### Component Independence
+
+* **No Direct Dependencies**: CLI and Controller are developed and released independently with no direct dependencies between them, though both share the common OCM library as their foundation
```
fabianburth, line 76: So, what are those conformance tests testing? Or are you also talking about the individual library modules? (maybe clarify this then) An example of such a component constructor would be awesome!

docs/adr/0009_versioning_branching_releasing.md
morri-son: You're right, makes no sense. I removed the line completely. I was not involved in the reasoning about a VERSION file over tags.

docs/adr/0009_versioning_branching_releasing.md
morri-son: Somehow Github merged this and the comment before and does not offer a way to comment the first one. Letz me do this together in this comment:

- In any case we require the ability to patch older releases. So, one thing to add in any case is our SLA for older release as shorter cycles directly lead to less admin effort.  I'm of the opinion that we can handle the overhead for release branches, even if we now have to create one for each component. That's why the ADR contains this proposal. But I find it valid to mention the option of cherry-picking and temporary branches for the same purpose. I'll bring this up in the meeting I plan for next week.

docs/adr/0009_versioning_branching_releasing.md
morri-son: Oh, do I still mention a direct dependency anywhere? That's wrong then. In line 76 I mention that there is no direct dependency, but only a sharing of the library: "**No Direct Dependencies**: CLI and Controller are developed and released independently with no direct dependencies between them, though both share the common OCM library as their foundation"

PR #882, opened 2025-09-13
fix(ctf): Add support for recognizing additional CTF archive files: tar.gz, tgz, tar (#651)

piotrjanik:

Previously, repository references ending with archive extensions (like `transport-archive.tar.gz`) could be misidentified as a hostname due to the presence of dots in the path. This fix ensures that archives such as tar.gz are correctly detected.

Fixes #651


cli/cmd/cmd_test.go
```diff
@@ -824,7 +824,7 @@ resources:
 				"--constructor", constructorYAMLFilePath,
 				"--repository", archiveFilePath,
 				"--working-directory", workingDir,
-			), test.WithErrorOutput(logs))
+			))
```
jakobmoellerdev, line 827: why did you remove this one?

cli/internal/reference/compref/compref.go
```diff
@@ -39,6 +39,9 @@ var (
 // DefaultPrefix is the default prefix used for component descriptors.
 const DefaultPrefix = "component-descriptors"
 
+// CtfArchiveExtensions is the list of archive file extensions that should be treated as CTF
+var CtfArchiveExtensions = [...]string{".tar.gz", ".tgz", ".tar"}
```
jakobmoellerdev, line 43:
```suggestion
var CtfArchiveExtensions = [...]string{".tar.gz", ".tgz", ".tar"}
```

why make this a fixed length array if its a public var? I suggest you either make it package private or leave it as slice for extensions by users.

jakobmoellerdev (commented):
The issue issue number you are linking points to a closed PR. Maybe you were trying to cross reference something else? Would you mind double checking that please?

cli/internal/reference/compref/compref.go
jakobmoellerdev: I think the reference parser is opinionated towards to CTF, OCI and CLI. I think this is fine for that reason

jakobmoellerdev (approved):

jakobmoellerdev:
@piotrjanik please sign your commits so this can be merged :)

PR #885, opened 2025-09-15
feat(plugin): introduce signing handler plugin support

jakobmoellerdev:
#### What this does
- Added a new `SigningRegistry` to manage signing handler plugins.
- Integrated signing handler support into the plugin manager and registries.
- Updated dependencies in `go.mod` and `go.sum` to include the signing library.
- Adjusted test cases to align with the new signing plugin functionality.

_Note: I have made it so that Signing and Verification expect the same Config Type right now. Technically this is a bit of a conflict with the SigStore ADR Proposal because it would have two different configs for signing and verifications but I think having one type that can be used with different fields would also work, and its significantly easier to implement here_

#### Why this change is needed

- Expands the plugin framework capabilities to include signing handlers for enhanced security and verification processes.

part of adopting the RSA handler implemented for https://github.com/open-component-model/ocm-project/issues/648 in https://github.com/open-component-model/open-component-model/pull/859. It would be registered as an internal handler via `RegisterInternalComponentSignatureHandler` in the CLI


bindings/go/plugin/manager/registries/signinghandler/endpoints_function_test.go
```diff
@@ -0,0 +1,108 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+	"ocm.software/open-component-model/bindings/go/plugin/internal/dummytype"
+	dummyv1 "ocm.software/open-component-model/bindings/go/plugin/internal/dummytype/v1"
+	"ocm.software/open-component-model/bindings/go/plugin/manager/contracts"
+	v1 "ocm.software/open-component-model/bindings/go/plugin/manager/contracts/signing/v1"
+	"ocm.software/open-component-model/bindings/go/plugin/manager/endpoints"
+	"ocm.software/open-component-model/bindings/go/plugin/manager/types"
+	"ocm.software/open-component-model/bindings/go/runtime"
```
Skarlso, line 15: Don't we usually order these in groups?

bindings/go/plugin/manager/registries/signinghandler/endpoints_function.go
```diff
@@ -0,0 +1,162 @@
...
+// handleSign handles the GetGlobalResource endpoint
+func handleSign[T runtime.Typed](plugin v1.SignerPluginContract[T]) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		var request v1.SignRequest[T]
+		if err := json.NewDecoder(r.Body).Decode(&request); err != nil {
+			handleError(w, err, http.StatusBadRequest, "failed to unmarshal request")
+			return
+		}
+
+		response, err := plugin.Sign(r.Context(), &request, make(map[string]string))
```
Skarlso, line 85: This is signing without credentials... is that right? 🤔

bindings/go/plugin/manager/registries/signinghandler/implementations.go
```diff
@@ -0,0 +1,145 @@
...
+	}
+
+	credHeader, err := toCredentials(credentials)
+	if err != nil {
+		return nil, fmt.Errorf("error converting credentials: %w", err)
+	}
+
+	var response v1.VerifyResponse
+	if err := plugins.Call(ctx, p.client, p.config.Type, p.location, Verify, http.MethodPost, plugins.WithPayload(request), plugins.WithResult(&response), plugins.WithHeader(credHeader)); err != nil {
+		return nil, fmt.Errorf("failed to sign from plugin %q: %w", p.ID, err)
```
Skarlso, line 119:
```suggestion
		return nil, fmt.Errorf("failed to verify from plugin %q: %w", p.ID, err)
```

bindings/go/plugin/manager/registries/signinghandler/endpoints_function.go
```diff
@@ -0,0 +1,162 @@
...
+			return
+		}
+
+		response, err := plugin.GetVerifierIdentity(r.Context(), &request)
+		if err != nil {
+			handleError(w, err, http.StatusInternalServerError, "failed to get identity")
+			return
+		}
+
+		handleJSONResponse(w, response)
```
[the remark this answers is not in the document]
jakobmoellerdev: I think there are a lot of things that belong to a shared logic but I dont want to do that here. I think all plugins would benefit from shared helpers for the endpoints_functions file no?

bindings/go/plugin/manager/registries/signinghandler/endpoints_function_test.go
jakobmoellerdev: fixed up manually

fabianburth (changes requested):

Skarlso (approved):

PR #846, opened 2025-09-09
feat: add remote repository support for helm input method

Skarlso:
#### What this PR does / why we need it

Related to https://github.com/open-component-model/ocm-project/issues/601

PR #902, opened 2025-09-19
chore: move loose parser into public oci package

Skarlso:

Needed for https://github.com/open-component-model/open-component-model/pull/846.

Since we need to parse Repository ref and we need the exposed Tag that this package provides. And it's also less strict than ORAS Ref Parsing.


Skarlso:
Errr.. what kind?

Skarlso:
Ah you mean a package comment?

fabianburth (approved):

PR #921, opened 2025-09-25
fix(rsa): update default signature encoding policy to SignatureEncodingPolicyDefault

jakobmoellerdev:

The default signature encoding policy was wrongly hard coded to PEM which caused empty Configs to incorrectly default to PEM even though the default is Plain

uses the default constant as originally intended


Skarlso:
I swear I commented this on the initial PR didn't I? Maybe I just wanted to...? I see it in my notes!!! FUCK :D I didn't comment it. :D

Issue #903, opened 2025-09-19
fix: dag is still failing sometimes


Skarlso:
**What happened**:

Here, multiple bugs piled up:
- dag race
- ctf index race: construct instantiated an individual ctf instance through the plugin manager for each components construction. These instances operated on the same ctf while having individual index locks. Since the construct calls were called concurrently, they raced. 
- ctf tagging issues (also surfacing as race condition): due to loose parser bug parsing references like `sha256:abc` which also surfaced as a race condition between `AddComponentVersion` and `ListComponentVersion`. `AddComponentVersion` initially tags with the wrongly parsed reference but later on overwrites that wrongly parsed reference with a valid version tag. So, the bug only surfaced when `ListComponentVersion` was called concurrently with `AddComponentVersion` and accessed the artifact index while it contained the invalid tags. 

error:

```
--- FAIL: Test_List_Component_Version_Variations_Recursive (0.01s)
    --- FAIL: Test_List_Component_Version_Variations_Recursive/Default_Options_(Table)_-_all_versions (0.00s)
panic: attribute descriptor not found in vertex name=ocm.software/leaf-a,version=0.0.1 [recovered, repanicked]

goroutine 121 [running]:
testing.tRunner.func1.2({0x9f3160, 0xc000347220})
	/opt/hostedtoolcache/go/1.25.0/x64/src/testing/testing.go:1872 +0x237
testing.tRunner.func1()
	/opt/hostedtoolcache/go/1.25.0/x64/src/testing/testing.go:1875 +0x35b
panic({0x9f3160?, 0xc000347220?})
	/opt/hostedtoolcache/go/1.25.0/x64/src/runtime/panic.go:783 +0x132
ocm.software/open-component-model/bindings/go/dag/sync.(*Vertex[...]).MustGetAttribute(0xbe4f40, {0xae0df0, 0xa})
	/home/runner/go/pkg/mod/ocm.software/open-component-model/bindings/go/dag@v0.0.4/sync/dag.go:543 +0xd9
ocm.software/open-component-model/cli/cmd/get/component-version.buildRenderer.buildTableFormatSerializer.func6({0xbda040, 0xc0004c2e58}, {0xc0005d0640, 0x5, 0xbdff18?})
	/home/runner/work/open-component-model/open-component-model/cli/cmd/get/component-version/cmd.go:282 +0x166
ocm.software/open-component-model/cli/internal/render/graph/list.ListSerializerFunc[...].Serialize(...)
[...]

PR #946, opened 2025-09-26
fix(graph)!: finally get rid of sync map

jakobmoellerdev:

- This commit **removes the old sync.Map–based DAG implementation** and replaces it with a standard map-backed DAG guarded by an external RWMutex wrapper, eliminating all conversion utilities between map and sync.Map variants. 
- The new synced_dag adds a thin SyncedDirectedAcyclicGraph that exposes **WithReadLock/WithWriteLock** around the existing DAG. 
- Discovery is redesigned: a GraphDiscoverer now resolves vertex values via a pluggable Resolver, discovers neighbors via a Discoverer, tracks per-vertex DiscoveryState and value attributes, orders edges, and ensures each vertex is processed at most once via a doneMap. 
- Processing is refactored into a GraphProcessor that performs Kahn-style topological processing with optional concurrency limits and a Processor interface applied to stored vertex values.
- The base DAG gets some fixes: refactoring to properly work with attributes during construction and cloining, a DeleteEdge API that is also used by DeleteVertex

this fixes https://github.com/open-component-model/open-component-model/issues/903

We do this by combining a few fixes:
- using atomic locking makes sure we never race
- adding nodes during discover centrally ensures we never lock incorrectly

We tested lock safety with `-race` and about 100_000 read/write tests on the graph

Because the resolver and the discoverer as well as the processor now have slightly changed interfaces, this is a breaking change. We think this is worth it though as the benefits from a cleaned up API are quite large


matthiasbruns (dismissed):
It took me some time to figure out that dag is wrapped by synced_dag - but that makes totally sense.

LGTM

bindings/go/dag/dag.go
```diff
@@ -89,12 +113,9 @@ func (d *DirectedAcyclicGraph[T]) DeleteVertex(id T) error {
 	}
 
 	// Remove all edges to this node
-	for _, node := range d.Vertices {
-		if _, exists := node.Edges[id]; exists {
-			// Decrement the in-degree of the node
-			d.Vertices[node.ID].InDegree--
-			// Remove the edge from the node
-			delete(node.Edges, id)
+	for edge := range d.Vertices[id].Edges {
```
Skarlso, line 116: Do we need to consider updating edges that point to the deleted id?

We don't, right? Because we don't want to destroy the dag, right?

bindings/go/dag/sync/discover.go
```diff
@@ -31,164 +31,155 @@ func (s DiscoveryState) String() string {
...
+func (d *GraphDiscoverer[K, V]) CurrentEdges(key K) []K {
+	var edges []K
+	_ = d.graph.WithReadLock(func(d *dag.DirectedAcyclicGraph[K]) error {
+		v, ok := d.Vertices[key]
+		if !ok {
+			return nil
+		}
+		edges = make([]K, len(v.Edges))
+		for k, edge := range v.Edges {
+			idx := edge[attributeOrderIndex]
```
Skarlso, line 89: Do an exists check here? Otherwise, it could end up being a problem in the below `idx.(int)`.

bindings/go/dag/dag.go
Skarlso, line 62: I know you haven't touched this, but... this clone is creating a new map yes, but the vertices are still pointers. If you modify the clone you modify the original. 🤔 

I'm pretty sure I can write a test for it to prove it. :D

Skarlso (commented):
couple of observations

bindings/go/dag/dag.go
jakobmoellerdev: yeah the idea here is to not modify the input graph anymore

bindings/go/dag/sync/discover.go
jakobmoellerdev: how should I react if we do not have the index?

bindings/go/dag/dag.go
Skarlso: yep... so is this intentional?

```og
func TestDAGClone(t *testing.T) {
	r := require.New(t)
	d := NewDirectedAcyclicGraph[string]()
	r.NoError(d.AddVertex("A", map[string]any{"key": "original"}))
	r.NoError(d.AddVertex("B"))
	r.NoError(d.AddEdge("A", "B", map[string]any{"edgeKey": "originalEdge"}))

	cloned := d.Clone()

	r.Equal("original", d.Vertices["A"].Attributes["key"])
	r.Equal("original", cloned.Vertices["A"].Attributes["key"])

	cloned.Vertices["A"].Attributes["key"] = "modified"

	r.Equal("modified", d.Vertices["A"].Attributes["key"])
	r.Equal("modified", cloned.Vertices["A"].Attributes["key"])
}
```

bindings/go/dag/sync/process.go
```diff
@@ -1,248 +1,121 @@
...
-	ctx context.Context,
-	topology *DirectedAcyclicGraph[T],
-	ids []T,
-	processor VertexProcessor[T],
-	doneMap *sync.Map,
-	opts *ProcessTopologyOptions,
-) ([]T, error) {
-	errGroup, ctx := errgroup.WithContext(ctx)
-	errGroup.SetLimit(opts.GoRoutineLimit)
+	for len(queue) > 0 {
```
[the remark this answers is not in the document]
jakobmoellerdev: checked errgroup, agreed! will move this into the queue iteration

bindings/go/dag/dag.go
jakobmoellerdev: agreed. will adjust

jakobmoellerdev: implemented a working (lol) clone

bindings/go/dag/sync/process.go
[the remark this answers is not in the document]
jakobmoellerdev: moved to per batch run

Skarlso (approved):

Skarlso (approved):

bindings/go/dag/sync/process_options.go
```diff
@@ -0,0 +1,33 @@
...
+	"context"
+)
+
+// GraphProcessorOptions configures the behavior of a GraphProcessor.
+type GraphProcessorOptions[K cmp.Ordered, V any] struct {
+	// Processor what is called for each node value during processing.
+	Processor Processor[V]
+	// Concurrency limits the number of concurrent processing goroutines.
+	// If <= 0, concurrency is unlimited.
+	Concurrency int
```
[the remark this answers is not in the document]
jakobmoellerdev: added a test (even though its not ideal, better than nothing)

PR #913, opened 2025-09-24
feat(575): resolver cli integration

matthiasbruns:
#### What this PR does / why we need it
This PR integrates the following repository changes
- use the new patchmatcher repo 
- use legacy fallback if no pathmatcher config is available

#### Which issue(s) this PR fixes
Contributes https://github.com/open-component-model/ocm-project/issues/575

PR #962, opened 2025-09-30
feat(575): make sure that we convert raw to typed in registries

matthiasbruns:
On-behalf-of: SAP <matthias.bruns@sap.com>

In https://github.com/open-component-model/open-component-model/pull/913 we found out that the plugin registries are not prepared to work with `runtime.Raw` specs. In the current resolver/glob implementation, we do not have a `runtime.Typed` repo anymore but a `runtime.Raw` repo. This PR allows the use of `runtime.Raw` in the plugin registries where needed.

Contributes: https://github.com/open-component-model/ocm-project/issues/575

### DIFF
--- a/bindings/go/plugin/manager/convert.go
+++ b/bindings/go/plugin/manager/convert.go
@@ -1,0 +1,21 @@
+package manager
+
+import (
+	"fmt"
+
+	"ocm.software/open-component-model/bindings/go/runtime"
+)
+
+// EnsureTyped ensures that a runtime.Typed is not a Raw object.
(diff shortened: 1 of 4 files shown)


matthiasbruns:
waiting for @fabianburth pr's

jakobmoellerdev (commented):
maybe I have this wrong again but didnt we agree that it would be acceptable to instead do this conversion only in the inbuilt types because not all registries have the ability for conversion?

matthiasbruns:
Yeah you are right - the first thing I did was rebasing all my PRs.
I wanted to can do it in the warroom today

PR #963, opened 2025-09-30
feat(oci): enhance repository provider with user agent, temp dir, and caching support

fabianburth:
**Core features:** 
- Implements a `storeCache`, enabling repository caching based on path to enable sharing a lock.
This (hopefully) fixes the current bugs related to race conditions.
- Prevents digests from being used as tag and correspondingly prevents list to attempt to resolve artifacts with empty tags. This should fix the race condition between `AddComponentVersion` and `ListComponentVersion`.

**Summary**:
- Introduces new options (`WithUserAgent`, `WithTempDir`) for more flexible configuration of repository providers.
- Adds a default `UserAgent` value (`DefaultCreator`).
- Updates `NewFromCTFRepoV1` to return both repository and store, allowing to cache the store.
- Refactors `OpenCTFByFileExtension` for better modularity by separating format detection into `DiscoverCTFFormatFromPath`.

#### Implications
- Improves control over temporary working directories and user agent settings.
- Enhances compatibility and performance when working with CTF archives.
- Existing repository logic adapts seamlessly to changes, maintaining backward compatibility for most use cases.

Contributes to https://github.com/open-component-model/ocm-project/issues/564


bindings/go/oci/repository/repository_test.go
```diff
@@ -79,7 +79,7 @@ func TestNewFromCTFRepoV1(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			repo, err := NewFromCTFRepoV1(t.Context(), tt.repository)
+			repo, _, err := NewFromCTFRepoV1(t.Context(), tt.repository)
```
matthiasbruns, line 82: Is there a reason why we do not check the store or have tests for that?

ikhandamirov (approved):

bindings/go/oci/ctf/store.go
```diff
@@ -88,7 +90,7 @@ func (s *Store) ComponentVersionReference(ctx context.Context, component, versio
 type Repository struct {
 	archive ctf.CTF
 	repo    string
-	indexMu sync.RWMutex
+	indexMu *sync.RWMutex
```
Skarlso, line 93: Are you sure about this? The mutex doesn't need to be a pointer if all the method receivers are pointers.

It is super rare that a mutex is a pointer and then it needs to be really explained as to why.

I see that you pass in the mutex from the Store. If I understand correctly, this means that the Repository CANNOT LIVE without a store. Because otherwise, this mutex will fail with a nil error since it's not initialized anywhere. And since it's an unexported field it means that this struct CANNOT BE USED outside of this package ever. Am I understanding this correctly?

Skarlso: If yes, we might as well just make this an unexported struct.

ikhandamirov (dismissed):

bindings/go/oci/ctf/store.go
```diff
@@ -68,7 +69,8 @@ func (s *Store) StoreForReference(_ context.Context, reference string) (spec.Sto
 	}
 	ref := rawRef.(looseref.LooseReference)
 
-	return &Repository{
+	return &repository{
+		indexMu: &s.mu,
```
jakobmoellerdev, line 73: maybe just rename to mu from indexMu at this point

bindings/go/oci/ctf/store.go
```diff
@@ -231,8 +234,20 @@ func (s *Repository) Tag(ctx context.Context, desc ociImageSpecV1.Descriptor, re
 	repo := s.repo
 
 	var meta v1.ArtifactMetadata
-
-	if ref, err := looseref.ParseReference(reference); err == nil {
+	// This is a workaround around currently covering two cases:
```
jakobmoellerdev, line 237: please add a todo item for this

bindings/go/oci/repository/provider/options.go
```diff
@@ -0,0 +1,22 @@
+package provider
+
+type Options struct {
```
jakobmoellerdev, line 3: this struct is missing documentation

bindings/go/oci/repository/provider/provider.go
```diff
@@ -18,32 +19,51 @@ import (
...
+const DefaultCreator = "ocm.software/open-component-model/bindings/go/oci"
+
 // CachingComponentVersionRepositoryProvider is a caching implementation of the repository.ComponentVersionRepositoryProvider interface.
 // It provides efficient caching mechanisms for repository operations by maintaining:
 // - A credential cache for authentication information
 // - An OCI cache for manifests and layers
 // - An authorization cache for auth tokens
 // - A shared HTTP client with retry capabilities
 type CachingComponentVersionRepositoryProvider struct {
+	creator            string
```
jakobmoellerdev, line 31: I think we should add docs to the individual variables here to make sure we still understand whats ahppening. I also think the struct doc is outdated

bindings/go/oci/repository/provider/provider.go
```diff
@@ -96,9 +118,23 @@ func (b *CachingComponentVersionRepositoryProvider) GetComponentVersionRepositor
 			Client:     b.httpClient,
 			Cache:      b.authorizationCache,
 			Credential: b.credentialCache.get,
+			Header: map[string][]string{
+				"User-Agent": {b.creator},
+			},
 		}, opts...)
 	case *ctfrepospecv1.Repository:
-		return ocirepository.NewFromCTFRepoV1(ctx, obj, opts...)
+		loadFunc := func(path string) (*ocictf.Store, error) {
```
jakobmoellerdev, line 126: we should explain what happens here tbh

bindings/go/oci/ctf/store.go
Skarlso: Now that it is private, let me ask this... do we even need it? Like, what would happen if the Store struct would be the one implementing all these methods?

If you say that it's there so it's a reasonable separation of concern, that's acceptable. :) WDYT?

jakobmoellerdev (approved):

Skarlso (approved):

PR #973, opened 2025-10-01
fix: rename resolver configuration version from v1 to v1alpha1

jakobmoellerdev:

the resolvers config had the wrong constant declaring it as "v1" when it was "v1alpha1".


morri-son (approved):
lgtm

matthiasbruns (approved):

PR #974, opened 2025-10-01
feat(construct): implement external component discovery and processing for recursion support

jakobmoellerdev:

this adds external reference neighbor processing and transfer to the target repository of construct

contributes to https://github.com/open-component-model/ocm-project/issues/666


matthiasbruns (approved):
LGTM

PR #993, opened 2025-10-05
feat: component lister plugin registry

ikhandamirov:

This PR provides component lister plug-in registry for CLI.
The following points from the task are covered:
* a new component lister plug-in registry
* plug-in handler, allowing 3-rd parties to provide component listers for other types of store

The "internal plug-in" will be part of a separate PR and part of the `cli` module.

Contributes to open-component-model/ocm-project#674

### DIFF
--- a/bindings/go/plugin/manager/registries/componentlister/registry.go
+++ b/bindings/go/plugin/manager/registries/componentlister/registry.go
@@ -198,9 +198,7 @@ func (r *ComponentListerRegistry) GetComponentLister(ctx context.Context,
 	defer r.mu.Unlock()
 
 	// look for an internal implementation that actually implements the interface
-	if _, err := r.scheme.DefaultType(repositorySpecification); err != nil {
-		return nil, fmt.Errorf("failed to default type for prototype %T: %w", repositorySpecification, err)
-	}
+	_, _ = r.scheme.DefaultType(repositorySpecification)
 	typ := repositorySpecification.GetType()
 	// if we find the type has been registered internally, we look for internal plugins for it.
 	if ok := r.scheme.IsRegistered(typ); ok {


ikhandamirov:
The PR currently addresses the points "new component lister plug-in registry" and "plug-in handler". It would be great to have the direction validated, especially by @jakobmoellerdev , @Skarlso and @piotrjanik 

Regarding the not yet touched "internal plug-in", I'm looking for suggestions, which module should it live in. In the CLI under "cli/internal/plugin/builtin"?

bindings/go/plugin/manager/contracts/componentlister/v1/contract.go
```diff
@@ -0,0 +1,23 @@
...
+	contracts.PluginBase
+	GetIdentity(ctx context.Context, typ *GetIdentityRequest[T]) (*GetIdentityResponse, error)
+}
+
+// ComponentListerPluginContract is a REST wrapper around the repository.ListComponents interface for communicating
+// with a plugin.
+type ComponentListerPluginContract[T runtime.Typed] interface {
+	contracts.PluginBase
+	IdentityProvider[T]
+	ListComponents(ctx context.Context, request *ListComponentsRequest[T], credentials map[string]string) ([]string, error)
```
jakobmoellerdev, line 22: how does this actually do pagination? I dont understand how one would indicate if a new call is required?

piotrjanik (dismissed):

bindings/go/plugin/manager/registries/componentlister/registry.go
```diff
@@ -0,0 +1,254 @@
...
+// GetComponentLister returns ComponentLister for a specific repository type.
+func (r *ComponentListerRegistry) GetComponentLister(ctx context.Context,
+	repositorySpecification runtime.Typed,
+	credentials map[string]string,
+) (repository.ComponentLister, error) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+
+	// look for an internal implementation that actually implements the interface
+	if _, err := r.scheme.DefaultType(repositorySpecification); err != nil {
```
jakobmoellerdev: this should not error.

ikhandamirov: Ok, ignoring the error with https://github.com/open-component-model/open-component-model/pull/993/commits/b5bfcbabfffe350bfa8aebae2464c0c849ffacdf

p.s. I'd be interested to see an example illustrating the problem with the error. Didn't get my head around it yet.

piotrjanik (approved):

PR #1012, opened 2025-10-07
feat(construct): add external component version policy to handle references

jakobmoellerdev:

this allows to turn on or off external reference copying in the constructor. we can use this for example in CLI by binding this to `--recursive` which will enable or disable copying of external component versions

contributes to https://github.com/open-component-model/ocm-project/issues/666


bindings/go/constructor/options.go
```diff
@@ -101,3 +105,18 @@ const (
...
+// ExternalComponentVersionCopyPolicy defines the policy for handling external component version references
+// when interacting with the target repository.
+// If the constructor library encounters an external component version reference that is not part of the constructor
+// specification, it will use this policy to determine how to handle the reference.
+type ExternalComponentVersionCopyPolicy int
+
+const (
+	// ExternalComponentVersionCopyPolicySkip will skip the copy of the component version to the target repository.
+	// This is the default policy.
+	ExternalComponentVersionCopyPolicySkip = iota
```
matthiasbruns, line 118: is this the default behavior? I would expect iota 0 to be default

jakobmoellerdev: yes that would be default, as in by default external references are not copied.

fabianburth (approved):

PR #1014, opened 2025-10-07
feat: create a migration infrastructure for context and configuration v1=>v2

Skarlso:

Towards https://github.com/open-component-model/ocm-project/issues/626


kubernetes/controller/go.mod
```diff
@@ -27,6 +27,12 @@ require (
 	k8s.io/client-go v0.34.1
 	k8s.io/utils v0.0.0-20250820121507-0af2bda4dd1d
 	ocm.software/ocm v0.29.1
+	ocm.software/open-component-model/bindings/go/configuration v0.0.9
```
jakobmoellerdev, line 30: can we make this a separate require block to not mix stuff up?

kubernetes/controller/internal/configuration/config.go
```diff
@@ -0,0 +1,123 @@
...
+
+	genericv1 "ocm.software/open-component-model/bindings/go/configuration/generic/v1/spec"
+	"ocm.software/open-component-model/bindings/go/runtime"
+	"ocm.software/open-component-model/kubernetes/controller/api/v1alpha1"
+)
+
+// GetConfigFromSecret extracts and decodes OCM configuration from a Kubernetes Secret.
+// It looks for configuration data under the OCMConfigKey.
+func GetConfigFromSecret(secret *corev1.Secret) (*genericv1.Config, error) {
+	data, ok := secret.Data[v1alpha1.OCMConfigKey]
```
jakobmoellerdev, line 23: i think if the key is not found we should error so people know they configured the wrong secret

kubernetes/controller/internal/configuration/config.go
```diff
@@ -0,0 +1,123 @@
...
+// It fetches the referenced Secrets/ConfigMaps from the cluster and extracts their configuration.
+// All configurations are merged using FlatMap into a single Config.
+func LoadConfigurations(ctx context.Context, k8sClient client.Reader, namespace string, ocmConfigs []v1alpha1.OCMConfiguration) (*genericv1.Config, error) {
+	if len(ocmConfigs) == 0 {
+		return &genericv1.Config{}, nil
+	}
+
+	// Best effort sort of the configs to ensure they are always processed in the same order.
+	sort.Slice(ocmConfigs, func(i, j int) bool {
+		key1 := ocmConfigs[i].Name + "." + ocmConfigs[i].Namespace
```
jakobmoellerdev, line 76: I think you could sort by UID no?

kubernetes/controller/internal/configuration/config.go
```diff
@@ -0,0 +1,123 @@
...
+		case "Secret":
+			obj = &corev1.Secret{}
+		case "ConfigMap":
+			obj = &corev1.ConfigMap{}
+		default:
+			return nil, fmt.Errorf("unsupported configuration kind: %s", ocmConfig.Kind)
+		}
+
+		key := client.ObjectKey{Namespace: ns, Name: ocmConfig.Name}
+		if err := k8sClient.Get(ctx, key, obj); err != nil {
```
jakobmoellerdev, line 100: are we sure this cannot cause perf issues when accessing many objects?

kubernetes/controller/internal/configuration/config.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	return genericv1.FlatMap(configs...), nil
+}
+
+// FilterForType is a convenience wrapper around genericv1.FilterForType
+// that filters configurations by type T.
+// Note: This isn't really used at the moment.
+func FilterForType[T runtime.Typed](scheme *runtime.Scheme, config *genericv1.Config) ([]T, error) {
+	return genericv1.FilterForType[T](scheme, config)
```
jakobmoellerdev, line 122: one liner func should just be removed imo

kubernetes/controller/internal/setup/repository.go
```diff
@@ -0,0 +1,53 @@
...
+
+	"ocm.software/open-component-model/bindings/go/credentials"
+	"ocm.software/open-component-model/bindings/go/plugin/manager"
+	"ocm.software/open-component-model/bindings/go/repository"
+	"ocm.software/open-component-model/bindings/go/runtime"
+)
+
+// RepositoryOptions configures OCM repository creation.
+type RepositoryOptions struct {
+	PluginManager   *manager.PluginManager
```
jakobmoellerdev, line 17: imo the plugin manager is not an option, but mandatory. WDYT?

kubernetes/controller/internal/setup/repository.go
```diff
@@ -0,0 +1,53 @@
...
+	CredentialGraph credentials.GraphResolver
+	Logger          logr.Logger
+}
+
+// NewRepository creates an OCM repository for the given repository specification.
+// This creates a direct repository connection without fallback support.
```
jakobmoellerdev, line 23: this needs a bit of a better description on behavior imho

kubernetes/controller/internal/configuration/config.go
Skarlso: it doesn't have one. 🤔

kubernetes/controller/internal/configuration/config.go
Skarlso: Hm, I could try and parallelise this.

kubernetes/controller/internal/configuration/config.go
jakobmoellerdev: huh?! every object in k8s has a uid no?

Skarlso: Okay, need to figure out how to deal with this effectively. This is important and critical code section called many times. Hashing and sorting all the time every config and individually getting all the items might be a bit too slow.

Skarlso:
>Also: Is flatmap stable?

That's a good question. :D

kubernetes/controller/internal/configuration/config.go
Skarlso: Okay, I _think_ I parallelized most of the critical paths here. It can be improved of course ( constant for a semaphore right now ), but I want your first eyes on it.

Skarlso:
I know I need to do some refactoring that method is huge. I'm going to do it in a bit.

kubernetes/controller/internal/setup/repository.go
Skarlso: Then yeah. It shouldn't. :D

Skarlso:
@jakobmoellerdev So now I added hashing to the sorting, but that's not what you mean here, right:
>

But this exists, what other kind of hashing did you have in mind?

PR #1040, opened 2025-10-14
feat: Implement GetBlobFromPath in blob module

morri-son:

Implement new function to create a blob from a path, either using a single file or directory as input. It is possible to specify multiple options like include/exclude filters for files.


bindings/go/blob/filesystem/blob_from_path.go
```diff
@@ -0,0 +1,313 @@
...
+//
+// Note on pattern option semantics: include/exclude patterns are matched using
+// `filepath.Match` against the full entry path.
+// No additional globbing extensions (like `**`) are performed.
+// Exclude patterns take precedence over include patterns.
+//
+// The function returns an error if the file path is empty, the specified path is outside the working directory
+// or if there are issues accessing the directory.
+
+func GetBlobFromPath(ctx context.Context, path string, opt DirOptions) (blob.ReadOnlyBlob, error) {
```
fabianburth:
```suggestion
// GetBlobFromPath creates a blob from a path using the provided options.
//
// It reads a path from the filesystem and offers include and exclude filters.
// Matching files are added to a TAR archive, and if configured also the parent directory.
// If configured, the final blob gets compressed using gzip.
//
// Note on pattern option semantics: include/exclude patterns are matched using
// `filepath.Match` against the full entry path.
// No additional globbing extensions (like `**`) are performed.
// Exclude patterns take precedence over include patterns.
//
// The function returns an error if the file path is empty, the specified path is outside the working directory
// or if there are issues accessing the directory.
func GetBlobFromPath(ctx context.Context, path string, opt DirOptions) (blob.ReadOnlyBlob, error) {
```

bindings/go/blob/filesystem/blob_from_path.go
```diff
@@ -0,0 +1,313 @@
...
+		default:
+			return fmt.Errorf("unsupported file type %s for entry %q in directory %q", ei.Mode().String(), entry.Name(), subPath)
+		}
+	}
+	return nil
+}
+
+// createTarFromSingleFile writes a single file to the TAR archive.
+// subPath is the relative path within the virtual filesystem.
+func createTarFromSingleFile(ctx context.Context, fileSystem FileSystem, subPath string, opt *DirOptions, tw *tar.Writer) error {
```
fabianburth: I'm fine with creating tars from single files. Just want to mention that I believe old ocm did not do that for input type file. So, while this might be a useful method in general, we cannot use it as a drop in replacement for v1 file input.

bindings/go/blob/filesystem/blob_from_path.go
```diff
@@ -0,0 +1,313 @@
...
+	pr, pw := io.Pipe()
+
+	go func() {
+		tw := tar.NewWriter(pw)
+
+		// use local gerr to capture errors from tar creation
+		var gerr error
+
+		// take care to close tar writer and pipe and capture any error
+		defer func() { gerr = errors.Join(gerr, tw.Close()) }()
```
matthiasbruns: we are not doing anything with the `gerr`, are we?

bindings/go/blob/filesystem/blob_from_path.go
```diff
@@ -0,0 +1,313 @@
...
+
+		// fail early for symlinks
+		if (ei.Mode() & fs.ModeSymlink) != 0 {
+			return fmt.Errorf("symlinks are not supported: found symlink %q in directory %q", entry.Name(), subPath)
+		}
+
+		// reconstruct the full path of the entry
+		entryPath := filepath.Join(subPath, entry.Name())
+		// Check if path is included based on include/exclude patterns from options.
+		ok, err := isPathIncluded(entryPath, opt.IncludeFiles, opt.ExcludeFiles)
```
matthiasbruns: could be merged into an

`if ok, err := isPathIncluded(entryPath, opt.IncludeFiles, opt.ExcludeFiles); err != nil {} else !ok {} `

bindings/go/blob/filesystem/blob_from_path.go
```diff
@@ -0,0 +1,313 @@
...
+
+	go func() {
+		tw := tar.NewWriter(pw)
+
+		// use local gerr to capture errors from tar creation
+		var gerr error
+
+		// take care to close tar writer and pipe and capture any error
+		defer func() { gerr = errors.Join(gerr, tw.Close()) }()
+		defer func() { _ = pw.CloseWithError(gerr) }()
```
fabianburth: I don't think we want 2 separate defers here

bindings/go/blob/filesystem/blob_from_path.go
```diff
@@ -0,0 +1,313 @@
...
+}
+
+// createTarFromDir writes the contents of a directory to the TAR archive.
+// subPath is the relative path within the virtual filesystem.
+func createTarFromDir(ctx context.Context, fileSystem FileSystem, subPath string, opt *DirOptions, tw *tar.Writer) error {
+	select {
+	case <-ctx.Done():
+		return fmt.Errorf("context cancelled while processing directory %q: %w", subPath, ctx.Err())
+	default:
+	}
```
fabianburth: I don't think that's necessary here, especially once you removed the async as matthias pointed out above

bindings/go/blob/filesystem/blob_from_path.go
```diff
@@ -0,0 +1,313 @@
...
+			// add parent directory if PreserveDir is set
+		} else {
+			baseDir = filepath.Dir(path)
+			subPath = filepath.Base(path)
+		}
+		// Single file case
+	} else {
+		baseDir = filepath.Dir(path)
+		subPath = filepath.Base(path)
+	}
```
fabianburth: Those cases are identical, I believe you could merge them

bindings/go/blob/filesystem/blob_from_path.go
```diff
@@ -0,0 +1,313 @@
...
+	fileSystem, err := NewFS(baseDir, os.O_RDONLY)
+	if err != nil {
+		return nil, fmt.Errorf("error creating virtual filesystem for path %q: %w", baseDir, err)
+	}
+
+	// Create a TAR archive from the directory contents.
+	// We differentiate between directory and single file cases.
+	pr, pw := io.Pipe()
+
+	go func() {
```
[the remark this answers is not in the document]
fabianburth: I don't think so, since we use the pipe here. But I think `pr` is a readcloser, so this function should also return a read closer

bindings/go/blob/filesystem/blob_from_path.go
fabianburth: I stand corrected, I think since this is called with recursion, it actually makes sense!

bindings/go/blob/filesystem/blob_from_path.go
```diff
@@ -0,0 +1,313 @@
...
+		tarBlob = compression.Compress(tarBlob)
+	}
+	return tarBlob, nil
+}
+
+// createTarStream creates a TAR archive from the contents of the specified path, either a directory or a single file.
+func createTarStream(ctx context.Context, path string, opt *DirOptions) (io.Reader, error) {
+	fi, err := os.Stat(path)
+	if err != nil {
+		return nil, fmt.Errorf("error stating path %q: %w", path, err)
```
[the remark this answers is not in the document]
morri-son: added more context

bindings/go/blob/filesystem/blob_from_path.go
```diff
@@ -0,0 +1,313 @@
...
+		} else {
+			gerr = createTarFromSingleFile(ctx, fileSystem, subPath, opt, tw)
+		}
+	}()
+	return pr, nil
+}
+
+// createTarFromDir writes the contents of a directory to the TAR archive.
+// subPath is the relative path within the virtual filesystem.
+func createTarFromDir(ctx context.Context, fileSystem FileSystem, subPath string, opt *DirOptions, tw *tar.Writer) error {
```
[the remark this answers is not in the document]
morri-son: used io/fs.WalkDir now

bindings/go/blob/filesystem/blob_from_path.go
```diff
@@ -0,0 +1,313 @@
...
+}
+
+// createTarFromSingleFile writes a single file to the TAR archive.
+// subPath is the relative path within the virtual filesystem.
+func createTarFromSingleFile(ctx context.Context, fileSystem FileSystem, subPath string, opt *DirOptions, tw *tar.Writer) error {
+	select {
+	case <-ctx.Done():
+		return fmt.Errorf("context cancelled while processing single file %q: %w", subPath, ctx.Err())
+	default:
+	}
```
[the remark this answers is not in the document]
morri-son: removed the second check and just kept the context check at the beginning of the dir walk.

bindings/go/blob/filesystem/blob_from_path.go
```diff
@@ -0,0 +1,342 @@
...
+		fileBlob = compression.Compress(fileBlob)
+	}
+	return fileBlob, nil
+}
+
+// createTarFromDir creates a TAR archive from the filesystem.
+// Uses the virtual filesystem to read the directory contents.
+// Uses fs.WalkDir to traverse the directory structure in a deterministic order.
+// This is required to ensure reproducible TAR archives.
+func createTarFromDir(ctx context.Context, fileSystem FileSystem, subPath string, opt DirOptions, tw *tar.Writer) error {
```
jakobmoellerdev: for me this function is too long and too complicated. FWIW this should be separated into smaller funcs imho. were not in C land

jakobmoellerdev (approved):

PR #1043, opened 2025-10-14
feat(encoding)!: replace legacy TAR encoding with extensible media type-based approach and default to JSON

jakobmoellerdev:

- Removed the hardcoded TAR/YAML-based encoder/decoder and replaced it with a media-type extensible framework for handling YAML, JSON, and TAR descriptor encodings.
- Introduced `SingleFileEncodeDescriptor` and `SingleFileDecodeDescriptor` functions to support multiple formats with legacy compatibility.
- Updated `store_descriptor` and `repository` logic to dynamically determine encoding/decoding via media type.
- Added test cases for validating encoding and decoding across all supported formats (YAML, JSON, TAR).
- Added an option that defaults to JSON to configure new encodings

This refactor allows interacting with all comp desc encodings used in previous OCM iterations. It also changes the default encoding policy to be JSON, not yaml+tar as before.

part of https://github.com/open-component-model/ocm-project/issues/680


bindings/go/oci/spec/descriptor/decode.go
```diff
@@ -0,0 +1,82 @@
...
+			return nil, fmt.Errorf("converting component descriptor: %w", err)
+		}
+		return desc, nil
+	default:
+		return nil, fmt.Errorf("unsupported descriptor media type %s", mediaType)
+	}
+}
+
+// descriptorFileFromTar reads the component descriptor from a tar.
+// The component is expected to be inside the tar in a file called descriptorFileHeader.
```
Skarlso: This comment isn't correct, right? You are looking for LegacyComponentDescriptorTarFileName.

Skarlso (commented):
Couple of remarks. :)

Skarlso (approved):

PR #1049, opened 2025-10-16
feat(ci): add workflow for automated release branch creation

jakobmoellerdev:

- Introduced a new GitHub Actions workflow to automate the creation of release branches, triggered manually via `workflow_dispatch`.
- Validates branch names to ensure compliance with the `release/0.[0-9]+` regex pattern.
- Supports specifying both source (`main` by default) and target branches for flexibility.
- Ensures idempotency by checking for existing branches before creation.

fixes https://github.com/open-component-model/ocm-project/issues/698 and preparation for https://github.com/open-component-model/ocm-project/issues/699

tested via https://github.com/jakobmoellerdev/open-component-model/actions/runs/18557717064 for successful creation
tested via https://github.com/jakobmoellerdev/open-component-model/actions/runs/18557692916 for mismatching pattern
tested via https://github.com/jakobmoellerdev/open-component-model/actions/runs/18557785257/job/52899336902 for already existing branch being skipped


.github/workflows/release-branch.yaml
```diff
@@ -0,0 +1,80 @@
...
+  workflow_dispatch:
+    inputs:
+      target_branch:
+        description: 'Branch to cut to, must match regex releases/v0.[0-9]+'
+        required: true
+      source_branch:
+        description: 'Branch to cut from (default main)'
+        required: true
+        default: 'main'
+
```
Skarlso: Needs a top level `permissions:` with Read and then the individual jobs need to specify their write request.

jakobmoellerdev: I believe it doesnt because default permission is read and we use app tokens

.github/workflows/release-branch.yaml
```diff
@@ -0,0 +1,80 @@
...
+        with:
+          github-token: ${{ steps.get_token.outputs.token || secrets.GITHUB_TOKEN }}
+          script: |
+            const target = process.env.TARGET_BRANCH;
+            const source = process.env.SOURCE_BRANCH;
+            
+            const repo = context.repo.repo;
+            const owner = context.repo.owner;
+
+            const regex = /^releases\/v0\.[0-9]+$/;
```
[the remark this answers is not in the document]
jakobmoellerdev: i will create a pr allowing that in a million years

.github/workflows/release-branch.yaml
jakobmoellerdev: if the scanner complains i will triage it to death!

jakobmoellerdev:
i have run this action a few times also with the new pattern. should be good to go (i have about 300 action runs that i went with for testing but im sure something will break anyways)

PR #1072, opened 2025-10-22
fix: fall back to TCP if on a read-only disk

Skarlso:

In a read-only disk ( like a scratch container ) the flow to detect connection fails because it can't create a folder. In that case it should fall back to TCP.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -287,15 +287,15 @@ jobs:
             .github/workflows/ci.yml
             ${{ matrix.project }}
       - name: Initialize CodeQL
-        uses: github/codeql-action/init@755f44910c12a3d7ca0d8c6e42c048b3362f7cec # v3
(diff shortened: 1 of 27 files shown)


jakobmoellerdev (dismissed):

bindings/go/plugin/manager/manager.go
```diff
@@ -305,10 +305,12 @@ func (pm *PluginManager) addPlugin(ctx context.Context, ocmConfig *genericv1.Con
 }
 
 func determineConnectionType(ctx context.Context) (mtypes.ConnectionType, error) {
+	// if we can't create a temp folder ( for example we are in a scratch container ) we default to TCP
 	tmp, err := os.MkdirTemp("", "")
```
piotrjanik: Maybe some slog.DebugContext? Just in case someone was wondering why TCP was chosen.

piotrjanik (approved):

PR #1078, opened 2025-10-24
feat: construct return graph

matthiasbruns:
This pr returns the calculated graph from construct

Contributes: https://github.com/open-component-model/ocm-project/issues/697


cli/cmd/add/component-version/cmd.go
```diff
@@ -290,7 +290,7 @@ func AddComponentVersion(cmd *cobra.Command, _ []string) error {
 	}
 	defer stop()
 
-	_, err = constructor.ConstructDefault(cmd.Context(), constructorSpec, opts)
+	_, _, err = constructor.ConstructDefault(cmd.Context(), constructorSpec, opts)
```
matthiasbruns: will be used in https://github.com/open-component-model/open-component-model/pull/1076

bindings/go/constructor/construct.go
```diff
@@ -68,7 +68,7 @@ type ConstructorOrExternalComponent struct {
 	ExternalComponent    *descriptor.Descriptor
 }
 
-func (c *DefaultConstructor) Construct(ctx context.Context, componentConstructor *constructor.ComponentConstructor) ([]*descriptor.Descriptor, error) {
+func (c *DefaultConstructor) Construct(ctx context.Context, componentConstructor *constructor.ComponentConstructor) ([]*descriptor.Descriptor, *syncdag.SyncedDirectedAcyclicGraph[string], error) {
```
jakobmoellerdev: why do we return a descriptor list on top of the graph? wont we need only the graph and can get the roots from there?

matthiasbruns: for the sake of the scope of this task, I will keep the descriptors return for now

bindings/go/constructor/construct_test.go
```diff
@@ -288,22 +261,72 @@ components:
...
+				return nil
+			}
+
+			// ocm.software/test-component
+			compOrExt := findInMap("ocm.software/test-component")
+			require.NotNil(t, compOrExt)
+			assert.Equal(t, "test-provider", compOrExt.ConstructorComponent.Provider.Name)
+			assert.Len(t, compOrExt.ConstructorComponent.Resources, 1)
+			assert.Len(t, compOrExt.ConstructorComponent.Sources, 1)
+			assert.Equal(t, "application/json", compOrExt.ConstructorComponent.Resources[0].Access.(*v2.LocalBlob).MediaType)
```
matthiasbruns: this fails since we do not have `Access` set in the graph yet

bindings/go/constructor/construct.go
```diff
@@ -419,11 +413,39 @@ func (c *DefaultConstructor) processResource(ctx context.Context, targetRepo Tar
 	}
 
 	logger.Debug("resource processed successfully")
+
+	// update resource in graph
+	if err := c.graph.WithWriteLock(func(d *dag.DirectedAcyclicGraph[string]) error {
```
matthiasbruns: trying to set additional metadata while constructing (doesn't work, only an experiment)

bindings/go/constructor/go.mod
```diff
@@ -18,6 +18,10 @@ require (
 	sigs.k8s.io/yaml v1.6.0
 )
 
+replace (
+	ocm.software/open-component-model/bindings/go/dag => ../dag
```
matthiasbruns: until https://github.com/open-component-model/open-component-model/pull/1106 is merged

bindings/go/constructor/construct.go
```diff
@@ -29,20 +29,25 @@ import (
...
 var ErrShouldSkipConstruction = errors.New("should skip construction")
 
+const AttributeDescriptor string = "constructor/descriptor"
+
 type Constructor interface {
 	// Construct processes a component constructor specification and creates the corresponding component descriptors.
 	// It validates the constructor specification and processes each component in topological order.
-	Construct(ctx context.Context, constructor *constructor.ComponentConstructor) ([]*descriptor.Descriptor, error)
-}
+	Construct(ctx context.Context) error
```
jakobmoellerdev: nit: is this protected against a double call? (i.e. if discover graph is empty?)

bindings/go/constructor/construct.go
```diff
@@ -29,20 +29,25 @@ import (
 // e.g. because the component version already exists in the target repository.
 var ErrShouldSkipConstruction = errors.New("should skip construction")
 
+const AttributeDescriptor string = "constructor/descriptor"
```
jakobmoellerdev: is there a particular reason we use attribute descriptor instead of the value attribute from dag?

jakobmoellerdev (approved):

bindings/go/constructor/construct.go
```diff
@@ -29,20 +29,25 @@ import (
...
-}
+	Construct(ctx context.Context) error
 
-// ConstructDefault is a convenience function that creates a new default DefaultConstructor and calls its Constructor.Construct method.
-func ConstructDefault(ctx context.Context, constructor *constructor.ComponentConstructor, opts Options) ([]*descriptor.Descriptor, error) {
-	return NewDefaultConstructor(opts).Construct(ctx, constructor)
+	// GetGraph returns the internal graph used during construction.
+	// It can be used to inspect the relationships between components.
+	// Construct must be called to ensure the graph is populated.
+	// TODO: this is temporary and needs to be replaced with proper abstraction
```
fabianburth: We do have kind of an inofficial policy that TODOs that are not tackled in the PR itself need an issue link.
IMO, you can omit the TODO here for now.

bindings/go/constructor/construct_test.go
```diff
@@ -606,3 +624,22 @@ func TestComponentVersionConflictPolicies(t *testing.T) {
 		})
 	}
 }
+
+func collectDescriptors(t *testing.T, graph *syncdag.SyncedDirectedAcyclicGraph[string]) []*descriptor.Descriptor {
```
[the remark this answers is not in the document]
matthiasbruns: this is a helper for the tests - what do mean by `part of descriptors struct as a method`?

fabianburth: Maybe, since we expose the graph now, we should test against it in a least one of the tests?
So, instead of merely collecting the descriptors and checking if they are there, check that the final graph structure looks as expected and all descriptor attributes are set?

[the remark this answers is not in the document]
jakobmoellerdev: I am saying that this helper probably could be replaced with a native function. but resolving since thats unimportant for now

jakobmoellerdev (approved):

fabianburth (approved):

PR #1106, opened 2025-10-28
feat: allow external graph in discover

matthiasbruns:
This pr allows to inject a graph into the dag discoverer
Used in construct
https://github.com/open-component-model/open-component-model/pull/1078/files#diff-f164090da9c1d33652290c4b5a386ca52e3a0916a521891c08aef662b261adc3R111

Contributes: https://github.com/open-component-model/ocm-project/issues/697

### DIFF
--- a/bindings/go/dag/sync/discover.go
+++ b/bindings/go/dag/sync/discover.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"log/slog"
 	"sync"
 
 	"golang.org/x/sync/errgroup"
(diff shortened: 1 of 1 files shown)


bindings/go/dag/sync/discover_test.go
```diff
@@ -61,6 +65,64 @@ func TestDAGDiscovery(t *testing.T) {
...
+		graph := map[string][]string{
+			"A": {"B", "C"},
+			"B": {"D"},
+			"C": {"D"},
+			"D": {},
+		}
+		externalGraph := NewSyncedDirectedAcyclicGraph[string]()
+		dag := NewGraphDiscoverer(&GraphDiscovererOptions[string, string]{
+			Roots: []string{"A", "B", "C", "D"},
+			Graph: externalGraph,
```
fabianburth: IMO, we can omit this test.

The main reason I've been hesitant with the passed in graph so far is that the way the discovery logic is currently implemented, it kind of assumes that the graph is empty.
If a node in the graph that has NOT been created as part of the discovery already exists, the `Discover` logic will fail.

I guess allowing to pass in a reference is still fine. But we should probably add a check like `if len(g.Vertices) > 0 { log.Info("discovery called with pre-populated graph") }`.

Issue #1143, opened 2025-11-04
Race condition in `Test_Provider_Smoke/access_provider_concurrently/same_version` test case


piotrjanik:
**What happened**:
The subtest `same version` in [`provider_test.go` ](https://github.com/open-component-model/open-component-model/blob/main/bindings/go/oci/repository/provider/provider_test.go) fails intermittently due to a race condition.

**What you expected to happen**:
The concurrent access tests should be deterministic and pass reliably.

**How to reproduce it (as minimally and precisely as possible)**:
- `go test ./bindings/go/oci/repository/provider -run Test_Provider_Smoke/access\ provider\ concurrently/same\ version -count=1000 -race`

PR #1185, opened 2025-11-10
fix: thread-safe writes for concurrent CTF access

piotrjanik:
What this PR does / why we need it:

Fixes race conditions in CTF stores that cause:
- "unexpected EOF" errors during concurrent blob reads/writes
- File corruption from partial writes when multiple goroutines access same blobs
- Race conditions between blob writing (O_TRUNC) and concurrent reads. 

Combines atomic file writes with simplified single-mutex locking to prevent CTF race conditions during concurrent blob operations. 
Implements atomic temp-file-rename pattern for all file writes and uses single sync.RWMutex per store.

Which issue(s) this PR fixes:
Fixes https://github.com/open-component-model/open-component-model/issues/1143


bindings/go/oci/repository_test.go
```diff
@@ -123,8 +123,8 @@ func TestRepository_GetComponentVersion(t *testing.T) {
 	r.NoError(err, "Failed to add component version when it should succeed")
 
 	desc, err = repo.GetComponentVersion(ctx, "test-component", "1.0.0")
-	r.NoError(err, "Expected error when getting non-existent component version")
-	r.NotNil(desc, "Expected nil descriptor when getting non-existent component version")
+	r.NoError(err, "No error expected when getting existing component version")
+	r.NotNil(desc, "Expected non-nil descriptor when getting existing component version")
```
piotrjanik: typo

jakobmoellerdev (approved):

bindings/go/oci/internal/lister/component/component_versions.go
```diff
@@ -91,10 +93,13 @@ func ReferenceTagVersionResolver(component string, store interface {
 				return "", fmt.Errorf("failed to decode index for tag %q: %w", tag, err)
 			}
```
Skarlso: By this time, after fetch has returned with a `data` but no error, but the decoding isn't working, you effectively lock the `store` in data, that you aren't freeing up in the NewDecoder's error now. So now, you need to add `data.Close` in there error cases as well if I understand the locking mechanism correctly.

bindings/go/oci/internal/lister/component/component_versions.go
Skarlso: Same here as below. Since data is not holding a lock successfully, you need to close it here. The below close after this is never called.

bindings/go/oci/repository.go
```diff
@@ -705,6 +705,7 @@ func getDescriptorOCIImageManifest(ctx context.Context, store spec.Store, refere
 		if descriptorManifest.MediaType != ociImageSpecV1.MediaTypeImageManifest {
 			return ociImageSpecV1.Manifest{}, nil, fmt.Errorf("index manifest is not an OCI image manifest")
 		}
+		err = errors.Join(err, manifestRaw.Close())
```
Skarlso: Here too, you now have the same problem. On an error you never close `manifestRaw`.

piotrjanik: Not necessarily :) line 692 :) notice manifestRaw is assigned to a new value.

piotrjanik: I am not sure if this looks better:
```go

    defer func() {
		err = errors.Join(err, manifestRaw.Close())
	}()

	switch base.MediaType {
	case ociImageSpecV1.MediaTypeImageIndex:
		if err := json.NewDecoder(manifestRaw).Decode(&index); err != nil {
			return ociImageSpecV1.Manifest{}, nil, err
		}
		if len(index.Manifests) == 0 {
			return ociImageSpecV1.Manifest{}, nil, fmt.Errorf("index has no manifests")
		}
		descriptorManifest := index.Manifests[0]
		if descriptorManifest.MediaType != ociImageSpecV1.MediaTypeImageManifest {
			return ociImageSpecV1.Manifest{}, nil, fmt.Errorf("index manifest is not an OCI image manifest")
		}
		indexManifestRaw, err := store.Fetch(ctx, descriptorManifest)
		defer func() {
			err = errors.Join(err, indexManifestRaw.Close())
		}()
		if err != nil {
			return ociImageSpecV1.Manifest{}, nil, fmt.Errorf("failed to fetch manifest: %w", err)
		}
		if err := json.NewDecoder(indexManifestRaw).Decode(&manifest); err != nil {
			return ociImageSpecV1.Manifest{}, nil, err
		}
	case ociImageSpecV1.MediaTypeImageManifest:
		if err := json.NewDecoder(manifestRaw).Decode(&manifest); err != nil {
			return ociImageSpecV1.Manifest{}, nil, err
		}
	default:
		return ociImageSpecV1.Manifest{}, nil, fmt.Errorf("unsupported media type %q", base.MediaType)
	}

return manifest, index, nil
```

bindings/go/oci/internal/lister/component/component_versions.go
```diff
@@ -76,21 +77,21 @@ func ReferenceTagVersionResolver(component string, store interface {
 		case ociImageSpecV1.MediaTypeImageManifest:
 			data, err = store.Fetch(ctx, desc)
 			if err != nil {
-				return "", fmt.Errorf("failed to fetch descriptor for tag %q: %w", tag, err)
+				return "", errors.Join(fmt.Errorf("failed to fetch descriptor for tag %q: %w", tag, err), data.Close())
```
jakobmoellerdev: store and most ReadCloser returning interface are defined in a way where if err != nill then data doesnt need a close. As such are we use this commit was needed?

bindings/go/oci/ctf/store.go
```diff
@@ -99,19 +105,44 @@ type repository struct {
...
+	var closeErr error
+	isFirstCall := true
+	return struct {
+		io.Reader
+		io.Closer
+	}{
+		Reader: rc,
+		Closer: closerFunc(func() error {
+			if !isFirstCall {
+				slog.Warn("Close called multiple times on locked reader.")
```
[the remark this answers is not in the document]
jakobmoellerdev: I think this is defensive programming and we should break on multiple closes

PR #1187, opened 2025-11-10
fix: LockedReader - thread-safe access to blobs

piotrjanik:
What this PR does / why we need it:

LockedReader provides thread-safe access to blob data by wrapping reads with mutex synchronization. It uses a pipe to stream data while holding a read lock on the underlying blob.

Which issue(s) this PR fixes:

Fixes https://github.com/open-component-model/open-component-model/issues/1143

### DIFF
--- a/bindings/go/blob/locked_reader.go
+++ b/bindings/go/blob/locked_reader.go
@@ -39,13 +39,6 @@ func NewLockedReader(ctx context.Context, mu *sync.RWMutex, rc io.ReadCloser) io
 				slog.ErrorContext(ctx, "failed to close reader", slog.String("error", err.Error()))
 			}
 		}
-		//// Get reader
-		//if rc, errs = blob.ReadCloser(); errs != nil {
-		//	if err := pw.CloseWithError(fmt.Errorf("unable to get reader: %w", errs)); err != nil {
-		//		slog.ErrorContext(ctx, "unable to close pipe with error", slog.String("error", err.Error()))
-		//	}
-		//	return
-		//}
 
 		go func() {
 			defer close(done)


bindings/go/blob/locked_reader.go
```diff
@@ -0,0 +1,62 @@
...
+		if err := pw.Close(); err != nil {
+			slog.ErrorContext(ctx, "unable to close reader", slog.String("error", err.Error()))
+		}
+	}()
+
+	return &LockedReader{pr: pr}, nil
+}
+
+func (lr *LockedReader) Read(p []byte) (n int, err error) {
+	return lr.pr.Read(p)
```
Skarlso: I'm stupid, but shouldn't the Read and the Close use the lock instead of the io.Copy in the go routine?

bindings/go/blob/locked_reader.go
```diff
@@ -0,0 +1,62 @@
...
+
+		defer func() {
+			if err := rc.Close(); err != nil {
+				slog.ErrorContext(ctx, "unable to close reader", slog.String("error", err.Error()))
+			}
+			mu.RUnlock()
+		}()
+
+		// Copy data through pipe
+		if _, err := io.Copy(pw, rc); err != nil {
```
Skarlso: You also need to check for context cancellation.

piotrjanik: in first case you are closing pipe with error and then wait till io.Copy is done. I doubt this will happen.

piotrjanik: Please check now @Skarlso  :)

bindings/go/blob/locked_reader.go
```diff
@@ -0,0 +1,86 @@
...
+// a read lock on the underlying blob.
+type LockedReader struct {
+	pr *io.PipeReader
+}
+
+// NewLockedReader creates a thread-safe reader for a blob that acquires a read lock
+// before streaming data. The lock is held during the entire copy operation to ensure
+// consistent access and prevent concurrent writes.
+func NewLockedReader(ctx context.Context, mu *sync.RWMutex, blob ReadOnlyBlob) (io.ReadCloser, error) {
+	mu.RLock()
```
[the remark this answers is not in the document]
piotrjanik: done

bindings/go/blob/locked_reader.go
```diff
@@ -0,0 +1,80 @@
...
+}
+
+// NewLockedReader creates a thread-safe reader for a blob that acquires a read lock
+// before streaming data. The lock is held during the entire copy operation to ensure
+// consistent access and prevent concurrent writes.
+func NewLockedReader(ctx context.Context, mu *sync.RWMutex, blob ReadOnlyBlob) (io.ReadCloser, error) {
+	pr, pw := io.Pipe()
+	// Copy goroutine - performs the actual data copy
+	go func() {
+		mu.RLock()
```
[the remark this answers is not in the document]
Skarlso: Also this needs to be a Lock and not an `RLock` since you are writing as well in  io.Copy.

bindings/go/blob/locked_reader.go
```diff
@@ -0,0 +1,80 @@
...
+// with mutex synchronization. It uses a pipe to stream data while holding
+// a read lock on the underlying blob.
+type LockedReader struct {
+	pr *io.PipeReader
+}
+
+// NewLockedReader creates a thread-safe reader for a blob that acquires a read lock
+// before streaming data. The lock is held during the entire copy operation to ensure
+// consistent access and prevent concurrent writes.
+func NewLockedReader(ctx context.Context, mu *sync.RWMutex, blob ReadOnlyBlob) (io.ReadCloser, error) {
```
Skarlso: There is no way this would return an error so you might as well just drop it.

bindings/go/blob/locked_reader.go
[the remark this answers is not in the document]
piotrjanik: I am "sure" I saw "Un" between R and lock :) Ok, done

bindings/go/blob/locked_reader.go
```diff
@@ -0,0 +1,80 @@
...
+	pr *io.PipeReader
+}
+
+// NewLockedReader creates a thread-safe reader for a blob that acquires a read lock
+// before streaming data. The lock is held during the entire copy operation to ensure
+// consistent access and prevent concurrent writes.
+func NewLockedReader(ctx context.Context, mu *sync.RWMutex, blob ReadOnlyBlob) io.ReadCloser {
+	pr, pw := io.Pipe()
+	// Copy goroutine - performs the actual data copy
+	go func() {
```
jakobmoellerdev: please make this nested function smaller

bindings/go/blob/locked_reader.go
```diff
@@ -0,0 +1,80 @@
...
+		done := make(chan struct{})
+		var copyErrs error
+		var err error
+		var rc io.ReadCloser
+		closePipe := func() {
+			if copyErrs != nil {
+				if err := pw.CloseWithError(fmt.Errorf("unable to copy data: %w", copyErrs)); err != nil {
+					slog.ErrorContext(ctx, "unable to close pipe with error", slog.String("error", err.Error()))
+				}
+			} else {
```
jakobmoellerdev: else if can be inlined

bindings/go/blob/locked_reader.go
```diff
@@ -0,0 +1,80 @@
...
+// before streaming data. The lock is held during the entire copy operation to ensure
+// consistent access and prevent concurrent writes.
+func NewLockedReader(ctx context.Context, mu *sync.RWMutex, blob ReadOnlyBlob) io.ReadCloser {
+	pr, pw := io.Pipe()
+	// Copy goroutine - performs the actual data copy
+	go func() {
+		mu.RLock()
+		defer mu.RUnlock()
+		done := make(chan struct{})
+		var copyErrs error
```
jakobmoellerdev: why do we need to track copyErrs and err separately?

bindings/go/blob/go.mod
```diff
@@ -20,6 +20,7 @@ require (
 	github.com/stretchr/objx v0.5.3 // indirect
 	github.com/wk8/go-ordered-map/v2 v2.1.8 // indirect
 	go.yaml.in/yaml/v2 v2.4.3 // indirect
+	golang.org/x/sync v0.18.0
```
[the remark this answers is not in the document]
piotrjanik: done

bindings/go/blob/locked_reader.go
piotrjanik: done

PR #1195, opened 2025-11-13
fix: publish run on submodule release workflow is incorrect

Skarlso:


.github/workflows/release-go-submodule.yaml
```diff
@@ -175,5 +175,8 @@ jobs:
     needs: [ version, release ]
     uses: ./.github/workflows/publish-helminput-plugin-component.yaml
     secrets: inherit
+    permissions:
+      id-token: write
```
jakobmoellerdev: dont you need package write?

jakobmoellerdev (approved):

PR #1223, opened 2025-11-18
chore(linter): run the verification sperately from the linter main run

Skarlso:

The previous _fix_ was still running the verification as part of the matrix job. This is now truly a separate job.


frewilhelm (changes requested):

.github/workflows/ci.yml
```diff
@@ -156,8 +156,38 @@ jobs:
...
+      - name: Read tool versions
+        id: versions
+        run: |
+          source .env
+          echo "golangci_lint=${GOLANGCI_LINT_VERSION}" >> $GITHUB_OUTPUT
+      - name: Verify the golangci-lint config
+        run:
+          task tools:golangci-lint/verify
+
+  golangci_lint_run:
```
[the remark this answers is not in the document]
Skarlso: Oh you're absolutely right. In fact, I won't rename this job instead to avoid renaming things.

.github/workflows/ci.yml
```diff
@@ -156,6 +156,36 @@ jobs:
...
+        run: |
+          source .env
+          echo "golangci_lint=${GOLANGCI_LINT_VERSION}" >> $GITHUB_OUTPUT
+      - name: Verify the golangci-lint config
+        run:
+          task tools:golangci-lint/verify
+
   golangci_lint:
     name: "golangci-lint"
     needs: discover_modules
```
frewilhelm: i think we should add ` golangci_lint_verify` as a requirement for the matrix to run. wdyt?

Skarlso: Sure 👍  Good idea.

Skarlso: you are on 🔥 brother

frewilhelm:
we need a rebase on this

PR #1280, opened 2025-11-28
chore: export digest functions to signing module

frewilhelm:


frewilhelm:
Need to merge this PR to fix the problems as the signing module is not published yet for the CLI

frewilhelm:
closed because upper-case letters in branch names are apparently not allowed

PR #1326, opened 2025-12-02
feat(cel): JSON schema CEL compatibility library

jakobmoellerdev:

This implements a partial port for translating JSON Schemas (based on https://github.com/santhosh-tekuri/jsonschema) into CEL type descriptors. Heavily inspired by the KRO and Kubernetes API Server Translation

The idea is that JSONSchema (draft 2020-12, draft 2019-09, draft-7, draft-6, draft-4) can be translated into a type descriptor that can be accessed via CEL expression. 

Assuming something like

```json
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "$id": "test/Formats.schema.json",
  "title": "Formats",
  "type": "object",
  "properties": {
    "duration": { "type": "string", "format": "duration" },
    "date": { "type": "string", "format": "date" },
    "datetime": { "type": "string", "format": "date-time" },
    "binary": { "type": "string", "format": "byte", "maxLength": 32 },
    "enumFmt": {
      "type": "string",
      "format": "date-time",
      "enum": ["2020-01-01T00:00:00Z", "2021-05-03T12:05:10Z"]
    }
  }
}
```

we can define a set of expressions in CEL that make an instance backed by such a scheme accessible:

Very simple accesses are field accesses such as `instance.duration`.

Furthermore, JSON Schema is very flexible in that we may allow "open" schemas with dynamic values:

```
{
  "$schema": "https://json-schema.org/draft/2020-12/schema",
  "$id": "test/OpenObject.schema.json",
  "title": "OpenObject",
  "type": "object"
}
```

In this case, CEL allows us to defer this object type to the runtime. 

This allows us to have a two stage object declaration language:

1. Static Type Analysis on provided JSON Schemas, while deferring unknown schemas to the runtime
2. Dynamic Run that also allows evaluating previously unknown schema values

This is the first major part of the transformation engine in https://github.com/open-component-model/ocm-project/issues/536 which will need a language specification that can actually access all of our OCM object specifications.

With CEL expressions we will be able to evaluate these expressions in the transfer specification effectively and allow a type checking system.


bindings/go/cel/jsonschema/santhosh-tekuri/v6/decl_type.go
```diff
@@ -0,0 +1,306 @@
...
+		}
+
+		return declTypeForSchema(
+			decl.NewListType(itemsType.Type, maxItems),
+			s,
+		)
+
+	case "object":
+		// If additionalProperties is itself a schema → treat as map<string, X>
+		// NEW: literal constraints of the additionalProperties schema must apply to all values
```
fabianburth: What did you change? How or why wasnt this the case before?

bindings/go/cel/jsonschema/santhosh-tekuri/v6/decl_type.go
```diff
@@ -0,0 +1,306 @@
...
+					maxProperties = estimateMaxAdditionalPropertiesFromMinSize(propsType.MinSerializedSize)
+				}
+				return declTypeForSchema(
+					decl.NewMapType(decl.StringType, propsType.Type, maxProperties),
+					s,
+				)
+			}
+			// Fallback: additionalProperties = true/false or unknown → Dyn
+			// NEW: this allows *any* key with dyn value
+			return declTypeForSchema(decl.NewMapType(decl.StringType, decl.DynType, decl.NoMaxLength), s)
```
fabianburth: in case of false, wouldnt we want to somehow not allow additional properties?

bindings/go/cel/jsonschema/santhosh-tekuri/v6/decl_type.go
jakobmoellerdev: tbh this comment is wrong, removing

bindings/go/cel/jsonschema/decl/value.go
```diff
@@ -0,0 +1,770 @@
...
+		combo[i+szRight] = oArr.Get(types.Int(i))
+	}
+	return types.DefaultTypeAdapter.NativeToValue(combo)
+}
+
+// Append adds another entry into the ListValue.
+func (lv *ListValue) Append(entry *DynValue) {
+	lv.Entries = append(lv.Entries, entry)
+	// The append resets all previously built indices.
+	lv.initValueSet = sync.Once{}
```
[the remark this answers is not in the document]
jakobmoellerdev: changed to pointer although im very sure this worked before. reason is that the sync once is stored in the listvalue pointer. and that gets a reset struct which works fine

PR #1330, opened 2025-12-03
feat(cel): CEL expression extraction and validation from strings

jakobmoellerdev:

```
// Package parser provides helpers to locate and validate CEL expressions
// embedded in strings.
//
// CEL expressions are delimited by the literal sequence `${` and a matching `}`.
// The parser recognizes string literals and escape sequences and enforces the
// rule that nested expressions are not allowed unless they appear inside a
// string literal. For example:
//
//   - `${outer("${inner}")}` is allowed
//   - `${outer(${inner})}` is not allowed and will cause ErrNestedExpression
//
// The primary behaviors implemented in this package:
//
//   - ParseSchemaless: returns all CEL expressions found in a schemaless map[string]any structure
```

```
// Package inspector provides static analysis inspection for CEL expressions.
//
// It parses CEL expressions in a given cel.Env and inspects the native AST to
// discover:
//   - resource dependencies (root variables and their accessed field paths),
//   - function and method calls, and
//   - unknown resources and functions.
//
// The analysis is purely structural: expressions are not evaluated, making
// inspector suitable for validation, dependency tracking, and tooling around
// CEL-based configuration.
```

This is the baseline that we need to actually interpret nested field dependencies between different elements in the transformation specification to allow dependency graph setup between two different nodes.

Its part of the bigger transformation engine in https://github.com/open-component-model/ocm-project/issues/536 which will need a language specification that can actually access all of our OCM object specifications.

This is basically a port of KROs expression parser so its behavior is almost identical

### DIFF
--- a/kubernetes/controller/go.mod
+++ b/kubernetes/controller/go.mod
@@ -393,10 +393,10 @@ require (
 	golang.org/x/time v0.14.0 // indirect
 	golang.org/x/tools v0.38.0 // indirect
 	gomodules.xyz/jsonpatch/v2 v2.5.0 // indirect
-	google.golang.org/api v0.253.0 // indirect
+	google.golang.org/api v0.255.0 // indirect
 	google.golang.org/genproto v0.0.0-20250922171735-9219d122eba9 // indirect
 	google.golang.org/genproto/googleapis/api v0.0.0-20250929231259-57b25ae835d4 // indirect
-	google.golang.org/genproto/googleapis/rpc v0.0.0-20251014184007-4626949a642f // indirect
+	google.golang.org/genproto/googleapis/rpc v0.0.0-20251029180050-ab9386a59fda // indirect
 	google.golang.org/grpc v1.76.0 // indirect
 	google.golang.org/protobuf v1.36.10 // indirect
 	gopkg.in/evanphx/json-patch.v4 v4.13.0 // indirect
--- a/kubernetes/controller/go.sum
+++ b/kubernetes/controller/go.sum
@@ -2378,8 +2378,8 @@ google.golang.org/api v0.108.0/go.mod h1:2Ts0XTHNVWxypznxWOYUeI4g3WdP9Pk2Qk58+a/
 google.golang.org/api v0.110.0/go.mod h1:7FC4Vvx1Mooxh8C5HWjzZHcavuS2f6pmJpZx60ca7iI=
 google.golang.org/api v0.111.0/go.mod h1:qtFHvU9mhgTJegR31csQ+rwxyUTHOKFqCKWp1J0fdw0=
 google.golang.org/api v0.114.0/go.mod h1:ifYI2ZsFK6/uGddGfAD5BMxlnkBqCmqHSDUVi45N5Yg=
-google.golang.org/api v0.253.0 h1:apU86Eq9Q2eQco3NsUYFpVTfy7DwemojL7LmbAj7g/I=
-google.golang.org/api v0.253.0/go.mod h1:PX09ad0r/4du83vZVAaGg7OaeyGnaUmT/CYPNvtLCbw=
+google.golang.org/api v0.255.0 h1:OaF+IbRwOottVCYV2wZan7KUq7UeNUQn1BcPc4K7lE4=
+google.golang.org/api v0.255.0/go.mod h1:d1/EtvCLdtiWEV4rAEHDHGh2bCnqsWhw+M8y2ECN4a8=
 google.golang.org/appengine v1.1.0/go.mod h1:EbEs0AVv82hx2wNQdGPgUI5lhzA/G0D9YwlJXL52JkM=
 google.golang.org/appengine v1.4.0/go.mod h1:xpcJRLb0r/rnEns0DIKYYv+WjYCduHsrkT7/EB5XEv4=
 google.golang.org/appengine v1.5.0/go.mod h1:xpcJRLb0r/rnEns0DIKYYv+WjYCduHsrkT7/EB5XEv4=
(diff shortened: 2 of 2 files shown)


bindings/go/cel/expression/parser/cel.go
```diff
@@ -0,0 +1,124 @@
...
+		// If we exit the loop and were still inside a string literal → error
+		if inString {
+			return nil, ErrNestedExpression
+		}
+
+		// Incomplete expression → skip one character after '${'
+		start = startIdx + 1
+
+	nextExpr:
+	}
```
fabianburth: nit: imho, a labeled loop / labeled continue is a bit cleaner here
```suggestion
func extractExpressions(str string) ([]string, error) {
	var expressions []string
	start := 0

    expression:
	for start < len(str) {
		// Find next "${"
		i := strings.Index(str[start:], exprStart)
		if i == -1 {
			break
		}
		startIdx := start + i

		endIdx := startIdx + len(exprStart)
		braces := 1
		inString := false
		escape := false

		for endIdx < len(str) {
			ch := str[endIdx]

			// Escape handling inside strings
			if escape {
				escape = false
				endIdx++
				continue
			}

			// Detect nested ${ BEFORE handling strings or braces
			if endIdx+1 < len(str) && str[endIdx:endIdx+2] == "${" {
				if inString {
					// Inside string literal → treated as normal text
					endIdx += 2
					continue
				}

				// Outside string literal → forbidden nested expression
				return nil, ErrNestedExpression
			}

			// Inside string literal
			if inString {
				if ch == '\\' {
					escape = true
					endIdx++
					continue
				}
				if ch == '"' {
					inString = false
				}
				endIdx++
				continue
			}

			// Outside string literal

			// Starting a new string literal
			if ch == '"' {
				inString = true
				endIdx++
				continue
			}

			// Normal brace processing
			switch ch {
			case '{':
				braces++
			case '}':
				braces--
				if braces == 0 {
					// Found matching end
					expressions = append(expressions, str[startIdx+len(exprStart):endIdx])
					start = endIdx + 1
					continue expression
				}
			}

			endIdx++
		}

		// If we exit the loop and were still inside a string literal → error
		if inString {
			return nil, ErrNestedExpression
		}

		// Incomplete expression → skip one character after '${'
		start = startIdx + 1
	}
```

bindings/go/cel/expression/variable/variable.go
```diff
@@ -0,0 +1,127 @@
...
+// Contains a variable named "spec.mycustomReplicasField". Variables can be
+// static or dynamic. Static variables are resolved at the beginning of the
+// execution and their value is constant. Dynamic variables are resolved at
+// runtime and their value can change during the execution.
+//
+// ResourceVariables are an extension of CELField, and they contain additional
+// information about the variable kind.
+type ResourceField struct {
+	// CELField is the object that contains the expression, the path, and
+	// the expected type (OpenAPI schema).
```
fabianburth:
```suggestion
	// the expected type (JSON schema).
```

bindings/go/cel/expression/parser/cel.go
```diff
@@ -0,0 +1,124 @@
...
+		}
+		startIdx := start + i
+
+		endIdx := startIdx + len(exprStart)
+		braces := 1
+		inString := false
+		escape := false
+
+		for endIdx < len(str) {
+			ch := str[endIdx]
```
matthiasbruns: nit: `ch` reads like channel - maybe `c` is better for a char?

bindings/go/cel/expression/parser/cel.go
```diff
@@ -0,0 +1,124 @@
+package parser
+
+import (
+	"errors"
+	"strings"
+)
+
+const (
+	// CEL expressions are enclosed between "${" and "}"
+	exprStart = "${"
```
matthiasbruns: this comes from kro right?

https://github.com/open-component-model/open-component-model/blob/24b8f8099d9e5e74c2aeec4b898b72374da04f98/docs/adr/0005_transformation.md?plain=1#L13

wouldn't it be easier to go with go-templates if we had other characters to wrap the expressions?

bindings/go/cel/expression/variable/variable.go
```diff
@@ -0,0 +1,127 @@
...
+//	  replicas: ${schema.spec.mycustomReplicasField + 5}
+//
+// Contains a variable named "spec.mycustomReplicasField". Variables can be
+// static or dynamic. Static variables are resolved at the beginning of the
+// execution and their value is constant. Dynamic variables are resolved at
+// runtime and their value can change during the execution.
+//
+// ResourceVariables are an extension of CELField, and they contain additional
+// information about the variable kind.
+type ResourceField struct {
```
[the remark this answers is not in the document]
Skarlso: I think it would be better. Just because it makes updating easier. If the KRO one changes it's just a copy. But if we alter it too much, it's going to be difficult to maintain upstream changes and figure out where goes what.

WDYT?

bindings/go/cel/expression/parser/cel.go
```diff
@@ -0,0 +1,124 @@
...
+					continue
+				}
+
+				// Outside string literal → forbidden nested expression
+				return nil, ErrNestedExpression
+			}
+
+			// Inside string literal
+			if inString {
+				if ch == '\\' {
```
Skarlso: CEL allows `"""` and `'''` do we care? 

```
STRING_LIT     ::= [rR]? ( "    ~( " | \r | \n )*  "
                         | '    ~( ' | \r | \n )*  '
                         | """  ~"""*              """
                         | '''  ~'''*              '''
                         )
```

I'm guessing we don't want to be bothered with the full range? :)

bindings/go/cel/expression/parser/cel.go
```diff
@@ -0,0 +1,124 @@
...
+// unescaped expression. Nested expressions are only allowed when the
+// inner expression appears inside a string literal, for example:
+//
+//	${outer("${inner}")}   // allowed
+//	${outer(${inner})}     // not allowed
+var ErrNestedExpression = errors.New("nested expressions are not allowed unless inside string literals")
+
+// extractExpressions extracts all non-nested CEL expressions from a string.
+// It returns an error if it encounters an illegal nested expression.
+func extractExpressions(str string) ([]string, error) {
```
Skarlso: This is a nit however, just want to put this out there. I would love for this to be on a struct with proper internal indexes keeping track for where they are and we could call things like `peekChar` and `skipWhitespace`. It would read so much better. And we always write code for reading, and not writing.

Hopefully, we write this once and then never look at it again, but I guarantee it that we will have to do that eventually. :)  For now, this isn't that complex to understand. Yet. :)

Skarlso (commented):
A few comments really, nothing major. This looks really great!

bindings/go/cel/expression/parser/cel.go
jakobmoellerdev: forgot unfinished expression handling in my refactored func, now tests are okidoki again

bindings/go/cel/expression/parser/cel.go
jakobmoellerdev: go templates are dynamic and a DSL, cel is a full expression language that is type safe by default.

PR #1335, opened 2025-12-04
feat(624): Implement apply sets from kro for deployer #624

matthiasbruns:

We want to use the apply-sets from 

- https://github.com/kubernetes/enhancements/blob/master/keps/sig-cli/3659-kubectl-apply-prune/README.md
- https://github.com/kubernetes-sigs/kro/tree/main/pkg/applyset 

to handle pruning without side-effects

Fixes: https://github.com/open-component-model/ocm-project/issues/624

### DIFF
--- a/kubernetes/controller/chart/templates/rbac/manager-role.yaml
+++ b/kubernetes/controller/chart/templates/rbac/manager-role.yaml
@@ -66,6 +66,7 @@ rules:
         - resourcegraphdefinitions
       verbs:
         - create
+        - delete
         - list
         - patch
         - update


kubernetes/controller/internal/controller/deployer/applyset.go
```diff
@@ -0,0 +1,194 @@
...
+	if gvk.Version == "" && mapping.GroupVersionKind.Version != "" {
+		logger.Info("apiVersion will be defaulted to match discovered rest mapping", "defaultAPIVersion", mapping.GroupVersionKind.Version)
+		gvk.Version = mapping.GroupVersionKind.Version
+		obj.SetGroupVersionKind(gvk)
+	}
+	return nil
+}
+
+// updateStatus replaces the deployed status with the new set of objects.
+func (a *ApplySet) updateStatus(objs []*unstructured.Unstructured) {
```
matthiasbruns: TODO rename function

matthiasbruns:
ApplySet flow chart

```mermaid
flowchart TD
    A[User runs kubectl apply<br/>--applyset NAME --prune] --> B[Parse applyset flag]

    B --> C{ApplySet parent exists}
    C -->|No| D[Create ApplySet parent<br/>Secret or ConfigMap or Custom]
    C -->|Yes| E[Load existing parent]

    D --> F[Build ApplySet identity<br/>and field manager]
    E --> F

    F --> G[Compute ApplySet membership selector]

    G --> H[Pre update parent<br/>store intended membership]
    H --> I[Apply desired manifests]

    I --> J[Label each applied object<br/>with applyset tracking labels]

    J --> K{Prune enabled}
    K -->|No| Z[Done]
    K -->|Yes| L[Discover prunable namespaces]
    
    L --> M[Discover prunable resource kinds]
    M --> N[List live objects<br/>matching ApplySet selector]

    N --> O{Object in desired set}
    O -->|Yes| P[Keep object]
    O -->|No| Q[Prune object<br/>delete resource]

    P --> R{More objects}
    Q --> R
    R -->|Yes| O
    R -->|No| Z[ApplySet complete]

```

matthiasbruns:
Screenshot 2026-01-02 at 16 59 33
deployer ApplySet labels

kubernetes/controller/internal/controller/deployer/deployer_controller.go
```diff
@@ -513,55 +515,90 @@ func digestSpec(s string) (ocmv1.DigestSpec, error) {
...
-// applied object reference.
-func (r *Reconciler) apply(ctx context.Context, resource *deliveryv1alpha1.Resource, deployer *deliveryv1alpha1.Deployer, obj *unstructured.Unstructured) error {
-	setOwnershipLabels(obj, resource, deployer)
-	setOwnershipAnnotations(obj, resource)
-	if err := controllerutil.SetControllerReference(deployer, obj, r.Scheme); err != nil {
-		return fmt.Errorf("failed to set controller reference on object: %w", err)
+	// Add all objects to the ApplySet
+	for _, obj := range objs {
+		// Clone the object to avoid modifying the original
+		objCopy := obj.DeepCopy()
```
jakobmoellerdev: isnt apply set object already deepcopied? whats the reason for the deep copy here?

kubernetes/controller/test/e2e/e2e_examples_test.go
```diff
@@ -117,6 +119,12 @@ var _ = Describe("controller", func() {
 					"pod", "-l", "app.kubernetes.io/name="+example.Name()+"-podinfo",
 				)).To(Succeed())
 
+				By("verifying ApplySet labels on deployer (parent)")
+				verifyDeployerApplySetLabelsAndAnnotations(ctx, example.Name())
+
+				By("verifying ApplySet labels on deployed resources")
```
jakobmoellerdev: This verification is not testing correctly because you are basing the test on the status. the issue here is that your test would be skipped if your status is incorrect. I was expecting a simple 

- deploy 2 kubernetes resources with 1 ocm resource cr and 1 deployer cr
- update the ocm resource
- expect that 1 kubernetes resource is pruned

kubernetes/controller/internal/controller/applyset/applyset.go
```diff
@@ -0,0 +1,833 @@
...
+	observed.SetGroupVersionKind(gvk)
+
+	// Set namespace if applicable
+	ns := obj.GetNamespace()
+	if ns == "" && restMapping.Scope.Name() == meta.RESTScopeNameNamespace {
+		ns = a.parent.GetNamespace()
+		if ns == "" {
+			ns = metav1.NamespaceDefault
+		}
+	}
```
Skarlso: We should call `SetNamespace` on the object if we have a namespace?

kubernetes/controller/internal/controller/applyset/applyset.go
```diff
@@ -0,0 +1,833 @@
...
+
+	logger.Info("found objects to prune", "count", len(pruneObjects))
+
+	deleteOptions := []client.DeleteOption{}
+	if dryRun {
+		deleteOptions = append(deleteOptions, client.DryRunAll)
+	}
+
+	for _, obj := range pruneObjects {
```
Skarlso: This is a serial operation right? The doc says pruning is concurrent. But this isn't. I guess you should fix the doc there. Finding to be pruned objects is concurrent, but the operation itself isn't.

kubernetes/controller/internal/controller/applyset/applyset.go
```diff
@@ -0,0 +1,848 @@
...
+
+	logger.Info("apply operation complete",
+		"applied", len(result.Applied),
+		"pruned", len(result.Pruned),
+		"errors", len(result.Errors))
+
+	return result, nil
+}
+
+func (a *applySet) apply(ctx context.Context, result *Result, dryRun bool) error {
```
Skarlso: This will never return an error. 🤔 What's the point of the return error?

kubernetes/controller/internal/controller/applyset/applyset.go
```diff
@@ -0,0 +1,848 @@
...
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			logger.Info("object does not exist in cluster")
+			observed = nil
+		} else {
+			return nil, fmt.Errorf("error getting object from cluster: %w", err)
+		}
+	}
+
+	// remove managed fields from obj
```
Skarlso: this comment is no longer true, right?

kubernetes/controller/internal/controller/applyset/applyset.go
```diff
@@ -0,0 +1,848 @@
...
+	for _, gk := range gks {
+		gk = strings.TrimSpace(gk)
+		if gk != "" {
+			result = append(result, gk)
+		}
+	}
+	return result
+}
+
+func (a *applySet) getNamespacesToCheck() sets.Set[string] {
```
Skarlso: @jakobmoellerdev 

As per:

>The value of this annotation is a comma-separated list of the names of the namespaces (other than the ApplySet namespace) in which objects are found, for example default,kube-system,ns1,ns2. We reserve the empty value for future use. As with applyset.k8s.io/contains-group-kinds, this list of namespaces must be sorted alphabetically, and should be minimal (other than transiently during ApplySet mutations).

Is this something we should do here or are we leaving this up to the user?

kubernetes/controller/internal/controller/applyset/applyset.go
matthiasbruns: fixed

kubernetes/controller/internal/controller/applyset/applyset.go
```diff
@@ -0,0 +1,848 @@
...
+		"gks", len(gks),
+		"namespaces", len(namespaces))
+
+	var pruneObjects []PrunableObject
+	var mu sync.Mutex
+	eg, egctx := errgroup.WithContext(ctx)
+
+	concurrency := a.concurrency
+	if concurrency <= 0 {
+		concurrency = 10
```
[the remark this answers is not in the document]
matthiasbruns: good catch

kubernetes/controller/internal/controller/applyset/applyset.go
```diff
@@ -0,0 +1,848 @@
...
+		parent.GetNamespace(),
+		gvk.Kind,
+		gvk.Group,
+	}, ApplySetIDPartDelimiter)
+
+	hashed := sha256.Sum256([]byte(unencoded))
+	b64 := base64.RawURLEncoding.EncodeToString(hashed[:])
+
+	// Label values must start and end with alphanumeric values
+	return fmt.Sprintf(V1ApplySetIdFormat, b64)
```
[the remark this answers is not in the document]
matthiasbruns: well this might happen, if we change V1ApplySetIdFormat's value. but in the current implementation it cannot

I can add a safe-guard and truncate the value always or add a warning case and truncate it there - but I will that we might use the uniqueness in that case.

the b64 should always be 43 chars long, so we have 10 chars left for the prefix.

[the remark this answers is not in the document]
matthiasbruns: and the prefix is `applyset-` - so only a code change could cause that case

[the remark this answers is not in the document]
matthiasbruns:
```go
	// Label values must start and end with alphanumeric values
	value := fmt.Sprintf(V1ApplySetIdFormat, b64)
	if len(value) > 63 {
		slog.Warn("ApplySet ID is too long, truncating to 63 characters", "originalLength", len(value))
		value = value[:63]
	}
```

matthiasbruns:
> This is still missing a case for delete. if the deployer is deleted, how does the apply set behave?
> 
> I think we have two options:
> 
> * prune all objects from the applier
> * orphan all objects
> 
> Currently im favoring pruning.

I updated the applyset  e2e test and tested that all children should be deleted

kubernetes/controller/internal/controller/deployer/deployer_controller.go
jakobmoellerdev: I was expecting a delete prune here: https://github.com/open-component-model/open-component-model/pull/1335/files#diff-b665849ca2a0e83c6779cbac619ec2d2fff2815a0cd58bfcc439bc447000cdafR247 

Why do you not need to delete any objects when the deployer is deleted?

matthiasbruns: My assumption was that since I have set an owner, the children will be removed automatically  - read up on the topic and learned otherwise.

I am testing my clean up finalizer right now.

matthiasbruns:
To unblock this PR, I added a broader permission to the deployer

kubernetes/controller/internal/controller/applyset/discovery.go
```diff
@@ -0,0 +1,104 @@
...
+		gk = strings.TrimSpace(gk)
+		if gk != "" {
+			result = append(result, gk)
+		}
+	}
+	return result
+}
+
+func (a *applySet) getNamespacesToCheck() sets.Set[string] {
+	namespaces := sets.New[string]()
```
Skarlso: Shouldn't this default to `default`? If we have a cluster scoped object namespace will be empty and it might not find objects in the default namespace. 🤔 I wonder how much kubernetes would see empty namespace as default. I'm not sure.

kubernetes/controller/internal/controller/applyset/applyset.go
```diff
@@ -0,0 +1,303 @@
...
+	// These labels are added on top of the required ApplySet labels.
+	ToolLabels map[string]string
+
+	// Concurrency controls the maximum number of concurrent apply and prune operations.
+	// If not provided or <= 0, defaults to the number of objects in the ApplySet.
+	Concurrency int
+}
+
+// Set represents an ApplySet that can be applied to a Kubernetes cluster.
+type Set interface {
```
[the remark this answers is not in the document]
matthiasbruns: it is applied now

jakobmoellerdev:
I think that means we can merge as is tbh

matthiasbruns:
> I think that means we can merge as is tbh

I am also for merging the new applyset - let me fix the kro and testing permissions and I will ping you for a review

matthiasbruns:
@jakobmoellerdev please re-re-re-re-re-review :)

jakobmoellerdev (approved):

kubernetes/controller/internal/controller/deployer/deployer_controller.go
```diff
@@ -227,7 +231,37 @@ func (r *Reconciler) Untrack(ctx context.Context, deployer *deliveryv1alpha1.Dep
...
-	controllerutil.RemoveFinalizer(deployer, resourceWatchFinalizer)
+	return nil
+}
+
+func (r *Reconciler) pruneWithApplySet(ctx context.Context, deployer *deliveryv1alpha1.Deployer) error {
+	logger := log.FromContext(ctx).WithValues("deployer", deployer.Name, "namespace", deployer.Namespace)
+
+	set := r.createApplySet(deployer, logger)
+
+	metaData, err := set.Project(nil)
```
fabianburth:
```suggestion
	metadata, err := set.Project(nil)
```
nit

kubernetes/controller/internal/controller/deployer/deployer_controller.go
```diff
@@ -227,7 +231,37 @@ func (r *Reconciler) Untrack(ctx context.Context, deployer *deliveryv1alpha1.Dep
...
+	}
+
+	// Log results
+	logger.Info("ApplySet prune operation complete", "pruned", len(result.Pruned))
+
+	// nothing more to prune, remove finalizer
+	if result.HasPruned() {
+		logger.Info("pruned resources, doing one more pruning until nothing more to prune")
+		return fmt.Errorf("waiting for all resources to be pruned")
+	}
```
fabianburth: nit: maybe a bit of documentation on why this is necessary would be helpful.

Maybe something like:
Prune calls delete on every resource found, even if its already being deleted. If we were to remove this check, the deployer might be deleted while a child is stuck in terminating state.

kubernetes/controller/internal/controller/deployer/deployer_controller.go
```diff
@@ -321,6 +354,42 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
+		if controllerutil.ContainsFinalizer(deployer, applySetPruneFinalizer) {
+			logger.Info("pruning ApplySet before removing finalizer")
+			if err := r.pruneWithApplySet(ctx, deployer); err != nil {
+				logger.Error(err, "waiting for ApplySet to be pruned before removing finalizer")
+				errs = append(errs, err)
+			} else {
+				logger.Info("successfully pruned ApplySet for deployer")
+				controllerutil.RemoveFinalizer(deployer, applySetPruneFinalizer)
+			}
+		}
```
fabianburth: I assume the exponential backoff behavior of `pruneWithApplySet` (which I pointed out in my other comment) can lead to very large exponential backoffs. 
Thus, shouldn't we wait with `Untrack`ing resources until they are actually pruned, so if someone fixes a stuck resource, we actually get an event and dont have to wait the very long backoff time?

fabianburth (commented):
Mostly nits.

PR #1339, opened 2025-12-05
fix: ignore 403 and 401 for pinging the registry base

Skarlso:


bindings/go/oci/resolver/url/resolver.go
```diff
@@ -83,6 +86,17 @@ func (resolver *CachingResolver) Ping(ctx context.Context) error {
 		r.Client = resolver.baseClient
 	}
 	if err := r.Ping(ctx); err != nil {
+		errResp := &errcode.ErrorResponse{}
+		if ok := errors.As(err, &errResp); ok {
+			switch errResp.StatusCode {
+			case http.StatusForbidden, http.StatusUnauthorized:
+				// According to the distribution spec the registry might opt for credentials
```
frewilhelm: We might reference the distribution spec because we discussed this for so long now :D

Skarlso: Sure 👍

Skarlso:
I tested this with our e2e test locally and it passed 🟢

frewilhelm:
> I tested this with our e2e test locally and it passed 🟢

With the repository migration and without the mock-creds?

PR #1341, opened 2025-12-05
fix: 743 add semver check cli tests

matthiasbruns:
Adds semver checks to `ocm add cv` analog to `ocm get cv`

- This PR will only update cli and add tests when https://github.com/open-component-model/open-component-model/pull/1317 is merged to verify, that the constructor validation works as expected
- This Pr also reduces the logging level in `get cv` when requesting an invalid version https://github.com/open-component-model/open-component-model/pull/1393 

Fixes: https://github.com/open-component-model/ocm-project/issues/743
Fixes: https://github.com/open-component-model/ocm-project/issues/752

### DIFF
--- a/cli/cmd/get/component-version/cmd.go
+++ b/cli/cmd/get/component-version/cmd.go
@@ -161,7 +161,8 @@ func GetComponentVersion(cmd *cobra.Command, args []string) error {
 
 	reference := args[0]
 
-	// We have a reference, check if it is a component reference.
+	// We want to allow non-semver compatible versions to be used in component references.
+	// Therefore, we ignore semver compatibility errors during parsing here.
 	ref, compErr := compref.Parse(reference, []compref.Option{
 		compref.IgnoreSemverCompatibility(),
 	}...)
--- a/cli/internal/reference/compref/compref.go
+++ b/cli/internal/reference/compref/compref.go
@@ -173,7 +173,6 @@ func Parse(input string, opts ...Option) (*Ref, error) {
 		versionPart = input[idx+1:]
 		input = input[:idx]
 		if !versionRegex.MatchString(versionPart) {
-
 			// If IgnoreSemverCompatibility is not set, return an error
 			// to ensure strict compliance with semantic versioning.
 			if !parsedOptions.IgnoreSemverCompatibility {


cli/integration/go.mod
```diff
@@ -102,7 +102,7 @@ require (
 	google.golang.org/protobuf v1.36.10 // indirect
 	gopkg.in/yaml.v3 v3.0.1 // indirect
 	ocm.software/open-component-model/bindings/go/configuration v0.0.9 // indirect
-	ocm.software/open-component-model/bindings/go/constructor v0.0.2 // indirect
+	ocm.software/open-component-model/bindings/go/constructor v0.0.3-0.20251205090143-40ef3dda2ab1 // indirect
```
Skarlso: Don't we need to have a proper release for the constructor?

cli/integration/plugin_registry_list_integration_test.go
```diff
@@ -210,43 +210,56 @@ provider:
...
+	if len(info) > 0 {
+		jsonBytes, _ := json.MarshalIndent(info, "", "  ")
+		jsonStr := string(jsonBytes)
+
+		fmt.Fprintf(&b, `
     labels:
       - name: %s
-        value:
-%s`, list.PluginInfoKey, labels)
+        value: |
```
matthiasbruns: This feels like I am breaking current plugins - as per `schem.json`, we are not allowed to use yaml objects but only strings for `label.value`

see https://github.com/open-component-model/open-component-model/blob/main/bindings/go/constructor/spec/v1/resources/schema-2020-12.json#L304

`label.value` ref https://github.com/open-component-model/open-component-model/blob/main/bindings/go/constructor/spec/v1/resources/schema-2020-12.json#L45

matthiasbruns: According to old ocm, label should be an object in v2
https://github.com/open-component-model/ocm/blob/main/resources/component-descriptor-v2-schema.yaml#L29

and v3
https://github.com/open-component-model/ocm/blob/main/resources/component-descriptor-ocm-v3-schema.yaml#L44

so I will move this into new ocm as well

Skarlso: Nice catch!

cli/integration/go.mod
```diff
@@ -22,6 +22,8 @@ require (
 	oras.land/oras-go/v2 v2.6.0
 )
 
+replace ocm.software/open-component-model/bindings/go/constructor => ../../bindings/go/constructor
```
Skarlso: Do you still need this replace? Or should this now be okay since the schema is merged?

fabianburth:
I think we also wanted to make validation failures a warning instead of an error for `get cv`, right?

cli/cmd/get/component-version/cmd.go
```diff
@@ -162,7 +162,9 @@ func GetComponentVersion(cmd *cobra.Command, args []string) error {
 	reference := args[0]
 
 	// We have a reference, check if it is a component reference.
-	ref, compErr := compref.Parse(reference)
+	ref, compErr := compref.Parse(reference, []compref.Option{
```
matthiasbruns: only did that for `get cv` with the semver ignore option

fabianburth (approved):

PR #1389, opened 2025-12-10
fix(registry): add error handling for empty types in Register method

jakobmoellerdev:

if we get called without a type to default, we should error. we can think of a builder pattern like kubernetes controllers have later.


bindings/go/runtime/registry.go
```diff
@@ -208,6 +208,10 @@ func (r *Scheme) RegisterWithAlias(prototype Typed, types ...Type) error {
 	rt := reflect.TypeOf(prototype).Elem()
 	defaultTyp, defaultExists := r.defaults.GetRight(rt)
 	if !defaultExists {
```
frewilhelm: So, if no types were passed and a default does exist, we will return nil and do nothing. That seems odd

PR #1392, opened 2025-12-10
fix: use the latest runtime package to fix the none deterministic failure

Skarlso:
Fix the runner: https://github.com/open-component-model/open-component-model/actions/runs/20093305963/job/57645974622

### DIFF
--- a/bindings/go/runtime/registry.go
+++ b/bindings/go/runtime/registry.go
@@ -202,6 +202,10 @@ func IsTypeAlreadyRegisteredError(err error) bool {
 // Note that if Scheme.RegisterWithAlias or Scheme.MustRegister were called before,
 // even the first type will be counted as an alias.
 func (r *Scheme) RegisterWithAlias(prototype Typed, types ...Type) error {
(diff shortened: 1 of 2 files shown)


kubernetes/controller/internal/configuration/config.go
```diff
@@ -61,44 +61,34 @@ func createConfigFromDockerConfig(data []byte) (*genericv1.Config, error) {
...
-	credScheme := runtime.NewScheme()
-	credentialsv1.MustRegister(credScheme)
-
-	credConfig := &credentialsv1.Config{}
-	if _, err := credScheme.DefaultType(credConfig); err != nil {
-		return nil, fmt.Errorf("failed to get default type for credential config type")
+	credConfig := &credentialsv1.Config{
+		Type:         runtime.NewUnversionedType(credentialsv1.ConfigType),
+		Repositories: []credentialsv1.RepositoryConfigEntry{{raw}},
 	}
```
Skarlso: This is _avoiding_ the problem I guess. Grumble grumble.

PR #1396, opened 2025-12-11
fix: trim trailing slashes from the base-path

Skarlso:

Avoids things like this:
```yaml
Component:         bob.poc.sap.com/root
...
    Repository Spec:
      Base URL:  mcp-blueprints.common.repositories.cloud.sap/ocm/
...
```

ending up like this:

```
Message:               failed to list versions: failed to get store for reference: invalid reference: invalid repository "ocm//component-descriptors/bob.poc.sap.com/root"
```

Because, basically oras has `parts := strings.SplitN(artifact, "/", 2)` in it's parser which resulted in the repository section being cut as `ocm//...` for the full reference of `mcp-blueprints.common.repositories.cloud.sap/ocm//component-descriptors/bob.poc.sap.com/root`.

### DIFF
--- a/bindings/go/oci/repository/repository.go
+++ b/bindings/go/oci/repository/repository.go
@@ -98,13 +98,13 @@ func buildResolver(client remote.Client, repository *ocirepospecv1.Repository) (
 		return nil, fmt.Errorf("a base url is required")
 	}
 
-	purl, err := runtime.ParseURLAndAllowNoScheme(repository.BaseUrl)
+	baseURL := strings.TrimSuffix(repository.BaseUrl, "/")
+	purl, err := runtime.ParseURLAndAllowNoScheme(baseURL)
 	if err != nil {
 		return nil, fmt.Errorf("could not parse OCI repository URL %q: %w", repository.BaseUrl, err)
 	}
 
 	// Extract SubPath from BaseUrl if not explicitly set
-	baseURL := repository.BaseUrl
 	subPath := repository.SubPath
 	if subPath == "" && purl.Path != "" && purl.Path != "/" {
 		// Use path as SubPath and Host as BaseURL.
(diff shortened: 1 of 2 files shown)


bindings/go/oci/repository/repository.go
```diff
@@ -98,13 +98,13 @@ func buildResolver(client remote.Client, repository *ocirepospecv1.Repository) (
 		return nil, fmt.Errorf("a base url is required")
 	}
 
-	purl, err := runtime.ParseURLAndAllowNoScheme(repository.BaseUrl)
+	baseURL := strings.TrimSuffix(repository.BaseUrl, "/")
+	purl, err := runtime.ParseURLAndAllowNoScheme(baseURL)
```
frewilhelm: What about fixing it in `ParseURLAndAllowNoScheme`?

Skarlso: I don't want to touch that because it's part of the identity matching.

Skarlso: Also, I'm not using the end result there. :) I need the right baseURL :)

Skarlso:
I'm not using the outcome of that. But I'll explain a bit better later.

Skarlso:
Okay, so, the problem with _just_ doing this in `ParseURLAndAllowNoScheme` is that I'm not actually using that result in the baseURL, right? Because, that results in the port being cut off if it exists and put into the port section. And also, the path is now cut. I could re-assemble that and recreate the baseURL again in its original format, but that is prone to error in some form or the other.

So what I want is to use the original format. How it was defined originally! Especially if path WAS set, because then the baseURL must remain as it was! So if it was `domain:5555/org` and subPath is set to `repo` then it needs to remain `domain:5555/org` for later the reference parser to work and `repo` being added correctly.

That said, of course we can try and refactor this entire thing and then construct the original URL back into how the baseurl was defined, but I feel like that's a hassle. :D 

And that is why. :)

PR #1404, opened 2025-12-12
chore: add type property to access

fabianburth:
For each property listed in the required section of a json schema, the json schema linter requires a corresponding property.


fabianburth:
Actually, checked our tests, should be covered by the descriptor tests.

PR #1437, opened 2025-12-16
fix: escape type names

fabianburth:
We traverse through the types based on segments. Those segments are parsed by `strings.Split(s, ".")`. Since our json schema types names are something like `ocm.software.MyType.schema.json.field`, this obviously leads to issues. `ocm`, `software`, .. all end up being segments. But we only want `ocm.software.MyType.schema.json` to be one segment and `field` to be another segment.

This fix is required to ensure correct structural type checking.

Contributes to https://github.com/open-component-model/ocm-project/issues/536

### DIFF
--- a/bindings/go/cel/jsonschema/santhosh-tekuri/v6/schema.go
+++ b/bindings/go/cel/jsonschema/santhosh-tekuri/v6/schema.go
@@ -2,6 +2,7 @@ package jsonschema
 
 import (
 	stjsonschemav6 "github.com/santhosh-tekuri/jsonschema/v6"
+
 	"ocm.software/open-component-model/bindings/go/cel/jsonschema"
 )
 


bindings/go/cel/jsonschema/santhosh-tekuri/v6/conversion_test.go
```diff
@@ -42,7 +42,7 @@ func TestSchemas(t *testing.T) {
 		{
 			file: "bindings/go/cel/jsonschema/santhosh-tekuri/v6/testdata/Config.schema.json",
 			verify: func(t *testing.T, decl *jsonschema.DeclType) {
-				require.Equal(t, "ocm.software/open-component-model/bindings/go/rsa/signing/v1alpha1/schemas/Config.schema.json", decl.TypeName())
+				require.Equal(t, "ocm__dot__software__slash__open__dash__component__dash__model__slash__bindings__slash__go__slash__rsa__slash__signing__slash__v1alpha1__slash__schemas__slash__Config__dot__schema__dot__json", decl.TypeName())
```
Skarlso: WHAT :D

PR #1440, opened 2025-12-17
fix: 799 broken constructor resource parsing for type, version & access

matthiasbruns:
This PR fixes the oversight in the schema.json integration for constructor parsing.
As per the [getting-started docs](https://ocm.software/dev/docs/getting-started/) we should be allowed to parse this constructor

```yml
# specify a schema to validate the configuration and get auto-completion in your editor
# yaml-language-server: $schema=https://ocm.software/schemas/configuration-schema.yaml
components:
- name: github.com/acme.org/helloworld
  # version needs to follow "relaxed" SemVer
  version: 1.0.0
  provider:
    name: acme.org
  resources:
    # local file resource
    - name: mylocalfile
      type: blob
      input:
        type: file
        path: ./my-local-resource.txt
    # remote image resource
    - name: image
      type: ociImage
      version: 1.0.0
      access:
        type: ociArtifact
        imageReference: ghcr.io/stefanprodan/podinfo:6.9.1
```

After checking old ocm [schema](https://github.com/open-component-model/ocm/blob/main/resources/component-descriptor-v2-schema.yaml#L237) we should require to suport these properties:

```yml
      - 'name'
      - 'version' # for local resources, this must match component's version
      - 'type'
      - 'relation'
      - 'access'
```

Fixes: https://github.com/open-component-model/ocm-project/issues/799

### DIFF
--- a/bindings/go/constructor/spec/v1/resources/schema-2020-12.json
+++ b/bindings/go/constructor/spec/v1/resources/schema-2020-12.json
@@ -333,8 +333,16 @@
       "$ref": "#/$defs/commonResourceProperties",
       "oneOf": [
         {
-          "required": ["version", "relation", "access"],
+          "required": [
+            "type",
+            "version",
+            "access"
+          ],
           "properties": {
+            "type": {
+              "description": "Type of the resource",
+              "$ref": "#/$defs/ocmType"
+            },
             "version": {
               "description": "Version of the resource",
               "$ref": "#/$defs/relaxedSemver"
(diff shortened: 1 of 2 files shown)


jakobmoellerdev (changes requested):

matthiasbruns:
updated the schema to either allow

- type, version access -> allows relations (external, local) with additional digest
- type, input -> local only

Skarlso (approved):

PR #1503, opened 2025-12-29
fix(controllers): do not return an error when digest is not available

frewilhelm:

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,7 +37,7 @@ jobs:
         with:
           version: 3.x
           repo-token: ${{ secrets.GITHUB_TOKEN }}
-      - uses: actions/checkout@93cb6efe18208431cddfb8368fd83d5badbf9bfd # v5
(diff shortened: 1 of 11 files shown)


kubernetes/controller/internal/controller/component/component_controller.go
```diff
@@ -426,21 +428,22 @@ func (r *Reconciler) DetermineEffectiveVersionFromRepo(ctx context.Context, comp
...
 		logger.Info("no verifications configured, skipping signature verification")
 
 		return nil
 	}
 
-	logger.Info("verifying component version signatures", "verifications", len(verifications))
-
 	if err := signing.IsSafelyDigestible(&desc.Component); err != nil {
-		return err
+		event.New(r.EventRecorder, component, nil, eventv1.EventSeverityInfo, err.Error())
```
jakobmoellerdev: is the error stable? if not it will spam the event log of k8s

matthiasbruns (approved):

PR #1521, opened 2026-01-02
fix: pin the kube version for kind in the .env file

Skarlso:

### DIFF
--- a/bindings/go/cel/jsonschema/santhosh-tekuri/v6/conversion_test.go
+++ b/bindings/go/cel/jsonschema/santhosh-tekuri/v6/conversion_test.go
@@ -66,6 +66,21 @@ func TestSchemas(t *testing.T) {
 			},
 		},
 		{
+			file: "bindings/go/cel/jsonschema/santhosh-tekuri/v6/testdata/TrueSchema.schema.json",
+			verify: func(t *testing.T, decl *jsonschema.DeclType) {
+				require.True(t, decl.IsObject())
+				sa := decl.Fields["trueProp"]
+				require.Equal(t, "dyn", sa.Type.TypeName())
+			},
+			exprTests: []struct {
+				expr  string
+				parse IssueAssertionFunc
+				check IssueAssertionFunc
+			}{
+				{"instance.trueProp", NoIssues, NoIssues},
+			},
+		},
+		{
 			file: "bindings/go/cel/jsonschema/santhosh-tekuri/v6/testdata/Formats.schema.json",
 			verify: func(t *testing.T, decl *jsonschema.DeclType) {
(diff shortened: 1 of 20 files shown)


Skarlso:
@frewilhelm That's not in the controller module. Meaning if I run the task without the root module it doesn't work.

Skarlso:
That said, I can't figure out a way for this since Taskfiles doesn't allow nested dotenv. :(

frewilhelm:
> @frewilhelm That's not in the controller module. Meaning if I run the task without the root module it doesn't work.

Maybe we need to move it to the controller then. The envtest version is defined in there as well

jakobmoellerdev:
instead of 

```
ENVTEST_K8S_VERSION:
    sh: 'source ../../.env && echo "$ENVTEST_K8S_VERSION"'
```

I think you can use `{{ .ROOT_DIR }}` so that it will always pick the correct project root on the source

Skarlso:
@jakobmoellerdev I'm sorry, I'm not sure what you mean. :D I'm having a brainfart. Do you mean on the values, something like this:
```taskfile
sh: 'test -n "$KIND_NODE_IMAGE_VERSION" && echo "$KIND_NODE_IMAGE_VERSION" || (source ../../.env && echo "$KIND_NODE_IMAGE_VERSION")'
```

Or do you mean on the ROOT_DIR? And how would ROOT_DIR not exist? Wouldn't it be just `.` if you launch it from the current folder or something? ROOT_DIR is always the location you are running the task from, isn't it?

Skarlso:
@jakobmoellerdev okay, I _think_ this is now okay...?

kubernetes/controller/.env
```diff
@@ -0,0 +1,16 @@
+# Controller-specific overrides
+# These values override the project root .env if needed
+# Leave empty to use project defaults managed by renovate
+
+# renovate: datasource=go depName=sigs.k8s.io/controller-tools
+CONTROLLER_TOOLS_VERSION=v0.19.0
```
jakobmoellerdev: this env file will not be updated by renovate if you dont fix the config!

Skarlso: Ah yes. Because previously this was in the root. 🤦 My bad, I'll fix the config.

Skarlso: @jakobmoellerdev I'm actually fairly certain that this:
```
      managerFilePatterns: [
        '/(^|/)\\.env(\\.local)?$/',
      ],
```

already matches subfolders... 

Like if I check this regex is matches the following:

```
.env
kubernetes/controller/.env
kubernetes/controller/test/e2e/hacks/.env
```

Are you referring to some other config?

PR #1530, opened 2026-01-02
feat: add support for inline struct flattening in JSON schema generation

jakobmoellerdev:

Introduced logic to handle `json:",inline"` tags, allowing nested struct fields to be flattened into parent schemas. Updated schema generation, required field determination, and added test cases to validate behavior. Included schema serialization for `InlineType`.

makes sure that we can correctly embed inline type information (needed if we encode this, for example v2 descriptor resources have ElementMeta embedded)


bindings/go/generator/jsonschemagen/schema.go
```diff
@@ -260,13 +271,21 @@ func (g *generation) buildStructProperties(st *ast.StructType, ti *universe.Type
 	return props
 }
 
-func (g *generation) buildStructRequired(st *ast.StructType) []string {
+func (g *generation) buildStructRequired(st *ast.StructType, ti *universe.TypeInfo) []string {
 	var req []string
-	for _, f := range st.Fields.List {
-		if len(f.Names) == 0 {
+	for _, field := range st.Fields.List {
+		name, opts := parseJSONTagWithFieldNameFallback(field)
```
Skarlso: Should we care about spotting cyclic inclusions? Aka, if type A inlines type B and type B inlines type A. In that case, this will run into an infinite loop. Maybe just add a visited map?

PR #1557, opened 2026-01-07
fix!: v2 schema parsing for OCI reference library

jakobmoellerdev:

this allows `version` to be nullable, as there are descriptors out there that have this value set to null

also previously we used upstream kubernetes k8s UnmarshalStrict for v2 descriptor decoding in the OCI lib. this is an issue because we should have relied on the OCM descriptor schema instead. Now, the function to actually decode the descriptor schema is exchangeable and we use the v2 validation by default

also adds the missing nested digests field that we never really cared about but still want to expose in the v2 lib


matthiasbruns (approved):

Skarlso:
@jakobmoellerdev Because of the schema and the version field type change, this is a breaking change, right? I think it should be marked as such.

PR #1561, opened 2026-01-08
chore(doc): add integrity-verification to the contract

frewilhelm:
We noticed in https://github.com/open-component-model/ocm-project/issues/777 that the resource-download should imply an integrity-verification of the downloaded blob. Thus, we add this to the contract.

In the current implementations of `DownloadResource` this functionality is already present. So, no need to change anything there.


bindings/go/plugin/manager/registries/resource/contract.go
```diff
@@ -13,7 +13,7 @@ import (
 type Repository interface {
 	// GetResourceCredentialConsumerIdentity resolves the identity of the given [descriptor.Resource] to use for credential resolution.
 	GetResourceCredentialConsumerIdentity(ctx context.Context, resource *descriptor.Resource) (runtime.Identity, error)
-	// DownloadResource downloads a [descriptor.Resource] from the repository.
+	// DownloadResource downloads and verifies the integrity of a [descriptor.Resource] from the repository.
```
jakobmoellerdev: please also add this to the local resource contract

jakobmoellerdev (approved):

PR #1563, opened 2026-01-09
feat: migrate the deployment controller to v2

Skarlso:

Fixes https://github.com/open-component-model/ocm-project/issues/777


jakobmoellerdev:
@Skarlso maybe its wise to merge this PR as is and then follow up with another one. WDYT?

Skarlso:
@jakobmoellerdev You might be right. I'm over a 1000 changes right now locally. I think I can do that separately. The unused code isn't hurting anything.

PR #1635, opened 2026-01-21
feat: add component reference resolvers to the controller

Skarlso:

Involved in https://github.com/open-component-model/ocm-project/issues/777

### DIFF
--- a/kubernetes/controller/internal/setup/integration_test.go
+++ b/kubernetes/controller/internal/setup/integration_test.go
@@ -416,6 +416,9 @@ func TestIntegration_ResolverProvider(t *testing.T) {
 		require.NoError(t, err)
 		require.NotNil(t, repo)
 
-		// TODO: Extend this if possible with actually trying to get the component.
+		cv, err := repo.GetComponentVersion(ctx, "github.com/test/myrepo", "v1.0.0")
+		require.NoError(t, err)
+		require.NotNil(t, cv)
+		assert.Equal(t, "github.com/test/myrepo", cv.Component.Name)
 	})
 }


kubernetes/controller/internal/setup/resolver.go
```diff
@@ -11,9 +11,13 @@ import (
...
 // GetResolvers extracts resolver configuration from the generic config.
 // Resolvers specify fallback repositories for component references.
-// Note: This is the legacy resolver code for now.
 //
-//nolint:staticcheck // no replacement for resolvers available yet https://github.com/open-component-model/ocm-project/issues/575
+// Deprecated: This function extracts deprecated fallback resolvers. Use GetResolversV1Alpha1
+// to get the modern path matcher resolvers instead. Fallback resolvers are kept for backward
+// compatibility but will be removed in a future version.
+//
+// TODO: Question. This is now unused because I don't support the old repo matching, should we or
```
Skarlso: Going to have to do this, because we expect people to use old configurations.

kubernetes/controller/internal/resolution/cached_repository.go
```diff
@@ -81,39 +100,66 @@ func (c *CacheBackedRepository) GetComponentVersion(ctx context.Context, compone
...
+// CheckHealth calls health check on the underlying repository.
+// Since we use a provider, we get the repository for a placeholder component to check health.
+// Returns nil if the repository does not support health checking.
 func (c *CacheBackedRepository) CheckHealth(ctx context.Context) error {
-	checkable, ok := c.repo.(repository.HealthCheckable)
-	if !ok {
-		c.logger.V(1).Info("repository is not health-checkable", "repository", c.spec)
+	// For health check, we use a placeholder - the actual repo returned should support health check
+	// if any of the underlying repos do. This is a limitation but acceptable for health checks.
+	repo, err := c.provider.GetComponentVersionRepositoryForComponent(ctx, "health-check", "v0.0.0")
```
Skarlso: This is definitely a problem humm, hum...

Skarlso: We could either not care if it's not found and use the original repo provided for health checking, or, I could do all of this in the workers instead but then technically, the health check wouldn't check the right repository that the resolvers are creating.

piotrjanik: Does this mean that each repo will need to have a "health-check" cv stored?

kubernetes/controller/internal/resolution/service.go
```diff
@@ -75,14 +76,23 @@ func (r *Resolver) NewCacheBackedRepository(ctx context.Context, opts *Repositor
...
+	return newCacheBackedRepository(r.logger, provider, cfg, r.workerPool, requesterFunc), nil
 }
 
-func (r *Resolver) createRepository(ctx context.Context, spec runtime.Typed, cfg *configuration.Configuration) (repository.ComponentVersionRepository, error) {
-	options := setup.RepositoryOptions{
+// createProvider creates a ComponentVersionRepositoryForComponentProvider based on the configuration.
+// The provider handles resolving the appropriate repository for each component.
+//
+//nolint:staticcheck // compatibility mode for deprecated resolvers
+func (r *Resolver) createProvider(ctx context.Context, spec runtime.Typed, cfg *configuration.Configuration) (setup.ComponentVersionRepositoryForComponentProvider, error) {
```
fabianburth: we do have such a function in the cli already, right? could we unify those codepaths?

Skarlso: For example, I don't have any of the switch statements. I don't have compref. I do some other return statements, and we use a different logger ofc. I guess I can create something that can be used by both and then altered post creation if needed. 🤔 Something like `bindings/go/configuration/resolvers`. That would, of course, mean a larger and longer taking approach, of creating a different package, releasing, using, yada yada.

Skarlso:
I refactored a bunch of things, aaand
```
  status:
    component:
      component: github.com/gardener/etcd-druid
      repositorySpec:
        baseUrl: ghcr.io
        subPath: platform-mesh
        type: OCIRegistry
      version: v0.34.0
    conditions:
    - lastTransitionTime: "2026-01-22T17:13:15Z"
      message: Applied version v0.34.0
      observedGeneration: 1
      reason: Succeeded
      status: "True"
      type: Ready
```

Still works 🎉

cli/internal/credentials/config.go
```diff
@@ -1,42 +1,11 @@
...
 )
 
-var scheme = runtime.NewScheme()
-
-func init() {
-	credentialsv1.MustRegister(scheme)
-}
-
 // LookupCredentialConfiguration creates a new ConfigCredentialProvider from a central V1 config.
 func LookupCredentialConfiguration(cfg *genericv1.Config) (*credentialsRuntime.Config, error) {
```
Skarlso: Technically, this is no longer needed... should I _completely_ remove it?

kubernetes/controller/internal/resolution/cached_repository.go
```diff
@@ -60,15 +72,19 @@ func (c *CacheBackedRepository) GetComponentVersion(ctx context.Context, compone
 		configHash = c.cfg.Hash
 	}
 
-	// create our key function that the cache will use to determine the key for this request
 	keyFunc := func() (string, error) {
-		return buildCacheKey(configHash, c.spec, component, version)
+		return buildCacheKey(configHash, c.baseRepoSpec, component, version)
```
Skarlso: So this will only ever cache with the base repository, are we sure this is okay now?

Because I had the idea that the caching should be based on the recent spec here. But because that is in the config in any case, this spec here might be redundant. But that said, this might be enough here. Just thinking about this out loud.

component-a resolves to repo-a and component-b resolves the repo-b it will be cached with repo-a regardless.

kubernetes/controller/internal/resolution/cached_repository.go
```diff
@@ -81,39 +97,63 @@ func (c *CacheBackedRepository) GetComponentVersion(ctx context.Context, compone
...
 
-// CheckHealth calls health check on the underlying repository. Returns an error if the repository does not support
-// health checking.
+// CheckHealth calls health check on the underlying base repository.
+// Returns nil if the repository does not support health checking.
 func (c *CacheBackedRepository) CheckHealth(ctx context.Context) error {
-	checkable, ok := c.repo.(repository.HealthCheckable)
-	if !ok {
-		c.logger.V(1).Info("repository is not health-checkable", "repository", c.spec)
+	repo, err := c.provider.GetComponentVersionRepositoryForSpecification(ctx, c.baseRepoSpec)
```
Skarlso: I think baseRepoSpec here can be nil, right? Just saying thta `GetComponentVersionRepositoryForSpecification` needs to prep for that.

bindings/go/repository/component/providers/pathmatcher.go
```diff
@@ -17,9 +20,13 @@ type pathMatcherProvider struct {
...
+
+	lock       sync.RWMutex
+	repoCache  map[string]repository.ComponentVersionRepository
+	validSpecs map[string]struct{}
 }
 
-var _ SpecResolvingProvider = (*pathMatcherProvider)(nil)
+var _ ComponentVersionRepositoryForComponentProvider = (*pathMatcherProvider)(nil)
 
 func (p *pathMatcherProvider) GetRepositorySpecForComponent(ctx context.Context, component, version string) (runtime.Typed, error) {
```
Skarlso: I _think_ you can delete this now?

bindings/go/repository/component/providers/pathmatcher.go
```diff
@@ -28,32 +35,86 @@ func (p *pathMatcherProvider) GetRepositorySpecForComponent(ctx context.Context,
...
 
-	repo, err := p.repoProvider.GetComponentVersionRepository(ctx, repoSpec, credMap)
+	repo, err := p.repoProvider.GetComponentVersionRepository(ctx, specification, credMap)
 	if err != nil {
-		return nil, fmt.Errorf("getting component version repository for %q failed: %w", repoSpec, err)
+		return nil, fmt.Errorf("getting component version repository for %q failed: %w", specification, err)
 	}
 
+	p.lock.Lock()
+	p.repoCache[cacheKey] = repo
```
Skarlso: How do we evict these? This will just grow, right...? The problem is that with 1000 objects, and a lot of repositories and vairous ocm configs this will get very full very fast unless I'm missing a delete somewhere?

Skarlso (commented):
I have a few questions about the new approach. :)

bindings/go/repository/component/providers/pathmatcher.go
```diff
@@ -28,32 +35,86 @@ func (p *pathMatcherProvider) GetRepositorySpecForComponent(ctx context.Context,
 	})
 }
 
-func (p *pathMatcherProvider) GetComponentVersionRepositoryForComponent(ctx context.Context, component, version string) (repository.ComponentVersionRepository, error) {
-	repoSpec, err := p.GetRepositorySpecForComponent(ctx, component, version)
+// getRepository returns a cached repository for the given specification, or creates a new one.
+// It handles credential resolution and caching internally.
+// Uses double-checked locking to prevent duplicate repository creation on concurrent cache misses.
+func (p *pathMatcherProvider) getRepository(ctx context.Context, specification runtime.Typed) (repository.ComponentVersionRepository, error) {
```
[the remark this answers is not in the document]
Skarlso: Ah, blind again. Okay.

kubernetes/controller/internal/resolution/cached_repository.go
fabianburth: Yeah, I get your point. But you outlined my thought process exactly. The actual repository is determined by the provider. The provider is with the `config` (containing the relevant set of resolvers) + `base repository`.
So I figured this is a suitable cache key here.

fabianburth: With your previous implementation of the fallback repository, you also always did exactly what I'm doing here now (as your repository for spec on the fallback provider always returned the base repo).

kubernetes/controller/internal/resolution/cached_repository.go
Skarlso: That works too.

PR #1682, opened 2026-01-29
feat: refactor providers

fabianburth:
This PR refactors providers to:
- Allow accessing their underlying repositories by specification. We believe this is generally useful, but particularly required to implement the health check in the controllers.
- Implement caching of the underlying repositories.

Contributes to https://github.com/open-component-model/ocm-project/issues/777


bindings/go/repository/component/fallback/v1/repository.go
```diff
@@ -13,6 +14,7 @@ import (
 	"strings"
 	"sync"
 
+	"github.com/cyberphone/json-canonicalization/go/src/webpki.org/jsoncanonicalizer"
```
Skarlso: Geez that module name. :D

Skarlso (changes requested):

PR #1684, opened 2026-01-30
feat: oci artifact cli

matthiasbruns:
This PR ties all the changes related to https://github.com/open-component-model/ocm-project/issues/796 together in the cli package and implement oci transfer with tests.
It also adds two new flags, which were taken from old ocm https://ocm.software/docs/reference/ocm-cli/transfer/componentversions/

- `FlagCopyResources      = "copy-resources"`

in addition, I've added an integration test, that:
- creates two oci registries
- uploads one artifact to the first registry
- creates a cv pointing to the artifact
- calls:

```go
transferCMD.SetArgs([]string{
		"transfer",
		"component-version",
		sourceRef,
		targetRef,
		"--config", cfgPath,
		"--copy-resources", // required, otherwise we wouldn't transfer oci artifacts
	})
```
- verifies, that the oci artifact is in the 2nd registry

Split PRs:
- https://github.com/open-component-model/open-component-model/pull/1730
- https://github.com/open-component-model/open-component-model/pull/1751
- https://github.com/open-component-model/open-component-model/pull/1736
- https://github.com/open-component-model/open-component-model/pull/1727
- https://github.com/open-component-model/open-component-model/pull/1718
- https://github.com/open-component-model/open-component-model/pull/1717

### DIFF
--- a/cli/go.mod
+++ b/cli/go.mod
@@ -23,12 +23,12 @@ require (
 	ocm.software/open-component-model/bindings/go/dag v0.0.6
 	ocm.software/open-component-model/bindings/go/descriptor/normalisation v0.0.0-20260212155305-86f8891d890c
 	ocm.software/open-component-model/bindings/go/descriptor/runtime v0.0.0-20260212155305-86f8891d890c
-	ocm.software/open-component-model/bindings/go/descriptor/v2 v2.0.1-alpha9
+	ocm.software/open-component-model/bindings/go/descriptor/v2 v2.0.3-alpha1
 	ocm.software/open-component-model/bindings/go/helm v0.0.0-20260212155305-86f8891d890c
 	ocm.software/open-component-model/bindings/go/input/dir v0.0.2
 	ocm.software/open-component-model/bindings/go/input/file v0.0.3
 	ocm.software/open-component-model/bindings/go/input/utf8 v0.0.0-20260212155305-86f8891d890c
-	ocm.software/open-component-model/bindings/go/oci v0.0.29
+	ocm.software/open-component-model/bindings/go/oci v0.0.30
 	ocm.software/open-component-model/bindings/go/plugin v0.0.14
 	ocm.software/open-component-model/bindings/go/repository v0.0.7
 	ocm.software/open-component-model/bindings/go/rsa v0.0.0-20260212155305-86f8891d890c
--- a/cli/go.sum
+++ b/cli/go.sum
@@ -406,8 +406,8 @@ ocm.software/open-component-model/bindings/go/descriptor/normalisation v0.0.0-20
 ocm.software/open-component-model/bindings/go/descriptor/normalisation v0.0.0-20260212155305-86f8891d890c/go.mod h1:4dYCF1jeY06Fi/ue6iPJhw/KQsTqKaHl6I8DOPmuJpY=
 ocm.software/open-component-model/bindings/go/descriptor/runtime v0.0.0-20260212155305-86f8891d890c h1:42RSanqj626J0rxF+IjzvJTe8Eflw6VWi+J4ssJSjlk=
 ocm.software/open-component-model/bindings/go/descriptor/runtime v0.0.0-20260212155305-86f8891d890c/go.mod h1:w8DLZVNfVWjXt1Z1wBW+0sbz0r87jj8mfrT5gHco+BA=
-ocm.software/open-component-model/bindings/go/descriptor/v2 v2.0.1-alpha9 h1:SdQm6w7Lj0g+d+MaNXQQkpKUqOhFwUSzoUjzRyeKL0I=
-ocm.software/open-component-model/bindings/go/descriptor/v2 v2.0.1-alpha9/go.mod h1:5ES8RkuCdrbqiG56WJE40qhFoC4QHAQn8BasL5aKlvM=
+ocm.software/open-component-model/bindings/go/descriptor/v2 v2.0.3-alpha1 h1:9bE4ZORrXO8alloUnm4T+PgrndYAqtdnk9WhphZamJc=
+ocm.software/open-component-model/bindings/go/descriptor/v2 v2.0.3-alpha1/go.mod h1:FTKrAPkR0865aUlDBrzmmHTdFxuy7AbqQZrgG6pB9YY=
 ocm.software/open-component-model/bindings/go/helm v0.0.0-20260212155305-86f8891d890c h1:bYTozauMrfY0tKsxlSilYMMi0bHrOtlUcgFtGuzA6nQ=
 ocm.software/open-component-model/bindings/go/helm v0.0.0-20260212155305-86f8891d890c/go.mod h1:1X4nAt3Tyzv15JlDOn3KobHImQcyzBHPIvZC7o5ZMWc=
 ocm.software/open-component-model/bindings/go/input/dir v0.0.2 h1:5n9k0J0W5jDWuzf2uhuzQKnNeqMwBDHop0mAfv4naiw=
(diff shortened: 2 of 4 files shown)


bindings/go/oci/spec/transformation/v1alpha1/oci_get_oci_artifact.go
```diff
@@ -0,0 +1,46 @@
...
+	// Resource is the resource descriptor from the component
+	Resource *v2.Resource `json:"resource"`
+}
+
+// OCIGetOCIArtifactSpec is the input specification for the
+// OCIGetOCIArtifact transformation.
+// +k8s:deepcopy-gen=true
+// +ocm:jsonschema-gen=true
+type OCIGetOCIArtifactSpec struct {
+	// Resource identifies the oci artifact to retrieve.
```
fabianburth: comment seems off

bindings/go/oci/transformer/get_oci_artifact.go
```diff
@@ -0,0 +1,99 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed creating get oci artifact transformation object: %w", err)
+	}
+	if err := t.Scheme.Convert(step, transformation); err != nil {
+		return nil, fmt.Errorf("failed converting generic transformation to get oci artifact transformation: %w", err)
+	}
+
+	var resource *v2.Resource
+	var outputPath string
+	var output interface{}
```
fabianburth: I don't think your `output` has to be an interface. My understanding is that @jakobmoellerdev only had to do this for the other transformers, because he had to type switch between oci and ctf.

fabianburth (changes requested):

bindings/go/oci/transformer/get_oci_artifact.go
```diff
@@ -0,0 +1,86 @@
...
+
+	blobContent, err := t.Repository.DownloadResource(ctx, targetResource)
+	if err != nil {
+		return nil, fmt.Errorf("failed downloading OCI artifact %v %w", resource, err)
+	}
+
+	// Determine output path
+	if outputPath == "" {
+		// Create a temporary file
+		tempFile, err := os.CreateTemp("", "oci-artifact-*.tar")
```
jakobmoellerdev: you dont know if this will be a tar before checking the media type, no? also it would be a tar.gz right? (please test this)

jakobmoellerdev (commented):
generally LGTM, nice stuff. just minor questions and clarifications for testing and usage

cli/cmd/transfer/component-version/internal/discovery.go
```diff
@@ -246,6 +235,131 @@ func fillGraphDefinitionWithPrefetchedComponents(d *dag.DirectedAcyclicGraph[str
...
+	// e.g. ghcr.io/open-component-model/helmexample/charts/mariadb:12.2.7
+	// strip the domain part and keep the rest
+	var referenceName string
+	if ociAccess.ImageReference != "" {
+		parts := strings.SplitN(ociAccess.ImageReference, "/", 2)
+		if len(parts) != 2 {
+			return fmt.Errorf("invalid OCI image reference: %s", ociAccess.ImageReference)
+		}
+		referenceName = parts[1]
+	}
```
fabianburth: our image references may look like this:
`https://ghcr.io/open-component-model/helmexample/charts/mariadb:12.2.7`

Thus, I don't think this parser is sufficient.

cli/cmd/transfer/component-version/cmd.go
```diff
@@ -203,6 +235,18 @@ func graphBuilder(pm *manager.PluginManager, credentialProvider credentials.Reso
...
+	resourcePlugin, err := pm.ResourcePluginRegistry.GetResourcePlugin(ctx, &v1.OCIImage{})
+	if err != nil {
+		return nil, fmt.Errorf("unable to get OCI resource plugin: %w", err)
+	}
+
+	ociGetOCIArtifact := &transformer.GetOCIArtifact{
+		Scheme:             transformerScheme,
+		Repository:         resourcePlugin,
+		CredentialProvider: credentialProvider,
+	}
```
fabianburth: doesnt the resource plugin implement the resource repository interface now?

cli/cmd/transfer/component-version/cmd.go
```diff
@@ -120,14 +127,39 @@ func TransferComponentVersion(cmd *cobra.Command, args []string) error {
...
+
+	copyLocalResources, err := cmd.Flags().GetBool(FlagCopyLocalResources)
+	if err != nil {
+		return fmt.Errorf("getting copy-local-resources flag failed: %w", err)
+	}
+
+	// do not allow both copy-resources and copy-local-resources to be set at the same time, as they are mutually exclusive
+	if copyResources && copyLocalResources {
+		return fmt.Errorf("flags --%s and --%s cannot be set at the same time, as they are mutually exclusive", FlagCopyResources, FlagCopyLocalResources)
+	}
```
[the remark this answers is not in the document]
matthiasbruns: i will check, I mean in the end, if both are set, the broader wins if we allow both being set

cli/cmd/transfer/component-version/internal/discovery.go
```diff
@@ -246,6 +222,126 @@ func fillGraphDefinitionWithPrefetchedComponents(d *dag.DirectedAcyclicGraph[str
...
+			"resource": map[string]any{
+				// TODO(matthiasbruns): figure out how to not hate yourself doing this
+				"name":     fmt.Sprintf("${%s.output.resource.name}", getResourceID),
+				"version":  fmt.Sprintf("${%s.output.resource.version}", getResourceID),
+				"type":     fmt.Sprintf("${%s.output.resource.type}", getResourceID),
+				"relation": fmt.Sprintf("${%s.output.resource.relation}", getResourceID),
+				"access": map[string]interface{}{
+					"type":          descriptor.GetLocalBlobAccessType().String(),
+					"referenceName": referenceName,
+				},
```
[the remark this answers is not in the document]
matthiasbruns: When I add them in, the graph fails with

`
time=2026-02-12T13:21:05.227+01:00 level=ERROR msg="transformation failed" id=transformOcmSoftwareTestComponentV100AddtransformTestResourceV100 error="failed to evaluate expression \"transformOcmSoftwareTestComponentV100GettransformTestResourceV100.output.resource.labels\": no such key: labels"
time=2026-02-12T13:21:05.227+01:00 level=ERROR msg="execution failed" error="failed to process value for node transformOcmSoftwareTestComponentV100AddtransformTestResourceV100 in graph: failed to evaluate expression \"transformOcmSoftwareTestComponentV100GettransformTestResourceV100.output.resource.labels\": no such key: labels"
`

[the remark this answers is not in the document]
Skarlso: That is either bad or really weird. :D

cli/cmd/transfer/component-version/cmd.go
```diff
@@ -49,6 +50,7 @@ a target repository using an internally generated transformation graph.
 This command constructs a TransformationGraphDefinition consisting of:
   1. CTFGetComponentVersion / OCIGetComponentVersion
   2. CTFAddComponentVersion / OCIAddComponentVersion
+  3. GetOCIArtifact / OCIAddLocalResource
```
jakobmoellerdev: this is nowhere near enough explanation of the command. pretty much all behavior is missing

cli/cmd/transfer/component-version/internal/helpers.go
```diff
@@ -64,3 +55,21 @@ func ChooseAddLocalResourceType(repo runtime.Typed) runtime.Type {
...
+		return "", fmt.Errorf("cannot get reference name from empty image reference")
+	}
+	imageRef, err := looseref.ParseReference(ociAccess.ImageReference)
+	if err != nil {
+		return "", fmt.Errorf("invalid OCI image reference %q: %w", ociAccess.ImageReference, err)
+	}
+	if imageRef.Repository == "" {
+		return "", fmt.Errorf("invalid image reference %q: repository is required", ociAccess.ImageReference)
+	}
+	referenceName := imageRef.Repository
```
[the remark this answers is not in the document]
matthiasbruns: we had IsLocal(access) in helpers but there is already one in 
`ocm.software/open-component-model/cli/cmd/download/shared` and fabian chose to use this one  - I see that picking one from another command is not optimal and in this case also the wrong code - let me fix this by rolling back that change

cli/cmd/transfer/component-version/internal/discovery.go
```diff
@@ -112,56 +120,24 @@ func fillGraphDefinitionWithPrefetchedComponents(d *dag.DirectedAcyclicGraph[str
 				return fmt.Errorf("cannot convert resource access to typed object: %w", err)
 			}
 
+			if copyMode == CopyModeLocalBlobResources && !shared.IsLocal(access) {
```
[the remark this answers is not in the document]
matthiasbruns: we had IsLocal(access) in helpers but there is already one in
ocm.software/open-component-model/cli/cmd/download/shared and fabian chose to use this one - I see that picking one from another command is not optimal - let me fix this bringing back an internal helper

cli/cmd/transfer/component-version/internal/helpers.go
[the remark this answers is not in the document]
matthiasbruns: ah instead of constructing it? But in Fabians implementation, we actually set the tag in the imageReference

```go
func GetReferenceName(ociAccess ociv1.OCIImage) (string, error) {
	if ociAccess.ImageReference == "" {
		return "", fmt.Errorf("cannot get reference name from empty image reference")
	}
	imageRef, err := looseref.ParseReference(ociAccess.ImageReference)
	if err != nil {
		return "", fmt.Errorf("invalid OCI image reference %q: %w", ociAccess.ImageReference, err)
	}
	if imageRef.Repository == "" {
		return "", fmt.Errorf("invalid image reference %q: repository is required", ociAccess.ImageReference)
	}
	referenceName := imageRef.Repository
	if imageRef.Tag != "" {
		referenceName += ":" + imageRef.Tag
	}
	return referenceName, nil
}
```

should the tag in the reference be removed?

cli/cmd/transfer/component-version/cmd.go
matthiasbruns: on it

matthiasbruns: updated and added examples

cli/cmd/transfer/component-version/internal/discovery.go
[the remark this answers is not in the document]
jakobmoellerdev: actually I already mentioned to fabian I believe that we need a method for this in the descriptor package. I think the need for this is very real but instead of introducing a helper everytime it should sit next to the definition of the localBlob spec. WDYT?

PR #1728, opened 2026-02-05
chore: Enhance CI workflow to always lint all modules in case of .env changes

morri-son:
Change module discovery to emit all modules in case of .env changes to always lint all modules in such a case.


.github/workflows/ci.yml
```diff
@@ -107,9 +112,11 @@ jobs:
               });
               console.log(`Filtered modules: ${JSON.stringify(filteredModules)}`);
               core.setOutput("modules_json", JSON.stringify(filteredModules));
+              core.setOutput('lint_modules_json', JSON.stringify(lintModulesFromEnv ?? filteredModules));
             } else {
               console.log("check_only_changed is false, no filtering applied.");
               core.setOutput("modules_json", JSON.stringify(modulesJson));
+              core.setOutput('lint_modules_json', JSON.stringify(lintModulesFromEnv ?? modulesJson));
```
piotrjanik: Maybe rename it to all_modules? In case it is needed elsewhere, not only for linter?

morri-son: its not always all modules, it could also be the filtered list, depending if modules have been touched in addition to .env changes. This is why I emit lint_modules_json in the "filtered" step, because the "golangci_lint" step depends on a valid matrix for both cases, all and filtered.

PR #1730, opened 2026-02-05
feat(796): oci artifact spec

matthiasbruns:
First PR to add oci artifact transformers in the bindings lib.
Part of https://github.com/open-component-model/open-component-model/pull/1684

GetOCIArtifactSpec will download a artifact and put it into a file defined by `OutputPath` or into tmp when `OutputPath` is empty.
The next step is to integrate the spec into the cli `transfer component-version`.

```go
type GetOCIArtifactSpec struct {
	// Resource is the resource descriptor to get the OCI artifact from.
	Resource *v2.Resource `json:"resource"`
	// OutputPath is the path where the artifact should be downloaded to..
	// If empty, a temporary file will be created.
	OutputPath string `json:"outputPath,omitempty"`
}
```

### DIFF
--- a/bindings/go/constructor/go.mod
+++ b/bindings/go/constructor/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/stretchr/testify v1.11.1
 	golang.org/x/sync v0.19.0
 	ocm.software/open-component-model/bindings/go/blob v0.0.11
-	ocm.software/open-component-model/bindings/go/credentials v0.0.6
+	ocm.software/open-component-model/bindings/go/credentials v0.0.7
 	ocm.software/open-component-model/bindings/go/dag v0.0.6
 	ocm.software/open-component-model/bindings/go/descriptor/normalisation v0.0.0-20260209114331-2c034a9e2912
 	ocm.software/open-component-model/bindings/go/descriptor/runtime v0.0.0-20260209114331-2c034a9e2912
--- a/bindings/go/constructor/go.sum
+++ b/bindings/go/constructor/go.sum
@@ -51,8 +51,8 @@ ocm.software/open-component-model/bindings/go/blob v0.0.11 h1:ASalU+M+PMgLGI6vQ1
 ocm.software/open-component-model/bindings/go/blob v0.0.11/go.mod h1:pZzaYptwTVBhicYp6z9vDY2K6KTwd5c+MuYoC3jr6s8=
 ocm.software/open-component-model/bindings/go/configuration v0.0.9 h1:bfgPD2G3fgmcbRYw0WXO3Fnx7+0G4gOV54JnYXgQY7w=
 ocm.software/open-component-model/bindings/go/configuration v0.0.9/go.mod h1:VZ8jQ3a6oTWyOpY09C7V3L0CfzKe9i1/Z4uJjLAVjN4=
-ocm.software/open-component-model/bindings/go/credentials v0.0.6 h1:h+B9rpt1VgWEEymE2EjbqpZA+Lj9Tb0ueLvPnWq+u58=
-ocm.software/open-component-model/bindings/go/credentials v0.0.6/go.mod h1:nWi5y4VhkAGaxfy6wNoLCde0saCu8ViHvjqBXHbe04U=
+ocm.software/open-component-model/bindings/go/credentials v0.0.7 h1:pxf8fKSwWFiuvCjFrEfMODE4Vkyi+QexM/8x+mNdhcQ=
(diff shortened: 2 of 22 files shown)


bindings/go/oci/integration/integration_test.go
```diff
@@ -60,6 +67,18 @@ const (
...
+	Password string
+}
+
+func (m mockCredentialsResolver) Resolve(ctx context.Context, identity ocmruntime.Identity) (map[string]string, error) {
+	return map[string]string{
+		"username": m.Username,
+		"password": m.Password,
+	}, nil
+}
+
```
fabianburth: whats the value add of this mock cred resolver, are you testing against it in any way? couldnt you just pass nil?

bindings/go/oci/transformer/get_oci_artifact.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+	}
+
+	blobContent, err := t.Repository.DownloadResource(ctx, targetResource, creds)
+	if err != nil {
+		return nil, fmt.Errorf("failed downloading OCI artifact %v %w", resource.ToIdentity(), err)
+	}
+
+	// Determine output path
+	if outputPath == "" {
```
jakobmoellerdev: youre duplicating code here I believe, best to use a shared method maybe?

jakobmoellerdev (commented):
mostly lgtm, just smaller corrections and questions I believe

bindings/go/oci/integration/integration_test.go
```diff
@@ -60,6 +67,18 @@ const (
 	userAgent                 = "ocm.software"
 )
 
+type mockCredentialsResolver struct {
```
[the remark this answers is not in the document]
matthiasbruns: Maybe you understood me wrong - I do not feel that we need this helper in production code just to have a static credentials helper for testing - I will add this now, but again, I do not feel that there is something broken that needs to be fixed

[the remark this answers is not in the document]
matthiasbruns: https://github.com/open-component-model/open-component-model/pull/1736
I am not sure if I should have used the DAG for this or added it into the existing resolver flow from the dag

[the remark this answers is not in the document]
jakobmoellerdev: This resolver you wrote is totally fine! I think you wont be the last person to need static credentials without wanting a full credential graph, so having this in the binding library is useful either way.

bindings/go/oci/integration/integration_test.go
```diff
@@ -285,6 +304,14 @@ func Test_Integration_OCIRepository(t *testing.T) {
 		t.Run("oci image digest processing", func(t *testing.T) {
 			processResourceDigest(t, repo, "ghcr.io/test:v1.0.0", reference("new-test:v1.0.0"))
 		})
+
+		t.Run("get oci artifact", func(t *testing.T) {
```
jakobmoellerdev: this should not be part of the backwards compatibility tests. also it seems like you structure the tests a lot differently than the others. Why is that? I was expecting you to setup a new test set just for the transformers here

matthiasbruns: So I had a second look why I did the setup

The existing integration tests in oci relied on either `repository.ComponentVersionRepository` or `oci.Repository`

The transformer that is implemented here does not - it uses `repository.ResourceRepository`. I cannot use the  `oci.Repository`directly with the resolver setup since we did not change its api to support credential injection during up/download. This is why I had to setup the tests differently and use the `credentials.Resolver` instead the `urlresolver.Resolver` path.
I moved the transformer tests into their own parent tests case that we can add more tests there in the future.

jakobmoellerdev: Makes sense 👍 I think all of these findings are valid. Ideally you will need to extract such follow ups into discussions on the warroom or at least point them out with a godoc comment on the test (rather the second as I dont think any of this is particularly important)

matthiasbruns: docs added :)

bindings/go/oci/spec/transformation/v1alpha1/get_oci_artifact.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+// GetOCIArtifactSpec is the input specification for the
+// GetOCIArtifact transformation.
+// +k8s:deepcopy-gen=true
+// +ocm:jsonschema-gen=true
+type GetOCIArtifactSpec struct {
+	// Resource is the resource descriptor to get the OCI artifact from.
+	Resource *v2.Resource `json:"resource"`
+	// OutputPath is the path where the artifact should be downloaded to..
```
fabianburth:
```suggestion
	// OutputPath is the path where the artifact should be downloaded to.
```

Skarlso (commented):
Generally okay, just a few little questions. :)

PR #1745, opened 2026-02-08
chore: add a CLAUDE.md file to specify generally working with the open-component-model repo

Skarlso:


jakobmoellerdev:
While I appreciate the idea, this cannot be merged as is. Claude is not open source, anthropic models are not open source. I am against merging anything that is related to non-open source ai.

We should focus on more generic integrations such as https://agents.md/ and then give instructions how to use it. WDYT?

PR #1752, opened 2026-02-10
feat!: make OCM configurations propagate through resource reference

Skarlso:

Closes https://github.com/open-component-model/ocm-project/issues/877

It fixes the referenced issue by modifying `GetEffectiveConfig` to check if "parent" exists and use it's effective config.
If, however, the object defines an explicit config, we use that instead.

There is a slight optimization here: Since the effective config is calculated if there is no parent, aka the base of anything coming after that in the chain, we can skip the resolution loop and assume that the effective config is correct only Secrets and ConfigMaps.

### DIFF
--- a/bindings/go/constructor/go.mod
+++ b/bindings/go/constructor/go.mod
@@ -13,7 +13,7 @@ require (
 	ocm.software/open-component-model/bindings/go/descriptor/normalisation v0.0.0-20260210065627-2b87daf2cb8a
 	ocm.software/open-component-model/bindings/go/descriptor/runtime v0.0.0-20260210065627-2b87daf2cb8a
 	ocm.software/open-component-model/bindings/go/descriptor/v2 v2.0.1-alpha9
-	ocm.software/open-component-model/bindings/go/oci v0.0.28
+	ocm.software/open-component-model/bindings/go/oci v0.0.29
 	ocm.software/open-component-model/bindings/go/repository v0.0.7
 	ocm.software/open-component-model/bindings/go/runtime v0.0.5
 	sigs.k8s.io/yaml v1.6.0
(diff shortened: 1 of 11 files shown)


kubernetes/controller/internal/ocm/ocm.go
```diff
@@ -22,11 +22,17 @@ import (
 // Furthermore, references to other ocm objects are resolved and their effective
 // configuration (so again, config map and secret references) with policy
 // propagate are returned.
-func GetEffectiveConfig(ctx context.Context, client ctrl.Client, obj v1alpha1.ConfigRefProvider) ([]v1alpha1.OCMConfiguration, error) {
+func GetEffectiveConfig(ctx context.Context, client ctrl.Client, obj v1alpha1.ConfigRefProvider, parent v1alpha1.ConfigRefProvider) ([]v1alpha1.OCMConfiguration, error) {
 	configs := obj.GetSpecifiedOCMConfig()
 
-	if len(configs) == 0 {
-		return nil, nil
+	if len(configs) == 0 && parent != nil {
```
frewilhelm: This changes the previous behaviour. For example the `Repository` doesn't have a parent. Hence, when no config is specified, this would be `len(configs) == 0 && parent == nil` which would now return (empty) `refs` instead of `nil`.
Did you notice any problem with that?

frewilhelm:
Can you please adjust one of the [e2e-tests](https://github.com/open-component-model/open-component-model/blob/main/kubernetes/controller/test/e2e/e2e_credentials_test.go), either `basic-auth` or `docker-config-json`, to use the new propagation. Then, we have it also tested in there and maybe people use the testdata as an example as well

Skarlso:
Okay, let's see if that worked. :D

PR #1759, opened 2026-02-11
feat: add oci spec

matthiasbruns:
This PR adds AddOCI spec to the oci package.

```go
type AddOCIArtifactSpec struct {
	// OCIArtifact is the OCI artifact to be added, represented as a blob.
	OCIArtifact blob.ReadOnlyBlob `json:"ociArtifact"`
	// TargetRegistry is the target OCI repository
	TargetRegistry string `json:"targetRef"`
}
```

**Open TODOS:**

- integration tests
- maybe the spec should allow setting `referenceName` directly

Contributes: https://github.com/open-component-model/ocm-project/issues/875


matthiasbruns:
closed in favor of https://github.com/open-component-model/open-component-model/pull/1765

PR #1763, opened 2026-02-11
chore(docs): Add doc for release process

morri-son:

Add doc for release process


matthiasbruns (commented):
mostly questions, the process is clear to me, but feels like it contains a lot of manual work for the responsible person

some overall improvement ideas

- the guide reads very choppily, mostly due to the nature of the lists
- think about a way to automate cli and controller releases instead of having to release both of them side by side

PR #1765, opened 2026-02-11
feat: oci artifact uploader

fabianburth:
Enhances the transfer command with a flag to upload oci artifacts as actual oci artifacts based on their reference name. 

We added a new flafg `upload-as` for oci artifacts to control, if we use the already implemented `lobalBlob` or the new `ociArtifact` uploader. The latter is defined as default.

Integration tests have been updated/added to test both paths.

In addition, this PR splits the cluttered internal files into several smaller to give a better overview of the system and where everything resided.

contributes to https://github.com/open-component-model/ocm-project/issues/875

Split PRs:
- https://github.com/open-component-model/open-component-model/pull/1769

### DIFF
--- a/kubernetes/controller/Taskfile.yml
+++ b/kubernetes/controller/Taskfile.yml
@@ -84,7 +84,6 @@ tasks:
     cmds:
       - task: helm/install
       - PROJECT_DIR='{{.TASKFILE_DIR}}' go test '{{.TASKFILE_DIR}}/test/e2e/' -v -timeout 30m -ginkgo.v
-      - task: helm/uninstall
 
   # Build
   build:
@@ -202,7 +201,7 @@ tasks:
     desc: "Install the Helm chart"
     cmds:
       - |
-        helm install ocm-k8s-toolkit chart/ \
+        helm upgrade --install ocm-k8s-toolkit chart/ \
           -f chart/test-values.yaml \
           --namespace ocm-k8s-toolkit-system \
           --create-namespace \


cli/cmd/transfer/component-version/internal/oci.go
```diff
@@ -0,0 +1,151 @@
...
+				"name":     fmt.Sprintf("${%s.output.resource.name}", getResourceID),
+				"version":  fmt.Sprintf("${%s.output.resource.version}", getResourceID),
+				"type":     fmt.Sprintf("${%s.output.resource.type}", getResourceID),
+				"relation": fmt.Sprintf("${%s.output.resource.relation}", getResourceID),
+				"access": map[string]interface{}{
+					"type":          descriptor.GetLocalBlobAccessType().String(),
+					"referenceName": referenceName,
+					"labels":        fmt.Sprintf("${has(%s.output.resource.labels) ? %s.output.resource.labels  : []}", getResourceID, getResourceID),
+					"extraIdentity": fmt.Sprintf("${has(%s.output.resource.extraIdentity) ? %s.output.resource.extraIdentity  : {}}", getResourceID, getResourceID),
+					"srcRefs":       fmt.Sprintf("${has(%s.output.resource.srcRefs) ? %s.output.resource.srcRefs  : []}", getResourceID, getResourceID),
```
Skarlso: Oh, sorry, these are at the wrong level. :) They should be on the `resource` level, not the `access` level.

cli/cmd/transfer/component-version/internal/oci.go
```diff
@@ -0,0 +1,151 @@
...
+// ociUploadAsArtifact creates an AddOCIArtifact transformation that uploads the OCI artifact to the target repository as an OCI artifact.
+// It constructs the target image reference from the toSpec and referenceName, and uses the output of the GetOCIArtifact transformation to populate the fields of the AddOCIArtifact transformation, ensuring that the same resource is referenced and uploaded.
+func ociUploadAsArtifact(toSpec runtime.Typed, addResourceID string, getResourceID string, referenceName string) transformv1alpha1.GenericTransformation {
+	// Construct target Image Reference from toSpec and referenceName
+	var targetImageRef string
+	// Default to referenceName if we can't determine the target repo
+	targetImageRef = referenceName
+
+	if toSpec != nil {
+		raw, err := json.Marshal(toSpec)
```
Skarlso: I understand that we only set the url if the spec is right, but this could mask unintended issues here. At the very least, I think if err != nil we should log a warning? WDYT?

matthiasbruns: Yeah if `toSpec` is not marshalable to raw, I guess something is fishy.

cli/integration/transfer_oci_artifact_integration_test.go
```diff
@@ -209,6 +210,225 @@ components:
...
+	registryAddress := internal.StartDockerContainerRegistry(t, containerName, htpasswd)
+	host, port, err := net.SplitHostPort(registryAddress)
+	r.NoError(err)
+
+	targetContainerName := fmt.Sprintf("target-oci-artifact-repository-%d", time.Now().UnixNano())
+	targetRegistryAddress := internal.StartDockerContainerRegistry(t, targetContainerName, htpasswd)
+	targetHost, targetPort, err := net.SplitHostPort(targetRegistryAddress)
+	r.NoError(err)
+
+	reference := func(ref string) string {
```
piotrjanik: This can be taken out of a test - just a helper function that can be used by other tests as well

cli/integration/transfer_oci_artifact_integration_test.go
```diff
@@ -209,6 +210,225 @@ components:
...
+	// We run this parallel as it spins up a separate container
+	t.Parallel()
+
+	// 1. Setup Local Registry
+	user := "ocm"
+	password := internal.GenerateRandomPassword(t, 20)
+	htpasswd := internal.GenerateHtpasswd(t, user, password)
+
+	containerName := fmt.Sprintf("source-oci-artifact-repository-%d", time.Now().UnixNano())
+	registryAddress := internal.StartDockerContainerRegistry(t, containerName, htpasswd)
```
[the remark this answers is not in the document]
matthiasbruns: Surely it could but I wanted to test them with two registries.

cli/integration/transfer_oci_artifact_integration_test.go
matthiasbruns: Good call with the helper.
We have that on many places. I could make this shared within this package, but wouldn't couple other packages too much by sharing that func. 

About the length - most of this is setup, I will try to extract and reuse stuff. 

I think there is also a ticket by either you and me about refactoring/cleaning up tests. 
I think we should add some juice to them and include these here as well.

cli/integration/transfer_oci_artifact_integration_test.go
```diff
@@ -209,6 +210,225 @@ components:
...
+
+	// Create source CTF
+	addCMD := cmd.New()
+	addCMD.SetArgs([]string{
+		"add",
+		"component-version",
+		"--repository", fmt.Sprintf("ctf::%s", sourceCTF),
+		"--constructor", constructorPath,
+		"--config", cfgPath,
+		// "--skip-reference-digest-processing",
```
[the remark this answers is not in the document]
matthiasbruns: that's why
https://github.com/open-component-model/open-component-model/issues/1789

cli/integration/transfer_oci_artifact_integration_test.go
matthiasbruns: updated tests

cli/integration/transfer_oci_artifact_integration_test.go
```diff
@@ -209,6 +197,213 @@ components:
...
+	// Check type
+	typeVal, ok := accessMap["type"].(string)
+	if !ok {
+		// nested type object
+		typeObj, ok := accessMap["type"].(map[string]interface{})
+		r.True(ok)
+		typeVal, ok = typeObj["type"].(string)
+		if !ok {
+			// check if it is "ociBlob" or "OCIImageLayer"
+			r.Contains([]string{"ociBlob", "OCIImageLayer", "ociBlob/v1", "OCIImageLayer/v1"}, typeVal)
```
Skarlso: typeVal here is just empty, isn't it? Also, this only happening if type is empty. So what's going on here? :D

Skarlso:
Just a test fix, otherwise, I think this is becoming okay.

PR #1770, opened 2026-02-12
feat: add HTTP client configuration type

piotrjanik:
Adds a new `http.config.ocm.software/v1alpha1` configuration type that allows users to configure HTTP client settings (currently timeout) via the OCM config file.

```yaml
configurations:
  - type: http.config.ocm.software/v1alpha1
    timeout: "5m"
```

Fixes: https://github.com/open-component-model/ocm-project/issues/1731


bindings/go/configuration/http/v1alpha1/spec/config.go
```diff
@@ -0,0 +1,129 @@
...
+	ConfigType = "http.config.ocm.software"
+)
+
+// DefaultTimeout is the default HTTP client timeout used when no
+// configuration is provided.
+const DefaultTimeout = Timeout(30 * time.Second)
+
+// Timeout wraps time.Duration to support JSON/YAML marshaling
+// of human-readable duration strings (e.g. "30s", "5m", "1h").
+type Timeout time.Duration
```
matthiasbruns: I don't know if we actually need a type alias here - in the Config the field is already named `Timeout`

Skarlso: @matthiasbruns It's a copy of the Duration type in other configuration like here: cli/internal/plugin/spec/config/v2alpha1/config.go

PR #1775, opened 2026-02-13
feat(884): add selective json value marshal in Unstructured.DeepCopy

matthiasbruns:

The goal of this PR is to add support for non-json values in runtime.Unstructured.
For this I tried two approaches
- selective marshal/unmarshal
- full marshal on every call
- I kept the original implementation for reference and benchmarking

I have tested three ways of handling non-json values in Unstructured.DeepCopy with benchmarks.
We are testing two scenarios:
- Pure Json data
- Mixed data with non-json values

The original DeepCopy was not able to handle non-json values and paniced.

```
goos: darwin
goarch: arm64
pkg: ocm.software/open-component-model/bindings/go/runtime
cpu: Apple M3 Max
BenchmarkDeepCopy_PureJSON_New
BenchmarkDeepCopy_PureJSON_New-16             	  844740	      1424 ns/op
BenchmarkDeepCopy_PureJSON_Original
BenchmarkDeepCopy_PureJSON_Original-16        	 1753347	       682.3 ns/op
BenchmarkDeepCopy_PureJSON_FullMarshal
BenchmarkDeepCopy_PureJSON_FullMarshal-16     	  212961	      5665 ns/op
BenchmarkDeepCopy_MixedData_New
BenchmarkDeepCopy_MixedData_New-16            	  382938	      3202 ns/op
BenchmarkDeepCopy_MixedData_Original
    unstructured_test.go:289: DeepCopy panics on mixed data, skipping benchmark. This is expected behavior since the original DeepCopy does not normalize non-JSON-native types.
--- SKIP: BenchmarkDeepCopy_MixedData_Original
BenchmarkDeepCopy_MixedData_FullMarshal
BenchmarkDeepCopy_MixedData_FullMarshal-16    	  273564	      4276 ns/op
PASS
```

This results in the following performance comparison:

 | Scenario | Implementation | ns/op | vs baseline |
  |---|---|---:|---|
  | **Pure JSON** | Original (direct DeepCopyJSON) | 682 | baseline |                                                                                                                                      
  | | New (selective normalize) | 1,424 | ~2x slower |
  | | Full Marshal (marshal every value) | 5,665 | ~8x slower |                                                                                                                                            
  | **Mixed data** | Original (direct DeepCopyJSON) | **panic** | cannot handle structs |                                                                                                                
  | | New (selective normalize) | 3,202 | baseline |
  | | Full Marshal (marshal every value) | 4,276 | ~1.3x slower |

Fixes: https://github.com/open-component-model/ocm-project/issues/884


jakobmoellerdev (commented):
I have a few concerns with this change — though I'm open to hearing if I'm missing something.

**Performance regression in the common case:** If I'm reading your benchmark table correctly, this introduces a ~5x runtime cost on one of our most frequently used methods for working with unstructured data — even in the standard scenario (no non-JSON values). That's a significant regression for the base case.

**Unclear motivation:** The stated goal is to add support for non-JSON values in `runtime.Unstructured`, but I'm not sure where we actually need this. `u.Data` is `map[string]any`, and working with unstructured objects only supports primitive types anyway. What's the concrete use case?

If the goal is to support arbitrary JSON-marshalled values, wouldn't it make more sense to add support specifically for `json.RawMessage` and provide a method for nested field access on that? That would be more targeted and avoid the cost for the common path.

bindings/go/runtime/unstructured.go
```diff
@@ -58,3 +69,45 @@ func (u *Unstructured) DeepCopy() *Unstructured {
...
+	case []interface{}:
+		if v == nil {
+			return v
+		}
+		out := make([]interface{}, len(v))
+		for i, elem := range v {
+			out[i] = normalizeJSONValue(elem)
+		}
+		return out
+	case string, int64, bool, float64, nil, json.Number:
```
Skarlso: Worth noting that the original only handles int64 because it expects JSON. Here, you could technically provide further help by adding `int` into the mix and avoiding one more roundtrip to normalize it to float64.

bindings/go/runtime/unstructured_test.go
```diff
@@ -92,3 +138,162 @@ func TestUnstructured(t *testing.T) {
...
+
+func BenchmarkDeepCopy_MixedData_Original(b *testing.B) {
+	defer func() {
+		if r := recover(); r != nil {
+			b.Skipf("Expected panic during DeepCopy of mixed data without normalization: %v", r)
+			// Test passes if we get here, as we expect a panic due to non-JSON-native types.
+			return
+		}
+	}()
+	b.ResetTimer()
```
Skarlso: b.Loop already does resetTimer.

bindings/go/runtime/unstructured.go
```diff
@@ -58,3 +69,45 @@ func (u *Unstructured) DeepCopy() *Unstructured {
 	out.Data = DeepCopyJSON(u.Data)
 	return out
 }
+
+// normalizeJSONValue recursively checks whether a value is a JSON-native type
+// (as handled by DeepCopyJSONValue). If it is, the value is returned as-is.
+// If not, it is normalized via JSON marshal/unmarshal.
+func normalizeJSONValue(x interface{}) interface{} {
```
Skarlso: This is recursing way too much. You can actually do this a lot faster I believe by doing a two pass instead.

What do I mean? Basically, if the value is already json, you do normalisation on it anyways and allocate a bunch of maps and slices which is super redundant. Instead, check if the value is normalised already, and only loop if not.

For example a simple check would be to do your loop but return on any error:

```go
  func isJSONNative(x any) bool {
        switch v := x.(type) {
        case string, int64, bool, float64, nil, json.Number:
                return true
        case map[string]any:
                for _, val := range v {
                        if !isJSONNative(val) {
                                return false
                        }
                }
                return true
        case []any:
                for _, elem := range v {
                        if !isJSONNative(elem) {
                                return false
                        }
                }
                return true
        default:
                return false
        }
  }
```

Do this before as a first step:

```go
  func normalizeJSONValue(x interface{}) interface{} {
        if isJSONNative(x) {
                return x
        }

        .... // rest
```

And re-run your benchmarks. Worst case, it loops everything twice, but that's still the same O operation. Best case, it will skip a BUNCH of recursions and allocations.

matthiasbruns: my are my hero - I will tackle all your reviews after finishing oci add

bindings/go/runtime/unstructured.go
```diff
@@ -58,3 +69,45 @@ func (u *Unstructured) DeepCopy() *Unstructured {
...
+	switch v := x.(type) {
+	case map[string]interface{}:
```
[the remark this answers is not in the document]
Skarlso: Aye this isn't done yet. :D

PR #1797, opened 2026-02-18
fix: use the right k8sClient in the repository suite

Skarlso:

Looks like the repository suite was using the wrong client. This _might_ not be the entire fix, but certainly is the odd one out.

Fixes https://github.com/open-component-model/open-component-model/actions/runs/22130037921/job/63968085140.


Skarlso:
Basically, this was always broken. :D But we saw it now. :D

PR #1801, opened 2026-02-18
fix: 896 remove file ext handling in oci transfer

matthiasbruns:
This PR removed the file ext handling in oci transfer.
It added unnecessary complexity and noise log messages.
Since we only store the files in temp and use them for transfer, we do not need to file ext anyway.

Fixes: https://github.com/open-component-model/ocm-project/issues/896

### DIFF
--- a/.github/config/wordlist.txt
+++ b/.github/config/wordlist.txt
@@ -594,3 +594,4 @@ warroom
 wgs
 kubermatic
 baseUrl
+ociimage
--- a/cli/cmd/add/component-version/cmd.go
+++ b/cli/cmd/add/component-version/cmd.go
@@ -129,6 +129,27 @@ If you provide a working directory, all paths in the %[1]q file will be resolved
 Otherwise the path to the %[1]q file will be used as the working directory.
 You are only allowed to reference files within the working directory or sub-directories of the working directory.
 
+Environment Variable Substitution:
+
+The %[1]q file supports environment variable substitution using Go template syntax.
+Variables can be referenced using ${VAR_NAME} or $VAR_NAME format.
+All environment variables are expanded before the file is processed, allowing for dynamic
+configuration of component versions, resource paths, image references, and other values.
+
+Example:
+  components:
+    - name: ${COMPONENT_NAME}
+      version: ${COMPONENT_VERSION}
+      provider:
+        name: ${PROVIDER_NAME}
+      resources:
+        - name: my-image
+          type: ociImage
+          version: ${COMPONENT_VERSION}
+          access:
+            type: ociArtifact
+            imageReference: ${REGISTRY_URL}/my-app:${IMAGE_TAG}
+
 Repository Reference Format:
 	[type::]{repository}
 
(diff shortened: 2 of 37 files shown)


bindings/go/oci/transformer/output.go
```diff
@@ -2,35 +2,17 @@ package transformer
...
-		}
-
-		if fileExt == "" {
-			slog.Warn("unable to determine file extension from media type, setting .bin extension")
-			fileExt = ".bin"
-		}
-
 		// Create a temporary file
-		tempFile, err := os.CreateTemp("", fmt.Sprintf("%s-*%s", filePrefix, fileExt))
+		tempFile, err := os.CreateTemp("", filePrefix+"-*")
```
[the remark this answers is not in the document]
matthiasbruns: huh? I only removed the MediaType detection since this was used for the file ext in the first place, the rest was not touched.

Skarlso (dismissed):

bindings/go/oci/transformer/output.go
```diff
@@ -2,35 +2,17 @@ package transformer
...
 	"path/filepath"
-
-	"ocm.software/open-component-model/bindings/go/blob"
-	"ocm.software/open-component-model/bindings/go/oci/spec/layout"
 )
 
 // DetermineOutputPath determines the output path for buffering the blob content.
-// If the outputPath is empty, it creates a temporary file with an appropriate extension based on the media type of the blob content.
+// If the outputPath is empty, it creates a temporary file.
 // If the outputPath is provided, it ensures that the directory exists.
```
fabianburth: Why do we need a temp file in case the output path does not exists, but a directory is sufficient in case it does? Feels like an inconsistent contract.

bindings/go/oci/transformer/output.go
```diff
@@ -39,9 +22,23 @@ func DetermineOutputPath(outputPath string, filePrefix string, blobContent blob.
...
 		dir := filepath.Dir(outputPath)
+
+		// create the directory if it does not exist
 		if err := os.MkdirAll(dir, 0o755); err != nil {
 			return "", fmt.Errorf("failed creating output directory: %w", err)
 		}
+
+		// check if outputPath contains a file name, if not create a file based on the prefix
+		if info, err := os.Stat(outputPath); err != nil {
+			return "", fmt.Errorf("failed accessing output path: %w", err)
```
fabianburth: wont this return an error if we would have
`/my/output/path.txt` where the `path.txt` does not exist yet?

fabianburth:
So, thought about this again. From the standpoint of how we want to use this with the cli, we want to pass a directory for temporary files, so I think, the correct cases would be:
```
// DetermineOutputPath determines the output path for buffering the blob content.
// If the outputPath is empty, WE CREATE a temporary file (based on the temp dir).
// If the outputPath is provided, it should check if the path exists
//   If the outputPath does not exists, WE FAIL
//   If the outputPath does exist and is a file, WE FAIL
//   If the outputPath does exist and is a directory, WE CREATE a temporary file in that directory with the filePrefix as a prefix.
```

Any concerns?

Skarlso (approved):

PR #1804, opened 2026-02-19
feat: add HTTP client configuration with timeout and user-agent

piotrjanik:

Adds configurable HTTP client timeout support across the CLI, bindings, and controller. Users in regions with slow or unstable networks can now customize HTTP timeouts to avoid premature failures.

**New configuration variables (http.config.ocm.software/v1alpha1):**
- timeout — overall HTTP client timeout (default: disabled)
- tcpDialTimeout — TCP connection establishment timeout (default: 30s)
- tcpKeepAlive — TCP keep-alive probe interval (default: 30s)
- tlsHandshakeTimeout — TLS handshake timeout (default: 10s)
- responseHeaderTimeout — time to wait for response headers (default: 10s)
- idleConnTimeout — idle connection timeout (default: 90s)

Defaults are applied when a value is not provided via config file or CLI flag. Setting a value to 0 explicitly disables that timeout.

**CLI changes:**
- New persistent flags that override config file values:
  - --timeout (default: 0, disabled)
  - --tcp-dial-timeout (default: 30s)
  - --tcp-keep-alive (default: 30s)
  - --tls-handshake-timeout (default: 10s)
  - --response-header-timeout (default: 10s)
  - --idle-conn-timeout (default: 90s)

**Controller changes:**
- New flags with the same names and defaults as above.

**OCI Bindings changes:**
- New provider.NewHTTPClient() factory with functional options (WithHttpConfig, WithHTTPUserAgent) to create HTTP clients with configurable timeouts and User-Agent header.
- NewComponentVersionRepositoryProvider accepts a WithHTTPClient option to inject a pre-configured HTTP client.
- NewResourceRepository now accepts an httpv1alpha1.Config to internally create an HTTP client with the provided config.

**Testing:**
- Unit tests for config parsing, lookup, merge, marshal round-trip, HTTP client creation, and context isolation.
- Integration tests using toxiproxy to verify timeout behavior end-to-end: configured timeout shorter/longer than latency, --timeout flag override, and --timeout 0 to disable timeout.

Fixes: https://github.com/open-component-model/open-component-model/issues/1731

Signed-off-by: Piotr Janik <piotr.janik@sap.com>


cli/internal/plugin/builtin/oci/register.go
```diff
@@ -16,7 +17,10 @@ import (
 	"ocm.software/open-component-model/bindings/go/plugin/manager/registries/resource"
 )
 
-const creator = "Builtin OCI Repository Plugin"
+const (
+	creator   = "Builtin OCI Repository Plugin"
+	userAgent = "ocm-cli/2"
```
matthiasbruns: do we want to set the actual version instead of a fixed `/2`?

kubernetes/controller/cmd/main.go
```diff
@@ -170,7 +173,11 @@ func main() {
 	pm := manager.NewPluginManager(ctx)
 
 	ocirepository.MustAddLegacyToScheme(ocirepository.Scheme)
-	repositoryProvider := provider.NewComponentVersionRepositoryProvider(provider.WithScheme(ocirepository.Scheme))
+	httpClient := provider.NewHTTPClient(provider.WithUserAgent(userAgent))
```
Skarlso: Hmmmmm since the controller itself relies on user provided ocm config this is problematic here. The client should be dynamically created, with the ocm config provided by the user's object that includes the OCM config. Though I think that would be rather problematic to set up. 🤔

cli/internal/plugin/builtin/oci/register.go
piotrjanik: I removed that

matthiasbruns:
lgtm, just the broken and failing ci is blocking this

PR #1805, opened 2026-02-19
fix(rsa): handle errors when loading RSA private and public keys from credentials

jakobmoellerdev:
### What this does / why we need it

- Enhanced `PrivateKeyFromCredentials` and `PublicKeyFromCredentials` to return detailed errors when key loading fails, improving observability and debugging.
- Updated calling functions to handle potential errors when key retrieval is unsuccessful, ensuring fault tolerance.
- Fixed cases where `nil` values for keys could lead to runtime issues by introducing appropriate error handling.

This fixes cases where improper error handling for key loading could result in misleading behavior that wasnt surfaced.

Part of my scenario testing in https://github.com/open-component-model/ocm-project/issues/842

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -47,6 +47,19 @@
       ],
     },
     {
+      matchDatasources: ['docker'],
+      matchPackageNames: ['golang'],
+      matchFileNames: ['kubernetes/controller/Dockerfile'],
+      groupName: 'Golang versions',
+      groupSlug: 'golang-versions',
+    },
+    {
+      matchManagers: ['github-actions'],
+      matchDepNames: ['go'],
+      groupName: 'Golang versions',
+      groupSlug: 'golang-versions',
+    },
+    {
       matchManagers: [
         'gomod',
         'custom.regex',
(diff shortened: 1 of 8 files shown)


frewilhelm (commented):
Just one question. The rest lgtm

bindings/go/rsa/signing/handler/internal/credentials/credentials.go
```diff
@@ -21,29 +22,38 @@ const (
...
-	if err != nil || len(b) == 0 {
-		return nil
+	if err != nil {
+		return nil, fmt.Errorf("failed loading private key PEM: %w", err)
+	}
+	if len(b) == 0 {
+		return nil, nil
 	}
-	return rsapem.ParseRSAPrivateKeyPEM(b)
+	return rsapem.ParseRSAPrivateKeyPEM(b), nil
```
piotrjanik: Both `ParseRSAPrivateKeyPEM` and `ParseRSAPublicKeyPEM` return `nil` if no key is found, so I wonder what would happen if someone provided a malformed PEM.

Skarlso (approved):

bindings/go/rsa/signing/handler/internal/credentials/credentials.go
frewilhelm: ah, sorry. I thought you were asking about parsing malformed key. Listen to @Skarlso :D

PR #1806, opened 2026-02-19
fix: oci upload behavior

fabianburth:
The behaviour for
localblobs:
- `oci-upload` set AND target ocm repo is oci AND localblob is oci artifact (media type) --> upload as oci
- otherwise -> upload as localblob

ociArtifacts:
- `oci-upload` set AND target ocm repo is oci --> upload as oci
- otherwise -> upload as localblob

### DIFF
--- a/cli/cmd/cmd_test.go
+++ b/cli/cmd/cmd_test.go
@@ -18,7 +18,6 @@ import (
 	"time"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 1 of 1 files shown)


matthiasbruns:
I feel that we start getting into an if-else condition hell and it's worth exploring https://github.com/open-component-model/ocm-project/issues/903 in the next sprints.

PR #1807, opened 2026-02-19
fix: various cli fixes on transfer and construction behavior

jakobmoellerdev:

- It fixes a resource leak by ensuring the stream used to read a component constructor is always closed, and it adds environment-variable expansion so constructor content can be parameterized at runtime.
- It reduces normal log noise by downgrading “transformation started/completed” messages from info to debug in the progress output.
- It makes credential lookup more tolerant by treating “credentials not found” as a non-fatal case when resolving optional credentials during resource download.
- It adjusts a resource download success message to be logged via `defer`, preventing “success” from being printed before later work can still fail.
- It improves OCI/CTF transformation handling by instantiating repository objects based on the destination type and selecting the correct upload transformation (with clearer errors for unsupported types).

part of my sovereign scenario testing (https://github.com/open-component-model/ocm-project/issues/842)


frewilhelm (approved):
lgtm

matthiasbruns (approved):

PR #1829, opened 2026-02-24
chore: update pr template to include testing instructions

matthiasbruns:
During the retrospective we figured out that we should test our changes manually before pushing the PR.
I have updated the PR template to cover that requirement.
I also added a "how to test" section for reviewers to get quickly onboarded on how the new change should be tested (if applicable). This helps the reviewers to understand what has changes and how the new change should be used.

While the example is targeted to CLI, this should be applicable for all kind of changes.

### DIFF
--- a/.github/workflows/trigger-blackduck-scan.yaml
+++ b/.github/workflows/trigger-blackduck-scan.yaml
@@ -15,7 +15,7 @@ on:
 
 jobs:
   trigger-scan:
-    uses: open-component-model/.github/.github/workflows/blackduck-scan.yaml@d51313fe4d02a2ae4617dd5f4777c71b0c7e97f6
+    uses: open-component-model/.github/.github/workflows/blackduck-scan.yaml@e2feabd944a76de399805ecc5414ad22e449ee56
     with:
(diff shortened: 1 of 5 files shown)


frewilhelm (commented):
I think a whitespace is required to create those checkboxes

frewilhelm (approved):

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -5,3 +5,36 @@
...
+ocm get cv xxx
+
+ocm transfer xxx
+```
+-->
+
+##### Verification
+
+- [ ] I have tested the changes locally by running `ocm`
+  -->
```
matthiasbruns:
```suggestion
```

matthiasbruns:
> I'm not sure whether we really want to have the explicit "behaviour is the same as in the old cli" / "compared behaviour" thing. I think more often than not, this is not applicable. Besides, we don't go for feature equivalence on every feature we implement.
> 
> I feel like this would be a bit confusing or off-putting for new contributors.

removed

PR #1832, opened 2026-02-24
feat(883): helm transfer

matthiasbruns:
This pull-request adds the "download classic helm chart" to the transformer api.

`This PR adds a bigger refactoring to the existing `input`and `oci-image-creation` - the concepts stayed the same, just technical changes.

`
**introduce helm access type from ocmv1**
It also adds the missing Helm access type to the new ocm.
```go
type Helm struct {
	// +ocm:jsonschema-gen:enum=Helm/v1,helm/v1
	// +ocm:jsonschema-gen:enum:deprecated=Helm,helm

	// HelmRepository is the URL of the helm repository to load the chart from.
	HelmRepository string `json:"helmRepository"`

	// HelmChart if the name of the helm chart and its version separated by a colon.
	HelmChart string `json:"helmChart"`

	// Version can either be specified as part of the chart name or separately.
	Version string `json:"version,omitempty"`
}

```
It is important to note what I dropped deprecated fields from the spec. CA and Keyring related fields should be loaded from the credentials as already done in the helm input.

**reuse blob download from helm input**
Additionally, this PR refactors the `download helm blob `from input to generalize the logic and make it reusable for helm access as well.

**Reuse chart to OCI from help input**
I also generalized the chart->OCI logic from `input` and made it reusable. It is now used in both `input` and the `convert transformer.`

**Refactor `CopyChartToOCILayout`**
I had to refactor `CopyChartToOCILayout` to be able to get the `ociImageSpecV1.Descriptor`. The problem was that the whole impl was async. Channels lead to blocking goroutines since `io.Pipe` was not done writing at the point where I needed the descriptor. I had to create

```go
type Result struct {
	*direct.Blob
	desc chan descriptorOrError
}

type descriptorOrError struct {
	Descriptor ociImageSpecV1.Descriptor
	Err        error
}
```
to delay the access to said data.

**GetHelmChart transformer implementation**

The get transformer does the following:
- validate input spec against new `v1alpha1.GetHelmChart`
- generate output paths for the spec output
- get credentials by requesting `ResourceConsumerIdentityProvider.GetResourceCredentialConsumerIdentity` which is implemented in the `HelmAccess`
- delegate download to generalized helm download from the input package (now downloads)
  - in theory, we could use this transformer also for oci helm download, since the downloader supports the feature
  - but we will use the native GetOCIArtifact from transformers (probably)
-  resulting blobs (chart and optionally a prov) will be copied into `*v1alpha1.File`
- the output contains 
  - the original `Resource`  
  - the chart file pointer with `v1alpha1.File`
  - (optionally) the prov file pointer with `*v1alpha1.File`

**ConvertHelmChartToOCI transformer implementation**

The convertion transformer does the following:
- expect chart, prov and resources as input
- reuse the oci creation logic from input
- calculate the correct `ImageReference` for the `OCIImage` access
- push the artifact as localblob as an output back the the graph

 **testing**
- The pr spawns a ~~repotest.NewTempServer()~~ mock http server pointing to the shared helm `testdata/mychart-0.1.0.tgz` dir
- In the test we create Resources with several variations of the `Helm` access
- Each test verifies that the files are correctly downloaded and equal the original file
- provenance file download added
  - **Open Question**: should we also verify during transform? Currently, we only download the .prov file
- helm to oci tests and validated the created oci image and the prov files 

Contributes: https://github.com/open-component-model/ocm-project/issues/883

##### How to test the changes

See https://github.com/open-component-model/open-component-model/pull/1846 for testing

**make sure helm input did not break**
```bash
#!/bin/zsh

alias OCM='go run ../../main.go'

# Test helm input with local path
CHART_PATH="../../../bindings/go/helm/testdata/mychart"

# stat the chart path to ensure it exists
if [ ! -d "$CHART_PATH" ]; then
  echo "Error: Chart path $CHART_PATH does not exist"
  exit 1
fi

cat <<EOF > constructor-input-local.yaml
components:
- name: ocm.software/helm-input-local
  version: 0.1.0
  provider:
    name: ocm.software
  resources:
  - name: mychart
    version: 0.1.0
    type: helmChart
    input:
      type: helm/v1
      path: $CHART_PATH
EOF

# create dir relative to current directory to avoid issues with absolute paths in the constructor
INPUT_CTF_LOCAL="input-ctf-local-$(date +%s)"
mkdir $INPUT_CTF_LOCAL
echo "Using temporary directory for helm input (local): $INPUT_CTF_LOCAL"

OCM add cv --repository ctf::$INPUT_CTF_LOCAL --constructor constructor-input-local.yaml

echo "--- Local helm input test passed ---"

# Test helm input with remote repository (uses same podinfo chart)
cat <<EOF > constructor-input-remote.yaml
components:
- name: ocm.software/helm-input-remote
  version: 6.9.1
  provider:
    name: ocm.software
  resources:
  - name: podinfo
    version: 6.9.1
    type: helmChart
    input:
      type: helm/v1
      helmRepository: https://stefanprodan.github.io/podinfo/podinfo-6.9.1.tgz
EOF

INPUT_CTF_REMOTE="input-ctf-remote-$(date +%s)"
mkdir $INPUT_CTF_REMOTE
echo "Using temporary directory for helm input (remote): $INPUT_CTF_REMOTE"

OCM add cv --repository ctf::$INPUT_CTF_REMOTE --constructor constructor-input-remote.yaml

echo "--- Remote helm input test passed ---"
```

##### Verification

## Summary by CodeRabbit

* **New Features**
  * Added Helm access spec, credential-aware identity resolution, and options for TLS/credentials.
  * Remote Helm chart retrieval (HTTP/OCI) with provenance support and configurable buffering/output path handling.
  * Conversion of Helm charts into OCI image layouts and transformers to fetch charts and produce OCI artifacts.

* **Tests**
  * Extensive unit and integration tests covering access parsing, identity/error cases, downloads, provenance, OCI conversion, and transformers.

### DIFF
--- a/bindings/go/helm/access/access.go
+++ b/bindings/go/helm/access/access.go
@@ -10,7 +10,10 @@ import (
 	"ocm.software/open-component-model/bindings/go/runtime"
 )
 
-const LegacyHelmChartConsumerType = "HelmChartRepository"
+const (
+	LegacyHelmChartConsumerType = "HelmChartRepository"
+	HelmRepositoryType          = "helmChart"
+)
 
 var Scheme = runtime.NewScheme()
 
--- a/bindings/go/helm/transformation/convert_helm_chart_to_oci.go
+++ b/bindings/go/helm/transformation/convert_helm_chart_to_oci.go
@@ -55,7 +55,12 @@ func (t *ConvertHelmChartToOCI) Transform(ctx context.Context, step runtime.Type
 		provSpec  *filesystem.Blob
 	)
 
-	chartPath := pathFromURI(transformation.Spec.ChartFile.URI)
+	var chartPath string
+	if p, err := url.Parse(transformation.Spec.ChartFile.URI); err != nil {
+		return nil, fmt.Errorf("invalid chart file URI: %w", err)
+	} else {
+		chartPath = p.Path
+	}
 	fs, err := filesystem.NewFS(filepath.Dir(chartPath), os.O_RDONLY)
 	if err != nil {
 		return nil, fmt.Errorf("failed creating filesystem from chart file spec: %w", err)
@@ -64,7 +69,12 @@ func (t *ConvertHelmChartToOCI) Transform(ctx context.Context, step runtime.Type
 	chartSpec = filesystem.NewFileBlob(fs, filepath.Base(chartPath))
 
 	if transformation.Spec.ProvFile != nil && transformation.Spec.ProvFile.URI != "" {
-		provPath := pathFromURI(transformation.Spec.ProvFile.URI)
+		var provPath string
+		if p, err := url.Parse(transformation.Spec.ProvFile.URI); err != nil {
+			return nil, fmt.Errorf("invalid prov file URI: %w", err)
+		} else {
+			provPath = p.Path
+		}
 		fs, err := filesystem.NewFS(filepath.Dir(provPath), os.O_RDONLY)
 		if err != nil {
 			return nil, fmt.Errorf("failed creating filesystem from prov file spec: %w", err)
@@ -105,7 +115,7 @@ func (t *ConvertHelmChartToOCI) Transform(ctx context.Context, step runtime.Type
 
 	transformation.Output.File = *spec
 	transformation.Output.Resource = transformation.Spec.Resource
-	transformation.Output.Resource.Type = "helmChart"
+	transformation.Output.Resource.Type = access.HelmRepositoryType
 
 	// Use digest of top-level-manifest
 	transformation.Output.Resource.Digest = &descv2.Digest{
@@ -157,12 +167,4 @@ func imageReferenceFromHelmAccess(helmAccess v1.Helm) (string, error) {
 	}
 
 	return ref, nil
-}
-
-func pathFromURI(uri string) string {
-	u, err := url.Parse(uri)
-	if err != nil {
-		return uri
-	}
-	return u.Path
 }
--- a/bindings/go/helm/input/blob.go
+++ b/bindings/go/helm/input/blob.go
@@ -164,7 +164,6 @@ func newReadOnlyChart(path, tmpDirBase string) (result *ReadOnlyChart, err error
 func newReadOnlyChartFromRemote(ctx context.Context, helmSpec v1.Helm, tmpDirBase string, credentials map[string]string) (result *ReadOnlyChart, err error) {
 	opts := []dlinternal.Option{
 		dlinternal.WithCredentials(credentials),
-		dlinternal.WithTargetDir(tmpDirBase),
 		//nolint:staticcheck // downward compatibility for helm input
 		dlinternal.WithVersion(helmSpec.Version),
 		//nolint:staticcheck // downward compatibility for helm input
@@ -172,7 +171,7 @@ func newReadOnlyChartFromRemote(ctx context.Context, helmSpec v1.Helm, tmpDirBas
 		//nolint:staticcheck // downward compatibility for helm input
 		dlinternal.WithCACertFile(helmSpec.CACertFile),
 	}
-	resultChart, err := dlinternal.NewReadOnlyChartFromRemote(ctx, helmSpec.HelmRepository, opts...)
+	resultChart, err := dlinternal.NewReadOnlyChartFromRemote(ctx, helmSpec.HelmRepository, tmpDirBase, opts...)
 	if err != nil {
 		return nil, fmt.Errorf("error downloading helm chart from repository %q: %w", helmSpec.HelmRepository, err)
 	}
--- a/bindings/go/helm/input/method.go
+++ b/bindings/go/helm/input/method.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"os"
 
 	"ocm.software/open-component-model/bindings/go/constructor"
 	constructorruntime "ocm.software/open-component-model/bindings/go/constructor/runtime"
@@ -84,6 +85,16 @@ func (i *InputMethod) ProcessResource(ctx context.Context, resource *constructor
 	helm := v1.Helm{}
 	if err := i.GetInputMethodScheme().Convert(resource.Input, &helm); err != nil {
 		return nil, fmt.Errorf("error converting resource input spec: %w", err)
+	}
+
+	if i.TempFolder == "" {
+		// we cannot delete the temp folder since it will hold the downloaded helm chart blobs
+		// and we do not want to break existing fallback behavior for users who do not set the TempFolder field before
+		temp, err := os.MkdirTemp("", "helm-input-*")
+		if err != nil {
+			return nil, fmt.Errorf("error creating temporary directory for helm input processing: %w", err)
+		}
+		i.TempFolder = temp
 	}
 
 	helmBlob, chart, err := GetV1HelmBlob(ctx, helm, i.TempFolder, WithCredentials(credentials))
--- a/bindings/go/helm/internal/download/doc.go
+++ b/bindings/go/helm/internal/download/doc.go
@@ -11,7 +11,6 @@
 //	if err != nil {
 //	    log.Fatal(err)
 //	}
-//	defer os.RemoveAll(chartData.ChartDir)
 //
 // To download a chart from an HTTP/S repository:
 //
--- a/bindings/go/helm/internal/download/download.go
+++ b/bindings/go/helm/internal/download/download.go
@@ -23,21 +23,25 @@ import (
 // NewReadOnlyChartFromRemote downloads a Helm chart from a remote repository and returns it as [helm.ChartData].
 // The helmRepo parameter accepts both OCI references (e.g. "oci://registry.example.com/charts/mychart:1.0.0")
 // and HTTP/S URLs (e.g. "https://example.com/charts/mychart-1.0.0.tgz").
-func NewReadOnlyChartFromRemote(ctx context.Context, helmRepo string, opts ...Option) (result *internal.ChartData, err error) {
+// The targetDir parameter specifies the directory where the chart will be downloaded and processed. The caller is responsible for cleaning up this directory after use.
+func NewReadOnlyChartFromRemote(ctx context.Context, helmRepo, targetDir string, opts ...Option) (result *internal.ChartData, err error) {
+	if helmRepo == "" {
+		return nil, errors.New("helm repository must be specified")
+	}
+	if targetDir == "" {
+		return nil, errors.New("target directory must be specified")
+	}
+
 	opt := &option{}
 	for _, o := range opts {
 		o(opt)
-	}
-
-	if opt.TargetDir == "" {
-		opt.TargetDir = os.TempDir()
 	}
 
 	if opt.Credentials == nil {
 		opt.Credentials = make(map[string]string)
 	}
 
-	chartDir, err := os.MkdirTemp(opt.TargetDir, "helmRemoteChart*")
+	chartDir, err := os.MkdirTemp(targetDir, "helmRemoteChart*")
 	if err != nil {
 		return nil, fmt.Errorf("error creating temporary directory: %w", err)
 	}
@@ -49,7 +53,7 @@ func NewReadOnlyChartFromRemote(ctx context.Context, helmRepo string, opts ...Op
 		withCACert(opt.CACert),
 		withCredentials(opt.Credentials),
 	}
-	tlsOption, err := constructTLSOptions(opt.TargetDir, tlsOpts...)
+	tlsOption, err := constructTLSOptions(targetDir, tlsOpts...)
 	if err != nil {
 		return nil, fmt.Errorf("error setting up TLS options: %w", err)
 	}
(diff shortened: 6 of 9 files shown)


matthiasbruns:
helm-to-oci transformation will be added today

bindings/go/helm/transformation/convert_helm_chart_to_oci.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+	if transformation.Spec == nil {
+		return nil, fmt.Errorf("spec is required for convert helm transformation")
+	}
+
+	var output *v1alpha1.ConvertHelmToOCIOutput
+	if transformation.Output == nil {
+		transformation.Output = &v1alpha1.ConvertHelmToOCIOutput{}
+	}
+	output = transformation.Output
```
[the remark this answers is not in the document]
matthiasbruns: we used that pattern in other transformers as well, maybe it's actually better to directly work on `transformation.Output`

bindings/go/helm/transformation/output.go
```diff
@@ -0,0 +1,42 @@
...
+//
+// TODO(matthiasbruns): Unify with OCI DetermineOutputPath if possible
+func DetermineOutputPath(outputPath string, filePrefix string) (string, error) {
+	if outputPath == "" {
+		// Create a temporary file in the default temp directory
+		tempFile, err := os.CreateTemp("", filePrefix+"-*")
+		if err != nil {
+			return "", fmt.Errorf("failed creating temporary file: %w", err)
+		}
+		_ = tempFile.Close() // Close immediately, caller will overwrite it
```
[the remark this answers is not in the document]
matthiasbruns: old code from input - no clue tbh - didn't want to change too much in this :D

bindings/go/helm/oci/oci.go
```diff
@@ -0,0 +1,215 @@
...
+
+type descriptorOrError struct {
+	Descriptor ociImageSpecV1.Descriptor
+	Err        error
+}
+
+// Descriptor returns the OCI image manifest descriptor.
+// It blocks until the blob-producing goroutine has finished.
+// The blob MUST be fully consumed before calling this method, otherwise it will deadlock.
+func (r *Result) Descriptor() (ociImageSpecV1.Descriptor, error) {
```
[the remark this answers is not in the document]
matthiasbruns: Yeah that whole bit destroyed by goroutine brain - let me update this bit

bindings/go/helm/transformation/spec/v1alpha1/convert_helm_to_oci.go
```diff
@@ -0,0 +1,59 @@
...
+// It contains the resource descriptor of the helm chart to be converted and an optional output path.
+// If no output path is given, a temporary file will be created for buffering the OCI artifact.
+// ChartFile is the file access specification for the downloaded helm chart, which can be obtained from the output of the GetHelmChart transformation.
+// ProvFile is the file access specification for the downloaded prov file.
+// If provided, the OCI artifact will be created with the prov layer with media-type "application/vnd.cncf.helm.chart.provenance.v1.prov"
+// +k8s:deepcopy-gen=true
+// +ocm:jsonschema-gen=true
+type ConvertHelmToOCISpec struct {
+	// ChartFile is the file access specification for the downloaded helm chart
+	ChartFile v1alpha1.File `json:"chartFile"`
```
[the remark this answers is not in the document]
matthiasbruns: that's how 
`chartFileSpec, err := filesystem.BlobToSpec(resultData.ChartBlob, chartOutputPath)` 
works 

```go
// URI must conform to RFC8089 and is a locator to the file.
// Note that implementations may still choose to reject URIs that if they do not fully
// implement RFC8089.
URI string `json:"uri"`
```

> According to the definition in [[RFC1738](https://www.rfc-editor.org/rfc/rfc1738)], a file URL always started
>    with the token "file://", followed by an (optionally blank) host name
>    and a "/".  The syntax given in [Section 2](https://www.rfc-editor.org/rfc/rfc8089#section-2) makes the entire authority
>    component, including the double slashes "//", optional.

bindings/go/helm/types.go
```diff
@@ -0,0 +1,19 @@
...
+import (
+	"ocm.software/open-component-model/bindings/go/blob/filesystem"
+)
+
+// ReadOnlyChart contains Helm chart contents as tgz archive, some metadata and optionally a provenance file.
+type ReadOnlyChart struct {
+	Name          string
+	Version       string
+	ChartBlob     *filesystem.Blob
+	ChartBlobPath string
```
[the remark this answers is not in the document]
matthiasbruns: good catch - this was used during dev but not used anymore - cleaned up

piotrjanik (approved):
I have tested implementation locally. LGTM

bindings/go/helm/transformation/convert_helm_chart_to_oci.go
```diff
@@ -0,0 +1,165 @@
...
+	// Now that the blob is consumed, we can safely retrieve the descriptor.
+	ociDesc, err := result.Descriptor(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get OCI manifest descriptor: %w", err)
+	}
+
+	transformation.Output.File = *spec
+	transformation.Output.Resource = transformation.Spec.Resource
+	// Update output.Resource.Type to ociImage
+	transformation.Output.Resource.Type = "ociImage"
```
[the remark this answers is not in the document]
matthiasbruns: we have talked about this in a meeting and I was very confused when this came up (maybe you remember this)- based on this
https://github.com/open-component-model/ocm-spec/blob/main/doc/01-model/02-elements-toplevel.md#artifact-types
it should be encoded like this, right?

```yaml
  resources:
  - name: mariadb-chart
    version: 12.2.7
    type: helmChart
    relation: external
    access:
      type: ociArtifact
      imageReference: ghcr.io/open-component-model/helmexample/charts/mariadb:12.2.7
```
(this is what I had initially)

bindings/go/helm/download/download.go
```diff
@@ -0,0 +1,229 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error creating temporary directory for helm operations: %w", err)
+	}
+
+	dl := &downloader.ChartDownloader{
+		Out:     os.Stderr,
+		Verify:  verify,
+		Getters: getterProviders(),
+		// set by ocm v1 originally.
+		RepositoryCache:  filepath.Join(helmTmp, ".helmcache"),
```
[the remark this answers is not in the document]
matthiasbruns: https://github.com/open-component-model/open-component-model/blob/main/bindings/go/helm/input/blob.go#L253

it is configurable now, the comment is a left over (opt.TmpDir can be provided into the method)

bindings/go/helm/download/download.go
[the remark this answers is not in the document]
matthiasbruns: its now

bindings/go/helm/access/access.go
```diff
@@ -0,0 +1,65 @@
...
+	descruntime "ocm.software/open-component-model/bindings/go/descriptor/runtime"
+	v1 "ocm.software/open-component-model/bindings/go/helm/access/spec/v1"
+	"ocm.software/open-component-model/bindings/go/runtime"
+)
+
+const LegacyHelmChartConsumerType = "HelmChartRepository"
+
+// ErrLocalHelmInputDoesNotRequireCredentials is returned when credential-related operations are attempted
+// on local helm inputs, since those are based on local filesystem and do not require authentication or authorization.
+var ErrLocalHelmInputDoesNotRequireCredentials = errors.New("local helm inputs do not require credentials")
```
fabianburth: This is quite the detailed error. I think, since the program calling a resource repository usually does not know the underlying representation, we cannot program against an access type specific error. And even if we had a general error, it could not cross the plugin boundary.

bindings/go/helm/transformation/convert_helm_chart_to_oci.go
```diff
@@ -0,0 +1,163 @@
...
+	if err = ociaccess.Scheme.Convert(&ociAccess, &updatedAccess); err != nil {
+		return nil, fmt.Errorf("failed converting ociv2.OCIImage access back to resource access format: %w", err)
+	}
+	transformation.Output.Resource.Relation = descv2.LocalRelation
+	transformation.Output.Resource.Access = &updatedAccess
+
+	return &transformation, nil
+}
+
+func referenceFromHelmAccess(helmAccess v1.Helm) (string, error) {
```
jakobmoellerdev: This function parser looks like it will break in prod. can we write tests for this please

bindings/go/helm/transformation/convert_helm_chart_to_oci.go
```diff
@@ -0,0 +1,163 @@
...
+	v1 "ocm.software/open-component-model/bindings/go/helm/access/spec/v1"
+	"ocm.software/open-component-model/bindings/go/helm/oci"
+	"ocm.software/open-component-model/bindings/go/helm/transformation/spec/v1alpha1"
+	"ocm.software/open-component-model/bindings/go/oci/looseref"
+	ociaccess "ocm.software/open-component-model/bindings/go/oci/spec/access"
+	ociv2 "ocm.software/open-component-model/bindings/go/oci/spec/access/v1"
+	"ocm.software/open-component-model/bindings/go/runtime"
+)
+
+type ConvertHelmChartToOCI struct {
```
jakobmoellerdev: this converter probably needs logging

bindings/go/helm/download/download.go
```diff
@@ -0,0 +1,229 @@
...
+
+	var plainHTTP bool
+	if strings.HasPrefix(helmRepo, "http://") {
+		slog.WarnContext(ctx, "using plain HTTP for chart download")
+		plainHTTP = true
+	}
+
+	getterOpts = append(getterOpts, getter.WithPlainHTTP(plainHTTP))
+
+	// TODO(fabianburth): check whether this needs additional configuration
```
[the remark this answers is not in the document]
fabianburth: Yeah, I added this during the helm v4 migration. The previous chart downloader API did not require a registry client. So, I just added a plain unconfigured one. I believe, my verdict was, that for the code path we used so far, the registry was unused anyway.
We might want to check again. But if you tested this sufficiently, I think we are good and you can just remove the comment.

bindings/go/helm/download/download.go
```diff
@@ -0,0 +1,250 @@
...
+		return nil, fmt.Errorf("error creating registry client: %w", err)
+	}
+
+	cacheDir, err := os.MkdirTemp(opt.TargetDir, "helm-cache*")
+	if err != nil {
+		return nil, fmt.Errorf("error creating temporary directory for helm operations: %w", err)
+	}
+	defer func(path string) {
+		_ = os.RemoveAll(path)
+	}(cacheDir)
```
fabianburth: you even declared a named returned parameter?

bindings/go/helm/download/download_test.go
```diff
@@ -0,0 +1,261 @@
...
+}
+
+func TestGetterProviders(t *testing.T) {
+	providers := getterProviders()
+	require.Len(t, providers, 2, "expected two getter providers")
+
+	assert.Contains(t, providers[0].Schemes, "http")
+	assert.Contains(t, providers[0].Schemes, "https")
+	assert.Contains(t, providers[1].Schemes, "oci")
+}
```
fabianburth: That is a weird test. If you wanted to test them individually, should you test against the actual providers API that is actually used by the downloader?

bindings/go/helm/oci/oci.go
```diff
@@ -0,0 +1,243 @@
...
+	}
+}
+
+// CopyChartToOCILayout takes a ChartData helper object and creates an OCI layout from it.
+// Three OCI layers are expected: config, tgz contents and optionally a provenance file.
+// The result is tagged with the helm chart version.
+// The returned Result contains the blob and provides access to the manifest descriptor
+// after the blob has been fully consumed.
+// See also: https://github.com/helm/community/blob/main/hips/hip-0006.md#2-support-for-provenance-files
+func CopyChartToOCILayout(ctx context.Context, chart *helm.ChartData) *Result {
```
fabianburth: Is there a particular reason why you decided to make `CopyChartToOCILayout` public API?

matthiasbruns: it is used in bindings/go/helm/transformation/convert_helm_chart_to_oci.go

bindings/go/helm/transformation/convert_helm_chart_to_oci.go
```diff
@@ -0,0 +1,163 @@
...
+		return "", fmt.Errorf("repository is required in Helm access to construct OCI image reference")
+	}
+
+	ref = lref.Repository
+	if lref.Tag != "" {
+		ref = ref + ":" + lref.Tag
+	}
+
+	return ref, nil
+}
```
fabianburth: Didn't you define another public API method called `HelmChartReference()` on the access type itself?

bindings/go/helm/types.go
```diff
@@ -0,0 +1,13 @@
...
+	"ocm.software/open-component-model/bindings/go/blob/filesystem"
+)
+
+// ChartData contains Helm chart contents as tgz archive, some metadata and optionally a provenance file.
+type ChartData struct {
+	Name      string
+	Version   string
+	ChartBlob *filesystem.Blob
+	ProvBlob  *filesystem.Blob
+}
```
fabianburth: do we need this as public API

matthiasbruns: moved into `internal`

bindings/go/helm/oci/oci.go
matthiasbruns: moved into `internal`

matthiasbruns:
@jakobmoellerdev I couldn't break helm input, maybe you run some ocm commands just to be safe

PR #1890, opened 2026-03-03
fix: register oci type instead of `AnyConsumerIdentityType`

frewilhelm:

Currently, the `OCICredentialRepository` is registered with the `AnyConsumerIdentityType`. As a result, any credential identity `*` would be matched against the `OCICredentialRepository`. However, the `OCICredentialRepository` can only handle credentials of type `DockerConfig` which requires certain fields, e.g. `hostname`.

This PR changes the registration of the `OCICredentialRepository` with `AnyConsumerIdentityType` to `v1.Type` which results in `OCIRepository`.

Fixes https://github.com/open-component-model/ocm-project/issues/786

##### Verification

## Summary by CodeRabbit

* **Refactor**
  * Modernized OCI credentials identity type infrastructure across controller components and test suites. Updated internal credential repository plugin registration mechanisms to use versioned identity specifications instead of generic consumer identity types. Changes enhance consistency, maintainability, and alignment with OCI credential handling standards, improving overall system architecture.


frewilhelm:
Previously failing integration tests were fixed with https://github.com/open-component-model/open-component-model/pull/1967 and https://github.com/open-component-model/open-component-model/pull/1970 (notice that we removed the typed error again in the second PR but kept returning the `ErrNoIndirectCredentials`)

PR #1896, opened 2026-03-04
fix: add failure on exisiting branch and dont swallow message.

morri-son:
Let the workflow fail in case the release branhc to create already exist. Also print a Summary in that case.

  

## Summary by CodeRabbit

* **Chores**
  * Updated release branch workflow to explicitly fail with a clear summary and message when the target branch already exists, replacing a previous silent skip.
  * Removed an unused environment configuration from the release branch workflow to simplify execution.

### DIFF
--- a/.github/workflows/release-branch.yml
+++ b/.github/workflows/release-branch.yml
@@ -61,7 +61,10 @@ jobs:
             // Check if target branch exists
             try {
               await github.rest.repos.getBranch({ owner, repo, branch: target });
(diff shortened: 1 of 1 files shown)


frewilhelm (dismissed):

frewilhelm (changes requested):
This PR has over 16k LoC. I think something went wrong on a merge/rebase. Please fix that because I do not know which of the 230 files I should review :D

frewilhelm (approved):

matthiasbruns (approved):

PR #1944, opened 2026-03-10
perf(oci)!: make OCILayoutWriter.Push fully write concurrent

### LINKED ISSUE #895: Re-visit `NewOCILayoutWriter` for efficiency during download
**Description**

During an OCM daily, we found that `NewOCILayoutWriter` is actually critically slow for downloading. Jakob was trying to download a postgres image and it took over minutes to download and get that. It eventually boiled down to this function and the comment on the function:

```go
// NewOCILayoutWriter creates a new oras.Target that writes to the given writer an oci-layout in tar format.
// The index and layout files are written to the storage when it is closed.
// This writer is bound to serialization of tar archives and thus cannot be concurrently copied to efficiently, however
// it is significantly more efficient in terms of I/O than writing an OCI Layout to the filesystem and then
// tarring the OCI Layout directory.
// As such, this writer is optimized for I/O and memory efficiency, not for concurrency.
// Note however that in most modern systems, I/O is generally fast enough (ssd) to keep up with
// network speeds and concurrent reads, so this should not be a problem in most cases,
// as long as your final goal is a tar of an oci layout and not just the oci layout itself.
func NewOCILayoutWriter(w io.Writer) *OCILayoutWriter {
	return &OCILayoutWriter{
		writer:      tar.NewWriter(w),
		tagResolver: newMemoryResolver(),
		index: &ociImageSpecV1.Index{
			Versioned: specs.Versioned{
				SchemaVersion: 2, // historical value
			},
			Manifests: []ociImageSpecV1.Descriptor{},
		},
	}
}
```

The goal of this issue is to introduce a file buffer, perform a parallel download to the file buffer and then create the tar.

**Done Criteria**
- [x] Test with larger images finishes in a reasonable time instead of several minutes ( postgres image for example )

jakobmoellerdev:

Makes it so that the Tar Writer we use for layout serialization is now using a file buffer so there is no contention on the tar and we can use our full network I/O. Before we accidentally serialized

Replace mutex-serialized tar.Writer with atomic offset reservation and concurrent WriteAt to non-overlapping regions in a scratch buffer. Previously Push held a mutex for the entire blob transfer (network read + tar write), serializing all goroutines. Now multiple goroutines write simultaneously with zero contention on the write path.

Key changes:

**Real-image benchmark shows ~7x speedup at concurrency 8 (our current concurrency setting) for transfers and copies of native OCI artifacts**

- Two-phase design: Push writes tar entries to reserved byte ranges via WriteAt; Close streams the completed tar to output
- Introduced OCILayoutBuffer interface (io.WriterAt + io.ReadSeeker)
- NewOCILayoutWriterWithTempFile now accepts a dir parameter
- Exists is O(1) via map[digest.Digest]Descriptor with RWMutex
- copyToWriterAt with pooled buffers for zero-alloc Push
- Wire TempDir from RepositoryOptions to Repository struct (was missing)

fix https://github.com/open-component-model/ocm-project/issues/895

#### Testing

I have ran manual tests and included a benchmark so I was able to actually run this. I also ran a go.work based test

##### How to test the changes

Run the benchmark locally.

##### Verification

## Summary by CodeRabbit

* **New Features**
  * Explicit temporary-directory option for OCI layout operations to improve resource handling.
  * Concurrent-safe OCI layout writing enabling parallel pushes and faster layout assembly.
* **Improvements**
  * Better error propagation during layout creation and improved copy/download logging.
* **Tests**
  * New and extended tests covering concurrent writes, scratch-file lifecycle, and benchmarks copying real OCI images at multiple concurrency levels.


matthiasbruns (approved):
Went through the code, checked it out and made cli integration somewhat to work (this will be fun).

All in all a great change

Benchmark goes brrr

```bash
BenchmarkOCILayoutWriter_Push_Sequential
BenchmarkOCILayoutWriter_Push_Sequential/blob_1KB
BenchmarkOCILayoutWriter_Push_Sequential/blob_1KB-16         	    2589	    461548 ns/op	 110.93 MB/s
BenchmarkOCILayoutWriter_Push_Sequential/blob_64KB
BenchmarkOCILayoutWriter_Push_Sequential/blob_64KB-16        	     427	   2837377 ns/op	1154.87 MB/s
BenchmarkOCILayoutWriter_Push_Sequential/blob_1024KB
BenchmarkOCILayoutWriter_Push_Sequential/blob_1024KB-16      	      31	  38096887 ns/op	1376.20 MB/s
BenchmarkOCILayoutWriter_Push_Concurrent
BenchmarkOCILayoutWriter_Push_Concurrent/blob_1KB
BenchmarkOCILayoutWriter_Push_Concurrent/blob_1KB-16         	    1264	   1011473 ns/op	  50.62 MB/s
BenchmarkOCILayoutWriter_Push_Concurrent/blob_64KB
BenchmarkOCILayoutWriter_Push_Concurrent/blob_64KB-16        	     559	   2220463 ns/op	1475.73 MB/s
BenchmarkOCILayoutWriter_Push_Concurrent/blob_1024KB
BenchmarkOCILayoutWriter_Push_Concurrent/blob_1024KB-16      	      42	  30249608 ns/op	1733.21 MB/s
```

I would want to wait for @Skarlso eyes on go perf stuff, but from my side, this looks great. Would you want to provide a PR that integrates the changes through the whole lib to see how it behaves when integrated? The whole temp handling in helm etc might need refactoring as well

Skarlso (approved):

PR #1971, opened 2026-03-13
fix: Correct calculation of commits for changelog for release workflows

morri-son:

Add new function to calculate the latest previous non-RC tag to determine the correct range of commits for the changelog.
  

## Summary by CodeRabbit

* **Chores**
  * Improved release versioning to compute and expose the previous release tag and a concrete changelog range for release-candidate flows, pass that range into the changelog generator, and surface these values in step summaries and workflow outputs (first-release behavior preserved).
* **Tests**
  * Added comprehensive tests covering prior-release detection, RC and edge-case scenarios to ensure correct changelog-range behavior.


matthiasbruns (dismissed):

piotrjanik (commented):
LGTM, one corner case to check

piotrjanik (approved):

matthiasbruns (approved):
ignore my nit - I do not want to block because of nits

PR #2028, opened 2026-03-20
chore: truncate versions for controller build that interfere with kubernetes 63 char limit

morri-son:

introduce new helper for controller chart to truncate Chart version
  

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected monitoring TLS configuration so certificate key settings are properly aligned and used.

* **Chores**
  * Enforced a maximum length for generated version labels to meet platform constraints; overly long pseudo-versions are truncated and a warning is emitted when truncation occurs.

* **Tests**
  * Added tests covering version truncation behavior, warning emission, and boundary conditions.


jakobmoellerdev (dismissed):

kubernetes/controller/chart/templates/_helpers.tpl
```diff
@@ -31,6 +31,13 @@ Always uses the Helm release namespace.
 {{- .Release.Namespace }}
 {{- end }}
 
+{{/*
+Create chart label value (chart name + version), truncated to 63 characters.
+*/}}
+{{- define "ocm-k8s-toolkit.chart" -}}
```
matthiasbruns: I do not think we should truncate within the chart - users need to make sure they provide valid resource names, it's now our duty to clean this up for them - the fix should not be here but where we build the charts in the cli

matthiasbruns (changes requested):
I think the fix is in the wrong place - the charts should not have value clean up - the values are provided by devops - they either know about the resource limit or will have to handle the errors themselves

the fix should be in the pipeline, where we build the controller and truncate the name there OR in create-version.js and truncate the output

matthiasbruns (dismissed):

.github/scripts/compute-version.test.js
```diff
@@ -201,3 +201,132 @@ assert.notStrictEqual(
...
 
 console.log("✅ All tests passed.");
+
+// ----------------------------------------------------------
+// Version truncation tests
+// ----------------------------------------------------------
+console.log("Testing version truncation...");
+
+// DEFAULT_MAX_VERSION_LENGTH should be exported and equal 57
+assert.strictEqual(
```
piotrjanik: This value is harcoded, so very little value of this test.

piotrjanik (approved):

jakobmoellerdev (approved):

PR #2070, opened 2026-03-25
fix: reset stale error when no digest processor is registered

morri-son:

## Summary
- When `GetDigestProcessor` returns no processor for an access type, the error was never cleared, causing `processResource` to fail with a stale error
- Refactored `GetPlugin`/`getPlugin` to return `(nil, nil)` instead of an error when no processor is registered
- Caller now checks `digestProcessor != nil` to skip digest processing 
## Changes
- `registry.go`: `GetPlugin` and `getPlugin` return `(nil, nil)` for unregistered types
- `construct.go`: rewritten digest block with proper error scoping (`err :=`) and nil-processor check
- `construct_resource_test.go`: added `TestConstructWithResourceDigestProcessorNotFound`


jakobmoellerdev (commented):
I dont understand the base assumption that the error was not cleared. where is the error not cleared in the previous code?

PR #2082, opened 2026-03-27
feat(cli): add self-update command via GitHub releases

jakobmoellerdev:
Adds `ocm update` command that checks GitHub releases for a newer version of the CLI and atomically replaces the running binary.

- Version discovery uses the GitHub Releases list API, filtering to `cli/v*` tags to handle the multi-component repository correctly
- Binary asset is matched by `ocm-{GOOS}-{GOARCH}` name, covering all 6 release platforms (linux, darwin, windows × amd64, arm64)
- Atomic replacement follows the go-rocket-update Patcher pattern: backup current binary to `.old`, rename new binary into place, auto-rollback on failure
- No new Go module dependencies (uses stdlib + existing github.com/Masterminds/semver/v3)

Flags:
  --check        exit 0 if update available, exit 1 if already up to date
  --version      update to a specific version
  --pre-release  include RC versions when finding latest
  --dry-run      print what would happen without downloading
  --yes / -y     skip confirmation prompt

this is a prototype for https://github.com/open-component-model/ocm-project/issues/576

##### Verification


morri-son:
reopened as asked for in #3235

morri-son:
@jakobmoellerdev reopened the PR as mentioned by you. What about dropping the use of GH releases and use OCM CVs instead, now that we have signed components. The CLI component is published on ghcr.io and signed with sigstore. We can verify the cv, then download via ocm download resources (which resolves to the same release assets now via Wget access). No gh CLI dependency, proper trust chain and we eat our own dog food 😊

For macOS we may massage the downloaded binary a bit, to avoid macOS quarantine.

jakobmoellerdev:
closing again as there is no movement on this :_(

PR #2128, opened 2026-03-31
feat!: helm resource repo bindings

### LINKED ISSUE #911: Helm ResourceRepository
**Description**
During the [helm transformer feature](https://github.com/open-component-model/ocm-project/issues/883) I added helm chart access (classic). Currently, resource download would fail since we do not have `ResourceRepository` support for the helm access type.

We need to provide a `ResourceRepository` abstraction and register that in the plugin system to allow cli based resolution for classic helm chart resources.

**Done Criteria**
- [ ] Create a resource repository in helm bindings
- [ ] Register the new repository in the plugin system
- [ ] Drop/Replace the temporary interface in `bindings/go/helm/interface.go` and use [repository.ResourceRepository](https://github.com/open-component-model/open-component-model/blob/main/bindings/go/repository/interface.go#L104)
- [ ] Add tests with helm access in cli
- [ ] Test that `download resource` also works
- [ ] Make sure it works with a resource created by ocm v1
- [ ] After its merged, update the helm transformer to use the new resource repository

matthiasbruns:

Introduces a Helm-based `ResourceRepository` implementing `repository.ResourceRepository` and  `bindings/go/plugin/manager/registries/resource/contract.go`.

Full usage can be seen here: https://github.com/open-component-model/open-component-model/pull/2130

Contributes open-component-model/ocm-project#911

#### CLI PR
https://github.com/open-component-model/open-component-model/pull/2094

#### Testing
Fully tested in https://github.com/open-component-model/open-component-model/pull/2130

#### Helm ResourceRepository — Changes compared to legacy OCM

- **CLI & transfer** will use the `ResourceRepository` interface instead of the deprecated `HelmAccess` struct for
  credential identity resolution and chart downloads
- **Deprecated types removed**: `helm.ResourceConsumerIdentityProvider` interface, `helmaccess.HelmAccess` struct, and
  their tests — functionality is covered by `ResourceRepository`
- **Blob format**: Downloads return a `ChartBlob` (tar archive wrapping `.tgz` + optional `.prov`), replacing the legacy
  pattern of separate blob accessors for chart and provenance
- **Plugin registration**: The helm `ResourceRepository` is registered as a builtin plugin alongside the existing OCI
  and digest processor plugins in https://github.com/open-component-model/open-component-model/pull/2130

##### Intentional differences from legacy

|                                      | Legacy OCM                                                        | New                                                          |
|--------------------------------------|-------------------------------------------------------------------|--------------------------------------------------------------|
| `CACert` / `Keyring` on access spec  | Supported as inline fields                                        | Omitted — use credential resolver instead                    |
| OCI-hosted helm charts (`oci://`)    | Handled by helm access method with separate OCI download path     | Should use the OCI ResourceRepository (separate access type) |
| Blob return format (only internally) | Raw `.tgz` blob (`ChartLayerMediaType`), prov accessed separately | `ChartBlob` tar wrapping both `.tgz` and `.prov`             |
| Local charts                         | Not in access method (`IsLocal() = false`)                        | Not in ResourceRepository — handled by helm input plugin     |

##### Test coverage

- Unit tests for `ResourceRepository` (credential identity, download, temp folder usage, nil guards)
- Unit tests for `ChartBlob` extraction (using real testdata charts)
- Integration test: `add cv` with helm access to OCI registry
- Integration test: `download resource` with helm access from CTF (byte-level verification)
- Shell test: end-to-end download of podinfo chart with SHA256 verification


jakobmoellerdev (commented):
This is a really cool change 🚀  First pass didnt make me flag anything but need to read through the repo changes more in detail

bindings/go/helm/repository/resource/resource_repository.go
```diff
@@ -0,0 +1,194 @@
...
+// for the given helm resource. For OCI-based helm repositories the identity type
+// is OCIRegistry; for HTTP/HTTPS repositories it is HelmChartRepository.
+// Returns nil if the resource has no remote repository (local chart).
+func (r *ResourceRepository) GetResourceCredentialConsumerIdentity(_ context.Context, resource *descriptor.Resource) (runtime.Identity, error) {
+	helm, err := r.convertAccess(resource)
+	if err != nil {
+		return nil, err
+	}
+
+	if helm.HelmRepository == "" {
```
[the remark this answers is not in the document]
frewilhelm: `helminternal.CredentialConsumerIdentity()` return an error instead of `nil, nil`. Nevertheless, it is redundant and could be handled by a custom error that is interpreted by the original caller (There is already such a defined error available)

matthiasbruns:
I will finish this tomorrow

bindings/go/helm/access/access.go
```diff
@@ -31,33 +25,3 @@ func MustAddToScheme(scheme *runtime.Scheme) {
 		runtime.NewUnversionedType(v1.LegacyType),
 	)
 }
-
-// HelmAccess is a ResourceCredentialConsumerIdentityProvider that provides consumer identities for Helm repository resources.
-// TODO(matthiasbruns): Introduce a helm based ResourceRepository and move this logic into the implementation. https://github.com/open-component-model/ocm-project/issues/911
-type HelmAccess struct{}
```
Skarlso, line 37: This is now a breaking change, right? This has been completely removed.

bindings/go/helm/internal/credentials.go
```diff
@@ -0,0 +1,37 @@
+package internal
+
+import (
+	"errors"
+	"fmt"
+
+	helmaccess "ocm.software/open-component-model/bindings/go/helm/access"
```
Skarlso: Yeah, this should definitely break. 🤔

bindings/go/helm/repository/resource/resource_repository.go
```diff
@@ -0,0 +1,228 @@
...
+// removes the temp directory once it has the resulting blob, so any lazy reading
+// from the filesystem would fail.
+func tarDirectoryRecursive(ctx context.Context, dir string) (blob.ReadOnlyBlob, error) {
+	var buf bytes.Buffer
+	tw := tar.NewWriter(&buf)
+	defer func(tw *tar.Writer) {
+		err := tw.Close()
+		if err != nil {
+			slog.WarnContext(ctx, "Error closing tar writer", "error", err)
+		}
```
Skarlso: Since you are also closing this at the tend to flush the writer, this would produce continuous warnings that you are closing an already closed writer. 🤔

bindings/go/helm/transformation/get_helm_chart.go
```diff
@@ -71,61 +76,58 @@ func (t *GetHelmChart) Transform(ctx context.Context, step runtime.Typed) (runti
...
-	// We omit backwards compatibility options like certificates and keyrings and rely on the credentials approach.
-	opts := []helmdownload.Option{
-		// If a version is specified in the Helm access spec, use it to ensure we get the correct chart version.
-		// If not specified, the downloader will use the default behavior which tries to get the version from helmAccess.HelmRepository
-		//nolint:staticcheck // downward compatibility for helm input
-		helmdownload.WithVersion(helmAccess.GetVersion()),
-		helmdownload.WithCredentials(creds),
-		// Override the default downloader behavior to always download the chart and prov files.
-		helmdownload.WithAlwaysDownloadProv(true),
+	chartBlob, ok := downloadedBlob.(*helmblob.ChartBlob)
```
Skarlso: Is this specifically needed? Can't this be a `blob.ReadOnlyBlob`?

bindings/go/helm/blob/chart_blob.go
```diff
@@ -0,0 +1,121 @@
...
+func (c *ChartBlob) ProvFile() (blob.ReadOnlyBlob, error) {
+	result, err := c.extract()
+	if err != nil {
+		return nil, err
+	}
+	return result.provBlob, nil
+}
+
+// extractFromTar reads a tar archive blob and extracts the chart tgz and optional prov file.
+func extractFromTar(tarBlob blob.ReadOnlyBlob) (chartBlob blob.ReadOnlyBlob, provBlob blob.ReadOnlyBlob, err error) {
```
Skarlso, line 79: I'm okay with this reading a chart into memory, but because it does that, I would like to explicitly comment that this should only be used for charts since it's reading the ENTIRE thing into memory.

bindings/go/helm/blob/chart_blob.go
```diff
@@ -0,0 +1,121 @@
...
+			return nil, nil, fmt.Errorf("error reading tar entry: %w", err)
+		}
+
+		data, err := io.ReadAll(tr)
+		if err != nil {
+			return nil, nil, fmt.Errorf("error reading tar entry %s: %w", header.Name, err)
+		}
+
+		switch {
+		case strings.HasSuffix(header.Name, ".tgz"):
```
Skarlso: Errr... what about `.tar.gz`.

Skarlso (commented):
Some questions. :)

bindings/go/helm/spec/access/scheme.go
fabianburth, line 1: We should probably align the package structure with [oci](https://github.com/open-component-model/open-component-model/blob/b9bab6be5974c7405274b5d65bba76d0c4a0a293/bindings/go/oci/spec/access/scheme.go) (and keep this aligned throughout the repository). 
So, `bindings/go/helm/spec/access/v1` and this file here should probably be called `scheme.go`.

I already pointed out some issues with the current setup [here](https://github.com/open-component-model/open-component-model/pull/2055#discussion_r2994581312). I think, those will be resolved now anyway, as you move the logic out of the `access` package.

bindings/go/helm/blob/chart_blob.go
```diff
@@ -0,0 +1,121 @@
...
+	"ocm.software/open-component-model/bindings/go/blob/compression"
+	"ocm.software/open-component-model/bindings/go/blob/filesystem"
+	"ocm.software/open-component-model/bindings/go/blob/inmemory"
+)
+
+// extractResult holds the lazily-extracted chart and provenance blobs.
+type extractResult struct {
+	chartBlob blob.ReadOnlyBlob
+	provBlob  blob.ReadOnlyBlob
+}
```
fabianburth, line 30: nit: I think this is obsolete.

```go
extract func() (chartBlob blob.ReadOnlyBlob, provBlob blob.ReadOnlyBlob, err error)
```
(or maybe even better than named return params, just document the extract function)

```go
func (c *ChartBlob) ChartArchive() (blob.ReadOnlyBlob, error) {
	chartBlob, _, err := c.extract()
	if err != nil {
		return nil, err
	}
	return chartBlob, nil
}
```

bindings/go/helm/blob/chart_blob.go
```diff
@@ -0,0 +1,121 @@
...
+			break
+		}
+		if err != nil {
+			return nil, nil, fmt.Errorf("error reading tar entry: %w", err)
+		}
+
+		data, err := io.ReadAll(tr)
+		if err != nil {
+			return nil, nil, fmt.Errorf("error reading tar entry %s: %w", header.Name, err)
+		}
```
fabianburth, line 103: nit: we would read all now even if the header is invalid (none of the switch cases match)

bindings/go/helm/repository/resource/resource_repository.go
```diff
@@ -0,0 +1,228 @@
...
+	return nil, fmt.Errorf("helm chart repositories do not support upload operations")
+}
+
+func (r *ResourceRepository) convertAccess(resource *descriptor.Resource) (*v1.Helm, error) {
+	if resource == nil {
+		return nil, fmt.Errorf("resource is required")
+	}
+	if resource.Access == nil {
+		return nil, fmt.Errorf("resource access is required")
+	}
```
fabianburth: I don't think we are usually this defensive. Would prefer inlining this.

bindings/go/helm/repository/resource/resource_repository.go
```diff
@@ -0,0 +1,228 @@
...
+		return nil, fmt.Errorf("error converting resource access to helm spec: %w", err)
+	}
+
+	identity, err := helminternal.CredentialConsumerIdentity(helm.HelmRepository)
+	switch {
+	case errors.Is(err, helminternal.ErrLocalHelmInputDoesNotRequireCredentials):
+		return nil, nil
+	case err != nil:
+		return nil, fmt.Errorf("error resolving credential consumer identity: %w", err)
+	}
```
fabianburth: I'd rather do this inline here than introducing an additional error type we have to check for.
```go
	if helmRepository == "" {
		return nil, ErrLocalHelmInputDoesNotRequireCredentials
	}
```

bindings/go/helm/repository/resource/resource_repository.go
```diff
@@ -0,0 +1,228 @@
...
+
+	return helmblob.NewChartBlob(tarBlob), nil
+}
+
+// tarDirectoryRecursive creates an in-memory tar archive from all files in the given
+// directory tree. The blob must be fully buffered in memory because the download
+// directory is cleaned up immediately after this function returns -- the caller
+// removes the temp directory once it has the resulting blob, so any lazy reading
+// from the filesystem would fail.
+func tarDirectoryRecursive(ctx context.Context, dir string) (blob.ReadOnlyBlob, error) {
```
fabianburth: We also have https://github.com/open-component-model/open-component-model/blob/1062fdf73fd4724133ca0fe0c81b5a9e58ff1f39/bindings/go/blob/filesystem/blob_from_path.go#L52. Would it be possible somehow to reuse this here? (I know it might cost a bit of effort due to the cleanup behaviour.

bindings/go/helm/transformation/get_helm_chart.go
```diff
@@ -38,6 +40,9 @@ func (t *GetHelmChart) Transform(ctx context.Context, step runtime.Typed) (runti
 	if transformation.Spec == nil {
 		return nil, fmt.Errorf("spec is required for get helm transformation")
 	}
+	if t.ResourceRepository == nil {
+		return nil, fmt.Errorf("ResourceRepository is required for get helm transformation")
+	}
```
fabianburth: Same here. I don't think we usually are programming this defensive.

bindings/go/helm/transformation/get_helm_chart.go
```diff
@@ -71,61 +76,58 @@ func (t *GetHelmChart) Transform(ctx context.Context, step runtime.Typed) (runti
...
+	// Load the chart from the written file to get the resolved name and version
+	chartURL, err := url.Parse(chartFileSpec.URI)
+	if err != nil {
+		return nil, fmt.Errorf("failed parsing chart file URI %q: %w", chartFileSpec.URI, err)
+	}
+	chartPath := chartURL.Path
+	loadedChart, err := loader.Load(chartPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed loading downloaded chart to read metadata: %w", err)
+	}
```
fabianburth: I think instead of loading the chart from filesystem again, we might be able to use https://github.com/helm/helm/blob/cf69a6ef70211946e31e6910e6b6d7662bbc7ff9/pkg/chart/loader/load.go#L161. The API is a bit interesting (especially the interface return value), but that's because of their interesting accessor API to be able to transparently access v2 and v3 charts.

bindings/go/helm/internal/credentials_test.go
```diff
@@ -0,0 +1,57 @@
...
+	t.Run("returns HelmChartRepository identity for HTTPS repository", func(t *testing.T) {
+		identity, err := internal.CredentialConsumerIdentity("https://charts.example.com/stable")
+		require.NoError(t, err)
+		require.NotNil(t, identity)
+		assert.Equal(t, runtime.NewUnversionedType(helmaccess.LegacyHelmChartConsumerType), identity.GetType())
+		assert.Equal(t, "https", identity["scheme"])
+		assert.Equal(t, "charts.example.com", identity["hostname"])
+	})
+
+	t.Run("returns HelmChartRepository identity for HTTP repository", func(t *testing.T) {
```
fabianburth: nit: I think this could be written as a table test.

bindings/go/helm/spec/access/scheme.go
matthiasbruns: yeah - since this pr is breaking anyway, it makes sense to do it here

bindings/go/helm/repository/resource/resource_repository.go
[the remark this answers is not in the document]
matthiasbruns: `helmRepository == nil` should result in an error as well right?

bindings/go/helm/blob/chart_blob.go
matthiasbruns: not in combination with sync.Values

```go
	cb.extract = sync.OnceValues(func() (*extractResult, error) {
		if tarBlob == nil {
			return nil, fmt.Errorf("tar blob is required")
		}
		chart, prov, err := extractFromTar(tarBlob)
		if err != nil {
			return nil, err
		}
		return &extractResult{chartBlob: chart, provBlob: prov}, nil
	})
```

`cb.extract` stores the state of the extraction, but I could also move this into the `ChartBlob` as properties.

This is the definition of sync.OnceValues

`func OnceValues[T1, T2 any](f func() (T1, T2)) func() (T1, T2) {`

`T1` is the result struct in this case
`T2` is the error

bindings/go/helm/transformation/get_helm_chart.go
matthiasbruns: I need chart and prov file access, which is handled by `ChartBlob `

bindings/go/helm/repository/resource/resource_repository.go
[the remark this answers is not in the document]
matthiasbruns: if we pass into

```go
func (r *ResourceRepository) GetResourceCredentialConsumerIdentity(ctx context.Context, resource *descriptor.Resource) (runtime.Identity, error) {
	helm, err := r.convertAccess(resource)
	if err != nil {
		return nil, fmt.Errorf("error converting resource access to helm spec: %w", err)
	}

	if helm.HelmRepository == "" {
		return nil, fmt.Errorf("no helm repository specified")
	}
// ....
```
it will fail with an error in `convertAccess`

bindings/go/helm/transformation/get_helm_chart.go
matthiasbruns: what I could do it simply return `blob.ReadOnlyBlob` from the `ResourceRepo` and call `chartBlob := helmblob.NewChartBlob(downloadedBlob)` in the transformer

bindings/go/helm/transformation/get_helm_chart.go
matthiasbruns: yeah I guess in this case we can panic

bindings/go/helm/repository/resource/resource_repository.go
matthiasbruns: on it

matthiasbruns: I cannot get it to work

When I change this to

```go
tarBlob, err := filesystem.GetBlobFromPath(ctx, downloadResult.ChartDir, filesystem.DirOptions{})
	if err != nil {
		return nil, fmt.Errorf("error creating tar archive from helm download: %w", err)
	}
```

The only header I get is `./`

Debugging https://github.com/open-component-model/open-component-model/blob/1062fdf73fd4724133ca0fe0c81b5a9e58ff1f39/bindings/go/blob/filesystem/blob_from_path.go#L327

```go
// createTarHeader creates a TAR header from the given FileInfo.
// If reproducible is true, it normalizes the header for reproducibility.
func createTarHeader(fi fs.FileInfo, linkTarget string, reproducible bool) (*tar.Header, error) {
	h, err := tar.FileInfoHeader(fi, linkTarget)
	if err != nil {
		return nil, fmt.Errorf("error creating tar header for file %q: %w", fi.Name(), err)
	}
	if reproducible {
		h.ModTime = time.Unix(0, 0)
		h.AccessTime = time.Unix(0, 0)
		h.ChangeTime = time.Unix(0, 0)
		h.Uid, h.Gid = 0, 0
		h.Uname, h.Gname = "", ""
		h.Mode &= 0o777 // remove all but permission bits
	}
	return h, nil
}
```

has `./` in FileInfoHeader when it detects a dir - I need a list of files though - maybe we can discuss this in the WR

jakobmoellerdev (commented):
I think this is great and looks ready for merge. Unfortunately now the website is in the monorepo so I would ask you to be the guinea pig to submit your docs change in this PR as well!

matthiasbruns:
> I think this is great and looks ready for merge. Unfortunately now the website is in the monorepo so I would ask you to be the guinea pig to submit your docs change in this PR as well!

Did you call me a pig? :D 
Will do :) 
🐹💉

bindings/go/helm/repository/resource/resource_repository.go
matthiasbruns: I will keep the private impl for now - we can have a look at this again, but its working as it should and doesnt hurt anybody

bindings/go/helm/blob/chart_blob.go
matthiasbruns: we need to read the page of data to get the next header, event if we do not support the header itself

matthiasbruns:
> I think this is great and looks ready for merge. Unfortunately now the website is in the monorepo so I would ask you to be the guinea pig to submit your docs change in this PR as well!

I don't even know what to update here.
I wanted to include the schema rendering for the new schemes, but this does not work, since the website references main and the moved files are not available yet where I expect them. Nothing has changed besides the ResourceRepo, but for that we do not have anything yet.

So I will write a reference and concept doc about resource repos

matthiasbruns:
@netlify deploy

matthiasbruns:
added 

- concept https://deploy-preview-2128--ocm-website.netlify.app/docs/concepts/resource-repositories/
- reference https://deploy-preview-2128--ocm-website.netlify.app/docs/reference/resource-repositories/

website/content/docs/reference/resource-repositories.md
```diff
@@ -0,0 +1,141 @@
...
+|----------------------------------------------------------------|---------|
+| [`Helm/v1`]({{< relref "input-and-access-types.md" >}}#helmv1) | `helm`  |
+
+### Capabilities
+
+| Operation         | Supported |
+|-------------------|-----------|
+| Download          | Yes       |
+| Upload            | No        |
+| Digest Processing | No        |
```
jakobmoellerdev: I thought HELM Repos support digest processing through the index.yaml. Could you double check?

bindings/go/helm/repository/resource/resource_repository.go
matthiasbruns: https://github.com/open-component-model/ocm-project/issues/1014

website/content/docs/reference/resource-repositories.md
Skarlso: Was this concluded...? 🤔

bindings/go/helm/blob/chart_blob.go
```diff
@@ -0,0 +1,134 @@
...
+		return nil, nil, fmt.Errorf("error opening tar blob: %w", err)
+	}
+	defer func() {
+		if err := rc.Close(); err != nil {
+			slog.Warn("error closing tar blob reader", "error", err)
+		}
+	}()
+
+	// make sure to get all from pipe (won't work :D )
+	tarBytes, err := io.ReadAll(rc)
```
Skarlso: Unchecked error here.

bindings/go/helm/repository/resource/resource_repository.go
Skarlso: Correct me if I'm wrong here, but you didn't address this yet?

Skarlso (commented):
Two questions.

bindings/go/helm/internal/credentials.go
```diff
@@ -0,0 +1,32 @@
...
+
+	helmaccess "ocm.software/open-component-model/bindings/go/helm/spec/access"
+	ocicredentialsspecv1 "ocm.software/open-component-model/bindings/go/oci/spec/credentials/identity/v1"
+	"ocm.software/open-component-model/bindings/go/runtime"
+)
+
+// CredentialConsumerIdentity resolves the credential consumer identity for the
+// given helm repository URL. For OCI-based repositories (oci:// scheme) the
+// identity type is OCIRegistry; for HTTP/HTTPS repositories it is
+// HelmChartRepository. Returns nil if helmRepository is empty (e.g. for local charts).
```
piotrjanik: well, it returns error too

website/content/docs/overview/core-model.md
```diff
@@ -62,7 +62,18 @@ A key design principle of OCM is that **identity is separate from storage locati
 
 This means you can transport component versions across boundaries — from a public registry to an air-gapped environment, or between cloud providers — without changing their identity or breaking their signatures.
 
+## How Resources Are Accessed
+
+A component descriptor lists resources, but the actual artifacts (container images, Helm charts, etc.) live in external storage backends. Each resource carries an **access specification** that describes where and how to retrieve it.
```
piotrjanik: "live in external storage backends" sounds as if that's always the case, but resources can also be embedded by value(input types).

bindings/go/helm/blob/chart_blob.go
matthiasbruns: removed

bindings/go/helm/spec/access/scheme.go
```diff
@@ -0,0 +1,27 @@
+package access
+
+import (
+	v2 "ocm.software/open-component-model/bindings/go/helm/spec/access/v1"
```
jakobmoellerdev: wrong import

bindings/go/helm/blob/chart_blob.go
jakobmoellerdev: why is that? if the header is invalid the archive is corrupt now? (i actually think code is correct, just unsure regarding the conversation here)

bindings/go/helm/cmd/main.go
```diff
@@ -76,7 +77,7 @@ func main() {
 		Level: slog.LevelDebug,
 	}))
 
-	capabilities := endpoints.NewEndpoints(helminput.Scheme)
+	capabilities := endpoints.NewEndpoints(inputspec.Scheme)
```
jakobmoellerdev, line 81: could we add an issue here to drop helm as external plugin? I dont think we plan on keeping this right?

bindings/go/helm/spec/access/v1/helm.go
```diff
@@ -13,8 +13,15 @@ const (
 	LegacyTypeVersion = "v1"
 )
 
-// Helm describes the access for a helm chart.
-// This spec is aligned with ocm v1 https://github.com/open-component-model/ocm/blob/main/api/ocm/extensions/accessmethods/helm/method.go#L41
+// Helm describes the access for a helm chart hosted in a classic HTTP/HTTPS
```
jakobmoellerdev, line 16: why did you drop the reference to the old spec here?

bindings/go/helm/transformation/get_helm_chart.go
```diff
@@ -71,61 +73,57 @@ func (t *GetHelmChart) Transform(ctx context.Context, step runtime.Typed) (runti
...
-	helmAccess.HelmChart = resultData.Name
-	helmAccess.Version = resultData.Version
-	slog.InfoContext(ctx, "Successfully retrieved helm chart", "chart", resultData.Name, "version", resultData.Version)
+	// Load the chart from the written file to get the resolved name and version
+	closer, err := chartArchive.ReadCloser()
+	if err != nil {
+		return nil, fmt.Errorf("error reading chart archive: %w", err)
+	}
+	defer func(closer io.ReadCloser) {
+		_ = closer.Close()
```
jakobmoellerdev: closing error should be logged or joined right? probably can be split to a method so that its easier to join the closing error

jakobmoellerdev (commented):
Mostly small corrections, after that good for merge

matthiasbruns:
> why is that? if the header is invalid the archive is corrupt now? (i actually think code is correct, just unsure regarding the conversation here) 
@jakobmoellerdev since we rely on the header to determine what kind we are working with (chart or prov), I think failing if we cannot read the header is the correct way to go and we should error out.

bindings/go/helm/transformation/get_helm_chart.go
matthiasbruns: I am for logging, but nor erroring out - if something is wrong with the blob, it will fail later anyway. Let me add a log

bindings/go/helm/spec/access/v1/helm.go
matthiasbruns: brought it back

jakobmoellerdev (dismissed):

fabianburth (approved):

jakobmoellerdev (approved):

PR #2148, opened 2026-04-02
POC: Typed credentials exploration for Helm binding

### LINKED ISSUE #980: ADR: Typed credentials / identities
**Description**

Write an ADR for introducing a type-safe credential and consumer identity system for OCM.

This builds on the exploration in #979 and supersedes the earlier proposal in #704. It also incorporates typed consumer identity types from #800.

### Scope

The ADR should cover decisions on:

- **Typed credentials** -- `Resolver` returns `runtime.Typed` instead of `map[string]string`, each binding owns its credential spec (e.g. `OCICredentials/v1`, `RSACredentials/v1`)
- **Typed consumer identities** -- central `ConsumerIdentityTypeScheme` registry replacing scattered string constants, with alias normalization
- **Credential and identity type registration** -- both use `runtime.Scheme`, registered per binding at init time
- **Plugin contract** -- plugins declare `ProducedCredentialType`, validated at startup
- **`DirectCredentials/v1`** -- role as generic fallback during and after migration

### Open Questions to Address

1. Should `DirectCredentials/v1` always be convertible to any typed credential, or require an explicit type?
2. How do we handle credential spec versioning (e.g. `OCICredentials/v1` -> `v2`) with old configs?
3. Should K8s verification keys go through the credential graph instead of bypassing it?
4. What should the error look like when a provider returns the wrong credential type?
5. All plugins are currently builtin -- should the external plugin contract use `runtime.Typed` from the start?

### Timebox: 2 day(s)

**Done Criteria**
- [ ] Estimation of impact on end-user documentation (if applicable)
- [ ] Estimation of impact on internal technical documentation (if applicable)

matthiasbruns:
> **This PR is for exploration purposes only. The code serves as a reference implementation to validate design decisions. It is not intended to be merged.**

Explores introducing typed credentials and consumer identity types into the OCM credential system, using the Helm binding as a reference implementation. The findings inform ADR 0015 in #2149.

https://github.com/open-component-model/ocm-project/issues/980
ADR: #2149

##### How to test the changes

This is an exploration POC. To run the typed credential tests:

```bash
go test ocm.software/open-component-model/bindings/go/helm/... -run "Typed|CredentialAcceptor" -v
go test ocm.software/open-component-model/cli/integration -run "Test_TypedCredentials" -v
```

##### Verification

- [x] All existing tests pass (no regressions)
- [x] New typed credential tests pass end-to-end
- [x] CredentialAcceptor validation warnings verified


matthiasbruns:
closed but kept as reference

PR #2149, opened 2026-04-02
docs: ADR Typed credentials and consumer identity types

### LINKED ISSUE #800: Introduce Typed Consumer Identity Types
**Description**
introduce central registry of well-known consumer identity types
- that are shared by credential graph, plugin manager and plugin implementations
- this allows normalizing aliases and type fields

TBD: 
- decide on dymanic consumer identity system based on json schemas (main issue here is overlapping identities)
- rename consumer identity type to consumer type

Schedule a design session! We need to discuss how we evolve this (we probably do not want to start with the fully fledged dynamic registration solution, but we need to know how it will look like, the plan the path towards that!)

**Done Criteria**

### LINKED ISSUE #980: ADR: Typed credentials / identities

matthiasbruns:

Adds ADR 0015 proposing typed credentials and consumer identity types for the OCM credential system. The ADR proposes evolving `map[string]string` credentials into typed Go structs registered in `runtime.Scheme` registries:

- **Typed credential specs** per binding (e.g., `HelmHTTPCredentials/v1`, `OCICredentials/v1`)
- **Typed consumer identity types** with `CredentialAcceptor` validation
- **`ResolveTyped`** method added to existing `Resolver` interface for gradual migration
- **6-phase migration path** respecting the multi-module monorepo publish order
- **Backward compatibility** via `DirectCredentials/v1` fallback at every stage

Design validated by POC: #2148 (exploration code, not for merge)

Contributes: https://github.com/open-component-model/ocm-project/issues/980

##### How to test the changes

This is a documentation-only PR (ADR). No code changes.

##### Verification

- [ ] Review ADR with team in design session
- [ ] Validate migration phases are feasible with module dependency graph


docs/adr/0015_typed_credentials.md
```diff
@@ -0,0 +1,151 @@
...
+
+Typed identity structs declare which credential types they accept:
+
+```go
+type CredentialAcceptor interface {
+    AcceptedCredentialTypes() []runtime.Type
+}
+```
+
+The graph validates during ingestion that configured credential types are compatible with the identity type. Incompatible pairs produce warnings. Consumers reject credentials of the wrong type with clear errors.
```
fabianburth: I'm wondering, is `identity type -> credential type` a sufficient mapping? 
Generally, it feels intuitive - but I'm still thinking whether this might be plugin specific (oci plugin returns an `oci` identity type and can deal with a `token`, but the helm plugin also returns an `oci` identity type but can only deal with `username` and `password`).

docs/adr/0016_typed_credentials.md
```diff
@@ -0,0 +1,151 @@
...
+
+**At the plugin boundary:** When the graph resolves credentials via a repository plugin, the resolved `runtime.Typed` is converted to `map[string]string` at the HTTP transport layer before being sent to the plugin. The plugin receives a flat map.
+
+**Inside the plugin:** The plugin converts the received `map[string]string` to its typed credential struct using `FromDirectCredentials`. This gives the plugin typed access to credential fields with compile-time safety. The existing Helm input plugin (`bindings/go/helm/cmd`) demonstrates this pattern.
+
+**Type naming for plugins:** External plugin credential and identity type names must be prefixed with a reverse-domain namespace derived from the plugin identity (e.g., `com.hashicorp.vault.VaultCredentials/v1`). Builtin types use short names (e.g., `HelmHTTPCredentials/v1`). This prevents collisions between plugins and builtins. `runtime.Scheme` enforces uniqueness via duplicate registration errors.
```
fabianburth: How do we intend to differentiate between a collision and same type?

matthiasbruns:
oh I see feedback already :D give me one more focussed day - I will undraft the ADR once I am happy with it @fabianburth 
but thank you for your thoughts - that helps already <3

matthiasbruns:
## POC Validation

Branch `feat/800_typed_credentials_poc` validates the design end-to-end using Helm as the reference binding:

- Typed `HelmHTTPCredentials/v1` and `HelmChartRepositoryIdentity/v1` specs
- Graph resolves typed credentials from config via `CredentialTypeScheme`
- `CredentialAcceptor` validation at the consumer level — incompatible types rejected with errors
- Full test coverage: spec unit tests, graph integration tests, CLI integration tests
- Backward compatibility verified with old `Credentials/v1` configs
- Download library fully migrated to typed credentials internally

docs/adr/0016_typed_credentials.md
```diff
@@ -0,0 +1,151 @@
...
+type CredentialAcceptor interface {
+    AcceptedCredentialTypes() []runtime.Type
+}
+```
+
+The graph validates during ingestion that configured credential types are compatible with the identity type. Incompatible pairs produce warnings. Consumers reject credentials of the wrong type with clear errors.
+
+### Resolver Evolution
+
+The existing `Resolver` interface gains a `ResolveTyped` method that returns `runtime.Typed` instead of `map[string]string`. The graph stores credentials as `runtime.Typed` internally and resolves typed credentials from config when a `CredentialTypeScheme` is provided. `DirectCredentials/v1` serves as the fallback for old configurations.
```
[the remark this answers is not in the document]
matthiasbruns: https://github.com/open-component-model/open-component-model/pull/2149#discussion_r3050348703

docs/adr/0016_typed_credentials.md
```diff
@@ -0,0 +1,151 @@
...
+- Unversioned identity types work through `runtime.Scheme` alias resolution
+- External plugin wire format stays `map[string]string`
+
+### External Plugin Integration
+
+External plugins (separate binaries) communicate with the plugin manager via HTTP using `map[string]string` as the credential wire format. This does not change. Typed credentials integrate with external plugins as follows:
+
+**At discovery time:** The plugin manager runs each plugin binary with `capabilities` and reads the capability JSON. Each `SupportedConsumerIdentityType` in the capability spec includes an `AcceptedCredentialTypes` field — the JSON equivalent of the Go `CredentialAcceptor` interface. This allows plugins written in any language to declare the identity → credential type mapping. The composition root reads these declarations and builds the identity/credential schemes for the graph.
+
+**At the plugin boundary:** When the graph resolves credentials via a repository plugin, the resolved `runtime.Typed` is converted to `map[string]string` at the HTTP transport layer before being sent to the plugin. The plugin receives a flat map.
```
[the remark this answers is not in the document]
matthiasbruns: Well that's just one idea of how we would pass the data to the plugin via http - in the end, `Raw` does exactly that - so we send the json data to the plugin - let me rephrase it a bit

docs/adr/0016_typed_credentials.md
```diff
@@ -0,0 +1,151 @@
...
+
+- Module publish order matters — each phase must be merged and published before downstream phases.
+- Phase 2 PRs can run in parallel across bindings.
+- Phase 4 blocks on Phase 2 completion.
+- Phase 6 is the only step that removes backward compatibility.
+- Old `.ocmconfig` files work at every stage.
+
+## POC Validation
+
+Branch `feat/800_typed_credentials_poc` validates the design end-to-end using Helm as the reference binding:
```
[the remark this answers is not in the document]
matthiasbruns: I moved that out of the ADR and into this pr as a comment - the draft pr can be found here https://github.com/open-component-model/open-component-model/pull/2148

docs/adr/0016_typed_credentials.md
matthiasbruns: The collision handling is only for adding new types to the system - e.g. we have OCIImage in our bindings and a plugin wants to register OCIImage as a new type from the plugin into the system

docs/adr/0017_typed_credentials.md
```diff
@@ -0,0 +1,268 @@
...
+self-documenting system that validates credential and identity types at both configuration time and consumption time.
+
+## Context and Problem Statement
+
+The credential graph (see [ADR 0002](0002_credentials.md)) resolves credentials for consumer identities through a DAG.
+The resolution model is sound, but credentials and identities are untyped:
+
+- **Credentials are `map[string]string`** — key names like `username`, `password`, `accessToken` are scattered string
+  constants with no compile-time guarantees.
+  A real bug exists where OCI resource downloads used `access_token` (snake_case) while docker config resolution used
```
jakobmoellerdev, line 17: link that bug!

docs/adr/0017_typed_credentials.md
```diff
@@ -0,0 +1,268 @@
...
+    CTS -->|passed to| Graph["Credential Graph"]
+    CITS -->|passed to| Graph
+```
+
+External plugins declare their produced types in their capability spec. After plugin discovery, the composition root
+reads the capabilities and registers types into the schemes. Consumers that need typed access to plugin credentials can
+register their own Go struct for the same type — `scheme.Convert` handles the `Raw` → struct conversion.
+
+**Plugin type naming convention:** External plugin type names must be prefixed with the plugin's reverse-domain ID (
+e.g., `com.hashicorp.vault.VaultCredentials/v1`). Builtin types use short names (e.g., `HelmHTTPCredentials/v1`).
```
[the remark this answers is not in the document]
matthiasbruns: done - only on this adr though

.github/config/wordlist.txt
```diff
@@ -681,14 +683,16 @@ steffen
 stichting
 storable
 struct
+structs
+structs
```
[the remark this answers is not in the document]
matthiasbruns: done

.github/config/wordlist.txt
```diff
@@ -22,6 +22,7 @@ RGD's
 ResourceGraphDefinitions
 Taskfiles
 Testcontainers
+Testcontainers
```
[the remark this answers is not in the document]
matthiasbruns: done

docs/adr/0016_typed_credentials.md
```diff
@@ -0,0 +1,151 @@
...
+
+### Identity → Credential Type Validation
+
+Typed identity structs declare which credential types they accept:
+
+```go
+type CredentialAcceptor interface {
+    AcceptedCredentialTypes() []runtime.Type
+}
+```
```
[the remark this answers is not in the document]
matthiasbruns: well, we could also load the AcceptedCredentialTypes during registration and hold them in the graph and access this or pass them from the cli during bootstrapping - I am pretty open for any of those three

docs/adr/0017_typed_credentials.md
```diff
@@ -0,0 +1,268 @@
...
+config when a `CredentialTypeScheme` is provided. `DirectCredentials/v1` serves as the fallback for old configurations.
+
+Adding a method to an interface only breaks implementors (all in our codebase), not consumers. Each binding migrates
+from `Resolve` to `ResolveTyped` independently.
+
+```go
+// Pseudocode — updated Resolver interface
+type Resolver interface {
+    Resolve(ctx context.Context, identity runtime.Identity) (map[string]string, error)    // existing
+    ResolveTyped(ctx context.Context, identity runtime.Identity) (runtime.Typed, error)   // new
```
jakobmoellerdev: Could we do `Resolve[T](ctx context.Context, identity runtime.Identity) (T, error)`

now that methods with generics are a thing?

docs/adr/0017_typed_credentials.md
```diff
@@ -0,0 +1,268 @@
...
+
+Typed identity structs implement `IdentityProvider` to produce `runtime.Identity` maps for graph lookup. They provide
+structured construction and validation while remaining compatible with the graph's existing matching system.
+
+### Scheme Wiring and Graph Independence
+
+The credential graph must remain independent of binding-specific types. It accepts two optional schemes via its
+configuration:
+
+- **`CredentialTypeScheme`** — knows how to create typed credential objects (e.g., `HelmHTTPCredentials/v1`)
```
jakobmoellerdev: Im uncertain on this. Shouldnt Credential Plugins offer these types instead of them being separately passed via schemes?

docs/adr/0017_typed_credentials.md
```diff
@@ -0,0 +1,268 @@
...
+    CTS -->|passed to| Graph["Credential Graph"]
+    CITS -->|passed to| Graph
+```
+
+External plugins declare their produced types in their capability spec. After plugin discovery, the composition root
+reads the capabilities and registers types into the schemes. Consumers that need typed access to plugin credentials can
+register their own Go struct for the same type — `scheme.Convert` handles the `Raw` → struct conversion.
+
+**Plugin type naming convention:** External plugin type names must be prefixed with the plugin's reverse-domain ID (
+e.g., `com.hashicorp.vault.VaultCredentials/v1`). Built-in types use short names (e.g., `HelmHTTPCredentials/v1`).
```
jakobmoellerdev: This is exactly the opposite definition of how we do it in config. We also say that conflicting Names within a subsystem (i.e. ConsumerIdentity) is an error case so far.

docs/adr/0017_typed_credentials.md
```diff
@@ -0,0 +1,268 @@
...
+
+Why enforce this? Without namespace prefixes, nothing prevents two independent plugins from registering the same type
+name — or a plugin from shadowing a built-in type. For example, a malicious or misconfigured plugin could register
```
jakobmoellerdev: a malicious plugin should have been caught long before and a misconfigured plugin for me is a buggy / malformed state

docs/adr/0017_typed_credentials.md
```diff
@@ -0,0 +1,268 @@
...
+- Built-in types are registered as Go structs, external plugin types as `runtime.Raw` — consumers use `scheme.Convert` to
+  get typed structs
+- Built-ins register first (at startup), plugins register after (at discovery) — built-ins always take precedence
+- Both schemes are optional (nil-safe) — the graph degrades to `DirectCredentials` behavior when no scheme is provided
+
+### Backward Compatibility
+
+- `.ocmconfig` format is unchanged — `Credentials/v1` with `properties` continues to work
+- `DirectCredentials/v1` is the universal fallback, registered with all aliases
+- Each typed credential provides a `FromDirectCredentials` converter that constructs the typed struct from a
```
jakobmoellerdev: Is this a must implements? When would it get called?

docs/adr/0017_typed_credentials.md
```diff
@@ -0,0 +1,268 @@
...
+  `map[string]string`
+- Unversioned identity types work through `runtime.Scheme` alias resolution
+- External plugin wire format stays `map[string]string`
+
+### External Plugin Integration
+
+External plugins (separate binaries) communicate with the plugin manager via HTTP using `map[string]string` as the
+credential wire format. This does not change. Typed credentials integrate with external plugins as follows:
```
jakobmoellerdev: Looking at this the conversion to the direct credentials probably needs to be moved to the central interface as it is required for on the wire serialization. Im wondering why our runtime.Typed dont just implement json marshal and then we can use that over the wire

docs/adr/0017_typed_credentials.md
```diff
@@ -0,0 +1,268 @@
...
+```go
+// Pseudocode — plugin-side credential conversion
+// The plugin receives credentials as map[string]string over HTTP.
+// FromDirectCredentials maps the well-known keys to typed struct fields.
+creds, err := HelmHTTPCredentials.FromDirectCredentials(credentials)
+// creds.CertFile, creds.KeyFile, creds.Keyring are now typed fields
+```
+
+**Type naming for plugins:** External plugin credential and identity type names must be prefixed with a reverse-domain
+namespace derived from the plugin identity (e.g., `com.hashicorp.vault.VaultCredentials/v1`). Built-in types use short
```
jakobmoellerdev: This problem exists globally and is an error case I think, see above. I believe since you anyhow enforce uniqueness the reverse-domain prefix actually doesnt help you here right?

jakobmoellerdev (commented):
Well done. The concept is very solid and the interfaces are very clear. great job on that.

docs/adr/0017_typed_credentials.md
matthiasbruns: Well, https://github.com/golang/go/milestone/408 is still in progress isn't it?
Or what are you imagining? 
My main goal with this ADR is to have as few breaking parts as possible until the migration is fully done - changing the signature of `Resolve` will break credentials in the first phase, wouldn't it?

docs/adr/0017_typed_credentials.md
matthiasbruns: They **do** offer the types - e.g. helm offers helm related credential and identity types. But since I want the graph to handle validation, it needs to know about it somehow, that's why we may register such types in the graph with `CredentialTypeScheme` and `ConsumerIdentityTypeScheme` (both in the graph)

docs/adr/0017_typed_credentials.md
matthiasbruns: I was inspired by what e.g. Jenkins does with their plugins. The last thing I want is that users cannot use two plugins in combination, when both of them create a conflict when they try to register their custom typed to ocm.

docs/adr/0017_typed_credentials.md
matthiasbruns: yeah - removing that part. this adds nothing to the credential changes

docs/adr/0017_typed_credentials.md
matthiasbruns: makes more sense - let's use our json conversion instead

docs/adr/0017_typed_credentials.md
matthiasbruns: that paragraph is doubling down on the unique name rule and that we should think about enforcing external plugins to have a unique identifier somehow that we can prevent this from happening in the first place - see https://github.com/open-component-model/open-component-model/pull/2149/changes#r3092697295 why I want this in the first place

docs/adr/0017_typed_credentials.md
matthiasbruns: https://github.com/open-component-model/ocm-project/issues/985 will be added to the adr

docs/adr/0017_typed_credentials.md
matthiasbruns: Not must-implement.
The graph never calls it. It's the fallback for old `.ocmconfig` files on `Credentials/v1`: the graph stores those as `*DirectCredentials{Properties: ...}` and a typed consumer uses   
  `FromDirectCredentials` to lift the nested map into its flat struct. Generic `scheme.Convert` can't do that lift because the JSON shapes differ.

docs/adr/0017_typed_credentials.md
matthiasbruns: gonna check if we can add a 2nd interface here

docs/adr/0017_typed_credentials.md
jakobmoellerdev: but the graph also knows about the credential plugins, so im not sure why you need to pass them explicitly, since the graph can ask the plugins directly instead of being passed a scheme right?

matthiasbruns: So instead of adding schemes, we could use `CredentialPluginProvider`->`CredentialPlugin` instead? 
Let me check this out

docs/adr/0017_typed_credentials.md
matthiasbruns: So I tested several ideas

## Validation

### Generic Interface with `ResolveAs`

```go
type Resolver[T any] interface {
	Resolve(ctx context.Context, identity runtime.Identity) (map[string]string, error)
	ResolveAs(ctx context.Context, identity runtime.Identity) (T, error)
}
```

breaks all bindings and call-sites due to the generic change in the interface and would break my migration plan. It would also simply break our generic way of providing providers as shown in the next paragraph.

### Second interface

```go
type TypedResolver[T any] interface {
	ResolveAs(ctx context.Context, identity runtime.Identity) (T, error)
}
```

looks pretty cool, but this has some changes, that I could accept during the migration. It wont work though to the fact how we provide credential resolvers in the first place (explained further down)

https://github.com/open-component-model/open-component-model/blob/593d7ce5dfb01a77485b55ef78eba4d81b4a3e8f/cli/internal/context/context.go#L136

```go
func (ctx *Context) CredentialGraph() credentials.Resolver {
	if ctx == nil {
		return nil
	}
	ctx.mu.RLock()
	defer ctx.mu.RUnlock()
	return ctx.credentialGraph
}
```

We need a second accessor here to return the generic resolver here. The problem is more about the generic nature of the interface:

The Graph doesn't know T at compile time. It resolves credentials from a DAG and returns runtime.Typed. It could implement TypedResolver[runtime.Typed].

So if we have a generic interface in the graph, how would 

```go
  helmResolver := credentials.NewTypedResolver[*helmcredsv1.HelmHTTPCredentials](graph)
```
actually work?

e.g. in the helm transformer, we would support multiple credentials, which would lead to 

```go
type GetHelmChart struct {
	Scheme *runtime.Scheme
	// ResourceConsumerIdentityProvider is used to get the consumer identity for the resource when resolving credentials.
	ResourceConsumerIdentityProvider helm.ResourceConsumerIdentityProvider
	// CredentialProviders to resolves credentials.
	HelmHTTPCredentialsProvider credentials.TypedResolver[*helmcredsv1.HelmHTTPCredentials]
	GenericCredentialsProvider  credentials.TypedResolver[*credentialsconfigv1.DirectCredentials]
}
```
during creation, we would now have to provide strong-typed objects that match the typing above - but we dont really have a generic way to get those imho.
```go
	getHelmChart := &helmtransformer.GetHelmChart{
		Scheme:                           transformerScheme,
		ResourceConsumerIdentityProvider: &helmaccess.HelmAccess{},
		HelmHTTPCredentialsProvider:               ???,
		GenericCredentialsProvider: ???
	}
```

In this approach, we would lose the generic nature of our interfaces compeltely.

### Alternative idea

Add a generic helper method instead of generic interfaces (pointer handling etc not done here - just for demo purpose)

```go
func ResolveAs[T runtime.Typed](ctx context.Context, r TypedResolver, identity runtime.Identity) (T, error) {
	var zero T
	typed, err := r.ResolveTyped(ctx, identity)
	if err != nil {
		return zero, err
	}
	out, ok := typed.(T)
	if !ok {
		return zero, fmt.Errorf("credential type mismatch: want %T, got %T (%s): %w", zero, typed, typed.GetType(), ErrUnknown)
	}
	return out, nil
}
```

With this we would

1. not have to update context.CredentialGraph - and all callsides to provide two providers e.g. in [cli](https://github.com/open-component-model/open-component-model/blob/5bf696d9b373eb0a2f650e11f5ed71e525a00692/cli/cmd/add/component-version/cmd.go#L255) 
2. we would keep the evaluated migration path and still get generic support
3. the consumer - in this case helm, could ask either for `runtime.Typed` or do

```go
helmHTTPCreds, err = credentials.ResolveAs[*helmcredsv1.HelmHTTPCredentials](ctx, t.CredentialProvider, consumerId)
```
This would work as an addition, not as a replacement and I am happy providing such util function.

## Result

- Keep one interface and add a second method to it as proposed
- Do not introduce a generic interface
- Add a util func with generic support for direct credential conversation

**Why not two interfaces**: We tested it. Changing Helm's field from `Resolver` to `TypedResolver` immediately broke builder.go, which cascades through CLI context, and every intermediate layer that threads a resolver. Each layer would need to carry and pass both interfaces. With a single `Resolver` that has both methods, intermediate layers (context, builder, transformers) are untouched — each binding only  changes its internal call from `Resolve()` to `ResolveTyped()` or `ResolveAs[T]()`.                                                                                                                            
                                                                                                                                                                                                         
**Why not a generic interface**: `TypedResolver[T any]` — the Graph returns `runtime.Typed` (type-erased), so it can only implement `TypedResolver[runtime.Typed]`. Go generics are invariant, so consumers wanting `TypedResolver[*HelmHTTPCredentials]` can't accept it. The generic parameter adds nothing.
                                                                                                                                                                                                           
**What works:** one interface + `ResolveAs[T]` helper func                                                                                                             
  - `Resolver` keeps both `Resolve` and `ResolveTyped` — zero signature changes at any intermediate layer                                                                                                        
  - `ResolveAs[T]` gives compile-time type safety at call sites via Go generic functions                                                                                                                   
  - Phase 2 migration per binding: replace `resolver.ResolveTyped()` + manual type switch with `credentials.ResolveAs[*MyCredType](ctx, resolver, id)` — one line, no field changes, no caller changes         
  - Phase 6: remove Resolve from the interface 
  
All of this was validated in the poc - the final result have been pushed there and the adr was updated to add optional support for generic type lookup with `ResolveAs`

docs/adr/0017_typed_credentials.md
matthiasbruns: Ok I don't think this is doable / favorable

When the graph sees type e.g. `HelmHTTPCredentials/v1` in the config, it doesn't call a plugin at all — there's no plugin to call. The user typed their credentials directly in the config file. The graph just needs to turn that JSON into a Go struct. That's a `runtime.Scheme` operation (type registry + deserialization).                                                                                          
                                                                                                                                                                                                         
The `CredentialPlugin` is only involved when the credential type references something that needs external resolution (Vault, AWS, etc.). For inline typed credentials, plugins are not in the picture. 

I wouldn't want to mix that up - so we should offer one way to provide typed.Credentials and typed.Identities, not one for internal bindings and one for plugins. This approach has benefits:

- less conceptional load. we wouldn't have to think about "is this internal" / "is this external"
- one way to register instead of two - simpler to follow and debug

docs/adr/0016_typed_credentials.md
```diff
@@ -0,0 +1,151 @@
...
+
+The graph validates during ingestion that configured credential types are compatible with the identity type. Incompatible pairs produce warnings. Consumers reject credentials of the wrong type with clear errors.
```
matthiasbruns: it now supports multiple credentials 
also as in the helm example in https://github.com/open-component-model/open-component-model/pull/2148/changes/e7af42bb08f1495b5db4aa837a71a561493dae20#diff-781fcd37a681fafd4c4b6709cd06d216ced51b189668bd41f877a5fffa5e368dR157 the system supports multiple credentials per identity

docs/adr/0017_typed_credentials.md
```diff
@@ -0,0 +1,259 @@
...
+
+```go
+type CredentialAcceptor interface {
+AcceptedCredentialTypes() []runtime.Type
+}
+```
+
+The graph validates during ingestion that configured credential types are compatible with the identity type.
+Incompatible pairs produce warnings. Consumers reject credentials of the wrong type with clear errors.
+
```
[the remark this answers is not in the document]
matthiasbruns: I want to only have warnings. we dont know what the consumers do internally and I do not want to block the flow if they have some internal wiring going on that handles those cases - consumer can still reject types at any time

matthiasbruns:
> Since you did not mention any adjustment to the credential matching mechanism in the ADR, I tried to skim through your prototype. I think `validateConsumerIdentityTypes` would not be sufficient to solve our current issues with matching on type aliases (mentioned it here).

This should now also have been tackled

docs/adr/0017_typed_credentials.md
jakobmoellerdev: I dont like a wrapper method for the sole reason of it extending the call stack unnecessarily. It is one of the main design mistakes in OCM v1 since we added helper on top of helper on top of helper. The fix must be to fix the underlying interfaces / original methods.

I think your original proposal

```
type Resolver interface {
  ResolveTyped(ctx context.Context, identity runtime.Identity) (runtime.Typed, error)
}
```

is good with go1.26 semantics.

docs/adr/0017_typed_credentials.md
jakobmoellerdev:
```
Each typed credential provides a `FromDirectCredentials` converter that constructs the typed struct
```

if this is not implemented, it cannot work I believe. thats why I was asking for Must implement.

docs/adr/0017_typed_credentials.md
jakobmoellerdev: I agree the CredentialPlugin is not the right fit because it deals with other typing structures

I think we are missing a type of abstraction here. 

Assume you have a Plugin for Signature Handling. In this case the Signature Handler exposes a set of supported Consumer Identity Types that must show up in a scheme, on top of credential types supported for these consumer identity types.

If you expect the CLI to provision the Scheme instead of the Plugins, how is a plugin ever supposed to expose that? Is the CLI expected to aggregate supported credential types manually into a scheme? Somehow it doesnt fit our current architecture from what Im seeing.

docs/adr/0017_typed_credentials.md
matthiasbruns: I think we can revisit this once 1.27 is out and will will check our codebase anyways for improvements

docs/adr/0017_typed_credentials.md
matthiasbruns: Ok, switched to the registry pattern we already use :)
Works quite well, cleans up the wiring code from cli a lot and matches our other patterns.

so builtin.go and manager.go now do the wiring of credentials

docs/adr/0017_typed_credentials.md
matthiasbruns: You're right. Updated to MUST. When ResolveTyped returns *DirectCredentials (legacy config), the consumer type-asserts and calls FromDirectCredentials to lift the properties map into the typed struct. It's not enforced by a framework interface, but it's required for backward compatibility to work.

morri-son (approved):

fabianburth (approved):

PR #2176, opened 2026-04-07
feat: transfer config adr

### LINKED ISSUE #926: Create transfer config ADR proposal
**Description**
This spike is about exploring the requirements for the transfer and creating an ADR for a transfer config which provides greater flexibility for controlling the transfer behaviour (especially also upload locations) than reference names.

**You don't have to set your mind on a `ocm config` configuration. Feel free to take some time to brainstorm another concept to configure a TGD for such scenarios.**

### Timebox: 3 day(s)

**Done Criteria**
- [ ] Create follow up issues to implement the proposed concept

fabianburth:

**Technical Story:** Design a user-facing configuration format for the
`ocm transfer cv` command that gives users fine-grained control over the
transfer behaviour — especially over resource upload locations — while
compiling down to the existing transformation specification.

fix https://github.com/open-component-model/ocm-project/issues/926

##### Verification


docs/adr/0015_transfer_configuration.md
```diff
@@ -0,0 +1,599 @@
+# Transfer Configuration
+
+* **Status**: proposed
+* **Deciders**: Fabian Burth
```
jakobmoellerdev: TSC or SIG Runtime could be deciders

docs/adr/0017_transfer_configuration.md
```diff
@@ -0,0 +1,599 @@
...
+compiles to transformation specification primitives.
+
+The CLI receives a new `--config` flag:
+
+```bash
+# Simple transfer — unchanged
+ocm transfer cv ghcr.io/src//comp:1.0.0 ghcr.io/dst
+
+# Config-based transfer
+ocm transfer cv --config transfer.yaml \
```
jakobmoellerdev, line 171: i may be missing things but from the read above it sounded like you wanted to support multiple targets. currently this would still not working because the transfer here has a positional argument to exactly one target `ghcr.io/dst` right?

docs/adr/0015_transfer_configuration.md
```diff
@@ -0,0 +1,599 @@
...
+
+All typed configs are bundled in a single file using the existing
+[generic configuration](../../bindings/go/configuration/generic/v1/spec/config.go)
+wrapper (`generic.config.ocm.software/v1`). This is the same pattern
+used for OCM resolver and HTTP client configuration:
+
+```yaml
+type: generic.config.ocm.software/v1
+configurations:
+  - type: ociImageOverwrite.transfer.config.ocm.software/v1alpha1
```
jakobmoellerdev: why do these types look so much different from our other typings? I believe we said we dont want to have groups or suffixes and that Types alone would be sufficient right?

docs/adr/0015_transfer_configuration.md
```diff
@@ -0,0 +1,603 @@
...
+Allows users to declare literal target image references for specific
+resources. This directly addresses [Limitation 1](#limitation-1-no-control-over-resource-upload-locations).
+
+```yaml
+type: ociImageOverwrite.transfer.config.ocm.software/v1alpha1
+overwrites:
+  # Resource in the root component — no referencePath needed
+  - resource:
+      name: my-pod
+    imageReference: ghcr.io/target-org/images/my-pod:1.0.0
```
matthiasbruns: While I overall like the idea, here is a downside
for every new version, the transfer config needs to be updated in this case

we also discussed partial field updated

```yaml
type: ociImageOverwrite.transfer.config.ocm.software/v1alpha1
overwrites:
  # Resource in the root component — no referencePath needed
  - resource:
      name: my-pod
    imageReference: 
      registry: ghcr.io
      path: target-org/images
```

would override the original imageReg e.g. `myrepo.org/project/my-pod:1.0.0` to `ghcr.io/target-org/images/my-pod:1.0.0` without having to update the version field on every update in the transfer config

wdyt?

docs/adr/0015_transfer_configuration.md
```diff
@@ -0,0 +1,603 @@
...
+```
+
+### Alternative: Global Conversion Flag Instead of HelmToOCIConversionOverwriteConfig
+
+`HelmToOCIConversionOverwriteConfig` bundles two concerns in one entry:
+*what* to do (convert Helm to OCI) and *where* to put it (the target
+image reference). These could be separated:
+
+1. A **global conversion policy** — a CLI flag like
+   `--convert-helm-to-oci` or a dedicated config type — enables
```
matthiasbruns: since we already have `--upload-as ociArtifact`, I wouldn't add another flag

docs/adr/0015_transfer_configuration.md
```diff
@@ -0,0 +1,603 @@
...
+2. `OCIImageOverwriteConfig` then overrides the target image reference
+   for specific resources — regardless of whether the source was
+   originally OCI or Helm. Once a Helm chart is converted to an OCI
+   artifact, it is indistinguishable from any other OCI resource from
+   the overwrite config's perspective.
+
+In this model:
+
+```yaml
+type: generic.config.ocm.software/v1
```
matthiasbruns: in that case, would we even need to define anything?
Can this be handled as a set or not-set flag somehow?
What I mean:
If `ociImageOverwrite` is set without resource filters, this could be the default behaviour for all transfers, where an OCI target is supported.

matthiasbruns (commented):
In general the ADR goes into the right direction.

What I would add is that we can generate versioned specs from such definitions which itself can be used as reference in our website.

docs/adr/0015_transfer_configuration.md
jakobmoellerdev: I believe transfer configs are similar to signing / verifying configs and they are runtime based but I agree taking over the version is a huge improvement to UX

docs/adr/0017_transfer_configuration.md
```diff
@@ -0,0 +1,599 @@
...
+overwrites:
+  # Resource in the root component — no referencePath needed
+  - resource:
+      name: my-pod
+    imageReference: ghcr.io/target-org/images/my-pod:1.0.0
+
+  # Resource in a referenced component, disambiguated by referencePath
+  - referencePath:
+      - name: db-stack
+    resource:
```
jakobmoellerdev, line 240: why do we separate out extraIdentity?

```
resource:
  name: monitoring-agent
  platform: linux/amd64
```

docs/adr/0015_transfer_configuration.md
```diff
@@ -0,0 +1,599 @@
...
+converted to OCI artifacts and uploaded to a given image reference.
+This config controls the *target location* of the conversion. A
+general "convert all Helm charts to OCI" behaviour could later be
+provided via a CLI flag or a separate config type — this config then
+overwrites the default target location for specific charts.
+
+```yaml
+type: helmToOCIConversionOverwrite.transfer.config.ocm.software/v1alpha1
+overwrites:
+  - resource:
```
jakobmoellerdev: I think an image override could be shared between OCI and Helm Charts. Why are they separate?

docs/adr/0015_transfer_configuration.md
jakobmoellerdev: Agreed, we should drop these flags and make them purely configuration based. Later on we should add something that allows to set config attributes generically by flags probably, but Out of Scope for this ADR imho

docs/adr/0017_transfer_configuration.md
```diff
@@ -0,0 +1,599 @@
...
+  evaluation surprises.
+* **Backwards compatible.** Existing flags and positional arguments
+  continue to work. Configs only override matching resources.
+* **Composable.** Multiple typed configs coexist in a single file via
+  the existing generic config wrapper. Each typed entry is
+  self-contained and independently versionable.
+
+**Cons:**
+
+* **Less powerful.** No pattern matching across resources. Each resource
```
jakobmoellerdev, line 459: not sure if this is really a con because we could easily add it once needed

docs/adr/0015_transfer_configuration.md
```diff
@@ -0,0 +1,599 @@
...
+* **Composable.** Multiple typed configs coexist in a single file via
+  the existing generic config wrapper. Each typed entry is
+  self-contained and independently versionable.
+
+**Cons:**
+
+* **Less powerful.** No pattern matching across resources. Each resource
+  must be listed explicitly. This is acceptable — we do not anticipate
+  most users needing dynamic, expression-based routing.
+* **More config types over time.** Each new use case requires a new
```
jakobmoellerdev: Is this really a downside? arguably its much easier to version off and control no?

docs/adr/0015_transfer_configuration.md
```diff
@@ -0,0 +1,599 @@
...
+  type. This is mitigated by the narrow scope of each type — they are
+  small and quick to implement.
+* **No cross-resource logic.** Cannot express "all Helm charts go to
```
jakobmoellerdev: we could do this if we unify some of these configs now probably right?

docs/adr/0017_transfer_configuration.md
fabianburth: right, the adr is now only focussed on the critical feature that we know we definitely need - could clarify this above

docs/adr/0015_transfer_configuration.md
fabianburth: i mean the camel case is different and should probably be changed, but other than that, it follows the config type naming scheme.

jakobmoellerdev: groups were never a thing in OCM outside of ocmconfig I believe. Looking at other similar things such as Signing and Verifying Config this is also not suffixed

docs/adr/0017_transfer_configuration.md
```diff
@@ -0,0 +1,599 @@
...
+  uploaded to in the target.
+* The compiler maps each overwrite to the appropriate Get → Add
+  transformation chain, embedding the literal `imageReference` in the
+  AddOCIArtifact transformation spec.
+* Resources not matched by any overwrite entry fall through to the
+  default behaviour (as determined by `--upload-as` / `--copy-resources`
+  flags).
+* Duplicate entries (same reference path + resource identity appearing
+  more than once) are rejected at parse time.
+
```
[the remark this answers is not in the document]
fabianburth: No longer relevant, HelmToOCIConversionOverwrite has been removed.

docs/adr/0015_transfer_configuration.md
fabianburth: Good catch. I also explicitly opted for reference path (which might be lengthy) instead of component name + version, so you might be able to use the same config over and over again. That would have broken my idea!

docs/adr/0015_transfer_configuration.md
fabianburth: Yeah, sorry - this really wasn't about the particular flag. I wanted to explain that I want to keep the conversion logic out of the config types. 

So, what I actually wanted to propose here is the following concept: If you want to specify the target `imageReference` for a helm chart you have to:
- specify per flag (`--upload-as`) or an equivalent "global" transfer config that all helm resources shall be converted to oci (so the **conversion logic is NOT controlled through the `imageOverwrite`**)
- `imageOverwrite` index is checked every time an `AddOCIImage` transformation constructed

I thought this might keep the implementation easier because `imageOverwrite` config would really only be relevant for the `AddOCIImage` transformation - so the logic would be very narrow scoped. I diverged from this now after your comments and because the UX might be unintuitive (your overwrite would do nothing for the helm chart, if you did not specify the global flag), but might be worth a quick discussion!

docs/adr/0015_transfer_configuration.md
fabianburth: I get the idea, but I think this feels clunky. I think some dedicated global config types might be better suited.

docs/adr/0015_transfer_configuration.md
fabianburth: this --> https://github.com/open-component-model/open-component-model/pull/2176#discussion_r3086006329

docs/adr/0017_transfer_configuration.md
```diff
@@ -0,0 +1,532 @@
...
+  # Partial overwrite — only registry and path, tag preserved from source.
+  # A resource originally at myrepo.org/project/my-pod:1.0.0 would be
+  # relocated to ghcr.io/target-org/images/my-pod:1.0.0.
+  - resource:
+      name: my-pod
+    imageReference:
+      registry: ghcr.io
+      repository: target-org/images/my-pod
+
```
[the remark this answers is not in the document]
fabianburth: should be resolved now

docs/adr/0015_transfer_configuration.md
```diff
@@ -0,0 +1,532 @@
...
+ImageReferenceOverwrite
+    │ index by (referencePath + resource identity)
+    ▼
+transfer.BuildGraphDefinition(...)
+    │
+    └─ processResource (for each resource):
+         │ look up resource in overwrite index
+         │
+         ├─ match found → infer chain from source access type
+         │   emit transformation chain with imageReference
```
fabianburth: The need for this inference is what I've been trying to avoid with what I described [here](https://github.com/open-component-model/open-component-model/pull/2176#discussion_r3086006329)

jakobmoellerdev (dismissed):

matthiasbruns:
lgtm - just make the rabbit happy :)

matthiasbruns (dismissed):

matthiasbruns (approved):

jakobmoellerdev (approved):

PR #2185, opened 2026-04-08
fix(controller)!: ApplySet conflict detection

frewilhelm:

ApplySet conflict detection checked the in-memory desired object's labels instead of the live cluster state, making the check dead code. Two Deployers targeting the same resource would silently overwrite each other's `applyset.kubernetes.io/part-of` label, causing an infinite flip-flop.

- Replace the in-memory label check in `applyResource()` with a GET against the live cluster object before applying
- If the live object's `part-of` label belongs to a different ApplySet, return `ApplySetConflictError` instead of proceeding with SSA
- Add unit tests covering conflict detection (different owner, same owner, new object, unlabeled object)

This is kind of a breaking change as we needed to adjust the RBAC.

Fixes #2090

#### Testing

- [x] All existing `applyset` package tests pass unchanged
  - [x] New `TestApply_ConflictDetection` covers:
    - Live object owned by different ApplySet → conflict error
    - Live object owned by same ApplySet → no error
    - Object does not exist yet → no error
    - Live object has no `part-of` label → no error


frewilhelm:
@coderabbitai review

website/content/docs/how-to/custom-rbac.md
```diff
@@ -50,6 +50,7 @@ rules:
     verbs:
       - create
       - delete
+      - get
```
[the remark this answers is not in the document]
frewilhelm: I think your markdown formatting got messed up. Besides that, would you also suggest to create `kubectl auth can-i get` for every verb/resource in that file?

frewilhelm:
@coderabbitai review

frewilhelm:
@coderabbitai review

kubernetes/controller/internal/controller/deployer/deployer_controller.go
```diff
@@ -831,7 +831,7 @@ func (r *Reconciler) applyWithApplySet(ctx context.Context, resource *deliveryv1
 	}
 
 	logger.Info("applying ApplySet")
-	applyResult, metadata, err := set.Apply(ctx, resourcesToAdd, applyset.ApplyMode{Concurrency: runtime.NumCPU()})
```
Skarlso, line 834: Huh....? Why are you ignoring the metadata here? I think that's a mistake?

frewilhelm: `set.Apply()` only returns batch-only metadata and is used to call `PruneScope()`. I believe it should not only include the batch but check "project" wide. But I am not sure here..

Skarlso: This increases the prune scope... It would remove project wise data and not just whatever apply returns. 🤔

kubernetes/controller/internal/controller/applyset/applyset.go
```diff
@@ -342,35 +342,58 @@ func (a *ApplySet) applyResource(
 ) ApplyResultItem {
 	item := ApplyResultItem{ID: r.ID}
 
-	// Inject applyset membership label (required for prune to find managed resources)
+	// Resolve namespace for the lookup key before checking live cluster state.
+	lookupNamespace := ""
```
jakobmoellerdev: this is an inefficient way to do a lookup i believe. shouldnt we aggregate all gvks and then do a list call for all of them with the parent id?

frewilhelm: Are you suggesting to gather all resources that shall be applied before `applyResource` is run, hold in memory, and then check inside `applyResource` if the resource is already applied from another `ApplySet`?
 
I believe this would increase the TOCTOU which could potentially result in overwriting resources again. Do you think the risk is real and acceptable?

kubernetes/controller/internal/controller/deployer/deployer_controller.go
frewilhelm: Imagine a `Deployer` applies a `configmap` and an `rgd`. Now you change the `Deployer` to only apply the `rgd`. The `metadata` returned from `Apply()` would only contain the `rgd` and would only look to prune the `rgd`. The `metadata` from `set.Project(resourcesToAdd)` unions with "parent annotations, so the `configmap` would still be available for pruning (if necessary).

Skarlso: What's the purpose of returning the metadata at all then if you aren't going to use it?

kubernetes/controller/internal/controller/applyset/applyset.go
jakobmoellerdev: TOCTOU risk is not a problem with apply sets because theyre idempotent mostly. There are some smaller gaps but I think for our design probably neglible

jakobmoellerdev (commented):
As discussed in Warroom, this probably needs a revisited optimized list based on project meta instead of individual get requests

kubernetes/controller/internal/controller/deployer/deployer_controller.go
frewilhelm: I don't know right now :D @matthiasbruns can you help me on this?

jakobmoellerdev: please read up on apply sets first @frewilhelm you should understand apply sets and make an opinion by reading the code

frewilhelm: I just reverted this and we will tackle this in another issue/pr

kubernetes/controller/internal/controller/applyset/applyset.go
```diff
@@ -417,6 +436,68 @@ func (a *ApplySet) applyResource(
...
+			ns = a.resolveNamespace(e.resource.Object.GetNamespace())
+		}
+		tasks.Insert(listTask{gvk: e.resource.Object.GroupVersionKind(), namespace: ns})
+	}
+
+	conflicts := make(map[conflictKey]string)
+
+	for task := range tasks {
+		list := &unstructured.UnstructuredList{}
+		list.SetGroupVersionKind(task.gvk)
```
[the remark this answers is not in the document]
frewilhelm: If I am not mistaken then the `List` suffix gets trimmed on the `List()` call (`unstructured_client.go:L234`: `gvk.Kind = strings.TrimSuffix(gvk.Kind, "List")`). However, I will add it to comply to the convention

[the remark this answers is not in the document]
frewilhelm: For anyone who is wondering:
- [K8s conventions](https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api-conventions.md#types-kinds)
- [Implementation](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/client/unstructured_client.go#L234)

frewilhelm:
@coderabbitai

concerning your nit "326-329: Conflicts are silently recorded without failing Apply().": This is not true because the deployer_controller.go checks for errors on L839 or did I over see something?

frewilhelm:
@coderabbitai review

kubernetes/controller/internal/controller/applyset/applyset.go
```diff
@@ -464,35 +464,46 @@ func (a *ApplySet) detectConflicts(ctx context.Context, entries []applyEntry) (m
 	}
 
 	conflicts := make(map[conflictKey]string)
+	var mu sync.Mutex
 
+	eg, egCtx := errgroup.WithContext(ctx)
```
matthiasbruns, line 469: we should probably set an upper limit here if `len(tasks)` goes our of hand

jakobmoellerdev (approved):

kubernetes/controller/internal/controller/applyset/applyset.go
matthiasbruns: just talked to @frewilhelm - this should be handled already with the cpu setting at a upper level

matthiasbruns (approved):

kubernetes/controller/internal/controller/deployer/deployer_controller.go
frewilhelm: here you go https://github.com/open-component-model/open-component-model/pull/2313

kubernetes/controller/internal/controller/applyset/applyset.go
frewilhelm: will do that!

frewilhelm: so, when `set.Apply()` is called it got the argument `runtime.NumCPU()` for concurrency. This is used for the `eg.Limit` that is set for the concurrent calls for `applyResource()`. We could use the same but I am still thinking about that

PR #2230, opened 2026-04-09
feat(ctf): implement garbage collection via deletion in CTF index

chrisbleyerSAP:
Replace tag-clearing with entry deletion when tags are moved to different digests. This prevents unbounded index growth and aligns CTF behavior with OCI registry garbage collection semantics.

When a tag is moved from one digest to another, the old entry is now completely removed instead of having its tag cleared. This provides automatic garbage collection without requiring external cleanup mechanisms.

Version aliasing (multiple tags per digest) and the untagged-first pattern (tagging artifacts after creation) continue to work as expected.


jakobmoellerdev:
Closing as discussed

PR #2248, opened 2026-04-10
feat(transfer): add file buffer cleanup transformation

jakobmoellerdev:
## Summary

Implements file buffer cleanup as a terminal node in the OCM transfer transformation graph, resolving the temporary file accumulation issue described in [#854](https://github.com/open-component-model/ocm-project/issues/854).

During OCM transfers, Get transformations (GetOCIArtifact, GetLocalResource, GetHelmChart, ConvertHelmToOCI) download blobs from source repositories and buffer them to temporary files on the local filesystem. These files are consumed by subsequent Add transformations, then uploaded to the target repository — but were never cleaned up, accumulating in `/tmp` until the OS cleared them.

## Changes

- **New `FileCleanup` transformation spec** (`bindings/go/oci/spec/transformation/v1alpha1/file_cleanup.go`): Defines `FileCleanup`, `FileCleanupSpec`, and `FileCleanupOutput` types with generated DeepCopy, type accessor, and JSON schema files.
- **New `FileCleanup` transformer** (`bindings/go/oci/transformer/file_cleanup.go`): Parses `file://` URIs from the spec, removes each file from the filesystem (best-effort — skips missing files and logs warnings rather than failing), and returns a count of cleaned files.
- **Graph integration** (`bindings/go/transfer/internal/graph.go`, `file_cleanup.go`): After all resource transformations are generated, the graph builder collects file buffer references and appends a single `FileCleanup` node referencing them all.
- **Scheme and builder registration**: Registered in the OCI transformation scheme and wired into the default transfer builder.

## Implicit dependency ordering

The cleanup node depends on all Add/Convert transformations implicitly — without any explicit `dependsOn` field. File references in the cleanup spec use **consumer spec** expressions (e.g., `${addResourceId.spec.file}`) rather than producer output expressions. The CEL dependency discovery system extracts the root identifier (`addResourceId`) from these expressions and adds a DAG edge `Add → FileCleanup`, guaranteeing cleanup runs only after all consumers have finished reading the buffered files.

Concretely, the edges that form are:
- `GetLocalResource → AddLocalResource` (via `${getId.output.file}` in Add's spec)
- `AddLocalResource → FileCleanup` (via `${addId.spec.file}` in Cleanup's spec)

The same pattern applies for OCI artifacts and Helm charts (which also create a `Convert → FileCleanup` edge via `${convertId.spec.chartFile}`).

## Test plan

- [x] `go test ./bindings/go/oci/transformer/ -run FileCleanup` — transformer unit tests (cleanup, skip missing, skip empty URIs, nil spec, file removal verification, URI parsing)
- [x] `go test ./bindings/go/transfer/internal/ -run "FileCleanup|Cleanup"` — graph helper unit tests (collector functions, expression format, consumer-vs-producer refs) and graph integration tests (cleanup node present, correct CEL expressions per resource type, absent when no resources)
- [x] `go test ./bindings/go/transfer/...` — no regressions in existing transfer tests
- [x] Verify no temporary files remain in `/tmp` after a transfer (manual)


bindings/go/transfer/internal/file_cleanup.go
```diff
@@ -0,0 +1,203 @@
...
+type FileCleanupTransformation struct {
+	// +ocm:jsonschema-gen:enum=FileCleanup/v1alpha1
+	Type   runtime.Type       `json:"type"`
+	ID     string             `json:"id"`
+	Spec   *FileCleanupSpec   `json:"spec"`
+	Output *FileCleanupOutput `json:"output,omitempty"`
+}
+
+func (t *FileCleanupTransformation) GetType() runtime.Type    { return t.Type }
+func (t *FileCleanupTransformation) SetType(typ runtime.Type) { t.Type = typ }
```
fabianburth: Why no `// +ocm:typegen=true`?

bindings/go/transfer/internal/file_cleanup.go
```diff
@@ -0,0 +1,203 @@
...
+		return "", fmt.Errorf("opaque file URI %q not supported, use file:///path form", uri)
+	}
+	if parsed.Host != "" && parsed.Host != "localhost" {
+		return "", fmt.Errorf("remote file URI %q not supported, host must be empty or localhost", uri)
+	}
+	if parsed.Path == "" {
+		return "", fmt.Errorf("file URI %q has no path", uri)
+	}
+	return parsed.Path, nil
+}
```
fabianburth: Maybe we should try centrally implementing the parsing logic, so it can be shared between  this and https://github.com/open-component-model/open-component-model/blob/1062fdf73fd4724133ca0fe0c81b5a9e58ff1f39/bindings/go/blob/filesystem/blob_spec.go#L30

bindings/go/transfer/internal/file_cleanup.go
```diff
@@ -0,0 +1,203 @@
...
+			// provFile is optional; cleanup transformer skips empty URIs.
+			fmt.Sprintf("${%s.spec.?provFile}", h.convertID),
+			fmt.Sprintf("${%s.spec.file}", h.addID),
+		}
+	default: // localBlob and OCIArtifact both use spec.file on the Add node
+		return []string{
+			fmt.Sprintf("${%s.spec.file}", h.addID),
+		}
+	}
+}
```
fabianburth: I get that we need to depend on the consumers, but it's a bit hard to wrap my head around it at the moment.
Just for my understanding of the logic. If a step would produce a file buffer that is not consumed (and thus, not returned as a file buffer hint by a consumer), it would never be cleaned up. BUT because we built the graph ourselves, we can assume by construction that - if we don't have a bug - every buffered file that needs to be cleanup up is also consumed?

bindings/go/transfer/internal/file_cleanup.go
```diff
@@ -0,0 +1,203 @@
...
+	fileBufferHelm
+)
+
+// fileBufferHint records which transformation IDs are responsible for consuming
+// a file buffer. The graph emits hints — it only knows IDs and resource kinds.
+// The path-to-spec-field mapping lives here, not in graph construction.
+type fileBufferHint struct {
+	kind      fileBufferKind
+	addID     string // Add transformation ID (consumer of the file buffer)
+	convertID string // Convert transformation ID (Helm only)
```
fabianburth: seems like a typically typed field, but might be overkill here

bindings/go/transfer/internal/file_cleanup.go
fabianburth: doesn't `processResource` or the helper functions within there already know about this and could immediately return the list of file cel expressions instead of the indirection layer through the `fileBufferHint`?

fabianburth (commented):
I think this works. Some code style comments and some conceptual questions - I'm especially wondering about whether we couldn't simplify it like [this](https://github.com/open-component-model/open-component-model/pull/2248/changes#r3079172865).

bindings/go/transfer/internal/file_cleanup.go
jakobmoellerdev: correct. the graph encodes everything so if we dont know about a specific spec that can contain a file I dont think we would be able to clean it up

jakobmoellerdev: sure can do that. I think the concept is still similar.

bindings/go/transfer/internal/file_cleanup.go
jakobmoellerdev: Sure can offer a unified method 👍 then have to go into blob module I believe

bindings/go/transfer/internal/file_cleanup.go
jakobmoellerdev: Because Jakob forgot about his own generators lol

bindings/go/transfer/internal/file_cleanup.go
jakobmoellerdev: https://github.com/open-component-model/open-component-model/pull/2275

fabianburth (dismissed):

bindings/go/transfer/internal/graph.go
```diff
@@ -265,30 +275,45 @@ func logSkippedResource(ctx context.Context, component, version string, resource
...
 		}
+		return []string{fmt.Sprintf("${%s.spec.file}", addResourceID)}, nil
 	case *helmv1.Helm:
+		convertResourceID := fmt.Sprintf("%sConvert%s", id, resourceID)
 		if err := processHelm(resource, id, val, tgd, toSpec, resourceTransformIDs, i, uploadAsArtifact); err != nil {
-			return fmt.Errorf("cannot process Helm Chart resource: %w", err)
+			return nil, fmt.Errorf("cannot process Helm Chart resource: %w", err)
 		}
+		return []string{
+			fmt.Sprintf("${%s.spec.chartFile}", convertResourceID),
```
matthiasbruns, line 306: ok this is fine - but I actually thought we would have something in the graph that would deal with (temp)/file creation and handles it there.

e.g. for helm, we have a lot of temp file handling and could use the filestemconfig to "inject" temp file dirs managed by the graph and clean them up once the graph execution is done

maybe this could be a good follow up on this problem

matthiasbruns (dismissed):
lgtm, but a possible follow-up

bindings/go/transfer/internal/graph.go
jakobmoellerdev: I think you have to opinionate on the graph execution then. I purposefully did not do any "magic" outside of the transformationgraphdefinition here, because it makes it very easy to serialize into. Arguably this is just one opinion on how to build this though, happy to revisit!

matthiasbruns (approved):

bindings/go/transfer/internal/graph.go
matthiasbruns: Ok after discussing this with you in a huddle:

- my mental model was that the graph should provide all file targets for each step (temp files needed during a transformation step)
- since transformers should handle all their files by themselves, we do not need such handling in the graph
- but what we should do is make sure to provide the filesystem config for each transformation - which can be easily done by adding another param to each constructor, that the transformation steps take user config into account

matthiasbruns: @jakobmoellerdev I created a investigation / follow up to make sure filesystem config is available where needed
https://github.com/open-component-model/ocm-project/issues/1023

fabianburth (approved):

PR #2267, opened 2026-04-13
fix(GHA): serialize labeler jobs to prevent race condition

frewilhelm:
#### Summary

- The `labeler` and `conventional-commit-labeler` jobs run in parallel and race against each other: `actions/labeler` can overwrite labels just added by the conventional-commit-labeler (observed in #2250)
- Chain `labeler` after `conventional-commit-labeler` via `needs:` to ensure labels are applied sequentially

#### Details

The `actions/labeler` action appears to use a replace-style API call (`setLabels`) internally. When it reads the current labels before the conventional-commit-labeler finishes writing, it overwrites them: removing `kind/*` labels and causing `verify-labels` to fail.


jakobmoellerdev:
god thank you for taking care of this!

jakobmoellerdev (approved):

morri-son (approved):

PR #2274, opened 2026-04-13
feat: add conversion webhook to the controller

Skarlso:

Adding a conversion webhook. The main problem is figuring out a Hub version. For now, this PR introduces the webhook, but doesn't introduce a Hub version. The Hub is the canonical version that spokes convert into. It should be AT LEAST v1. If we decide alpha to be a Hub version that's a volatile version that becomes load bearing, that's not desired. 

https://github.com/open-component-model/ocm-project/issues/983

##### Verification


Skarlso:
Hm, the kustomize fix I'm going to leave for a separate PR actually. I realised that it's more involved then just deleting it.

Skarlso:
Selecting v1alpha1 for Hub for now. Once the new version comes we can implement the Conversion methods.

kubernetes/controller/api/v1alpha1/conversion.go
```diff
@@ -0,0 +1,26 @@
+package v1alpha1
+
+import (
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+func (*Component) Hub()  {}
+func (*Deployer) Hub()   {}
+func (*Repository) Hub() {}
+func (*Resource) Hub()   {}
```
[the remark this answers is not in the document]
Skarlso: We will just upgrade it when we release v1.

kubernetes/controller/Taskfile.yml
```diff
@@ -45,6 +45,11 @@ tasks:
     cmds: [ task: build ]
 
   # Development
+  # TODO: Rework CRD generation pipeline to eliminate kustomize dependency.
```
jakobmoellerdev, line 48: why dont we do this right now? if we dont, lets create an issue?

Skarlso: I want to do it in another PR. It will really mess with this one. I'm going to do it in a separate one....

I can do it here... buuut... that adds the script, deleted the kustomize, redoes the task targets... And some more things.

jakobmoellerdev: All im asking for in this case is for you to create an issue so its not lost out. its quite the effort

Skarlso: It's not that it's complicated it will just be a lot in the diff. 😆 I didn't want to pollute it.

I will create a separate issue. Something like a cleanup.

jakobmoellerdev (approved):

Skarlso:
Added followup for the cleanup of kustomization https://github.com/open-component-model/ocm-project/issues/1030.

morri-son (approved):

PR #2275, opened 2026-04-13
feat(oci)!: make GlobalAccessPolicy optional and toggleable

jakobmoellerdev:
## Summary
- Add `GlobalAccessPolicy` to `RepositoryOptions` with two modes: `Never` (disabled) and `Auto` (force global access on all local blobs if the repo supports it)
- Wire through OCI repository spec (`globalAccessPolicy` JSON field) and provider
- Fix bug: `ResourceLocalBlobOCILayout` was ignoring `EnforceGlobalAccess` (unlike `ResourceLocalBlobOCILayer`)
- Remove `GlobalAccessPolicy` from CTF spec — CTF is local storage, global access references would be invalid

## Test plan
- [x] New test: `ResourceLocalBlobOCILayout` with `EnforceGlobalAccess: true`
- [x] All existing OCI tests pass
- [x] Code generators (`task generate`) run clean
- [x] Build compiles without errors


bindings/go/oci/internal/pack/pack.go
```diff
@@ -28,6 +28,17 @@ import (
...
+// The zero value is GlobalAccessPolicyNever, suppressing global access by default.
+type GlobalAccessPolicy int
+
+const (
+	// GlobalAccessPolicyNever suppresses global access even on remote registries. This is the default.
+	GlobalAccessPolicyNever GlobalAccessPolicy = iota
+	// GlobalAccessPolicyAuto auto-detects based on whether the storage backend is globally reachable.
+	GlobalAccessPolicyAuto
+)
+
```
fabianburth: The `iota` feels to complicate the option handling here.

bindings/go/oci/spec/repository/v1/oci/repository.go
```diff
@@ -57,8 +57,28 @@ type Repository struct {
 	//     BaseUrl="ghcr.io/open-component-model/ocm" + SubPath=""
 	//     → Auto-extracts to: BaseUrl="ghcr.io", SubPath="open-component-model/ocm"
 	SubPath string `json:"subPath,omitempty"`
+	// GlobalAccessPolicy controls whether global access references are added to local blobs
+	// when adding local resources or sources.
+	// If not set (empty), global access is never added to discourage reliance on global access references.
+	// Set to "auto" to auto-detect based on the storage backend.
+	GlobalAccessPolicy GlobalAccessPolicy `json:"globalAccessPolicy,omitempty"`
```
fabianburth: I don't really feel confident about extending the oci spec with this option at this point in time. While I get the rationale, I'm not really confident if repository is the right level to control this.

bindings/go/oci/internal/pack/pack.go
jakobmoellerdev: why? for me it is very natural to model enums as iota

bindings/go/oci/spec/repository/v1/oci/repository.go
jakobmoellerdev: where else should I put it? just on the transformers?

fabianburth: Yeah, I would have offered it as an option on transfer / add

jakobmoellerdev: done, please double check

fabianburth (dismissed):

bindings/go/oci/internal/pack/pack.go
```diff
@@ -96,7 +106,7 @@ func ResourceLocalBlobOCILayer(ctx context.Context, storage content.Storage, b *
 	annotations := maps.Clone(layer.Annotations)
 	maps.Copy(annotations, opts.ManifestAnnotations)
 
-	global := backedByGlobalStore(storage) || opts.EnforceGlobalAccess
+	global := resolveGlobalAccess(opts.GlobalAccessPolicy, storage)
```
Skarlso, line 101: This is a behavior change, right...? Or do I misunderstand the code here.

Previously it was Auto, now it will be Never by default.

jakobmoellerdev: correct. this is a behavior change. I will mark it as breaking.

Skarlso: Perfect. Thanks.

Skarlso (dismissed):

Skarlso (dismissed):

fabianburth (approved):

bindings/go/oci/internal/policy/global_access.go
```diff
@@ -0,0 +1,19 @@
...
+	// GlobalAccessPolicyNever suppresses global access on all local blobs, even when the storage
+	// backend is globally reachable. This is the default policy to discourage reliance on
+	// global access references.
+	GlobalAccessPolicyNever GlobalAccessPolicy = iota
+	// GlobalAccessPolicyAuto auto-detects based on the storage backend. Global access is only
+	// added when the storage backend is globally reachable (e.g. a remote OCI registry).
+	//
+	// Experimental: This policy is carried over from OCM v1 for backwards compatibility.
+	// Its future availability is being evaluated by the community.
+	GlobalAccessPolicyAuto
```
frewilhelm, line 18: I might be missing something but these policies do not match the PR description that mention an `Always` policy

> Add GlobalAccessPolicy to RepositoryOptions with two modes: Never (disabled) and Always (force global access on all local blobs)

jakobmoellerdev: I have decided to drop always because the policy really didnt have a good use case and CTF global accesses are anyhow broken

jakobmoellerdev: adjusted the PR description to match

matthiasbruns (approved):
lgtm

PR #2276, opened 2026-04-13
feat: switch transformations to GetCredentialConsumerIdentities-based credential flow

### LINKED ISSUE #847: Switch transformations to a `GetConsumerIdentity` based credential flow
**Description**
To switch to a `GetCredentialConsumerIdentities` based credential flow that can also be implemented by plugins (which is not possible in the current transformation implementation), we have to:
- Extend transformation interface (and implementations) with `GetCredentialConsumerIdentities(ctx, step) (map[string]runtime.Identity, error)` — named slots allow a single transformer to declare multiple consumer identities
- Adjust the Transform to `Transform(ctx context.Context, step runtime.Typed, credentials map[string]map[string]string) (runtime.Typed, error)` — outer key is the slot name, inner map the resolved credentials for that slot

Caveat:
- ~~There might be transformers that need access to multiple repositories. Then, the proposed solution would not be sufficient. We need a design session (and eventually an ADR for this).~~ Resolved by the named-slot design: multi-credential transformers use distinct keys (e.g. `"source"`, `"target"`).

**Done Criteria**

morri-son:
## Summary

Resolves https://github.com/open-component-model/ocm-project/issues/847

Moves credential resolution out of individual transformers into the graph runtime, enabling external plugin transformers that can't hold Go interfaces across process boundaries.

### Changes

- Add `GetCredentialConsumerIdentities(ctx, step) (map[string]runtime.Identity, error)` to `Transformer` interface
- Update `Transform` signature to receive pre-resolved `credentials map[string]map[string]string`
- Runtime resolves credentials centrally between identity lookup and transform call
- Thread `credentials.Resolver` through `Builder` → `Graph` → `Runtime`
- Update all 8 production transformers (6 OCI, 2 Helm)
- Update spec-level interface in `transform/spec/v1alpha1`
- Wire resolver in `transfer/internal/builder.go` at builder level instead of per-transformer

### Design

- Named credential slots via `map[string]runtime.Identity` — keys like `"repository"`, `"resource"` identify purpose
- Single-cred transformers return 1 entry, multi-cred return N, no-cred return `nil`
- `credentials.ErrNotFound` swallowed during resolution (missing creds = nil value, not error)
- All types fully serializable across plugin boundary

## Test plan

- [x] 6 new subtests in `runtime_test.go` covering: single/multiple identities, nil identities, nil provider, ErrNotFound handling, error propagation
- [x] `go build ./...` passes for transform, oci, helm modules
- [x] `go test ./...` passes for transform module
- [x] Transfer module builds with local replace directives (depends on published module versions)

### DIFF
--- a/bindings/go/credentials/keys.go
+++ b/bindings/go/credentials/keys.go
@@ -1,0 +1,8 @@
+package credentials
+
+const (
+	// CredentialKeyRepository identifies credentials for a component version repository.
+	CredentialKeyRepository = "repository"
+	// CredentialKeyResource identifies credentials for a resource (e.g. OCI artifact).
+	CredentialKeyResource = "resource"
+)
--- a/bindings/go/helm/transformation/convert_helm_chart_to_oci.go
+++ b/bindings/go/helm/transformation/convert_helm_chart_to_oci.go
@@ -25,7 +25,7 @@ type ConvertHelmChartToOCI struct {
 	Scheme *runtime.Scheme
 }
 
-func (t *ConvertHelmChartToOCI) Transform(ctx context.Context, step runtime.Typed, _ map[string]map[string]string) (runtime.Typed, error) {
+func (t *ConvertHelmChartToOCI) Transform(ctx context.Context, step runtime.Typed) (runtime.Typed, error) {
 	var transformation v1alpha1.ConvertHelmToOCI
 	if err := t.Scheme.Convert(step, &transformation); err != nil {
 		return nil, fmt.Errorf("failed converting generic transformation to convert helm transformation: %w", err)
--- a/bindings/go/helm/transformation/convert_helm_chart_to_oci.go
+++ b/bindings/go/helm/transformation/convert_helm_chart_to_oci.go
@@ -25,11 +25,6 @@ type ConvertHelmChartToOCI struct {
 	Scheme *runtime.Scheme
 }
 
-func (t *ConvertHelmChartToOCI) GetCredentialConsumerIdentities(_ context.Context, _ runtime.Typed) (map[string]runtime.Identity, error) {
-	// No credentials needed — this transformer converts local files only.
-	return nil, nil
-}
-
 func (t *ConvertHelmChartToOCI) Transform(ctx context.Context, step runtime.Typed, _ map[string]map[string]string) (runtime.Typed, error) {
 	var transformation v1alpha1.ConvertHelmToOCI
 	if err := t.Scheme.Convert(step, &transformation); err != nil {
--- a/bindings/go/helm/transformation/convert_helm_chart_to_oci.go
+++ b/bindings/go/helm/transformation/convert_helm_chart_to_oci.go
@@ -26,6 +26,7 @@ type ConvertHelmChartToOCI struct {
 }
 
 func (t *ConvertHelmChartToOCI) GetCredentialConsumerIdentities(_ context.Context, _ runtime.Typed) (map[string]runtime.Identity, error) {
+	// No credentials needed — this transformer converts local files only.
 	return nil, nil
 }
 
--- a/bindings/go/helm/transformation/convert_helm_chart_to_oci.go
+++ b/bindings/go/helm/transformation/convert_helm_chart_to_oci.go
@@ -25,7 +25,11 @@ type ConvertHelmChartToOCI struct {
 	Scheme *runtime.Scheme
 }
 
-func (t *ConvertHelmChartToOCI) Transform(ctx context.Context, step runtime.Typed) (runtime.Typed, error) {
+func (t *ConvertHelmChartToOCI) GetCredentialConsumerIdentities(_ context.Context, _ runtime.Typed) (map[string]runtime.Identity, error) {
+	return nil, nil
+}
+
+func (t *ConvertHelmChartToOCI) Transform(ctx context.Context, step runtime.Typed, _ map[string]map[string]string) (runtime.Typed, error) {
 	var transformation v1alpha1.ConvertHelmToOCI
 	if err := t.Scheme.Convert(step, &transformation); err != nil {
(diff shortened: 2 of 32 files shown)


morri-son:
@coderabbitai review

bindings/go/transform/graph/runtime/runtime_test.go
```diff
@@ -95,3 +132,130 @@ func TestProcessValueEvents(t *testing.T) {
...
+		}
+		rt := &Runtime{
+			EvaluatedExpressionCache: map[string]any{},
+			EvaluatedTransformations: map[string]any{},
+			Transformers:             map[runtime.Type]Transformer{testutils.MockGetObjectV1alpha1: mock},
+			CredentialProvider:       resolver,
+		}
+		require.NoError(t, rt.ProcessValue(t.Context(), transformation))
+		require.NotNil(t, mock.gotCreds)
+		require.Nil(t, mock.gotCreds["repository"])
```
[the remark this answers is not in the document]
morri-son: fixed

bindings/go/helm/transformation/get_helm_chart.go
```diff
@@ -27,10 +25,33 @@ type GetHelmChart struct {
...
+
+	targetResource := descriptor.ConvertFromV2Resource(transformation.Spec.Resource)
+	identity, err := t.ResourceConsumerIdentityProvider.GetResourceCredentialConsumerIdentity(ctx, targetResource)
+	if err != nil {
+		return nil, fmt.Errorf("failed getting resource consumer identity for credential resolution: %w", err)
+	}
+	if identity == nil {
+		return nil, nil
+	}
+	return map[string]runtime.Identity{"resource": identity}, nil
```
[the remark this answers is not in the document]
morri-son: fixed

bindings/go/oci/transformer/doc.go
```diff
@@ -3,3 +3,10 @@
 // This package enables extraction and transformation of OCI artifacts with
 // media-type-specific handling for various content types, including Helm charts.
 package transformer
+
+const (
+	// CredentialSlotRepository identifies credentials for a component version repository.
+	CredentialSlotRepository = "repository"
```
matthiasbruns: What is a "slot"?

bindings/go/helm/transformation/convert_helm_chart_to_oci.go
```diff
@@ -25,7 +25,11 @@ type ConvertHelmChartToOCI struct {
 	Scheme *runtime.Scheme
 }
 
-func (t *ConvertHelmChartToOCI) Transform(ctx context.Context, step runtime.Typed) (runtime.Typed, error) {
+func (t *ConvertHelmChartToOCI) GetCredentialConsumerIdentities(_ context.Context, _ runtime.Typed) (map[string]runtime.Identity, error) {
+	return nil, nil
```
matthiasbruns: why is this empty?

bindings/go/oci/transformer/doc.go
morri-son: Renamed to CredentialKey* to follow the existing convention (CredentialKeyUsername, CredentialKeyPassword, etc.).

bindings/go/helm/transformation/convert_helm_chart_to_oci.go
morri-son: Required by the new Transformer interface. nil, nil signals that this transformer doesn't need credentials, as this transformer only converts local files (no network access), so there's nothing to authenticate against. GetHelmChart which sits before and AddOCIArtifact which sits after it, both require creds.

matthiasbruns: Maybe its worth thinking about an optional interface instead of forcing credentials onto every transformation in this case - @fabianburth  wdyt? this time we could prevent stuff like we have in e.g. ctf :)

bindings/go/oci/transformer/credentials.go
```diff
@@ -0,0 +1,8 @@
+package transformer
+
+const (
+	// CredentialKeyRepository identifies credentials for a component version repository.
+	CredentialKeyRepository = "repository"
+	// CredentialKeyResource identifies credentials for a resource (e.g. OCI artifact).
+	CredentialKeyResource = "resource"
```
frewilhelm, line 7: Why are these constants defined in the `oci` module. Would it make sense to put them in a shared package?

bindings/go/helm/transformation/convert_helm_chart_to_oci.go
fabianburth: Yes, discussed this with @jakobmoellerdev a few times already. I think its tech debt that we currently "force" `GetConsumerIdentity` on our other interfaces, leading to this weird credential flow we need everywhere now.

So, you are absolutely right! We might want to start doing it right here! An optional interface sounds like the intuitive approach to me.

bindings/go/credentials/keys.go
```diff
@@ -0,0 +1,8 @@
+package credentials
+
+const (
+	// CredentialKeyRepository identifies credentials for a component version repository.
+	CredentialKeyRepository = "repository"
+	// CredentialKeyResource identifies credentials for a resource (e.g. OCI artifact).
+	CredentialKeyResource = "resource"
+)
```
fabianburth, line 8: I just thought about this a bit. Tbh, I think we should just wait with this PR until we have the typed credentials and consumer identities. This `key` concept introduces all the same issues (being an implicit undocumented interface) that we currently have for credentials and consumer identities. So, we'd have to refactor this rather soon as well again.

matthiasbruns: I wouldnt mind merging this - I can add the migration to typed into the credentials epic, if that helps

matthiasbruns:
I will take over this one as discussed in the refinement yesterday

PR #2281, opened 2026-04-14
feat(controller): move configuration from internal to pkg

fabianburth:
Move internal/configuration to pkg/configuration to make the
configuration loading API available to external consumers. Add package
documentation with an API stability disclaimer. Update all internal
import paths accordingly.

##### Verification


kubernetes/controller/pkg/configuration/doc.go
```diff
@@ -0,0 +1,12 @@
+// Package configuration provides functions for loading and merging OCM
+// configurations from Kubernetes Secrets and ConfigMaps.
+//
+// It supports native OCM config entries as well as Docker-style registry
+// credentials, flattens multiple configuration sources into a single
+// [Configuration] with a content-addressable hash, and enforces a strict
```
matthiasbruns: could you go a bit more into detail with this one? 
I think we should make sure it's clear that items are dropped and what items are whitelisted by default or at least where to find them

matthiasbruns (dismissed):

kubernetes/controller/pkg/configuration/doc.go
```diff
@@ -0,0 +1,13 @@
+// Package configuration provides functions for loading and merging OCM
+// configurations from Kubernetes Secrets and ConfigMaps.
+//
+// It supports native OCM config entries as well as docker-style registry
+// credentials and
+//   - flattens multiple configuration sources into a single [Configuration],
+//   - calculates a content-addressable hash, and
+//   - enforces a hardedcoded strict allowlist of accepted config types (config
```
matthiasbruns:
```suggestion
//   - enforces a hardcoded strict allowlist of accepted config types (config
```

matthiasbruns (approved):

jakobmoellerdev (approved):

PR #2282, opened 2026-04-14
feat: make event subscriber buffer size configurable

morri-son:
## Problem

Subscriber channel buffer in `WorkerPool.Subscribe()` is hardcoded to 10. Under CI resource constraints, this can fill up and silently drop resolution events. No way to tune without code changes.

## Fix

- New `SubscriberBufferSize` field in `PoolOptions` (default: 10)
- New CLI flag `--resolver-subscriber-buffer-size` to configure it
- `Subscribe()` uses the configured value instead of a hardcoded constant

This enables tuning for testing (e.g. `=1` to force drops) or production (e.g. `=100` to reduce drop probability).


Skarlso (dismissed):

kubernetes/controller/internal/controller/component/component_controller.go
```diff
@@ -297,13 +297,14 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
-		// Resolution is in progress, the controller will be re-triggered via event source when resolution completes
+		// Resolution is in progress, the controller will be re-triggered via event source when resolution completes.
+		// RequeueAfter acts as a safety net in case the event notification is dropped (e.g. channel buffer full).
 		status.MarkNotReady(r.EventRecorder, component, v1alpha1.ResolutionInProgress, err.Error())
 		logger.Info("component version resolution in progress, waiting for event notification",
 			"component", component.Spec.Component,
 			"version", version)
 
-		return ctrl.Result{}, nil
+		return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
```
jakobmoellerdev: I think this should be configurable no? did you already align on that?

morri-son: good point. we didn't align on that. @Skarlso, what do you think?

Skarlso: Sure. But at that point, aren't we putting back `interval`? :)

jakobmoellerdev: I thought we already have interval on components anyways

Skarlso: ah so sorry. I thought I'm looking at resource... 🤦

kubernetes/controller/internal/controller/resource/resource_controller.go
```diff
@@ -381,11 +382,12 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
 	switch {
 	case errors.Is(err, workerpool.ErrResolutionInProgress):
-		// Resolution is in progress, the controller will be re-triggered via event source when resolution completes
+		// Resolution is in progress, the controller will be re-triggered via event source when resolution completes.
+		// RequeueAfter acts as a safety net in case the event notification is dropped (e.g. channel buffer full).
 		status.MarkNotReady(r.EventRecorder, resource, v1alpha1.ResolutionInProgress, err.Error())
 		logger.Info("reference path resolution in progress, waiting for event notification")
 
-		return ctrl.Result{}, nil
+		return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
```
matthiasbruns: either make it configurable as proposed or at least out this timeout into a const :)

kubernetes/controller/internal/resolution/workerpool/workerpool.go
```diff
@@ -146,7 +146,7 @@ func (wp *WorkerPool) Subscribe() <-chan []RequesterInfo {
 	wp.subscribersMu.Lock()
 	defer wp.subscribersMu.Unlock()
 
-	ch := make(chan []RequesterInfo, 10)
+	ch := make(chan []RequesterInfo, 100)
```
Skarlso: Let's start with this ( or make it configurable through the command line instead of hardcoding ).

kubernetes/controller/internal/controller/resource/resource_controller.go
```diff
@@ -382,12 +381,11 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
...
 	switch {
 	case errors.Is(err, workerpool.ErrResolutionInProgress):
-		// Resolution is in progress, the controller will be re-triggered via event source when resolution completes.
-		// RequeueAfter acts as a safety net in case the event notification is dropped (e.g. channel buffer full).
+		// Resolution is in progress, the controller will be re-triggered via event source when resolution completes
 		status.MarkNotReady(r.EventRecorder, resource, v1alpha1.ResolutionInProgress, err.Error())
 		logger.Info("reference path resolution in progress, waiting for event notification")
 
-		return ctrl.Result{RequeueAfter: 30 * time.Second}, nil
+		return ctrl.Result{}, nil
```
matthiasbruns, line 388: where did the `RequeueAfter` go? did I miss something?

kubernetes/controller/internal/controller/resource/resource_controller.go
morri-son: removed, we just go with raising the max channels

kubernetes/controller/internal/controller/resource/resource_controller.go
morri-son: we decided to just start with raising the max channels and make them configurable

kubernetes/controller/internal/resolution/workerpool/workerpool.go
morri-son: raised the default and made it configurable

matthiasbruns (dismissed):

frewilhelm:
@coderabbitai review

frewilhelm (changes requested):
we should add it for the helm values too

kubernetes/controller/chart/templates/crd/components.delivery.ocm.software.yaml
```diff
@@ -4,23 +4,8 @@ kind: CustomResourceDefinition
...
-        strategy: Webhook
-        webhook:
-            clientConfig:
-                service:
-                    name: {{ include "ocm-k8s-toolkit.resourceName" (dict "suffix" "webhook-service" "context" $) }}
-                    namespace: {{ .Release.Namespace }}
-                    path: /convert
-            conversionReviewVersions:
-                - v1
-    {{- end }}
```
frewilhelm, line 23: what are these changes about?

kubernetes/controller/cmd/main.go
```diff
@@ -109,6 +110,9 @@ func main() {
 		"This is the number of active resolver workers.")
 	flag.IntVar(&resolverWorkerQueueLength, "resolver-worker-queue-length", 1000, //nolint:mnd // no magic number
 		"The maximum number of work items in the queue for the workers to pick up component versions to resolve from.")
+	flag.IntVar(&resolverSubscriberBuffer, "resolver-subscriber-buffer-size", 10, //nolint:mnd // no magic number
```
frewilhelm, line 113: Should we add an upperbound guard?

frewilhelm (changes requested):
> Subscriber channel buffer in WorkerPool.Subscribe() is hardcoded to 10. Under CI resource constraints, this can fill up and silently drop resolution events. No way to tune without code changes.

I am missing a change that adjusts the channel buffer? Am I missing something? :D

kubernetes/controller/chart/templates/crd/components.delivery.ocm.software.yaml
morri-son: Hmmmm. Could it be that after #2274 was merged and I rebased the PR branch, executing `helm/sync-manifests` task removes it? I brought the stuff back now.

kubernetes/controller/cmd/main.go
morri-son: You mean to avoid an OOM? The flag documents the memory implication in the flag help text. I now added to reject if subscriberBufferSize > queueSize. A subscriber cannot receive more events than the queue can produce.

morri-son:
@frewilhelm 
>Subscriber channel buffer in WorkerPool.Subscribe() is hardcoded to 10. Under CI resource constraints, this can fill up and silently drop resolution events. No way to tune without code changes.
>>I am missing a change that adjusts the channel buffer? Am I missing something? :D

no, you don't miss anything, I did 😂. Corrected now. I also added the new param to the log and commented it.

kubernetes/controller/cmd/main.go
```diff
@@ -127,6 +131,12 @@ func main() {
 		os.Exit(1)
 	}
 
+	if resolverSubscriberBuffer > resolverWorkerQueueLength {
+		setupLog.Error(nil, "invalid flag value", "flag", "resolver-subscriber-buffer-size",
+			"value", resolverSubscriberBuffer, "reason", "must not exceed resolver-worker-queue-length")
+		os.Exit(1)
+	}
```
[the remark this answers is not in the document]
morri-son: corrected

kubernetes/controller/chart/templates/manager/manager.yaml
```diff
@@ -66,6 +66,9 @@ spec:
                     {{- if .workerQueueLength }}
                     - --resolver-worker-queue-length={{ .workerQueueLength }}
                     {{- end }}
+                    {{- if .subscriberBufferSize }}
```
frewilhelm:
```suggestion
                    {{- if hasKey . "subscriberBufferSize" }}
```

I believe if a user would set the value to `0` the `if .subscriberBufferSize` would be [false](https://helm.sh/docs/chart_template_guide/control_structures/#ifelse), so omitting the flag instead of setting it to `0`. Using `hasKey` instead would pass the `0` as CLI argument and we would error out.


However, this is only a nit since nobody should consider setting this to `0` (hopefully).

Skarlso (dismissed):
Looks okay. :)

Skarlso (approved):
Looks okay. :)

frewilhelm (approved):
I think that fine and does make the subscriber buffer size configurable. However, do we want to adjust the default since we think they are the root-cause for the flakes?

PR #2286, opened 2026-04-14
feat(blob-filesystem): expose FilePathFromURI with locality validation

jakobmoellerdev:
## Summary

- Extracts `FilePathFromURI` into `blob/filesystem` package so the URI-to-path logic can be shared
- Validates locality: rejects opaque URIs (`file:relative/path`), remote hosts (`file://host/...`), and empty paths; accepts `file:///path` and `file://localhost/path`
- Replaces the inline `url.Parse` block in `GetBlobFromSpec` with a call to the new function
- Fixes existing tests that used malformed `file://relative` URIs

## Related

Requested in https://github.com/open-component-model/open-component-model/pull/2248 — `transfer` module will use this once `blob` is released with the new function.

## Test plan

- [ ] `go test ./filesystem/...` passes in `bindings/go/blob`
- [ ] `TestFilePathFromURI` covers: valid absolute URI, localhost, opaque, remote host, empty path, wrong scheme
- [ ] Existing `blob_spec_test.go` tests pass with updated assertions


bindings/go/blob/filesystem/blob_spec.go
```diff
@@ -33,28 +32,19 @@ func GetBlobFromSpec(ctx context.Context, spec runtime.Typed) (blob.ReadOnlyBlob
...
-		if err != nil {
-			return nil, fmt.Errorf("cannot resolve absolute path: %w", err)
-		}
-		return GetBlobFromPath(ctx, absPath, DirOptions{
-			MediaType:    file.MediaType,
-			Reproducible: true,
-		})
-	default:
-		return nil, fmt.Errorf("unsupported URI scheme: %s", parsed.Scheme)
+	absPath, err := filepath.Abs(path)
```
fabianburth, line 40: do we need a way to consider a configured working directory here?

fabianburth (approved):

piotrjanik (approved):

PR #2293, opened 2026-04-15
chore(GHA): use arm runner

frewilhelm:
This PR migrates GitHub Actions CI/CD workflows from ubuntu-latest (x86_64) to ubuntu-24.04-arm (ARM64) runners for improved performance and cost efficiency. It also:
- Adds CI workflow self-change detection to trigger full rebuilds
- Introduces multi-architecture matrix testing for conformance and E2E jobs
- Makes the Dockerfile and Taskfile architecture-aware


morri-son (dismissed):

jakobmoellerdev:
do you mind doing an actual change in the modules or forcing a check on everything (i think the CI trigger for pull request events should be changed also if the .github file was changed) so we can see if stuff breaks?

.github/workflows/kubernetes-controller.yml
```diff
@@ -220,6 +220,7 @@ jobs:
 
   E2E:
     name: E2E Tests
+    # Loads amd64 images into Kind cluster via skopeo --override-arch amd64.
```
jakobmoellerdev: I think we can parameterize this override arch though right? should not be hard to push this into the E2E tests

.github/workflows/jsonschema.yml
```diff
@@ -19,6 +19,8 @@ permissions:
 jobs:
   lint:
     name: Lint
+    # sourcemeta/jsonschema install script uses uname -m (returns aarch64 on Linux ARM64)
```
jakobmoellerdev, line 22: this looks like a very easy thing to fix here

.github/workflows/conformance.yml
```diff
@@ -65,6 +65,7 @@ env:
 jobs:
   conformance:
     name: Sovereign Scenario
+    # Loads amd64 images into Kind cluster via skopeo --override-arch amd64.
```
jakobmoellerdev: This should be parameterized I believe

.github/workflows/jsonschema.yml
frewilhelm: yeah, but it would be on upstream https://github.com/sourcemeta/jsonschema/blob/main/install#L31

frewilhelm: i can write it into my todos but this would not be something i want to tackle right now

.github/workflows/markdown.yml
```diff
@@ -15,7 +15,7 @@ permissions:
 jobs:
   markdown-lint:
     name: Lint Markdown (website)
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-24.04-arm
```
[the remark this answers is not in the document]
frewilhelm: did you see the comments i added?

.github/workflows/jsonschema.yml
frewilhelm: we could download it manually for arm instead

frewilhelm:
@coderabbitai review

.github/workflows/kubernetes-controller.yml
frewilhelm: i hardcoded it to arm64 because its running on that runner image. We can think about making paramerized when we actually add a os/arch matrix

jakobmoellerdev (dismissed):

.github/workflows/kubernetes-controller.yml
```diff
@@ -254,7 +254,10 @@ jobs:
         uses: docker/setup-docker-action@1a6edb0ba9ac496f6850236981f15d8f9a82254d # v5
 
       - name: Setup ocm
-        uses: open-component-model/ocm-setup-action@655f3525fd283ca7d22e6b2ceec42324331df401 # v1.0.0
+        run: |
+          VERSION="$(basename "$(curl -Ls -o /dev/null -w %{url_effective} https://github.com/open-component-model/ocm/releases/latest)")"
+          curl -Ls "https://github.com/open-component-model/ocm/releases/download/${VERSION}/ocm-${VERSION#v}-linux-arm64.tar.gz" | tar xz -C /usr/local/bin ocm
+          ocm version
```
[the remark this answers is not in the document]
frewilhelm: I understand the concern but we maintain the project where we download the binary from. So this should be fine

.github/workflows/ci.yml
```diff
@@ -153,7 +157,7 @@ jobs:
           script: |
             const modules = JSON.parse(`${{ steps.discover.outputs.modules_json || '[]' }}`);
             const filtered = JSON.parse(`${{ steps.filtered.outputs.modules_json || '[]' }}`);
-            const envChanged = `${{ steps.env_changes.outputs.env || 'false' }}` === 'true';
+            const infraChanged = `${{ steps.infra_changes.outputs.infra || 'false' }}` === 'true';
```
Skarlso: How is this not just literally 
~~~
const infraChanged = `${{ steps.infra_changes.outputs.infra }}`
~~~

?

.github/workflows/ci.yml
```diff
@@ -91,26 +92,29 @@ jobs:
...
+              core.setOutput('lint_modules_json', JSON.stringify(modulesJson));
+            } else if (process.env.check_only_changed === "true") {
               console.log(`Filtering modules based on changes: ${JSON.stringify(changeJson)}`);
               const filteredModules = modulesJson.filter(module => {
                 return changeJson.some(change => change.includes(module));
               });
               console.log(`Filtered modules: ${JSON.stringify(filteredModules)}`);
               core.setOutput("modules_json", JSON.stringify(filteredModules));
-              core.setOutput('lint_modules_json', JSON.stringify(lintModulesFromEnv ?? filteredModules));
+              core.setOutput('lint_modules_json', JSON.stringify(filteredModules));
```
Skarlso: was removing lintModulesFromEnv here necessary? Aren't we now always running? I thought the purpose of lintModulesFromEnv was that we only run for changed modules. Or I'm misreading this. :)

Skarlso:
Conformance has a hardcoded platform:
```Taskfile
  PLATFORMS: linux/amd64
```

I think we need to make it dynamic like the other task targets that we have around the Task files that do stuff like:
```
GOARCH: [ "amd64", "arm64" ]
```

Or something to that end, don't we?

.github/workflows/ci.yml
frewilhelm: > I thought the purpose of lintModulesFromEnv was that we only run for changed modules. 

I don't think that how it worked.

`lintModulesFromEnv = envChanged ? modulesJson : null;` was all modules when `envChanged` was true and `null` if false. `envChanged` was true if any `.env` file changed.

frewilhelm:
I will address the remaining comments tomorrow

Skarlso (dismissed):

.github/workflows/ci.yml
frewilhelm: I made the whole logic more explicit. please check again :)

.github/workflows/ci.yml
frewilhelm: As above - please check again :)

frewilhelm:
@coderabbitai review

morri-son (approved):

matthiasbruns (approved):

PR #2296, opened 2026-04-15
fix: make sure we check the correct tag during checking oci image push

Skarlso:
##### Verification


Skarlso:
Closed in favor of https://github.com/open-component-model/open-component-model/pull/2297.

PR #2309, opened 2026-04-16
feat(website): add RSS feed support

fabianburth:
Add feed auto-discovery via <link> tag in <head> and a custom RSS template that scopes the homepage feed to blog content.

##### Verification


jakobmoellerdev (dismissed):

website/layouts/_default/rss.xml
fabianburth, line 1: what this is supposed to do is essentially point the rss feed of our landing page `ocm.software` (.IsHome) to blog, while all other sides should still point to their own rss feed.

on that note, did you know that hugo exposes an rss feed for most sections by default (e.g. ocm.software/index.xml), that's why we only have to redirect.

matthiasbruns:
isn't an rss feed descending-sorted by publish date? because https://deploy-preview-2309--ocm-website.netlify.app/index.xml isn't

website/layouts/_default/rss.xml
morri-son: uh, really? didn't now that...

fabianburth:
> isn't an rss feed descending-sorted by publish date? because https://deploy-preview-2309--ocm-website.netlify.app/index.xml isn't

Good catch, the default is weight + date. I now changed it to only consider date. Still, the question is whether we might want to remove the weight from the controller blog post, I assume then it would also be shown below on the blog page.

website/layouts/_partials/head/custom-head.html
```diff
@@ -0,0 +1,3 @@
+{{- with .OutputFormats.Get "rss" -}}
+  {{ printf `<link rel=%q type=%q href=%q title=%q>` .Rel .MediaType.Type .Permalink site.Title | safeHTML }}
+{{- end -}}
```
morri-son: The rss.xml template works correctly — homepage feed is scoped to blog content.

But custom-head.html causes a duplicate <link rel="alternate" type="application/rss+xml"> in the homepage <head>.

The Thulite SEO plugin (@thulite/seo/alternatives.html) already emits RSS discovery links for every page by iterating .AlternativeOutputFormats. The new custom-head.html then emits the same tag via .OutputFormats.Get "rss", resulting in two identical entries.

Since the SEO plugin already handles auto-discovery and the rss.xml override handles the content scoping, layouts/_partials/head/custom-head.html can be removed. I tested this locally, so I propose to remove and test again.

morri-son (approved):

jakobmoellerdev (approved):

matthiasbruns (approved):

PR #2313, opened 2026-04-16
fix(ocm-k8s-toolkit): fix pruning scope

frewilhelm:
## Summary

- Fix ApplySet prune scope bug: `applyWithApplySet` passed the batch-only metadata from `Apply()` to `Prune()` instead of the projected metadata from `Project()`. This caused orphaned resources to survive when a resource type was completely removed between versions (e.g., a Secret in v1 removed in v2 would never be pruned because `Secret` was not in the batch-only scope).
- Remove `Metadata` from `Apply()` return signature to prevent the same class of bug. The batch-only metadata was a footgun - same type as `Project()` output but narrower scope - and the variable shadowing at the call site (`applyResult, metadata, err := set.Apply(...)`) silently overwrote the correct projected `metadata`.
- Add regression test that deploys ConfigMap + Secret via OCM resource, updates to ConfigMap-only, and asserts the Secret is pruned.

## Test plan

- [x] New regression test `"prunes removed resources when the OCM resource content changes"` passes (previously failed with 30s timeout waiting for Secret deletion)
- [x] All 17 deployer controller tests pass
- [x] All applyset unit tests pass
- [x] `go vet` clean


frewilhelm:
I let the test [fail](https://github.com/open-component-model/open-component-model/actions/runs/24509498185/job/71636553555?pr=2313) on purpose to show the problem with the current pruning scope:

```
[FAILED] Timed out after 30.000s.
[kubernetes/controller:test]   Secret should have been pruned after being removed from the YAML stream, but it still exists. This indicates the prune scope is too narrow (batch-only instead of projected).
```

frewilhelm:
@coderabbitai review

frewilhelm:
@coderabbitai review

frewilhelm:
@coderabbitai review

frewilhelm:
@coderabbitai review

frewilhelm:
@coderabbitai review

frewilhelm:
@coderabbitai review

jakobmoellerdev (dismissed):

matthiasbruns (approved):
lgtm

jakobmoellerdev (approved):

PR #2316, opened 2026-04-16
feat: add replication controller ADR

### LINKED ISSUE #953: Revisit replication controller ADR
**Description**
The replication controller [ADR](https://github.com/open-component-model/open-component-model/blob/main/kubernetes/controller/docs/adr/replication.md) is outdated. We want to use the chance to learn from our mistakes and to re-imagine the replication controller.

**Note**
@jakobmoellerdev already mentioned that storing the [replication history](https://github.com/open-component-model/open-component-model/blob/main/kubernetes/controller/docs/adr/replication.md#problem-3-how-many-historical-records-does-a-single-replication-resource-need-to-keep-in-the-status) in the status of the replication CR might not have been ideal. (@jakobmoellerdev please add some context on why you think so? what alternative would you propose? a similar concept as deployment and replicaset?)

### Timebox: 2 day(s)

**Done Criteria**

Skarlso:

Fixes https://github.com/open-component-model/ocm-project/issues/953.

##### Verification


docs/adr/0017_replication_controller.md
```diff
@@ -0,0 +1,233 @@
...
+  componentRef:
+    name: podinfo-component
+    namespace: default
+
+  targetRef:
+    name: target-repository
+
+  transferConfig:
+    recursive: false
+    copyMode: localBlob     # localBlob | allResources
```
fabianburth: I assume you tried to avoid the bool flag from the cli here?

docs/adr/0017_replication_controller.md
```diff
@@ -0,0 +1,233 @@
...
+    uploadAs: default        # default | localBlob | ociArtifact
```
fabianburth:
```suggestion
```

Let's omit this from the controller API initially. So far, after presenting the new oci native local blob concept, none of our stakeholders needed this.
We should even discuss whether we want to keep supporting this for the cli.

docs/adr/0017_replication_controller.md
```diff
@@ -0,0 +1,233 @@
...
+
+  ocmConfig:
+    - name: my-ocm-config
+      kind: Secret
+      policy: Propagate
+
+  interval: 10m
```
fabianburth: Why we (initially) need an interval API here?

docs/adr/0017_replication_controller.md
```diff
@@ -0,0 +1,233 @@
...
+
+* Component CR digest differs from `status.lastTransferredDigest`.
+* Target repository does not contain the expected component version.
+* Replication CR spec changes (via `observedGeneration`).
+
+### Worker Pool
+
+Dedicated transfer worker pool, separate from resolution. Non-blocking submission.
+On completion, emits an event to retrigger the reconciler. Results cached by
+composite key (source digest + target spec hash + config hash) with TTL.
```
fabianburth: What happens if the CR gets deleted while the resolution is already scheduled?

docs/adr/0017_replication_controller.md
fabianburth: Are we generally confident, we want to expose a potentially diverging API on the CRD? Or should this be our trigger to map our flags to an existing transfer config after all?

docs/adr/0017_replication_controller.md
fabianburth: if we end up actually specifying those enums, we should include validating admission policies

Skarlso: Not so keen on this one being this much exposed though. 🤔 Are you/we sure that this is a good way to go here?

docs/adr/0017_replication_controller.md
Skarlso: Absolutely.

docs/adr/0017_replication_controller.md
Skarlso: We always had one to retransfer if there is a new version. However, now that we have watches on the component version I think this can be removed, yeah. :)

docs/adr/0017_replication_controller.md
Skarlso: you're fucked. :DDD

that said, I think that's an existing problem for other stuff that's in progress as well... 🤔  I'll analyze.

docs/adr/0017_replication_controller.md
Skarlso: Same question for me as well. I'm not sure this is a good way to expose the config. 🤔

docs/adr/0017_replication_controller.md
Skarlso: Okay, I did some deep digging. So, right now, for resolution, the worker are working with their own context. This is to prevent the reconciliation done cancelling the context. So far so good. This is okay, if the process just takes a couple minutes TOPS.

For a long running process, this is unfortunate. We can design something like... we store tha relevant context, and once the object is deleted, the finalizer goes through transfer processes that belong to this item and sends a cancel signal to the workers. This will cancel the worker, but on the receiving side, there is no way to avoid having partial or corrupted data already. I think that's fine.

I can outline this is the ADR. Does that sound okay?

Skarlso: I added this in detail.

docs/adr/0017_replication_controller.md
```diff
@@ -0,0 +1,272 @@
...
+* State sharing adds complexity.
+* Job lifecycle management is non-trivial.
+
+## Future Evolution
+
+1. **Multi-version replication**: extend `spec` with `versionConstraint`.
+2. **CRD split**: internal phases map directly to Replication/Transfer CRDs.
+3. **Transfer policies**: `replaceIfPresent`, `skipExisting`, etc.
+4. **Multiple targets**: `ReplicationSet` CRD for fan-out.
+
```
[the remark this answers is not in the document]
Skarlso: humpf. @fabianburth is the bot right? :D

docs/adr/0017_replication_controller.md
```diff
@@ -0,0 +1,272 @@
...
+  namespace: default
+spec:
+  componentRef:
+    name: podinfo-component
+    namespace: default
+
+  targetRef:
+    name: target-repository
+
+  transferConfig:
```
frewilhelm, line 64: Looking in the future and don't want to limit ourselves by introducing breaking changes later, I believe we should rather use this field to either use inline-values or a reference to a configmap or the like that will contain a "real" transfer config. Using specific fields could limit us hard in the future. Or could we have like two exclusive fields, e.g. `transferOptions` with your fields and a `transferConfig` that could would have a transfer config like @fabianburth initially designed it?

Skarlso: I updated the comment to ping the right Fabian. :) 

Also, I agree. Should either be an apiextension.JSON or a ref to a configmap.

docs/adr/0017_replication_controller.md
```diff
@@ -0,0 +1,272 @@
...
+metadata:
+  name: replicate-podinfo
+  namespace: default
+spec:
+  componentRef:
+    name: podinfo-component
+    namespace: default
+
+  targetRef:
+    name: target-repository
```
frewilhelm, line 58: I am wondering. Should this be a list if people want to transfer 1:m?

docs/adr/0017_replication_controller.md
```diff
@@ -0,0 +1,272 @@
...
+
+#### Phase 1: Plan (build TGD)
+
+```mermaid
+flowchart TD
+    A[Reconcile] --> A1[Read Component CR status]
+    A1 --> A2{Component Ready and digest present?}
+    A2 -->|No| A3[Requeue, wait for Component event]
+    A2 -->|Yes| C{Source digest == lastTransferredDigest?}
+    C -->|Yes| D{Target has matching digest?}
```
frewilhelm: What do you mean with `Target has matching digest`? How would we check this?

docs/adr/0017_replication_controller.md
```diff
@@ -0,0 +1,272 @@
...
+### Trigger Conditions
+
+* Component CR digest differs from `status.lastTransferredDigest`.
+* Target repository does not contain the expected component version.
+* Replication CR spec changes (via `observedGeneration`).
+
+### Worker Pool
+
+Dedicated transfer worker pool, separate from resolution. Non-blocking submission.
+On completion, emits an event to retrigger the reconciler. Results cached by
```
frewilhelm: What are the results that are cached?

docs/adr/0017_replication_controller.md
```diff
@@ -0,0 +1,272 @@
...
+Transient source/target errors (network, 5xx, rate limit) retry with exponential
+backoff inside the worker; terminal errors surface immediately as `Ready=False`.
+
+### Transfer Spec Storage
+
+TGDs are written to a scratch volume:
+
+* Default: `emptyDir`. TGDs regenerate cheaply on pod restart.
+* Optional: PVC for operators who need persistence across restarts.
+* Path: `/var/run/ocm/transfer-specs/{namespace}-{name}-{version}.json`.
```
frewilhelm: Do you mean the `namespace`, `name`, and `version` of the `Replication` CR? I believe this combination is not unique enough for TGDs depending on the transfer config/options and passed config (container resolvers or the like).

docs/adr/0017_replication_controller.md
```diff
@@ -0,0 +1,272 @@
...
+kind: Replication
+metadata:
+  name: replicate-podinfo
+  namespace: default
+spec:
+  componentRef:
+    name: podinfo-component
+    namespace: default
+
+  targetRef:
```
frewilhelm: Is this a repository ref to a `Repository` CR?

docs/adr/0017_replication_controller.md
```diff
@@ -0,0 +1,272 @@
...
+* Minimal complexity for users.
+* Resolution service already works.
+* Filesystem negates the limit from etcd.
+* Internal plan/execute split enables future CRD separation.
+
+### Cons
+
+* Long transfers block a worker pool slot. Mitigated by configurable pool size.
+* Filesystem TGD storage lost on pod restart. Acceptable since TGDs regenerate
+  on next reconciliation.
```
frewilhelm: I believe this is only partially true. If the component version already exists in the target repository, then we would loose the TGD as well because its not created anymore (looking at the workflow in Phase 1)

docs/adr/0017_replication_controller.md
```diff
@@ -0,0 +1,272 @@
...
+* If filesystem isn't available because of read-only disks, this does not work.
```
frewilhelm: this seems like a showstopper :D do we have any workaround for this?

docs/adr/0017_replication_controller.md
Skarlso: Ah, yes, we talked about having multiple targets.

docs/adr/0017_replication_controller.md
Skarlso: Hm, I'm going to have to think about this a bit. I imagined that since there are multiple versions we could do some matching. But thinking about this, it might be unfeasible.

docs/adr/0017_replication_controller.md
Skarlso: Ah ops,  I wanted to remove this, because I was thinking it through. There is nothing to cache really. Thanks.

docs/adr/0017_replication_controller.md
Skarlso: Good point! I thought it might be enough. But yeah, you can use different transfer configs. Ahh, it should contain something that is specific to the config. Maybe source - target digest. 🤔  But I'll have to think about this once we define the digest part above.

docs/adr/0017_replication_controller.md
Skarlso: Yes.

docs/adr/0017_replication_controller.md
Skarlso: Isn't it generated on the fly always? We agreed that we would keep it in memory for now, right?

docs/adr/0017_replication_controller.md
Skarlso: No, right? This is the problem with _this_ approach. :) I just wanted to highlight that.

docs/adr/0017_replication_controller.md
frewilhelm: If it is an OCI registry, we could create the respective manifest (with component constructor and resources if specified) and compare it to the manifest in the registry I believe

docs/adr/0017_replication_controller.md
frewilhelm: Maybe its time for another config-hash key XD

docs/adr/0017_replication_controller.md
frewilhelm: Not according to the Phase 1 diagram right now, because if "Target has matching digest?" we would "No-op" without "BuildGraphDefinition".

docs/adr/0017_replication_controller.md
frewilhelm: If there is only a RO FS I believe we should just keep everything in memory or compute it on every reconcile. Nice catch!

docs/adr/0017_replication_controller.md
Skarlso: Uuugghh :DDDDD

docs/adr/0017_replication_controller.md
Skarlso: Ugh :D Okay.

Skarlso: Yeah I re-read this and yes. I agree it's partially true.

Skarlso: I reworded it saying it's only regenerated on the next actual transfer not on the next reconcile.

docs/adr/0017_replication_controller.md
Skarlso: You know what, this whole thing is bullshit. I over thought this. Besides, any existing thing will be handled by the OCI Layering logic anyways.

docs/adr/0017_replication_controller.md
frewilhelm: I would also add a `namespace` to make it a bit clearer

docs/adr/0017_replication_controller.md
Skarlso: Redesigned this to say it's an undefined schema and/or a configmap ref that contains the actual config.

docs/adr/0017_replication_controller.md
Skarlso: Fair.

Skarlso: Added :)

docs/adr/0017_replication_controller.md
fabianburth: So my call would be to go ahead and implement one or multiple transfer config types (as indicated in my adr) that are capable of mapping our current cli flags (so never introduce dedicated API fields such as recursive and copyMode here).
We should keep this field aligned with how we want to extend the `ocmConfig` to allow inline config https://github.com/open-component-model/ocm-project/issues/869#issue-3915580454.

fabianburth (dismissed):

docs/adr/0017_replication_controller.md
Skarlso: So we talked with Frederic and actually these are rather transfer Options as depicted in the CLI instead of the config described by your ADR.

So I renamed this field to transferOptions instead. WDYT?

Skarlso:
Gimme a sec, I'll revert the multi target part.

Skarlso:
Okay, I overhauled this thing. Because the graph didn't really depict the async flow of the transfer and the worker pool integration.

I also added some notes about dangling, not described things that I only touched on. And made some further clarifications.

fabianburth (dismissed):

docs/adr/0017_replication_controller.md
```diff
@@ -0,0 +1,317 @@
...
+  namespace: default
+spec:
+  componentRef:
+    name: podinfo-component
+    namespace: default
+
+  # Ref of the `Repository` CRs that the transfer should happen to.
+  targetRepositoryRef:
+    name: target-repository
+    namespace: default
```
frewilhelm, line 59:
```suggestion
    - name: target-repository
      namespace: default
```
this should be a list right?

docs/adr/0017_replication_controller.md
```diff
@@ -0,0 +1,317 @@
...
+  # Transfer options. Single field with a ref-or-inline union, aligned with the
+  # planned OCMConfiguration extension (ocm-project#869). Either a ref:
+  transferOptions:
+    kind: ConfigMap
+    name: configmap-for-transferoptions
+  # or inline data:
+  # transferOptions:
+  #   inline:
+  #     recursive: false
+  #     copyMode: localBlob     # localBlob | allResources
```
frewilhelm: I believe we misunderstood us here. I thought it would make sense to have a `transferOption` similar to the one from the CLI:

```yaml
transferOptions:
  recursive: true
  copyMode: localBlob
```

And later we could add another field to the CR that contains a "real" transfer configuration 
```yaml
transferConfig:
  kind: ConfigMap
  name: configmap-for-transfer-configuration
OR
  inline: ...
```

But after looking at the [ADR](https://github.com/open-component-model/open-component-model/blob/main/docs/adr/0017_transfer_configuration.md#decision-outcome), I see that the transfer config would actually be part of the OCM config in general.

Accordingly, from my point of view no additional `transferConfig` would be needed.

Nevertheless, I think in `transferOptions` we should rather just offer the flags we do in the cli without offering referencing it in a configmap or such because I assume the amount of flags isn't won't be so high

docs/adr/0017_replication_controller.md
```diff
@@ -0,0 +1,317 @@
...
+* `allResources`: transfer every resource as a standalone artifact; keep external references intact.
+
+`recursive` controls whether referenced component versions are transferred alongside the root.
+
+Source and target credentials are resolved through `ocmConfig`; no separate credential fields on the CR.
+
+### Status Design
+
+```yaml
+status:
```
frewilhelm, line 115: Maybe I am out of the loop but I believed we would have a list of target repositories in the spec and then in the status we would show the transfer-status and transferred component for each repository. Did I miss something?

docs/adr/0017_replication_controller.md
```diff
@@ -0,0 +1,317 @@
...
+#### Phase 2: Execute (run TGD)
+
+Phase 2 is asynchronous. Submission returns immediately; the submitting reconcile exits and waits for the worker pool
+to emit a completion event that triggers a new reconcile, which reads the result and updates status.
+
+```mermaid
+flowchart TD
+    A[Submit TGD to worker pool] --> B{ErrTransferInProgress?}
+    B -->|Yes| C[Exit reconciliation, wait for event]
+    B -->|No, submission accepted| C
```
frewilhelm: This looks kind of odd. Do we do the same for either `Yes` and `No`?

docs/adr/0017_replication_controller.md
```diff
@@ -0,0 +1,317 @@
...
+```
+
+Both terminal branches clear `TransferInProgress=False`.
+
+Stale condition on any reconcile entry, if `TransferInProgress=True` but the worker pool has no in-flight 
+key for this CR's UID (after pod crash or leader change), the condition is treated as stale, cleared, and Phase 1 runs again.
+
+### Trigger Conditions
+
+* Component CR digest differs from `status.lastTransferredDigest`.
```
frewilhelm, line 199: I believe this is not actually a trigger for reconciliation but rather a gate if the reconciliation proceeds or not.

docs/adr/0017_replication_controller.md
Skarlso: Yes. So that's why this is called options, right? Or... what are you saying here? :D

docs/adr/0017_replication_controller.md
Skarlso: Yes, you did. :)

docs/adr/0017_replication_controller.md
frewilhelm: Does it make sense to put transferOptions into a configmap or should one just specify it on the CR?

Skarlso: Good question. Since it's rather small number of options, it doesn't really.

Skarlso: Done. :) Only appears as inlined apiextensionsv1.JSON object now. :)

docs/adr/0017_replication_controller.md
Skarlso: Ops, I think I messed that up.

docs/adr/0017_replication_controller.md
Skarlso: Actually, I realised that the graph and the old PR is still talking about writing the TGD to the filesystem.... :D

frewilhelm (dismissed):

docs/adr/0017_replication_controller.md
Skarlso: >I believe this is not actually a trigger for reconciliation but rather a gate if the reconciliation proceeds or not.

I'm not sure I get this. Any change in the component CR is basically a trigger for a reconcile. Which is why, if the digest changes in the component CR, because it reconciled a different blob, I think that should be a trigger for a new transfer, right?

frewilhelm: Okay, I misunderstood you here. I thought the trigger conditions were trigger for the reconciliation. But you meant trigger for transfer, right?

Skarlso: Correct. :)

frewilhelm (dismissed):

docs/adr/0017_replication_controller.md
```diff
@@ -0,0 +1,296 @@
...
+* Replication CR spec changes (via `observedGeneration`).
+* Interval elapsed.
+
+The controller does not do drift detection. Transfers are content-addressed at the blob level so the registry takes care
+of duplicates. Manual changes are resolved by bumping the Replication spec or waiting for the source digest to move.
+
+### Worker Pool
+
+Dedicated transfer worker pool, separate from resolution. Non-blocking submission backed by a bounded queue. If full
```
matthiasbruns:
```suggestion
Dedicated transfer worker pool, separate from resolution. Non-blocking submission backed by a bounded queue. If full, 
```

matthiasbruns (dismissed):

frewilhelm (dismissed):

docs/adr/0017_replication_controller.md
fabianburth: I think eventually, we'd want the transfer config to be able to include the cli options

fabianburth (approved):

frewilhelm (approved):

PR #2321, opened 2026-04-17
docs(community): document RSS feed availability

fabianburth:
Document the availability of an RSS feed to stay updated

##### Verification


website/content/community/community.md
```diff
@@ -55,6 +55,10 @@ You can bring anything, including:
 - Concepts you would like to work on together
 - Questions you have about the community or upcoming work
 
+### RSS Feed
+
+Stay up to date with the latest OCM blog posts and announcements by subscribing to our [RSS feed](https://ocm.software/blog/index.xml). The feed is available in any RSS reader and provides a convenient way to follow project news without needing to check the website manually.
```
fabianburth, line 60: Not sure whether we should link to the actual feed url or to our reconstructed `ocm.software/index.xml`. If we document it like this, we could eventually just revert the complex rss template and keep the default, WDYT?

morri-son: I would link to /blog/index.xml as it is self-describing. Then we can drop the custom template entirely. Users can subscribe to the blog feed directly. The /index.xml would revert to Hugo's default behavior and include all site content (docs pages, community pages, etc.). That could potentially be noisy, but that should be ok since nobody would be directed to it.

If we would like to avoid the issue with the noisy /index.xml at all, we may change the hugo.toml and remove RSS (home = ["HTML", "searchIndex"]). That would stop Hugo 
from generating /index.xml entirely and remove the <link> auto-discovery tag from the homepage. RSS readers visiting the homepage won't find any feed and have to visit /blog/ to discover it.

fabianburth: removed the custom `index.xml`

morri-son (dismissed):

frewilhelm (approved):

matthiasbruns (approved):

Skarlso (approved):

PR #2322, opened 2026-04-17
fix(website): fix install script version detection

frewilhelm:
## Summary

- Fix `install-cli.sh` picking pre-release tags (e.g. `cli/v0.4.0-rc.2`) instead of the latest stable release (`cli/v0.3.0`)
- Add a `gh auth status` check before attempting attestation verification
- Add live integration test workflow that installs a real release and verifies attestation

## Problem

The version extraction in `get_release_version()` matched any `cli/v`-prefixed tag. Since the API returns releases sorted by creation date, a pre-release like `cli/v0.4.0-rc.2` was picked over the latest stable `cli/v0.3.0`.

Additionally, `verify_binary` would fail with a confusing error when `gh` was installed but not authenticated.

## Fix

- Changed the grep filter to a strict stable-version pattern (`cli/vX.Y.Z` only - no `-rc`, no suffixes)
- Added `gh auth status` check with user-friendly warning when not authenticated

## Testing

Added `website-live-test-install-script.yml` that:

- Resolves the latest stable CLI release via `actions/github-script` (Octokit)
- Runs the real install script end-to-end with attestation verification on linux/amd64 and linux/arm64
- Verifies the installed binary version matches the resolved release
- Tests the unauthenticated path (no `gh` auth) to ensure graceful skip
- Triggers on PRs, new releases, and manual dispatch

## Test plan

- [x] Push and verify the live test workflow passes on this PR
- [x] Run `bash website/static/install-cli.sh --help` locally


jakobmoellerdev (dismissed):

frewilhelm:
I added a check to see if the user is authenticated or not which is required to check the attestation. If not an appropriate warning is returned.

IMO it is discussable if this is only a warning (that does not abort the script!). Meaning if the script is called from another script or the user oversees the warning for another reason, it would still install an unverified ocm binary

morri-son (dismissed):

morri-son (dismissed):

frewilhelm:
@coderabbitai review

frewilhelm:
@coderabbitai review

morri-son (dismissed):

morri-son:
> I added a check to see if the user is authenticated or not which is required to check the attestation. If not an appropriate warning is returned.
> 
> IMO it is discussable if this is only a warning (that does not abort the script!). Meaning if the script is called from another script or the user oversees the warning for another reason, it would still install an unverified ocm binary

The script is mainly intended to be used as installation method for an enduser. If he does not care about warnings, he gets potentially an unverified version. We may add some more information into the comments at the beginning of the script, but I would not add more safe guards during execution.

jakobmoellerdev:
> > I added a check to see if the user is authenticated or not which is required to check the attestation. If not an appropriate warning is returned.
> > IMO it is discussable if this is only a warning (that does not abort the script!). Meaning if the script is called from another script or the user oversees the warning for another reason, it would still install an unverified ocm binary
> 
> The script is mainly intended to be used as installation method for an enduser. If he does not care about warnings, he gets potentially an unverified version. We may add some more information into the comments at the beginning of the script, but I would not add more safe guards during execution.

I think the only way to get rid of the verification dependency to gh is by attaching the signature to the release artifact and verifying with this directly instead of going through gh attestations.

This is I think also a requirement to get the OpenSSF score higher

.github/workflows/website-test-install-script.yml
```diff
@@ -0,0 +1,25 @@
...
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    name: Test install-cli.sh
+    steps:
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+
+      - name: Run tests
+        run: bash website/static/install-cli-test.sh
```
jakobmoellerdev: if you already write tests, imho they should be js / github script. The test fixture is way too extensive for what it does I think.

jakobmoellerdev (dismissed):

frewilhelm:
@coderabbitai review

.github/workflows/website-test-install-script.yml
frewilhelm: I agree that shell tests are "a bit" cumbersome :D. Do you mean by `github script` that we could actually test against our releases? I kinda agree with this :/ let me see what I can do

frewilhelm: I deleted the testscript and added some tests to the workflow. Thanks a lot for the suggestion

.github/workflows/website-live-test-install-script.yml
```diff
@@ -0,0 +1,145 @@
...
+                  core.setOutput("tag", release.tag_name);
+                  core.info(`Resolved latest stable CLI release: ${release.tag_name}`);
+                  return;
+                }
+              }
+
+              if (releases.data.length < 100) break;
+            }
+
+            core.setFailed("No stable CLI release found");
```
fabianburth: don't we have smth like this in our js scripts yet?

frewilhelm: nice catch. will try it out

Skarlso (approved):

morri-son (approved):

frewilhelm:
@coderabbitai review

fabianburth (approved):

frewilhelm:
> > > I added a check to see if the user is authenticated or not which is required to check the attestation. If not an appropriate warning is returned.
> > > IMO it is discussable if this is only a warning (that does not abort the script!). Meaning if the script is called from another script or the user oversees the warning for another reason, it would still install an unverified ocm binary
> > 
> > 
> > The script is mainly intended to be used as installation method for an enduser. If he does not care about warnings, he gets potentially an unverified version. We may add some more information into the comments at the beginning of the script, but I would not add more safe guards during execution.
> 
> I think the only way to get rid of the verification dependency to gh is by attaching the signature to the release artifact and verifying with this directly instead of going through gh attestations.
> 
> This is I think also a requirement to get the OpenSSF score higher

I added https://github.com/open-component-model/open-component-model/issues/2348 as a follow up

PR #2343, opened 2026-04-20
feat: typed credential bindings

### LINKED ISSUE #704: Make Credential Graph returned typed objects instead of plain maps
**Description**

Most credentials can be fairly complex:

`OCIRegistry` for example has several fields. These fields in OCM today are undocumented. Because they are undocumented they are hard to use. We should switch the credential interfaces away from simple string maps to fully typed definitions. This allows any plugin to implement a credential plugin and return these credentials natively. While a bit more to implement than a simple consumer identity endpoint, it allows plugins to be fully type safe and control their credential behavior from E2E.

Where before the flow was: 

```mermaid
sequenceDiagram
    participant Runtime
    participant Graph
    participant Plugin

    Runtime->>Plugin: getConsumerIdentity()
    Plugin->>Runtime: runtime.Identity

    Runtime->>Graph: Resolve(runtime.Identity)
    Graph->>Runtime: map[string]string

    Runtime->>Plugin: Function(e.g. Get CV): function(..., credentials map[string]string)
```

we now have

```mermaid
sequenceDiagram
    participant Plugin
    participant Runtime
    participant Graph

    Runtime->>Plugin: getConsumerIdentity()
    Plugin->>Runtime: runtime.Identity

    Runtime->>Graph: Resolve(runtime.Identity)
    Graph<<-->>Runtime: GetCredentialPlugin(PluginManager)
    Graph-->>Plugin: Plugin.Resolve(identity runtime.Identity, creds runtime.Typed)
    Graph-->>Runtime: runtime.Typed(Credentials)

    Runtime->>Plugin: Function(e.g. Get CV): function(..., credentials runtime.Typed)
```

Note that previous credentials of type Properties still can work, but are not really type safe. Also note that custom types only work when there is a plugin registered for it.

**Done Criteria**
- [ ] ...

matthiasbruns:
**Phase 1** of the typed credentials [ADR](https://github.com/open-component-model/open-component-model/blob/main/docs/adr/0017_typed_credentials.md) 

This PR only touches the credential bindings and should not break anything.

Contributes: https://github.com/open-component-model/ocm-project/issues/704

You can see the whole picture in https://github.com/open-component-model/open-component-model/pull/2148

##### How to test the changes

Unit & integration tests only.

I ran `task init/go.work` with the updated `credentials` binding to see if the udate breaks ocm - it does not :)

##### Verification


bindings/go/credentials/ingest.go
```diff
@@ -157,26 +163,127 @@ func processRepositoryConfigurations(g *Graph, config *cfgRuntime.Config, repoTy
...
 
-		typed := v1.DirectCredentials{}
-		if err := scheme.Convert(cred, &typed); err != nil {
-			remaining = append(remaining, cred)
+		// Try the credential type scheme first (e.g. HelmCredentials/v1, OCICredentials/v1).
+		if g.credentialTypeScheme() != nil {
+			if typed, err := g.credentialTypeScheme().NewObject(cred.GetType()); err == nil {
+				if err := g.credentialTypeScheme().Convert(cred, typed); err == nil {
+					resolved = typed
+					continue
```
Skarlso: We are losing creds here. 🤔 Previously, we returned it and it was a problem, yadayada, it's described in the ADR. However, now, you continue and resolved will a single type. That's not correct, because now we are dropping certain types.

So `[Typed1, Typed2]` -> in old it was in remaning. Now it's overwritten and Typed1 is lost.

bindings/go/credentials/resolve_direct.go
```diff
@@ -56,8 +62,54 @@ func (g *Graph) resolveFromGraph(ctx context.Context, identity runtime.Identity)
...
+		return nil
+	}
+	result := make(map[string]string, len(rawAny))
+	for k, v := range rawAny {
+		if k == "type" {
+			continue // Don't leak the type field into credential properties
+		}
+		if s, ok := v.(string); ok && s != "" {
+			result[k] = s
+		}
```
Skarlso: So we intentionally drop everything that isn't string... right? Should we log that? 🤔

bindings/go/credentials/ingest.go
```diff
@@ -157,26 +163,127 @@ func processRepositoryConfigurations(g *Graph, config *cfgRuntime.Config, repoTy
...
 			return nil, nil, fmt.Errorf("credential type is empty")
 		}
 
-		typed := v1.DirectCredentials{}
-		if err := scheme.Convert(cred, &typed); err != nil {
-			remaining = append(remaining, cred)
+		// Try the credential type scheme first (e.g. HelmCredentials/v1, OCICredentials/v1).
+		if g.credentialTypeScheme() != nil {
+			if typed, err := g.credentialTypeScheme().NewObject(cred.GetType()); err == nil {
+				if err := g.credentialTypeScheme().Convert(cred, typed); err == nil {
```
Skarlso: Shouldn't we at least log this error? 🤔 Although I see it was ignored previously as well.

bindings/go/credentials/resolve_direct.go
```diff
@@ -56,8 +62,54 @@ func (g *Graph) resolveFromGraph(ctx context.Context, identity runtime.Identity)
...
+// For DirectCredentials, it returns the Properties map directly.
+// For other typed credentials (e.g. HelmHTTPCredentials), it falls back to a JSON
+// round-trip, extracting only string-valued fields (excluding the "type" field).
+// TODO(matthiasbruns): Remove once plugin interfaces migrate to runtime.Typed https://github.com/open-component-model/ocm-project/issues/980
+func typedToMap(cred runtime.Typed) map[string]string {
+	if cred == nil {
+		return nil
+	}
+	if dc, ok := cred.(*v1.DirectCredentials); ok {
+		return dc.Properties
```
Skarlso:
```suggestion
		return maps.Clone(dc.Properties)
```

bindings/go/credentials/ingest.go
```diff
@@ -157,26 +163,127 @@ func processRepositoryConfigurations(g *Graph, config *cfgRuntime.Config, repoTy
 	return nil
 }
 
-// extractDirect extracts and separates a slice of raw credentials into two groups:
-// 1. Direct credentials of type CredentialsTypeV1 (which are decoded into a merged map).
-// 2. All remaining credentials that require plugin-based resolution.
-// Returns the merged direct credentials and the slice of remaining credentials.
-func extractDirect(creds []runtime.Typed) (map[string]string, []runtime.Typed, error) {
-	direct := map[string]string{}
```
Skarlso, line 165: If I read this correctly, this now, can store empty credentials on the node. If we have a cred with empty properties that will get stored as something like `DirectCredentials{Properties: nil}`. Or something like that.

Maybe not a concern, just wanted to point it out.

bindings/go/credentials/ingest.go
```diff
@@ -157,26 +163,127 @@ func processRepositoryConfigurations(g *Graph, config *cfgRuntime.Config, repoTy
...
+		return false
+	}
+	for _, a := range accepted {
+		aObj, err := credentialTypeScheme.NewObject(a)
+		if err != nil {
+			continue
+		}
+		if reflect.TypeOf(credObj) == reflect.TypeOf(aObj) {
+			return true
+		}
```
Skarlso: So I stepped through this. If you have a type that's Raw and another with WithAllowUnknown they both are Raw and this will say they are the same, even though they aren't.

I had to write a little test to make sure I'm not halucinating. :)

```go
      scheme := runtime.NewScheme(runtime.WithAllowUnknown())
      accepted := []runtime.Type{runtime.NewVersionedType("Foo", "v1")}
      credType := runtime.NewVersionedType("Bar", "v1")
      got := isAccepted(scheme, credType, accepted)
      // got will be true even though it shouldn't
```

Skarlso (commented):
Some initial findings! :) Nothing too major.

bindings/go/credentials/interface.go
```diff
@@ -14,9 +14,13 @@ var ErrNotFound = errors.New("credentials not found")
...
-// It provides a method to resolve credentials and returns them as a map of strings.
+//
 // In case of an error it will either return ErrNotFound when no credentials could be found
 // or another error indicating the failure reason wrapped by ErrUnknown.
 type Resolver interface {
+	// TODO(matthiasbruns): Remove once all consumers use ResolveTyped https://github.com/open-component-model/ocm-project/issues/980
+	//
+	// Deprecated: Migrate to ResolveTyped instead for typed credential support
 	Resolve(ctx context.Context, identity runtime.Identity) (map[string]string, error)
+	ResolveTyped(ctx context.Context, identity runtime.Identity) (runtime.Typed, error)
```
frewilhelm: I believe this addition makes this PR a breaking change

bindings/go/credentials/resolve_direct.go
frewilhelm: In theory it can only be a `string` I believe since this is a fallback from `map[string]string`.

bindings/go/credentials/ingest.go
```diff
@@ -157,26 +163,127 @@ func processRepositoryConfigurations(g *Graph, config *cfgRuntime.Config, repoTy
...
 		}
 
-		typed := v1.DirectCredentials{}
-		if err := scheme.Convert(cred, &typed); err != nil {
-			remaining = append(remaining, cred)
+		// Try the credential type scheme first (e.g. HelmCredentials/v1, OCICredentials/v1).
+		if g.credentialTypeScheme() != nil {
+			if typed, err := g.credentialTypeScheme().NewObject(cred.GetType()); err == nil {
+				if err := g.credentialTypeScheme().Convert(cred, typed); err == nil {
+					resolved = typed
```
frewilhelm: The func-doc says `Returns the first successfully resolved typed credential.` but if any other type matches that it would overwrite `resolved`

bindings/go/credentials/ingest.go
```diff
@@ -157,26 +163,127 @@ func processRepositoryConfigurations(g *Graph, config *cfgRuntime.Config, repoTy
...
+	}
+
+	return resolved, remaining, nil
+}
+
+// validateConsumerIdentityTypes checks that all consumer identity types in the config
+// are registered in the ConsumerIdentityTypeScheme and that credential types are
+// compatible with the identity type. Unknown types produce warnings — they are not
+// rejected because plugins loaded later may introduce new types.
+func validateConsumerIdentityTypes(ctx context.Context, g *Graph, config *cfgRuntime.Config) {
```
frewilhelm, line 246: Do we need this validation or could we just "warn" when we do not find the credentials later?

I am a bit worried about the double for-loop here. I guess depending on the config this could take some time. So, I am unsure if the validation is worth it if we only display warnings and continue either way.

bindings/go/credentials/interface.go
matthiasbruns: no its not :) thats the whole idea of the ADR

bindings/go/credentials/ingest.go
matthiasbruns: how did the graph manage multiple credentials? I missed that bit, before I can provider a good answer, i might need some insights @fabianburth can you help me here?

bindings/go/credentials/ingest.go
matthiasbruns: let me change the order of things, good finding

bindings/go/credentials/interface.go
matthiasbruns: only the implementor will break, since it's in the credentials binding, that fine. other bindings importing this interface will still work using the graph, since it will also implement the old interface - this is tested in the poc already

bindings/go/credentials/ingest.go
matthiasbruns: thinking about this again, I do not feel this will take that much of cpu/ram at all - I mean how "big" can a config get that this might be a problem?

bindings/go/credentials/ingest.go
matthiasbruns: added logging

bindings/go/credentials/ingest.go
matthiasbruns: this should now being caught

bindings/go/credentials/ingest.go
matthiasbruns: should be handled

bindings/go/credentials/ingest.go
matthiasbruns: good catch - until https://github.com/open-component-model/open-component-model/pull/2343#discussion_r3123779019 is discussed, I am taking the first match and sending the others down the road

bindings/go/credentials/interface.go
frewilhelm: oki, thanks for clarifying!

bindings/go/credentials/ingest.go
frewilhelm: lets assume a user has invalid consumer identity types in his config. The validation runs and displays him some warnings, but the program continues and probably errors out at the end because we cannot find the valid consumer identity types in his config. Hence, the value of the warning and validation is zero because he gets informed about the invalid consumer identity types when the program errors out.

Lets assume a user has valid and invalid consumer identity types in his config. The validation runs and displays the warnings for the invalid ones but the program continues. When the invalid consumer identity types were not relevant and the program finishes successfully, we just produced noise with the warnings.

Does this make sense? I am probably overseeing something in the bigger picture :D

matthiasbruns: Ok now I get what you are aiming at.
The problem is, that the graph does not know for sure, if the credential <-> identity mapping can be handled by the plugin. In the end, the plugin should decide what do do with the data, the graph can only give hints, that a plugin did not register a certain type - we could lower this warning to a debug, but I woudln't want the graph to decide what to do with not-matching combinations.

matthiasbruns: talked about this with @frewilhelm in the WR - we will keep it for now and if its not needed / too costly, we either hide it behind the verbose flag or drop it

bindings/go/credentials/ingest.go
Skarlso: This has been resolved now by putting it into remaining which is what I meant. :D Thank you! :)

Skarlso (dismissed):
Largely okay with this now. :) 👍

bindings/go/credentials/type_registry.go
```diff
@@ -0,0 +1,19 @@
...
+type CredentialAcceptor interface {
+	AcceptedCredentialTypes() []runtime.Type
+}
+
+// TypeSchemeProvider provides read access to a runtime.Scheme of known types.
+// The credential graph uses this during ingestion to deserialize typed credentials,
+// resolve type aliases, and validate identity ↔ credential compatibility.
+type TypeSchemeProvider interface {
+	Scheme() *runtime.Scheme
+}
```
fabianburth: Is this interface really needed?

bindings/go/credentials/graph.go
```diff
@@ -43,21 +48,55 @@ func ToGraph(ctx context.Context, config *cfgRuntime.Config, opts Options) (Reso
...
 	repositoryConfigurationsMu sync.RWMutex    // Mutex to protect access to repository configurations
 	repositoryConfigurations   []runtime.Typed // List of repository configurations parsed
 
 	*syncedDag // The underlying DAG structure for managing dependencies
 
-	repositoryPluginProvider RepositoryPluginProvider // injection for resolving custom repository types
-	credentialPluginProvider CredentialPluginProvider // injection for resolving custom credential types
+	repositoryPluginProvider     RepositoryPluginProvider // injection for resolving custom repository types
+	credentialPluginProvider     CredentialPluginProvider // injection for resolving custom credential types
+	identityTypeSchemeProvider   TypeSchemeProvider       // validates consumer identity types from config
```
fabianburth: Is this even set or used anywhere at the moment?

bindings/go/credentials/ingest.go
```diff
@@ -157,26 +174,149 @@ func processRepositoryConfigurations(g *Graph, config *cfgRuntime.Config, repoTy
...
+	for _, consumer := range config.Consumers {
+		for _, identity := range consumer.Identities {
+			identityType, err := identity.ParseType()
+			if err != nil {
+				slog.WarnContext(ctx, "consumer identity has unparseable type",
+					"identity", identity.String(), "error", err)
+				continue
+			}
+
+			identityObj, err := g.consumerIdentityTypeScheme().NewObject(identityType)
```
fabianburth: If we have a go type, shouldn't we also try to convert to see if it actually fullfills that type?

bindings/go/credentials/ingest.go
```diff
@@ -157,26 +174,149 @@ func processRepositoryConfigurations(g *Graph, config *cfgRuntime.Config, repoTy
...
+			accepted := acceptor.AcceptedCredentialTypes()
+
+			for _, cred := range consumer.Credentials {
+				credType := cred.GetType()
+				if credType.IsEmpty() || scheme.IsRegistered(credType) {
+					// DirectCredentials/Credentials are always accepted
+					continue
+				}
+				if !isAccepted(g.credentialTypeScheme(), credType, accepted) {
+					slog.WarnContext(ctx, "credential type not accepted by identity type",
```
piotrjanik, line 281: ADR says that warning messages must be produced. I could not find any test verifying that.

bindings/go/credentials/ingest.go
```diff
@@ -157,26 +174,149 @@ func processRepositoryConfigurations(g *Graph, config *cfgRuntime.Config, repoTy
...
+				continue
+			}
+
+			// If the identity type declares accepted credential types, validate
+			// This does not fail on purpose - credentials should still be passed as the user configured them
+			acceptor, ok := identityObj.(CredentialAcceptor)
+			if !ok {
+				continue
+			}
+			accepted := acceptor.AcceptedCredentialTypes()
```
fabianburth: I'm still not completely sold on this static credential acceptor method on the identity. What was our conclusion in the ADR here?

bindings/go/credentials/ingest.go
```diff
@@ -157,26 +174,149 @@ func processRepositoryConfigurations(g *Graph, config *cfgRuntime.Config, repoTy
...
+		aObj, err := credentialTypeScheme.NewObject(a)
+		if err != nil {
+			continue
+		}
+		if reflect.TypeOf(credObj) == reflect.TypeOf(aObj) {
+			return true
+		}
 	}
-	return direct, remaining, nil
+	return false
```
fabianburth: I'm almost certain we have a simpler way to do this

bindings/go/credentials/graph.go
matthiasbruns: we will need this in the state where we wire up cli etc and register the type schemes - so if we do not want to touch the credentials 3 times, we should add this now

https://github.com/open-component-model/open-component-model/pull/2149/changes#diff-40679641fb1796269e09168430c666ce10ad913f5810caf9091357c496b0a2a0R128

I merged both of them into one interface since its basically the same 

Here is the part for the registries
https://github.com/open-component-model/open-component-model/pull/2149/changes#diff-40679641fb1796269e09168430c666ce10ad913f5810caf9091357c496b0a2a0R184

bindings/go/credentials/type_registry.go
matthiasbruns: as for the registration phase https://github.com/open-component-model/open-component-model/pull/2343#discussion_r3137593229

bindings/go/credentials/ingest.go
matthiasbruns: https://github.com/open-component-model/open-component-model/pull/2149/changes#diff-40679641fb1796269e09168430c666ce10ad913f5810caf9091357c496b0a2a0R57
but I can see if we can somehow omit this interface and provide the mapping somehow else. The ADR states this is required, but I am still open for input :)

matthiasbruns:
Open topics after the WR:
- migrate runtime.Identity to runtime.Typed in new ResolveTyped func (and update ADR)
- check again if we really need the CredentialAcceptor or if we can do this via the registry in a later phase
- check if we can move this into runtime https://github.com/open-component-model/open-component-model/pull/2343#discussion_r3137556463 or if there is already one - create a follow up issue if there is nothing like that

moving the issue back into draft until I am done

matthiasbruns:
> **Dependency:** This PR now depends on #2399 (`Scheme.ResolveCanonicalType`) which must be merged first. The runtime helper was cherry-picked out of this branch to keep the runtime change minimal and independently reviewable.

bindings/go/credentials/ingest.go
matthiasbruns: we now have https://github.com/open-component-model/open-component-model/pull/2399

bindings/go/credentials/ingest.go
matthiasbruns: we are now solely relying on the registries :)

matthiasbruns:
> Waiting for https://github.com/open-component-model/open-component-model/actions/runs/25013143531/job/73253951362

bindings/go/credentials/go.mod
```diff
@@ -4,9 +4,9 @@ go 1.26.1
 
 require (
 	github.com/stretchr/testify v1.11.1
-	ocm.software/open-component-model/bindings/go/configuration v0.0.12
+	ocm.software/open-component-model/bindings/go/configuration v0.0.13
 	ocm.software/open-component-model/bindings/go/dag v0.0.6
-	ocm.software/open-component-model/bindings/go/runtime v0.0.7
+	ocm.software/open-component-model/bindings/go/runtime v0.0.8-0.20260427174225-155791088e10
```
fabianburth: why are we relying on a dev version here?

bindings/go/credentials/graph.go
fabianburth: I finally remembered why this bothered me. I think @jakobmoellerdev and me had a discussion about essentially this interface already. We decided to instead do more specific interfaces (check the contracts of each registry [here](https://github.com/open-component-model/open-component-model/tree/bb60545f8c1c9f23d1f05527699508cb78d70608/bindings/go/plugin/manager/registries)).

Back then, the main reason was that a central `TypeSchemeProvider` (that could have been embedded by all of those) would prevent a single type from implementing multiple of those interfaces.

bindings/go/credentials/doc.go
```diff
@@ -189,26 +230,38 @@
...
+// Basic usage with typed credential resolution:
+//
 //	config := &Config{...} // alternatively parse from yaml via serialization
 //	opts := Options{
-//	    GetCredentialPluginFn: myCredPluginResolver, // optional but needed if repository fallbacks are wanted
-//	    GetRepositoryPluginFn: myRepoPluginResolver, // optional but needed if type support other than DirectCredentials are needed
+//	    CredentialPluginProvider:       myCredPluginProvider,       // optional: needed for custom credential types (e.g. Vault)
+//	    RepositoryPluginProvider:       myRepoPluginProvider,       // optional: needed for repository fallbacks
+//	    CredentialTypeSchemeProvider:   myCredTypeSchemeProvider,   // optional: enables typed credential deserialization
+//	    IdentityTypeSchemeProvider:     myIdentityTypeSchemeProvider, // optional: enables identity ↔ credential validation
```
Skarlso: Errr.... didn't you remove this field? 🤔

Skarlso: Ah it's `IdentityTypeRegistry` now?

bindings/go/credentials/graph.go
```diff
@@ -43,24 +49,53 @@ func ToGraph(ctx context.Context, config *cfgRuntime.Config, opts Options) (Reso
...
+// Compile-time interface check.
+var _ Resolver = (*Graph)(nil)
+
+// Resolve implements Resolver. It returns credentials as map[string]string for backward compatibility.
+// Consumers that need typed credentials should use ResolveTyped instead.
 func (g *Graph) Resolve(ctx context.Context, identity runtime.Identity) (map[string]string, error) {
-	if _, err := identity.ParseType(); err != nil {
-		err = errors.Join(ErrUnknown, err)
-		return nil, fmt.Errorf("to be resolved from the credential graph, a consumer identity type is required: %w", err)
+	typed, err := g.ResolveTyped(ctx, identity)
```
Skarlso, line 80: Hm, should we really do this? The old resolve is still called by the plugins I think with not runtime.Typed. This could panic with the wrong type.

Skarlso (commented):
Sorry, just a couple more questions. :D

matthiasbruns:
> I finally remembered why this bothered me. I think @jakobmoellerdev and me had a discussion about essentially this interface already. We decided to instead do more specific interfaces (check the contracts of each registry [here](https://github.com/open-component-model/open-component-model/tree/bb60545f8c1c9f23d1f05527699508cb78d70608/bindings/go/plugin/manager/registries)).
> 
> Back then, the main reason was that a central TypeSchemeProvider (that could have been embedded by all of those) would prevent a single type from implementing multiple of those interfaces.

create two interfaces with fancy func names to prevent interface collisions

bindings/go/credentials/go.mod
matthiasbruns: will be v0.0.8

bindings/go/credentials/graph.go
matthiasbruns: https://github.com/open-component-model/open-component-model/pull/2343

bindings/go/credentials/graph.go
```diff
@@ -26,12 +26,12 @@ type Options struct {
...
 	IdentityTypeRegistry *IdentityTypeRegistry
 	// CredentialTypeSchemeProvider provides access to known credential types (e.g. HelmHTTPCredentials/v1).
-	CredentialTypeSchemeProvider TypeSchemeProvider
+	CredentialTypeSchemeProvider CredentialTypeSchemeProvider
 }
 
 // ToGraph creates a new credential graph from the provided configuration and options.
 // It initializes the graph structure and ingests the configuration into the graph.
-func ToGraph(ctx context.Context, config *cfgRuntime.Config, opts Options) (*Graph, error) {
+func ToGraph(ctx context.Context, config *cfgRuntime.Config, opts Options) (Resolver, error) {
```
fabianburth: why did you change that one back to the interface?

matthiasbruns: I compared my changes to main and reverted too mush - give me a sec

bindings/go/credentials/ingest.go
```diff
@@ -157,26 +173,146 @@ func processRepositoryConfigurations(g *Graph, config *cfgRuntime.Config, repoTy
...
+						// to plugin-based resolution to avoid silent loss.
+						remaining = append(remaining, cred)
+					}
+					continue
+				} else {
+					slog.WarnContext(ctx, "credential type scheme conversion failed",
+						"type", cred.GetType().String(), "error", err)
+				}
+			}
+		}
```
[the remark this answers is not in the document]
matthiasbruns: done

fabianburth (approved):

Skarlso (approved):
👍

PR #2344, opened 2026-04-20
feat(sigstore): add cosign signing handler [1/2]

morri-son:
## Summary

Adds the `bindings/go/sigstore` module — a signing handler wrapping Cosign CLI for keyless signing and verification via Sigstore (Fulcio, Rekor, OIDC). Adds integration test under `bindings/go/sigstore` which are fully based on [sigstore's scaffolding](https://github.com/sigstore/scaffolding) using a kind cluster with a full sigstore stack.

- **v1alpha1 types:** `CosignSigningConfig`, `OIDCIdentityToken`, `TrustedRootIdentity`, algorithm constants
- **Handler:** sign/verify operations using Cosign CLI subprocess, config resolution, cosign binary auto-download with checksum verification
- **Module infra:** `go.mod`, `Taskfile.yml`, root Taskfile include, renovate config for new module

## Merge order

This is **PR 1 of 2** — must be merged and tagged first:

1. **This PR** → sigstore handler module (`bindings/go/sigstore`)
2. `feat/cosign-signing-handler-3-cli` → CLI sign/verify commands (`cli/`)

PR 2 depend on this module being published

## Test plan

- [ ] `task tools:golangci-lint/module ITEM=bindings/go/sigstore/ -- --fix` passes
- [ ] `task bindings/go/sigstore:test` passes
- [ ] CI module discovery picks up the new module

PR #2346, opened 2026-04-20
feat(cli): integrate cosign signing handler [2/2]

### LINKED ISSUE #998: Wire sigstore handler and OIDC credential plugin into the OCM CLI
## Description

Wire the `bindings/go/sigstore` handler and an OIDC credential plugin into the CLI.

Scope:
- Register `ComponentSignatureHandler` in the CLI's signing handler registry
- Implement `SigstoreOIDC/v1alpha1` credential plugin in `cli/internal/plugin/builtin/sigstore/`:
  - Checks `SIGSTORE_ID_TOKEN` env var first (CI/automation path)
  - Falls back to interactive browser-based OIDC flow
  - Default issuer: `https://oauth2.sigstore.dev/auth`, client ID: `sigstore`
- Wire plugin into `CredentialPluginProvider` in `cli/cmd/setup/`
- Add `bindings/go/sigstore` to `cli/go.mod`

## Done Criteria

- [ ] `ocm sign componentversion --signer sigstore` works for keyless and key-based credentials
- [ ] `ocm verify componentversion --verifier sigstore` works with and without explicit trusted root
- [ ] OIDC credential plugin resolves token from env var and interactive browser flow
- [ ] `.ocmconfig` example with `SigstoreOIDC/v1alpha1` credential spec provided
- [ ] Unit tests cover plugin resolution and handler registration
- [ ] Code reviewed by other team members
- [ ] Enduser documentation updated (if applicable)

### LINKED ISSUE #1060: docs: Enhance how-tos and add tutorial for signing and verification with Sigstore and GPG
## Description

Enhance the existing how-to guides with new tabs for Sigstore signing and verifying, interactive and in CI, to the OCM website.

Scope:
- `content/docs/how-to/Sign and Verify/sign-component-versions.md` (ENHANCE)
- `content/docs/how-to/Sign and Verify/verify-component-versions.md` (ENHANCE)
- `content/docs/tutorials/signing/sigstore.md` (NEW)

## Done Criteria

- [x] How-tos enhanced 
- [x] Tutorial created
- [x] Covers keyless signing and verify
- [x] Code reviewed by other team members

morri-son:
## Summary

Integrates the sigstore cosign signing handler (#2344) into the OCM CLI:

- **Credential plugin registry** (`bindings/go/plugin/manager/registries/credentialplugin`) — generic registry that discovers and resolves `credentials.CredentialPlugin` implementations by their runtime type; plugins self-declare their scheme so callers never need to pre-register types manually
- **OIDC credential plugin** (`cli/internal/plugin/builtin/oidc`) — first concrete credential plugin; acquires OIDC identity tokens via interactive browser flow (`authorization-code`) 
- **OIDC browser flow** (`cli/internal/oidcflow`) — opens browser for interactive OIDC token acquisition, runs local callback server, uses a copy of the sigstore success page, but OCM branded

> **Note:** Contains `replace` directives for `bindings/go/sigstore` — expected to fail in CI until PR 1 is merged and tagged.

## Merge order

This is **PR 2 of 2** — merge after PR 1 is tagged:

1. #2344 → sigstore handler module (**merge + tag first**)
2. **This PR** → CLI sign/verify commands

## TODO

- [ ] Enhance `sigstore-integration.yml` workflow with CLI sigstore integration tests as a single job that reuses the scaffolding environment: spin up scaffolding once, run handler integration tests, then run CLI integration tests (`cli/integration/`) against the same env
- [ ] Ensure `--signer-spec` with `SigstoreSigningConfiguration/v1alpha1` correctly threads `signingConfig` to the handler

## Test plan

- [ ] Remove `replace` directives, point to published tag
- [ ] Manual: `ocm sign component-version --signing-config <config>` triggers browser OIDC flow
- [ ] CI: `sigstore-integration.yml` runs handler + CLI E2E tests against scaffolding cluster
- [ ] CI: CLI integration tests sign and verify a CTF using `--signer-spec` and `--verifier-spec` with scaffolding OIDC token
- [ ] `credentialplugin.Registry` resolves `OIDCIdentityTokenProvider/v1alpha1` to the OIDC plugin
- [ ] Token exchange flow resolves subject token from env var, file, or inline value


cli/internal/plugin/builtin/oidc/oidc_plugin.go
```diff
@@ -0,0 +1,109 @@
...
+type oidcConfig struct {
+	issuer   string
+	clientID string
+}
+
+func parseOIDCConfig(typed runtime.Typed) (*oidcConfig, error) {
+	var raw struct {
+		Issuer   string `json:"issuer"`
+		ClientID string `json:"clientID"`
+	}
```
fabianburth: I know, until @matthiasbruns PR is merged, we do not have typed credentials. Should we still validate the type?

cli/internal/plugin/builtin/sigstore/oidc_plugin.go
```diff
@@ -0,0 +1,109 @@
...
+// variable or an interactive browser-based OIDC flow.
+//
+// Example .ocmconfig entry:
+//
+//	consumers:
+//	- identity:
+//	    type: OIDCIdentityToken/v1alpha1
+//	    name: sigstore
+//	  credentials:
+//	  - type: SigstoreOIDC/v1alpha1
```
fabianburth:
```suggestion
//	  - type: OIDCProvider/v1alpha1
```
Not sigstore specific

cli/internal/plugin/builtin/cosign/oidc_plugin.go
```diff
@@ -0,0 +1,109 @@
...
+// OIDCPlugin implements credentials.CredentialPlugin for interactive OIDC
+// token acquisition. It resolves tokens via the SIGSTORE_ID_TOKEN environment
+// variable or an interactive browser-based OIDC flow.
+//
+// Example .ocmconfig entry:
+//
+//	consumers:
+//	- identity:
+//	    type: OIDCIdentityToken/v1alpha1
+//	    name: sigstore
```
fabianburth: I believe this is not accurate anymore. Should include `algorithm` and `signature`.

cli/internal/plugin/builtin/sigstore/oidc_plugin.go
```diff
@@ -0,0 +1,110 @@
...
+	id.SetType(OIDCPluginTypeVersioned)
+	return id, nil
+}
+
+// Resolve acquires an OIDC identity token. It checks SIGSTORE_ID_TOKEN first,
+// then falls back to an interactive browser-based OIDC flow.
+func (p *OIDCPlugin) Resolve(ctx context.Context, identity runtime.Identity, _ map[string]string) (map[string]string, error) {
+	if tok := os.Getenv("SIGSTORE_ID_TOKEN"); tok != "" {
+		return map[string]string{credentialKeyToken: tok}, nil
+	}
```
fabianburth: Should we really ignore the entire identity `SIGSTORE_ID_TOKEN` is set.
Might lead to unexpected behaviours.

cli/cmd/setup/setup.go
```diff
@@ -111,9 +112,11 @@ func CredentialGraph(cmd *cobra.Command) error {
...
 		RepositoryPluginProvider: pluginManager.CredentialRepositoryRegistry,
 		CredentialPluginProvider: credentials.GetCredentialPluginFn(
-			// TODO(jakobmoellerdev): use the plugin manager to get the credential plugin once we have some.
-			func(ctx context.Context, typed runtime.Typed) (credentials.CredentialPlugin, error) {
-				return nil, fmt.Errorf("no credential plugin found for type %s", typed)
+			func(_ context.Context, typed runtime.Typed) (credentials.CredentialPlugin, error) {
+				if typed.GetType().GetName() == sigstore.OIDCPluginType {
+					return &sigstore.OIDCPlugin{}, nil
+				}
+				return nil, fmt.Errorf("no credential plugin found for type %s", typed.GetType())
```
fabianburth: This is definitely not the way ;)
I assume we need to register a credential plugin registry here.

cli/cmd/sign/component-version/cmd.go
```diff
@@ -11,6 +11,8 @@ import (
 	"slices"
 	"strings"
 
+	_ "ocm.software/open-component-model/bindings/go/sigstore/signing/v1alpha1"
+
```
fabianburth: why do we need an empty import here?

cli/cmd/verify/component-version/cmd.go
```diff
@@ -112,6 +116,33 @@ verify component-version ghcr.io/open-component-model/ocm//ocm.software/ocmcli:0
...
+
+    type: SigstoreVerificationConfiguration/v1alpha1
+    certificateOIDCIssuer: https://fulcio.example.com
+    certificateIdentity: ci-user@example.com
+    trustedRoot: /path/to/trusted_root.json
+    privateInfrastructure: true
+
+# Verify with Sigstore verifier spec:
+verify component-version ghcr.io/open-component-model/ocm//ocm.software/ocmcli:0.23.0 --verifier-spec ./sigstore-verify.yaml
+
```
fabianburth, line 165: nit: might be too long for cli documentation

cli/cmd/verify/component-version/cmd.go
```diff
@@ -124,7 +155,7 @@ verify component-version ghcr.io/open-component-model/ocm//ocm.software/ocmcli:0
 
 	cmd.Flags().Int(FlagConcurrencyLimit, 4, "maximum amount of parallel requests to the repository for resolving component versions")
 	cmd.Flags().String(FlagSignature, "", "name of the signature to verify. If not set, all signatures are verified.")
-	cmd.Flags().String(FlagVerifierSpec, "", "path to an optional verifier specification file. If empty, defaults to an empty RSASSA-PSS configuration.")
+	cmd.Flags().String(FlagVerifierSpec, "", "path to a verifier specification file. If empty, defaults to RSASSA-PSS. For Sigstore keyless verification, use type SigstoreVerificationConfiguration/v1alpha1 (identity constraints required).")
```
fabianburth: I don't think "identity constraints required" would help me at all as a user

cli/internal/oidcflow/oidcflow.go
```diff
@@ -0,0 +1,363 @@
...
+	}
+	return DefaultIssuer
+}
+
+func (o *Options) clientID() string {
+	if o.ClientID != "" {
+		return o.ClientID
+	}
+	return DefaultClientID
+}
```
fabianburth: nit: That's not how we usually do defaulting in this codebase. Usually, we would just do that in the function consuming the options.

cli/internal/oidcflow/oidcflow.go
```diff
@@ -0,0 +1,363 @@
...
+		Handler:           callbackHandler(state, codeCh, errCh),
+	}
+	go func() {
+		if sErr := srv.Serve(listener); sErr != nil && !errors.Is(sErr, http.ErrServerClosed) {
+			errCh <- sErr
+		}
+	}()
+	defer func() {
+		go func() { _ = srv.Shutdown(context.Background()) }()
+	}()
```
fabianburth: This looks suspicious. We might be risking leaking go routines here.

cli/internal/oidcflow/oidcflow.go
fabianburth, line 1: Is this a 1:1 copy from sigstore? If so, mark it as such. If not, why not?

cli/internal/plugin/builtin/oidc/oidc_plugin.go
fabianburth, line 1: Seems to be a generic oidc plugin now, not sure whether sigstore is the correct location

cli/internal/plugin/builtin/builtin.go
```diff
@@ -51,14 +51,12 @@ func Register(manager *manager.PluginManager, filesystemConfig *filesystemv1alph
 	); err != nil {
 		return fmt.Errorf("could not register helm digest processor plugin: %w", err)
 	}
-	if err := manager.ResourcePluginRegistry.RegisterInternalResourcePlugin(
-		helmresource.NewResourceRepository(filesystemConfig),
-	); err != nil {
-		return fmt.Errorf("could not register helm resource repository plugin: %w", err)
-	}
```
fabianburth, line 58: Why are we removing helm here?

tools/mcp-knowledge-server/testdata/bindings/go/blob/doc.go
fabianburth, line 1: what are these mcp server changes doing in here?

fabianburth (commented):
Please be more careful with your PR hygiene!

I did not review the oidc flow in detail. I'd like to understand how detailed of a review I need to do here. Is this a battle tested copy from another established project that we copied to avoid the dependencies or is this a custom implementation?

cli/internal/plugin/builtin/oidc/oidc_plugin.go
morri-son: I would keep it for now and refactor the code after we have typed credentials

cli/internal/plugin/builtin/sigstore/oidc_plugin.go
morri-son: done.

cli/internal/oidcflow/oidcflow.go
morri-son: not a copy — independent implementation of the same oidc authorization code + pkce flow, avoids the github.com/sigstore/sigstore dependency. Only the success html page is a 1:1 copy (attributed in the code comment).

cli/internal/plugin/builtin/oidc/oidc_plugin.go
morri-son: agreed — renamed the folder from builtin/sigstore to builtin/oidc, package is now oidc.

cli/internal/plugin/builtin/sigstore/oidc_plugin.go
morri-son: the handler no longer reads SIGSTORE_ID_TOKEN from the environment (removed in pr #2344). the token comes exclusively from the credential graph. updated the help text accordingly.

cli/cmd/sign/component-version/cmd.go
morri-son: removed — the sigstore v1alpha1 package isn't available on this branch and registration happens explicitly via builtin/oidc/register.go.

cli/cmd/verify/component-version/cmd.go
morri-son: left the Long description as is (it's structured and ~30 lines). trimmed the flag help text.

cli/cmd/verify/component-version/cmd.go
morri-son: changed to '(set CertificateOIDCIssuer and CertificateIdentity)' — hopefully more actionable.

cli/internal/oidcflow/oidcflow.go
morri-son: fixed — moved defaults into GetIDToken directly, removed the methods.

cli/internal/oidcflow/oidcflow.go
morri-son: fixed — replaced the background goroutine shutdown with synchronous srv.Shutdown(ctx).

cli/internal/plugin/builtin/builtin.go
morri-son: restored — the helmresource registration was accidentally removed in the initial commit. added it back.

tools/mcp-knowledge-server/testdata/bindings/go/blob/doc.go
morri-son: removed — rebased the branch to drop all mcp knowledge server commits. they don't belong in this pr. really need to use git worktrees. Parallel actions doing checkouts in multiple windows is too confusing 🙂

cli/internal/plugin/builtin/cosign/oidc_plugin.go
morri-son: fixed — the example now includes algorithm and signature fields.

cli/cmd/setup/setup.go
morri-son: agreed — the proper approach is a CredentialPluginRegistry on the plugin manager (same pattern as SigningRegistry), but that doesn't exist yet. jakob's TODO was a placeholder for exactly this. will restore the TODO and keep the inline function until the registry is implemented. He explicitly said `once we have some`, now we have exactly one :-)

cli/cmd/setup/setup.go
```diff
@@ -111,9 +112,13 @@ func CredentialGraph(cmd *cobra.Command) error {
...
-			// TODO(jakobmoellerdev): use the plugin manager to get the credential plugin once we have some.
-			func(ctx context.Context, typed runtime.Typed) (credentials.CredentialPlugin, error) {
-				return nil, fmt.Errorf("no credential plugin found for type %s", typed)
+			// TODO(jakobmoellerdev): replace with a CredentialPluginRegistry on the plugin manager
+			// (same pattern as SigningRegistry) once the registry infrastructure exists.
+			func(_ context.Context, typed runtime.Typed) (credentials.CredentialPlugin, error) {
+				if typed.GetType().GetName() == oidc.OIDCPluginType {
+					return &oidc.OIDCPlugin{}, nil
+				}
+				return nil, fmt.Errorf("no credential plugin found for type %s", typed.GetType())
```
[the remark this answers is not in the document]
morri-son: addressed

cli/internal/oidcflow/oidcflow.go
```diff
@@ -0,0 +1,357 @@
...
+	}
+
+	var cmd *exec.Cmd
+	switch runtime.GOOS {
+	case "darwin":
+		cmd = exec.CommandContext(ctx, "open", rawURL)
+	case "linux":
+		cmd = exec.CommandContext(ctx, "xdg-open", rawURL)
+	case "windows":
+		cmd = exec.CommandContext(ctx, "cmd", "/c", "start", "", rawURL)
```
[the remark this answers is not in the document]
morri-son: addressed

cli/docs/reference/ocm_sign_component-version.md
```diff
@@ -116,6 +117,46 @@ sign component-version ghcr.io/open-component-model/ocm//ocm.software/ocmcli:0.2
...
+
+# With a local signing config file (private infrastructure, endpoint discovery):
+
+    type: SigstoreSigningConfiguration/v1alpha1
+    signingConfig: /path/to/signing_config.json
+
+## Example Credential Config (.ocmconfig) — Sigstore OIDC token
+#
+# The OIDCProvider plugin checks SIGSTORE_ID_TOKEN env var first (CI/automation),
+# then falls back to an interactive browser-based OIDC flow.
```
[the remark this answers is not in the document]
morri-son: addressed

morri-son:
@coderabbitai review

cli/internal/oidcflow/oidcflow.go
```diff
@@ -0,0 +1,388 @@
...
+
+// Options configures the OIDC flow.
+type Options struct {
+	Issuer   string
+	ClientID string
+}
+
+// GetIDToken performs an interactive OIDC authorization code flow with PKCE.
+// It opens the user's browser for authentication and waits for the callback.
+func GetIDToken(ctx context.Context, opts Options) (*Token, error) {
```
fabianburth, line 83: Since we discussed whether this can be a general OIDC token provider, I was wondering - can we always assume that the service we forward the token to performs the token verification (as fulcio does)?

morri-son: OIDCProvider/v1alpha1 together with our workflow just gets a credential from an OIDC provider. In the sigstore case Fulcio verifies if this token provided is an OIDC token and I would assume that other service that accept OIDC tokens do the same.

cli/cmd/setup/setup.go
matthiasbruns: no need for that right now - we will do the wiring later in the migration phase - you can keep working with credentials as before, just provide typed once and the util functions as in e.g. the rsa or helm binding prs

morri-son: addressed in the PR with this [commit](a9a1845a38752bfba1cfb3024c823d74aca6b2b3)

cli/internal/plugin/builtin/sigstore/oidc_plugin.go
morri-son: after team discussion we now accept cosign related env vars. If they are set, they take precedence over existing credential config.

bindings/go/plugin/manager/registries/signinghandler/registry.go
```diff
@@ -85,6 +85,12 @@ func (r *SigningRegistry) GetPlugin(ctx context.Context, spec runtime.Typed) (si
 			return nil, fmt.Errorf("no internal plugin registered for type %v", typ)
 		}
 
+		if e, ok := p.(interface{ Ensure(context.Context) error }); ok {
```
matthiasbruns: https://github.com/open-component-model/open-component-model/pull/2344#discussion_r3199115984
also the error message here explains nothing

cli/cmd/cmd_test.go
```diff
@@ -1514,11 +1514,18 @@ configurations:
...
+`
+	signerSpecFilePath := filepath.Join(tmp, "signer-spec.yaml")
+	r.NoError(os.WriteFile(signerSpecFilePath, []byte(signerSpecYAML), 0o600))
+
 	reference := archiveFilePath + "//" + name + ":" + version
 
 	_, err = test.OCM(t, test.WithArgs("sign", "component-version",
 		reference,
 		"--signature", signatureName,
+		"--signer-spec", signerSpecFilePath,
```
matthiasbruns: did the original test now disappear?

cli/docs/reference/ocm_verify_component-version.md
```diff
@@ -36,8 +36,10 @@ Verify component version(s) inside an OCM repository based on signatures.
 - Without --signature: verify all signatures  
 - Fail fast on first invalid signature  
 - Default verifier: RSASSA-PSS plugin  
-  - Supports config-less verification
```
matthiasbruns, line 39: did we drop the old behaviour?

cli/integration/sigstore_integration_test.go
```diff
@@ -0,0 +1,192 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"ocm.software/open-component-model/cli/cmd"
+)
+
+type sigstoreStack struct {
+	OIDCToken        string
+	SigningConfig     string
```
matthiasbruns: formatting

bindings/go/plugin/manager/registries/signinghandler/registry.go
morri-son: as discussed made the ensure implicit now so that callers don't have to take care. Made error more verbose.

cli/integration/sigstore_integration_test.go
```diff
@@ -0,0 +1,190 @@
...
+trustedRoot: %s
+privateInfrastructure: true
+`, stack.OIDCIssuer, stack.OIDCIdentity, stack.TrustedRoot)
+		privateSpecPath := filepath.Join(t.TempDir(), "private-verifier.yaml")
+		r.NoError(os.WriteFile(privateSpecPath, []byte(privateSpec), 0o600))
+
+		verifyCMD := cmd.New()
+		verifyCMD.SetArgs([]string{"verify", "cv", ref, "--config", cfgPath, "--verifier-spec", privateSpecPath})
+		r.NoError(verifyCMD.ExecuteContext(t.Context()))
+	})
```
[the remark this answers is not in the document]
morri-son: added signRef helper and call it in the three verify-only subtests

cli/docs/reference/ocm_verify_component-version.md
morri-son: Now that we have multiple signing algorithms, I changed the text. the latest version already had `Uses public key from discovered credentials (config-less)`. I now even made it more verbose: `Supports config-less verification: public key is resolved from discovered credentials (no --verifier-spec required)`.

cli/integration/sigstore_integration_test.go
morri-son: corrected

cli/cmd/cmd_test.go
morri-son: Good catch. Since we now have two signing algorithms I added the required spec to the test. But it makes sense to keep testing the default/happy path. I added Test_Sign_And_Verify_Component_Version_DefaultSpec which tests the happy path with no--signer-spec/--verifier-spec. Since we want to change the default after merge, it acts as a canary: if the default is changed from RSASSA-PSS to Sigstore, this test will fail explicitly and signal the credentials config needs updating.

Skarlso:
Screenshot 2026-05-14 at 13 17 56

Ran Gerald's massive test suite, they all passed. Well done. :)

cli/integration/go.mod
```diff
@@ -4,12 +4,15 @@ go 1.26.1
...
+// TODO: remove replace once bindings/go/sigstore is published (#2344).
+replace ocm.software/open-component-model/bindings/go/sigstore => ../../bindings/go/sigstore
+
 require (
 	github.com/opencontainers/go-digest v1.0.0
 	github.com/opencontainers/image-spec v1.1.1
 	github.com/stretchr/testify v1.11.1
-	github.com/testcontainers/testcontainers-go v0.42.0
-	github.com/testcontainers/testcontainers-go/modules/registry v0.42.0
+	github.com/testcontainers/testcontainers-go v0.41.0
```
Skarlso: Ah okay, I think these downgrades are because of the replace statement above, so should be fine, once reverted.

cli/internal/oidcflow/oidcflow.go
```diff
@@ -0,0 +1,402 @@
...
+
+func randomString(n int) (string, error) {
+	b := make([]byte, n)
+	if _, err := rand.Read(b); err != nil {
+		return "", err
+	}
+	return base64.RawURLEncoding.EncodeToString(b), nil
+}
+
+const successHTML = `<!DOCTYPE html>
```
Skarlso: Please put this next to this go file and use embed.FS instead. 🙇

Skarlso:
We'll merge https://github.com/open-component-model/open-component-model/pull/2512 today, then refresh this one and I'll run the tests again.

cli/integration/go.mod
Skarlso: Actually I think none of these changes should be here.

cli/internal/plugin/builtin/oidc/oidc_plugin.go
```diff
@@ -0,0 +1,78 @@
...
+	if credential == nil {
+		return nil, fmt.Errorf("credential must not be nil")
+	}
+	if credential.GetType().IsEmpty() {
+		return nil, fmt.Errorf("credential type must not be empty")
+	}
+	id := runtime.Identity{}
+	id.SetType(OIDCPluginTypeVersioned)
+	return id, nil
+}
```
Skarlso, line 75: This returns a bare identity. 🤔 Don't we want to add issuer / clientID here? That would allow registering multiple OIDC providers. AS this is right now it will only register ONE and then that will catch basically everything.

Skarlso: This will be discussed next time.

Skarlso (dismissed):

cli/internal/plugin/builtin/oidc/oidc_plugin.go
matthiasbruns: @morri-son Can you confirm that this is the desired state or do we need to wait for the update after the fix from @fabianburth in the credential graph?

jakobmoellerdev: since this is the identity of the OIDC plugin, I believe it should be able to return an identity that is unique per oidc provider.

cli/cmd/cmd_test.go
jakobmoellerdev: this doesnt really make sense here. You should not touch the default test unless you change the default signing / verify behavior no? since cosign is not the default it should not be the changing this test

cli/docs/reference/ocm_sign_component-version.md
```diff
@@ -116,6 +117,45 @@ sign component-version ghcr.io/open-component-model/ocm//ocm.software/ocmcli:0.2
     signatureAlgorithm: RSASSA-PSS
     signatureEncodingPolicy: PEM
 
+## Example Signer Spec File — Sigstore keyless (SigstoreSigningConfiguration/v1alpha1)
```
jakobmoellerdev, line 120: I feel like we should add a specific how to for the signing / verification into the website to accompany this as it makes it much easier to work with

cli/docs/reference/ocm_verify_component-version.md
jakobmoellerdev: this is not what is now written in the new text to be honest

cli/docs/reference/ocm_verify_component-version.md
```diff
@@ -102,7 +131,7 @@ verify component-version ghcr.io/open-component-model/ocm//ocm.software/ocmcli:0
       --concurrency-limit int   maximum amount of parallel requests to the repository for resolving component versions (default 4)
   -h, --help                    help for component-version
       --signature string        name of the signature to verify. If not set, all signatures are verified.
-      --verifier-spec string    path to an optional verifier specification file. If empty, defaults to an empty RSASSA-PSS configuration.
+      --verifier-spec string    path to a verifier specification file. If empty, defaults to RSASSA-PSS. For Sigstore keyless verification, use type SigstoreVerificationConfiguration/v1alpha1 (set CertificateOIDCIssuer and CertificateIdentity).
```
jakobmoellerdev: why do we mention CertificateOIDCIssuer and Identity here? I dont feel like it adds value to the flag help because it is unclear how to set them without reading the exact examples

cli/integration/Taskfile.yml
```diff
@@ -5,10 +5,11 @@ includes:
 
 vars:
   INTEGRATION_TEST_IDENTIFIER: 'Integration'
+  SIGSTORE_ENV_FILE: '../../bindings/go/sigstore/integration/tmp/sigstore-env.sh'
```
jakobmoellerdev: is this a leftover? it looks unused

cli/internal/oidcflow/oidcflow.go
```diff
@@ -0,0 +1,328 @@
...
+
+// Options configures the OIDC flow.
+type Options struct {
+	Issuer   string
+	ClientID string
+}
+
+// GetIDToken performs an interactive OIDC authorization code flow with PKCE.
+// It opens the user's browser for authentication and waits for the callback.
+func GetIDToken(ctx context.Context, opts Options) (*Token, error) {
```
jakobmoellerdev, line 83: this function is too long & not sufficiently documented on behavior tbh

cli/internal/oidcflow/oidcflow.go
```diff
@@ -0,0 +1,328 @@
...
+		return nil, fmt.Errorf("verify id token: %w", err)
+	}
+	if idToken.Nonce != nonce {
+		return nil, errors.New("nonce mismatch in id token")
+	}
+
+	return &Token{RawToken: rawIDToken}, nil
+}
+
+func callbackHandler(expectedState, expectedIssuer string, codeCh chan<- string, errCh chan<- error) http.Handler {
```
jakobmoellerdev, line 193: missing docs

cli/internal/oidcflow/oidcflow.go
```diff
@@ -0,0 +1,328 @@
...
+	}
+}
+
+// pkce implements Proof Key for Code Exchange (RFC 7636).
+type pkce struct {
+	challenge string
+	verifier  string
+}
+
+func newPKCE(provider *oidc.Provider) (*pkce, error) {
```
jakobmoellerdev: this should be documented better and located / tested in a separate file

cli/internal/oidcflow/oidcflow.go
```diff
@@ -0,0 +1,328 @@
...
+
+//go:embed assets/success.html
+var successHTML string
+
+const (
+	DefaultIssuer   = "https://oauth2.sigstore.dev/auth"
+	DefaultClientID = "sigstore"
+
+	callbackPath    = "/auth/callback"
+	callbackTimeout = 120 * time.Second
```
jakobmoellerdev: this timeout is a magic constant, it should probably be configurable

cli/internal/plugin/builtin/oidc/register.go
```diff
@@ -0,0 +1,20 @@
...
+import (
+	filesystemv1alpha1 "ocm.software/open-component-model/bindings/go/configuration/filesystem/v1alpha1/spec"
+	"ocm.software/open-component-model/bindings/go/plugin/manager/registries/credentialplugin"
+	"ocm.software/open-component-model/bindings/go/plugin/manager/registries/signinghandler"
+	"ocm.software/open-component-model/bindings/go/sigstore/signing/handler"
+)
+
+// Register registers the Sigstore signing handler with the signing registry.
+func Register(signingHandlerRegistry *signinghandler.SigningRegistry, filesystemConfig *filesystemv1alpha1.Config) error {
+	return signingHandlerRegistry.RegisterInternalComponentSignatureHandler(
```
jakobmoellerdev, line 12: this should be revisited. I dont see how a fluent interface with builder style arguments is actually useful

docs/adr/0017_sigstore_integration.md
```diff
@@ -81,6 +81,41 @@ The `cosign` binary will be downloaded on-demand and cached in the user's home d
 The version of `cosign` will be pinned and managed by Renovate. The user will interact with the feature through the standard `ocm sign` and `ocm verify` commands,
 with `sigstore` as the algorithm name.
 
+## OIDC Authentication Model
+
+When the `.ocmconfig` does not contain a pre-provisioned OIDC token, the CLI performs an interactive
+OIDC Authorization Code flow with PKCE (RFC 7636, S256) to acquire an identity token from the configured
```
jakobmoellerdev: please link all referenced RFCs

docs/adr/0017_sigstore_integration.md
```diff
@@ -81,6 +81,41 @@ The `cosign` binary will be downloaded on-demand and cached in the user's home d
...
+The loopback redirect URI (`http://127.0.0.1:<random-port>`) limits code interception to same-machine processes,
+which PKCE fully mitigates.
+
+### Enterprise Provider Compatibility
+
+The implementation supports arbitrary OIDC providers via the `issuer` and `clientID` fields in `.ocmconfig`.
+Enterprise deployments (Keycloak, Azure AD, Okta, etc.) are supported with the following considerations:
+
+- **RFC 9207 (Issuer Identification):** The callback validates the `iss` parameter when present (permissive mode).
+  Providers that omit `iss` are accepted without issuer verification. This protects against mix-up attacks
```
jakobmoellerdev: i dont think this is correct. we would have to verify and reject non-set iss fields to avoid mix-up no?

cli/internal/plugin/builtin/oidc/oidc_plugin.go
Skarlso: Okay, then this needs to be fixed and at least the issue and clientID should be added into the returned identity.

cli/cmd/cmd_test.go
```diff
@@ -1527,12 +1534,86 @@ configurations:
...
 		"--signature", signatureName,
+		"--verifier-spec", signerSpecFilePath,
 		"--config", ocmConfigFilePath),
 		test.WithOutput(logs),
 	)
 	r.NoError(err, "failed to verify component version")
 }
 
+// Test_Sign_And_Verify_Component_Version_DefaultSpec guards the default sign/verify path without --signer-spec/--verifier-spec; acts as a canary if the default algorithm changes.
+func Test_Sign_And_Verify_Component_Version_DefaultSpec(t *testing.T) {
```
matthiasbruns: wasn't this already present in the original test? I would expect this new test to test the sigstore changes

cli/docs/reference/ocm_sign_component-version.md
matthiasbruns: +1

matthiasbruns: but this is generated isnt it?

cli/integration/Taskfile.yml
```diff
@@ -5,10 +5,11 @@ includes:
...
 
 tasks:
   test/integration:
     cmds:
       - task: reuse:run-go-test
         vars:
-          GO_TEST_FLAGS: '-run "{{ .INTEGRATION_TEST_IDENTIFIER }}"'
\ No newline at end of file
+          GO_TEST_FLAGS: '-run "{{ .INTEGRATION_TEST_IDENTIFIER }}"'
```
matthiasbruns: I think the whole file can be rolled back

cli/internal/oidcflow/oidcflow.go
```diff
@@ -0,0 +1,328 @@
...
+
+	srv := &http.Server{
+		ReadHeaderTimeout: 2 * time.Second,
+		BaseContext:       func(_ net.Listener) context.Context { return ctx },
+		Handler:           callbackHandler(state, opts.Issuer, codeCh, errCh),
+	}
+	go func() {
+		if sErr := srv.Serve(listener); sErr != nil && !errors.Is(sErr, http.ErrServerClosed) {
+			select {
+			case errCh <- sErr:
```
matthiasbruns: is the `select` needed at all? if we are here, cant we just push `sErr` into the channel?

docs/adr/0017_sigstore_integration.md
morri-son: you're right, mix-up is not avoided in our current setup. I enhanced the doc and explain the behavior in detail and also mention that we cannot enforce iss by default as the public sigstore.dev Dex instance omits iss.

cli/internal/oidcflow/oidcflow.go
```diff
@@ -0,0 +1,328 @@
...
+			default:
+			}
+		}
+	}()
+	defer srv.Shutdown(ctx) //nolint:errcheck // best-effort shutdown
```
matthiasbruns: shows IDE warning - error not handled/ignored

docs/adr/0017_sigstore_integration.md
morri-son: added links to all RFCs

cli/internal/oidcflow/oidcflow.go
morri-son: moved code and tests to `pkce.go` and `pace_test.go`.

cli/integration/Taskfile.yml
morri-son: leftover, removed.

cli/docs/reference/ocm_verify_component-version.md
morri-son: removed details

cli/docs/reference/ocm_sign_component-version.md
morri-son: will add two how tos

cli/internal/oidcflow/oidcflow.go
morri-son: Tried to refactor it. Pls check again.

cli/internal/plugin/builtin/oidc/register.go
morri-son: hmmm...you mean I should revisit the setup in sigstore handler, move to a config struct and then adopt here?

cli/integration/Taskfile.yml
morri-son: rolled back

cli/cmd/cmd_test.go
morri-son: removed that test. functional tests need to happen against the integration environment, not in the cli, as a sigstore env is required.

cli/internal/oidcflow/oidcflow.go
morri-son: fixed, took over your proposal

cli/internal/oidcflow/oidcflow.go
morri-son: added

cli/internal/oidcflow/oidcflow.go
morri-son: made configurable

cli/internal/oidcflow/oidcflow.go
morri-son: removed select

cli/internal/oidcflow/oidcflow.go
morri-son: corrected

cli/internal/oidcflow/oidcflow.go
matthiasbruns: huh? where

cli/internal/oidcflow/oidcflow.go
jakobmoellerdev: not added yet. can you please double check your comments more indepth? its been a few times now that you said something was done when it was still missing when I reviewed it.

cli/internal/plugin/builtin/oidc/register.go
jakobmoellerdev: I think we could simply make the new handler work with fields instead of options, its a lot less code. fluent interfaces with option args only make sense for very large constructors with lots of options imho. I think this can be done in a separate PR, just a general learning :+1:

website/content/docs/how-to/Sign and Verify/sign-component-version.md
```diff
@@ -0,0 +1,354 @@
...
+
+Concretely, depending on which provider you logged in with:
+
+| Login provider | Issuer recorded in the certificate |
+| --- | --- |
+| Google | `https://accounts.google.com` |
+| GitHub | `https://github.com/login/oauth` |
+| Microsoft | `https://login.microsoftonline.com` |
+
+This is what the verifier's `certificateOIDCIssuer` constraint must match (see [How-to: Verify Component Versions]({{< relref "verify-component-version.md" >}})).
```
frewilhelm: looks a bit skewed 

Image

website/content/docs/how-to/Sign and Verify/sign-component-version.md
```diff
@@ -0,0 +1,354 @@
...
+**Estimated time:** ~5 minutes
+
+## Prerequisites
+
+- [OCM CLI installed]({{< relref "ocm-cli-installation.md" >}})
+- A browser available on the machine (the OIDC flow opens a browser to authenticate you)
+- A component version in a CTF archive or OCI registry (we'll use `github.com/acme.org/helloworld:1.0.0` from the [getting started guide]({{< relref "create-component-version.md" >}}); any component you can write to works)
+
+{{< callout context="tip" >}}
+You do **not** need to generate or distribute keys for this flow. Identity is the signer.
```
frewilhelm: > Identity is the signer.

What does this mean?

website/content/docs/how-to/Sign and Verify/sign-component-version.md
```diff
@@ -0,0 +1,354 @@
...
+```text
+time=2026-05-18T10:12:03.118+02:00 level=INFO msg="acquiring OIDC identity token" issuer=https://oauth2.sigstore.dev/auth
+time=2026-05-18T10:12:08.402+02:00 level=INFO msg="OIDC identity token acquired"
+time=2026-05-18T10:12:08.601+02:00 level=INFO msg="signing via Sigstore" fulcio=https://fulcio.sigstore.dev rekor=https://rekor.sigstore.dev
+time=2026-05-18T10:12:11.972+02:00 level=INFO msg="signed successfully" name=default digest=91dd197868907487e62872695db1fa7b397fde300bcbae23e24abc188fb147ad hashAlgorithm=SHA-256 normalisationAlgorithm=jsonNormalisation/v4alpha1
+```
+{{< /details >}}
+
+{{< callout context="tip" >}}
+The first run downloads and caches the `cosign` binary into `~/.cache/ocm/cosign/...`. Subsequent runs skip the download.
```
frewilhelm: Is this always true or only when `cosign` is not available?

> Subsequent runs skip the download.

What about outdated `cosign` binaries?

website/content/docs/how-to/Sign and Verify/sign-component-version.md
```diff
@@ -0,0 +1,354 @@
...
+  --signer-spec ./sigstore-sign.yaml \
+  /tmp/helloworld/transport-archive//github.com/acme.org/helloworld:1.0.0
+```
+{{< /tab >}}
+{{< tab "Remote OCI Registry" >}}
+
+```bash
+ocm sign cv \
+  --signer-spec ./sigstore-sign.yaml \
+  ghcr.io/<your-namespace>//github.com/acme.org/helloworld:1.0.0
```
frewilhelm, line 266: This failed unfortunetaly. Is this expected?

```text
ocm sign cv \
  --signer-spec ./sigstore-sign.yaml \
  /tmp/helloworld/transport-archive//github.com/acme.org/helloworld:1.0.0
{"time":"2026-05-19T15:02:13.851174+02:00","level":"INFO","msg":"capabilities sent"}
Error: setup credential graph: could not create credential graph: failed to process edges based on plugins: failed to process consumer credential: getting credential plugin failed: no credential plugin found for type {"type":"OIDCIdentityTokenProvider/v1alpha1"}
```

website/content/docs/how-to/Sign and Verify/sign-component-version.md
```diff
@@ -0,0 +1,354 @@
...
+
+**Cause:** The component version already has a signature with this name.
+
+**Fix:** Use a different signature name with `--signature newname`.
+
+### Symptom: Permission denied on registry
+
+**Cause:** Missing write access to the OCI registry.
+
+**Fix:** Ensure you're `.ocmconfig` file is configured with credentials for the registry.
```
frewilhelm:
```suggestion
**Fix:** Ensure your `.ocmconfig` file is configured with credentials for the registry.
```

website/content/docs/how-to/Sign and Verify/verify-component-version.md
```diff
@@ -0,0 +1,301 @@
...
+| Signer logged in via | `certificateOIDCIssuer` value |
+| --- | --- |
+| Google | `https://accounts.google.com` |
+| GitHub | `https://github.com/login/oauth` |
+| Microsoft | `https://login.microsoftonline.com` |
+
+{{< /callout >}}
+
+{{< callout context="caution" >}}
+Without identity constraints, verification cannot assert **who** signed the component. The verifier rejects specs that omit them.
```
frewilhelm: > The verifier rejects specs that omit them.

What does this mean? It sounds a bit complicated

cli/internal/plugin/builtin/oidc/register.go
morri-son: ok, let me note this down for a follow-up PR

frewilhelm (commented):
This review is only about the pages changed on the website. Semantically it is mostly okay. But as a person that has no prior experience with sigstore and has only a vague concept about how it works, it was very, very hard to understand what this is about. 

I don't believe we should explain sigstore in general in this How To. But maybe we could link pages from sigstore that explain the concept behind cosign, fulcio, rekor, etc.pp.

cli/internal/oidcflow/oidcflow.go
morri-son: hmmm...strange. that was already done yesterday with [this commit](https://github.com/open-component-model/open-component-model/pull/2346/changes/1f87261033d604f368e1325d51eb7b5bed3709fe)
```
// callbackHandler terminates the OAuth redirect on the loopback listener:
// validates state (constant-time) and iss (RFC 9207, permissive when absent),
// then forwards the auth code on codeCh or an error on errCh. Channels are
// non-blocking; duplicate callbacks get 409.
```

website/content/docs/how-to/Sign and Verify/sign-component-version.md
morri-son: nothing that I can quickly influence. This is how the current setup renders links in callouts. It always wraps it right before the link:

```
This is what the verifier's `certificateOIDCIssuer` constraint must match (see [How-to: Verify Component Versions]({{< relref "verify-component-version.md" >}})).
{{< /callout >}}

```

We should create an issue to get this fixed. You're right, it looks ugly :)

website/content/docs/how-to/Sign and Verify/sign-component-version.md
morri-son: removed the complete callout

website/content/docs/how-to/Sign and Verify/sign-component-version.md
morri-son: I added "If the Cosing CLI is not on your PATH or has a too old version, ..." to the callout.

website/content/docs/how-to/Sign and Verify/sign-component-version.md
morri-son: no, I tested this again and it works. Check your .ocmconfig, sigstore-sign.yaml and a properly built OCM CLI. Then a browser should pop up on "com sign cv"

website/content/docs/how-to/Sign and Verify/verify-component-version.md
morri-son: changed to 
```
Without the identity constraints, verification cannot assert **who** signed the component. The signature could be valid but from an untrusted party,
or the component could have been tampered with after signing. Always set constraints that match your trust requirements.
```

morri-son:
@frewilhelm 
> But as a person that has no prior experience with sigstore and has only a vague concept about how it works, it was very, very hard to understand what this is about.

Let me go through the docs again using the glasses of an enduser with no idea of what sigstore is and adopt it accordingly.

jakobmoellerdev:
could we separate out the concept of keyless signing and the how tos? I would love to have the ability to say "this is how keyless signing in OCM works with cosign"

website/content/docs/how-to/Sign and Verify/sign-component-version.md
```diff
@@ -0,0 +1,351 @@
...
+```text
+time=2026-05-18T10:12:03.118+02:00 level=INFO msg="acquiring OIDC identity token" issuer=https://oauth2.sigstore.dev/auth
+time=2026-05-18T10:12:08.402+02:00 level=INFO msg="OIDC identity token acquired"
+time=2026-05-18T10:12:08.601+02:00 level=INFO msg="signing via Sigstore" fulcio=https://fulcio.sigstore.dev rekor=https://rekor.sigstore.dev
+time=2026-05-18T10:12:11.972+02:00 level=INFO msg="signed successfully" name=default digest=91dd197868907487e62872695db1fa7b397fde300bcbae23e24abc188fb147ad hashAlgorithm=SHA-256 normalisationAlgorithm=jsonNormalisation/v4alpha1
+```
+{{< /details >}}
+
+{{< callout context="tip" >}}
+If the Cosing CLI is not on your PATH or has a too old version, the first run downloads and caches the `cosign` binary into `~/.cache/ocm/cosign/...`. Subsequent runs skip the download.
```
frewilhelm:
```suggestion
If the `cosign` CLI is not on your `PATH` variable or has a too old version, the first run downloads and caches the `cosign` binary into `~/.cache/ocm/cosign/...`. Subsequent runs skip the download.
```

website/content/docs/how-to/Sign and Verify/sign-component-version.md
frewilhelm: We were able to reproduce the bug together

website/content/docs/how-to/Sign and Verify/sign-component-version.md
morri-son: I fixed this inside of this PR

website/content/docs/how-to/Sign and Verify/sign-component-version.md
morri-son: ok, my conflict resolution after my last rebase against upstream/main introduced this bug. Fixed now.

morri-son:
@jakobmoellerdev 
> could we separate out the concept of keyless signing and the how tos? I would love to have the ability to say "this is how keyless signing in OCM works with cosign"

After the merge of RSA and Cosign (and your planned GPG) into single docs, my plan was to create an additional tutorial for Sigstore. Would this fit?

cli/docs/reference/ocm_verify_component-version.md
morri-son: now looks like this
```
## Behavior

- --signature selects a single signature by name; without it, every signature on the descriptor is verified
- Signatures are verified concurrently (--concurrency-limit); the command exits non-zero on the first failure
- Default verifier: RSASSA-PSS, resolves the public key from credentials in .ocmconfig
- For Sigstore keyless verification, pass --verifier-spec with SigstoreVerificationConfiguration/v1alpha1 config, resolves trusted root from credentials in .ocmconfig
  ```

File: website/content/docs/how-to/Sign and Verify/sign-component-version.md (hunk 41 above)
Comment (frewilhelm) [reply website/content/docs/how-to/Sign and Verify/sign-component-version.md:266]: can confirm. it worked now!

File: website/content/docs/how-to/Sign and Verify/sign-component-version.md [hunk 45]
```diff
@@ -0,0 +1,355 @@
...
+
+That's the whole signing flow. No private key was generated, none was loaded from disk, and nothing needs to be distributed to verifiers. Your OIDC identity is what the verifier will check.
+
+{{< details "Expected output from signing" >}}
+
+```text
+time=2026-05-18T10:12:03.118+02:00 level=INFO msg="acquiring OIDC identity token" issuer=https://oauth2.sigstore.dev/auth
+time=2026-05-18T10:12:08.402+02:00 level=INFO msg="OIDC identity token acquired"
+time=2026-05-18T10:12:08.601+02:00 level=INFO msg="signing via Sigstore" fulcio=https://fulcio.sigstore.dev rekor=https://rekor.sigstore.dev
+time=2026-05-18T10:12:11.972+02:00 level=INFO msg="signed successfully" name=default digest=91dd197868907487e62872695db1fa7b397fde300bcbae23e24abc188fb147ad hashAlgorithm=SHA-256 normalisationAlgorithm=jsonNormalisation/v4alpha1
```
frewilhelm: This looks "a bit" different on my machine:

```text
$> ocm sign cv   --signer-spec ./sigstore-sign.yaml   /tmp/helloworld/transport-archive//github.com/acme.org/helloworld:1.0.0
digest:
  hashAlgorithm: SHA-256
  normalisationAlgorithm: jsonNormalisation/v4alpha1
  value: 91dd197868907487e62872695db1fa7b397fde300bcbae23e24abc188fb147ad
name: default
signature:
  algorithm: sigstore
  issuer: https://github.com/login/oauth
  mediaType: application/vnd.dev.sigstore.bundle.v0.3+json
  value: ...

time=2026-05-19T17:43:28.524+02:00 level=INFO msg="signed successfully" name=default digest=91dd197868907487e62872695db1fa7b397fde300bcbae23e24abc188fb147ad hashAlgorithm=SHA-256 normalisationAlgorithm=jsonNormalisation/v4alpha1
```

Where are the log statements coming from?

website/content/docs/how-to/Sign and Verify/sign-component-version.md
```diff
@@ -0,0 +1,355 @@
...
+| Aspect | RSA | Sigstore |
+| --- | --- | --- |
+| Before you sign | Generate key pair, configure `.ocmconfig` with file paths | Nothing — just log in when prompted |
+| What proves authorship | Possession of the private key | Your OIDC login (e.g. corporate email) |
+| What the verifier needs | Your public key, distributed somehow | Your expected identity (email + provider) |
+| Audit trail | None unless you build one | Public, automatic |
+
+**Mental model:** your identity is the key. When you sign, you log in with your OIDC provider (Google, GitHub, Microsoft, …). Sigstore issues a short-lived certificate that binds the signature to that identity. The verifier doesn't need a public key from you — they just check that the identity in the certificate is one they trust.
+
+## You'll end up with (Sigstore)
```
frewilhelm: All these headings have `(Sigstore)` as suffix. I believe this is a leftover? (also on the verify guide)

website/content/docs/how-to/Sign and Verify/verify-component-version.md
```diff
@@ -0,0 +1,324 @@
...
+
+**Fix:** Check your network can reach `rekor.sigstore.dev`. If the signature is recent and the network is fine, the signing flow likely failed to record the entry — ask the signer to re-sign.
+
+### Symptom: "certificate expired"
+
+**Cause:** Sigstore certificates are intentionally short-lived (~10 minutes) — they're meant to last only as long as the signing operation. This error means the proof that ties the signature back to the certificate's valid window is missing or broken in the bundle.
+
+**Fix:** Re-fetch the component version (the bundle may be partial). If it's still failing, ask the signer to re-sign.
+
+### Symptom: "identity constraints required"
```
frewilhelm: The actual error is 

```text
Error: SIGNATURE VERIFICATION FAILED: invalid verification config: keyless verification requires both an issuer constraint (CertificateOIDCIssuer or CertificateOIDCIssuerRegexp) and an identity constraint (CertificateIdentity or CertificateIdentityRegexp)
```

I don't believe we should add the whole error as symptom but I was surprised that the term "identity constraints required" was not part of it

website/content/docs/how-to/Sign and Verify/verify-component-version.md
```diff
@@ -0,0 +1,324 @@
...
+```bash
+ocm get cv /tmp/helloworld/transport-archive//github.com/acme.org/helloworld:1.0.0 -o yaml \
+  | yq '.[0].signatures[] | select(.signature.algorithm == "sigstore")'
+```
+
+Look for the signer email and the OIDC issuer URL. Those are exactly what `certificateIdentity` and `certificateOIDCIssuer` are matched against.
+
+## Troubleshooting (Sigstore)
+
+### Symptom: "no matching identity in signing certificate"
```
frewilhelm:
```suggestion
### Symptom: "no matching CertificateIdentity found"
```

website/content/docs/how-to/Sign and Verify/verify-component-version.md
```diff
@@ -0,0 +1,324 @@
...
+
+### Verify a specific signature
+
+If the component carries multiple signatures (e.g. an RSA signature and a Sigstore signature), select one with `--signature`:
+
+```bash
+ocm verify cv \
+  --verifier-spec ./sigstore-verify.yaml \
+  --signature sigstore \
+  ghcr.io/myorg/components//github.com/acme.org/helloworld:1.0.0
```
frewilhelm: In the step for the remote OCI registry for signing we use `ghcr.io/<your-namespace>//github.com/acme.org/helloworld:1.0.0`. I would use this too to make it consistent to use 
```suggestion
  ghcr.io/<your-namespace>//github.com/acme.org/helloworld:1.0.0
```

frewilhelm (commented):
Just some smaller remarks. But the how-to works. Well done! :)

jakobmoellerdev (dismissed):

website/content/docs/how-to/Sign and Verify/sign-component-version.md
morri-son: yeah, that is required, to avoid the markdownlnter errors that report duplicate header titles with the same header size in the same document. Since we're using tabs and steps that check is not meaningful in my opinion. I keep the headers distinct for the time being to get the PR green, but from my point of view we should simply switch off this markdown linter check.

But for the time being I added `<!-- markdownlint-disable-next-line MD024 -->`above the header lines. Then we can omit `(Sigstore)`

website/content/docs/how-to/Sign and Verify/verify-component-version.md
morri-son: ok, adopted the titles and content of the troubleshooting section.

website/content/docs/how-to/Sign and Verify/verify-component-version.md
morri-son: done

website/content/docs/how-to/Sign and Verify/verify-component-version.md
morri-son: done

website/content/docs/how-to/Sign and Verify/sign-component-version.md
morri-son: copied yours into the output field.

cli/cmd/cmd_test.go
matthiasbruns: just checked the test again, the original test is still not back

website/content/docs/how-to/Sign and Verify/verify-component-version.md
```diff
@@ -0,0 +1,336 @@
...
+
+### Symptom: "no matching CertificateIdentity found"
+
+**Cause:** The signature was made by a different identity than what your spec expects — or the same identity but via a different OIDC provider.
+
+The full error names which side did not match. For an identity mismatch:
+
+```text
+Error: SIGNATURE VERIFICATION FAILED: cosign verify-blob failed: exit status 1
+stderr: Error: failed to verify certificate identity: no matching CertificateIdentity found, last error: expected SAN value "nobody@nowhere.invalid", got "gerald.gm.morrison@gmail.com"
```
frewilhelm: > gerald.gm.morrison@gmail.com

Is that your private mail? I would not use it in the docs. Rather use `nobody@nowhere.valid`

website/content/docs/how-to/Sign and Verify/verify-component-version.md
```diff
@@ -0,0 +1,336 @@
...
+  --verifier-spec ./sigstore-verify.yaml \
+  /tmp/helloworld/transport-archive//github.com/acme.org/helloworld:1.0.0
+```
+{{< /tab >}}
+{{< tab "Remote OCI Registry" >}}
+
+```bash
+ocm verify cv \
+  --verifier-spec ./sigstore-verify.yaml \
+  ghcr.io/myorg/components//github.com/acme.org/helloworld:1.0.0
```
frewilhelm:
```suggestion
  ghcr.io/<your namespace>//github.com/acme.org/helloworld:1.0.0
```

website/content/docs/how-to/Sign and Verify/sign-component-version.md
frewilhelm: Nice!

frewilhelm (commented):
The very last nits

website/content/docs/how-to/Sign and Verify/verify-component-version.md
morri-son: ah, thx. copy+pasted the result of my last test. Removed my email.

website/content/docs/how-to/Sign and Verify/verify-component-version.md
morri-son: done

website/content/docs/how-to/Sign and Verify/verify-component-version.md
```diff
@@ -0,0 +1,336 @@
...
+{{< /step >}}
+
+{{< step >}}
+
+### Verify a specific signature (optional)
+
+If the component has multiple signatures, specify which one to verify:
+
+```bash
+ocm verify cv --signature prod ghcr.io/myorg/components//github.com/acme.org/helloworld:1.0.0
```
frewilhelm:
```suggestion
ocm verify cv --signature prod ghcr.io/<your-namespace>//github.com/acme.org/helloworld:1.0.0
```

website/content/docs/how-to/Sign and Verify/verify-component-version.md
```diff
@@ -0,0 +1,336 @@
...
+**Local CTF Archive:**
+
+```bash
+ocm verify cv /tmp/helloworld/transport-archive//github.com/acme.org/helloworld:1.0.0
+```
+
+**Remote OCI Registry:**
+
+```bash
+ocm verify cv ghcr.io/myorg/components//github.com/acme.org/helloworld:1.0.0
```
frewilhelm:
```suggestion
ocm verify cv ghcr.io/<your-namespace>//github.com/acme.org/helloworld:1.0.0
```

website/content/docs/how-to/Sign and Verify/verify-component-version.md
```diff
@@ -0,0 +1,336 @@
...
+  --verifier-spec ./sigstore-verify.yaml \
+  /tmp/helloworld/transport-archive//github.com/acme.org/helloworld:1.0.0
+```
+{{< /tab >}}
+{{< tab "Remote OCI Registry" >}}
+
+```bash
+ocm verify cv \
+  --verifier-spec ./sigstore-verify.yaml \
+  ghcr.io/<my-namespce>//github.com/acme.org/helloworld:1.0.0
```
frewilhelm:
```suggestion
  ghcr.io/<your-namespace>//github.com/acme.org/helloworld:1.0.0
```

frewilhelm (commented):
just found some other instances

cli/cmd/cmd_test.go
morri-son: reset the test to before this PR#: now again happy-path without any config. I added another one that uses the signer-spec.

cli/internal/plugin/builtin/oidc/oidc_plugin.go
morri-son: This is a broader issue we have to tackle first on credential graph side, then consume the changes in the credential plugin.

website/content/docs/how-to/Sign and Verify/verify-component-version.md
morri-son: done

website/content/docs/how-to/Sign and Verify/verify-component-version.md
morri-son: done

website/content/docs/how-to/Sign and Verify/verify-component-version.md
morri-son: done

matthiasbruns (dismissed):
If you nee

matthiasbruns (approved):

jakobmoellerdev (approved):

cli/internal/plugin/builtin/oidc/oidc_plugin.go
jakobmoellerdev: agreed, today is a sync on this, bring it in there please

PR #2352, opened 2026-04-21
fix: update module ocm.software/open-component-model/bindings/go/oci to v0.0.40 (#2331)

frewilhelm:

Adapts all consumers to the breaking change introduced in #2275: `GlobalAccessPolicy` now defaults to `Never`, meaning `localBlob` resources no longer have a `globalAccess` field populated after transfer.

This upgrades `bindings/go/oci` from v0.0.39 to v0.0.40 and updates `go.mod`/`go.sum` across 8 consuming modules (constructor, helm, oci/integration, transfer, transfer/integration, cli, cli/integration, kubernetes/controller).

**Test fixes:**
- `transfer/integration`: Rewrites the `CopyModeAllResources` test — instead of verifying `globalAccess` is non-nil and resolvable in the target registry, it now asserts `globalAccess` is nil and verifies the blob is directly readable from the target repository via `GetLocalResource`
- `cli/integration`: Replaces weak output assertions (checking registry addresses from `globalAccess` image references) with a stronger negative test that proves component-b is unreachable from registry-a without resolver config, confirming resolver routing actually works

**Conformance scenario fixes:**
- RGD: Replaces `resource.access.globalAccess.imageReference.toOCI()` with `resource.access.toOCI()`, which constructs the OCI reference directly from the component's repository spec and local reference digest
- RGD: Uses `image.digest` instead of `image.tag` in HelmRelease values, since `toOCI()` on a `localBlob` returns only a digest reference (no tag)
- RGD: Drops the `tag` field from `additionalStatusFields` — no longer available without `globalAccess`
- Helm chart templates (notes + postgres): Switches image references from `:tag` to `@digest` format
- Helm chart values: Replaces `tag` field with `digest` field

**Documentation:**
- Sovereign cloud ADR: Updates CEL expression examples from `resource.access.globalAccess.imageReference.toOCI()` to `resource.access.toOCI()`
- OCM controllers concept page: Updates CEL example from `resource.access.globalAccess.imageReference.split('/')[0]` to `resource.access.toOCI().registry`

#### How it was tested

- Unit tests: all modules pass
- Integration tests: `transfer/integration` and `cli/integration` pass
- E2E tests: 12/12 pass on Kind cluster
- Conformance scenario: passes with both the current CLI (`0.0.0-main`, which still populates `globalAccess`) and a locally-built CLI using OCI v0.0.40 (which does not)

Fixes #2331


frewilhelm:
@coderabbitai review

morri-son (dismissed):

morri-son (dismissed):

morri-son (dismissed):

matthiasbruns (dismissed):

matthiasbruns (dismissed):

conformance/scenarios/sovereign/components/product/deploy/rgd.yaml
```diff
@@ -172,9 +172,9 @@ spec:
               resource:
                 name: helm-chart
           additionalStatusFields:
-            registry: resource.access.globalAccess.imageReference.toOCI().registry
-            repository: resource.access.globalAccess.imageReference.toOCI().repository
-            digest: resource.access.globalAccess.imageReference.toOCI().digest
+            registry: resource.access.toOCI().registry
+            repository: resource.access.toOCI().repository
+            digest: resource.access.toOCI().digest
```
fabianburth: we should adjust those examples to
```
additionalStatusFields:
  oci: resource.access.toOCI()
```

and then access the fields with `additional.oci.registry`, `additional.oci.repository`, ... .

conformance/scenarios/sovereign/components/notes/deploy/chart/values.yaml
```diff
@@ -3,7 +3,7 @@ replicaCount: 2
 
 image:
   repository: localhost:5001/acme.org/sovereign/notes
-  tag: "1.0.0"
+  digest: ""
```
fabianburth, line 7: why no digest here?

frewilhelm: the notes app depends on the source code (which probably won't change that much), the go module dependencies (which change probably more often), and the Dockerfile which use `golang:1.26-alpine@sha256:f853308...` and `lpine:3.23@sha256:5b10f43..` (which change quite often, see [git history](https://github.com/open-component-model/open-component-model/commits/main/conformance/scenarios/sovereign/components/notes/Dockerfile)). Hence, we would need to update this quite often i think.

conformance/scenarios/sovereign/components/product/deploy/rgd.yaml
frewilhelm: nice catch - I forgot about this

fabianburth (approved):

matthiasbruns (approved):

PR #2360, opened 2026-04-22
feat: typed oci credentials and identity (Phase 2)

### LINKED ISSUE #986: Update oci bindings to work with typed credentials
**Description**

We currently rely on these [consts](https://github.com/open-component-model/open-component-model/blob/2efef844237d3ca01651f92fdeeab12188c055bf/bindings/go/oci/credentials/docker_config.go) to access and store credential values
```go
// CredentialKey constants define the standard keys used in credential maps.
// These keys are used to store and retrieve different types of credentials.
const (
	// CredentialKeyUsername is the key for storing username credentials
	CredentialKeyUsername = "username"
	// CredentialKeyPassword is the key for storing password credentials
	CredentialKeyPassword = "password"
	// CredentialKeyAccessToken is the key for storing access token credentials
	CredentialKeyAccessToken = "accessToken"
	// CredentialKeyRefreshToken is the key for storing refresh token credentials
	CredentialKeyRefreshToken = "refreshToken"
)
```

Those will need to become its own spec within the oci bindings.

You can find the current scheme in here: [bindings/go/oci/spec/credentials/v1](https://github.com/open-component-model/open-component-model/tree/main/bindings/go/oci/spec/credentials/v1)

We also need to update the incoming credentials structure from map[string]string to `runtime.Typed`

***AC***
- [ ] Credentials for oci like Docker/

**Done Criteria**
- [ ] replace key-map access with proper scheme/spec access and parsing
- [ ] replace `map[string]string` credentials with `runtime.Typed` in `bindings/oci`
- [ ] Verify that we have all oci credentials as typed scheme

matthiasbruns:
**Phase 2** typed oci bindings [ADR](https://github.com/open-component-model/open-component-model/blob/main/docs/adr/0018_typed_credentials.md) 

Contributes: https://github.com/open-component-model/ocm-project/issues/986

You can see the whole picture in https://github.com/open-component-model/open-component-model/pull/2148

##### How to test the changes

Unit & integration tests only.

I ran `task init/go.work` with the updated `credentials` binding to see if the update breaks ocm - it does not :)

##### Verification


bindings/go/oci/credentials/docker_config.go
```diff
@@ -17,26 +17,26 @@ import (
...
-	CredentialKeyUsername = "username"
-	// CredentialKeyPassword is the key for storing password credentials
-	CredentialKeyPassword = "password"
-	// CredentialKeyAccessToken is the key for storing access token credentials
-	CredentialKeyAccessToken = "accessToken"
-	// CredentialKeyRefreshToken is the key for storing refresh token credentials
-	CredentialKeyRefreshToken = "refreshToken"
-
+	// CredentialKeyUsername is the key for basic auth username.
+	CredentialKeyUsername = credentialsv1.CredentialKeyUsername
```
matthiasbruns, line 23: moved to another location for reuse, but exporting it again here to not have a breaking change - all of this will be removed sooner or later

bindings/go/oci/spec/credentials/v1/oci_credentials.go
```diff
@@ -0,0 +1,74 @@
...
+	AccessToken  string       `json:"accessToken,omitempty"`
+	RefreshToken string       `json:"refreshToken,omitempty"`
+}
+
+// MustRegisterCredentialType registers OCICredentials/v1 in the given scheme.
+func MustRegisterCredentialType(scheme *runtime.Scheme) {
+	scheme.MustRegisterWithAlias(&OCICredentials{},
+		runtime.NewVersionedType(OCICredentialsType, Version),
+		runtime.NewUnversionedType(OCICredentialsType),
+	)
```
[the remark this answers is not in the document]
matthiasbruns: Fixed! Added the deprecated unversioned enum value to the Type field annotation to match the alias registration in MustRegisterCredentialType. The generated JSON schema now accepts the unversioned alias, which is already supported by the runtime scheme.

```diff
 // +ocm:jsonschema-gen:enum=OCICredentials/v1
+// +ocm:jsonschema-gen:enum:deprecated=OCICredentials
 Type         runtime.Type `json:"type"`
```

Thanks for catching this!

bindings/go/oci/spec/credentials/v1/oci_credentials_test.go
```diff
@@ -0,0 +1,141 @@
...
+	require.NoError(t, scheme.Convert(original, raw))
+
+	restored := &OCICredentials{}
+	require.NoError(t, scheme.Convert(raw, restored))
+
+	assert.Equal(t, original.Username, restored.Username)
+	assert.Equal(t, original.Password, restored.Password)
+	assert.Equal(t, original.AccessToken, restored.AccessToken)
+	assert.Equal(t, original.RefreshToken, restored.RefreshToken)
+}
```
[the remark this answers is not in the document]
matthiasbruns: Fixed — added `assert.Equal(t, original.Type, restored.Type)` to verify the Type field survives the round-trip conversion.

bindings/go/oci/transformer/add_local_resource.go
```diff
@@ -82,11 +81,10 @@ func (t *AddLocalResource) Transform(ctx context.Context, step runtime.Typed) (r
...
 	}
 
-	// Resolve credentials if provider available
 	var creds map[string]string
 	if t.CredentialProvider != nil {
 		if consumerId, err := t.RepoProvider.GetComponentVersionRepositoryCredentialConsumerIdentity(ctx, repoSpec); err == nil {
-			if creds, err = t.CredentialProvider.Resolve(ctx, consumerId); err != nil && !errors.Is(err, credentials.ErrNotFound) { //nolint:staticcheck // SA1019: tracked migration to ResolveTyped in ocm-project#702
+			if creds, err = resolveCredentialsMap(ctx, t.CredentialProvider, consumerId); err != nil {
 				return nil, fmt.Errorf("failed resolving credentials: %w", err)
 			}
```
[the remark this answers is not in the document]
matthiasbruns: This is already handled — `resolveCredentialsMap` returns `nil, nil` when `ResolveTyped` returns `credentials.ErrNotFound` (see credentials.go lines 20-22). The anonymous/public fallback is preserved because `creds` stays nil and `GetComponentVersionRepository` proceeds without credentials.

bindings/go/oci/transformer/add_oci_artifact_test.go
```diff
@@ -65,12 +66,14 @@ func (m *mockResourceRepositoryForAddOCI) GetResourceCredentialConsumerIdentity(
...
-func (m *mockCredentialResolver) Resolve(ctx context.Context, id runtime.Identity) (map[string]string, error) {
+func (m *mockCredentialResolver) Resolve(_ context.Context, _ runtime.Identity) (map[string]string, error) {
 	return map[string]string{"username": "test-user"}, nil
 }
 
 func (m *mockCredentialResolver) ResolveTyped(_ context.Context, _ runtime.Typed) (runtime.Typed, error) {
-	return runtime.Identity{"username": "test-user"}, nil
+	return &ocicredsv1.OCICredentials{
+		Username: "test-user",
+	}, nil
```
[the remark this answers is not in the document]
matthiasbruns: Fixed — `Resolve()` now returns `errors.New("unexpected legacy Resolve call")` so the test will fail if the transformer accidentally uses the deprecated path instead of `ResolveTyped`.

bindings/go/oci/transformer/credentials.go
```diff
@@ -0,0 +1,43 @@
...
+	}
+
+	switch c := typed.(type) {
+	case *ocicredsv1.OCICredentials:
+		return map[string]string{
+			ocicredsv1.CredentialKeyUsername:     c.Username,
+			ocicredsv1.CredentialKeyPassword:     c.Password,
+			ocicredsv1.CredentialKeyAccessToken:  c.AccessToken,
+			ocicredsv1.CredentialKeyRefreshToken: c.RefreshToken,
+		}, nil
```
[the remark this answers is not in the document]
matthiasbruns: Fixed — the `OCICredentials` branch now only inserts non-empty fields into the map. Added `TestResolveCredentialsMap_OCICredentials_SkipsEmptyFields` to verify empty fields are omitted.

bindings/go/oci/transformer/get_local_resource.go
```diff
@@ -72,11 +71,10 @@ func (t *GetLocalResource) Transform(ctx context.Context, step runtime.Typed) (r
...
 	}
 
-	// Resolve credentials if provider available
 	var creds map[string]string
 	if t.CredentialProvider != nil {
 		if consumerId, err := t.RepoProvider.GetComponentVersionRepositoryCredentialConsumerIdentity(ctx, repoSpec); err == nil {
-			if creds, err = t.CredentialProvider.Resolve(ctx, consumerId); err != nil && !errors.Is(err, credentials.ErrNotFound) { //nolint:staticcheck // SA1019: tracked migration to ResolveTyped in ocm-project#702
+			if creds, err = resolveCredentialsMap(ctx, t.CredentialProvider, consumerId); err != nil {
 				return nil, fmt.Errorf("failed resolving credentials: %w", err)
 			}
```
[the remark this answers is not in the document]
matthiasbruns: Same as the `add_local_resource.go` comment — this is already handled. `resolveCredentialsMap` returns `nil, nil` for `ErrNotFound` (credentials.go lines 20-22), so `creds` stays nil and the anonymous access path works.

bindings/go/oci/transformer/credentials.go
```diff
@@ -0,0 +1,51 @@
...
+	if err != nil {
+		if errors.Is(err, credentials.ErrNotFound) {
+			return nil, nil
+		}
+		return nil, err
+	}
+
+	switch c := typed.(type) {
+	case *ocicredsv1.OCICredentials:
+		result := map[string]string{}
```
frewilhelm, line 29: I believe this should be `var result map[string]string` because if c has no field set, `result` would not be `nil`. The same is true below

fabianburth:
The package structure is triggering me a bit xD. Could me make identity a neighbour of credentials?
```
spec
  |
  credentials
      | - identity
      |       |- v1
      |- v1
```

bindings/go/oci/transformer/credentials.go
```diff
@@ -0,0 +1,51 @@
...
+		if c.RefreshToken != "" {
+			result[ocicredsv1.CredentialKeyRefreshToken] = c.RefreshToken
+		}
+		return result, nil
+	case *credconfigv1.DirectCredentials:
+		result := make(map[string]string, len(c.Properties))
+		for k, v := range c.Properties {
+			result[k] = v
+		}
+		return result, nil
```
fabianburth: nit: could instead use `map.Clone` I guess (but then you might have to treat the nil case and return an empty map)

fabianburth (dismissed):

bindings/go/oci/transformer/credentials.go
matthiasbruns: Fixed — now returns `nil` when all fields are empty (using `len(result) == 0` check after building the map).

bindings/go/oci/transformer/credentials.go
matthiasbruns: Fixed — switched to `maps.Clone(c.Properties)` for the DirectCredentials case.

matthiasbruns:
> The package structure is triggering me a bit xD. Could me make identity a neighbour of credentials?
> 
> ```
> spec
>   |
>   credentials
>       | - identity
>       |       |- v1
>       |- v1
> ```

done

bindings/go/oci/transformer/credentials.go
matthiasbruns: done

bindings/go/oci/transformer/credentials.go
matthiasbruns: done

bindings/go/oci/transformer/credentials.go
jakobmoellerdev: ultra nit (really not required to fix) - you now allocate the map on the heap which is usually not desired.

bindings/go/oci/spec/credentials/v1/oci_credentials.go
```diff
@@ -0,0 +1,75 @@
...
+	// CredentialKeyPassword is the key for basic auth password.
+	CredentialKeyPassword = "password"
+	// CredentialKeyAccessToken is the key for OAuth2/bearer access tokens.
+	CredentialKeyAccessToken = "accessToken"
+	// CredentialKeyRefreshToken is the key for OAuth2 refresh tokens.
+	CredentialKeyRefreshToken = "refreshToken"
+
+	// LegacyCredentialKeyAccessToken is the legacy snake_case key for access tokens.
+	//
+	// Deprecated: Use CredentialKeyAccessToken instead.
```
jakobmoellerdev: this should have a todo for removal in the next minor no?

jakobmoellerdev: Now that I think about it, these keys are not persisted anywhere right now and I dont think anyone is using them. I think currently it would be safe to break them. One thing to double check is how ocmv1 keyed these credentials if you didnt do that already.

matthiasbruns: https://github.com/open-component-model/ocm-project/issues/1037
It's a follow up within this epic

bindings/go/oci/transformer/credentials.go
matthiasbruns: this will be removed in a later phase during migration once all of the apis support typed credentials

fabianburth (approved):

bindings/go/oci/spec/identity/v1/register.go
```diff
@@ -0,0 +1,11 @@
...
+
+import "ocm.software/open-component-model/bindings/go/runtime"
+
+// MustRegisterIdentityType registers OCIRegistry/v1 (with unversioned alias) in the given scheme.
+func MustRegisterIdentityType(scheme *runtime.Scheme) {
+	scheme.MustRegisterWithAlias(&OCIRegistryIdentity{},
+		VersionedType,
+		Type, // backward-compat alias
+	)
+}
```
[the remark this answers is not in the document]
matthiasbruns: will be done later in the adr https://github.com/open-component-model/open-component-model/blob/main/docs/adr/0018_typed_credentials.md

bindings/go/oci/transformer/credentials.go
```diff
@@ -0,0 +1,51 @@
...
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"maps"
+
+	"ocm.software/open-component-model/bindings/go/credentials"
+	credconfigv1 "ocm.software/open-component-model/bindings/go/credentials/spec/config/v1"
+	ocicredsv1 "ocm.software/open-component-model/bindings/go/oci/spec/credentials/v1"
```
Skarlso, line 11: Errr, shouldn't then this be including the new identity package then?

matthiasbruns: why? 

```go
func resolveCredentialsMap(ctx context.Context, resolver credentials.Resolver, identity runtime.Typed) (map[string]string, error) {
	typed, err := resolver.ResolveTyped(ctx, identity)
	if err != nil {
		if errors.Is(err, credentials.ErrNotFound) {
			return nil, nil
		}
		return nil, err
	}

```

`resolver.ResolveTyped(ctx, identity)` takes care of identity fu

or what do you mean?

Skarlso (approved):

PR #2361, opened 2026-04-22
fix(workflow): change artifact download 

morri-son:
fix controller workflow and download the chart from the RC release instead of the uploaded GH workflow artifacts.


frewilhelm (dismissed):

.github/workflows/controller-release.yml
```diff
@@ -311,15 +305,25 @@ jobs:
...
           mkdir -p "$work/rc" "$work/final"
           trap 'rm -rf "$work"' EXIT
 
-          tar -xzf "${{ runner.temp }}/rc-chart/chart-${RC_VERSION}.tgz" -C "$work/rc" --strip-components=1
+          # Download RC chart from GitHub release (not workflow artifact)
+          # to avoid retention-period expiry when the environment gate
+          # approval takes longer than one day.
+          gh release download "${RC_TAG}" \
+            --repo "${{ github.repository }}" \
+            --pattern "chart-*.tgz" \
```
Skarlso: Why not download a specific version? Like, should this be the thing that you want to download? `chart-${RC_VERSION}.tgz` Why not just stick that in there as a pattern in order to avoid downloading something that is, in fact, NOT that pattern that you are then later trying to extract:

```
          tar -xzf "$work/chart-${RC_VERSION}.tgz" -C "$work/rc" --strip-components=1
```

piotrjanik (dismissed):
Lgtm

.github/workflows/controller-release.yml
morri-son: oki, fixed like [this](https://github.com/open-component-model/open-component-model/pull/2361/changes/3a0786ef3987596fd28150634827f654190dc090)

Skarlso (approved):

frewilhelm (approved):

PR #2364, opened 2026-04-22
docs: Create docs describing all CI and release workflows and scripts and the website architecture

morri-son:
Our CI / Release and also our website setup is not trivial, but rather complex. This PR adds to documents that describe the CI / release setup and all exieting JS scripts and the architecture of the website setup, including layouts and SCSS details. These docs should also enable us to point AI agents to these specifics when working with those parts.


fabianburth (commented):
Seems way too long to be meaningful.
Humans will never read this, AI agents context will flood if they do. Besides I'm wondering whether this should rather be explicit AGENT.md's.

PR #2368, opened 2026-04-22
feat(oci): push ownership referrer on local artifact upload

### LINKED ISSUE #1024: Ownership Referrer Creation
**Description**

Allow tracing an OCI resource back to the component version that owns it. When a resource is stored by value, attach an ownership referrer to it. The referrer is a metadata-only OCI artifact that points to the resource via `subject` and carries the owning component's name, version, and artifact identity in its annotations.

`pack.ArtifactBlob` in `uploadAndUpdateLocalArtifact` already uploads the resource and returns its OCI descriptor — but currently discards it. This descriptor is what the referrer needs as its `subject`.

**Implementation notes**

- Add annotation constants:
  - `software.ocm.component.name` — component name (plain string)
  - `software.ocm.component.version` — component version (plain string)
  - `software.ocm.artifact` — artifact identity and kind (JCS-canonicalized JSON, e.g. `{"identity":{"name":"my-resource"},"kind":"resource"}`)
  - Artifact type `application/vnd.ocm.ownership.v1+json` — for Referrers API filtering
- Add a helper to build JCS-canonicalized `software.ocm.artifact` JSON from artifact identity and kind.
- The referrer has no real content — empty config, empty layer, a `subject` pointing to the resource, and ownership annotations. The `subject` field makes registries index it as a referrer. The `artifactType` distinguishes it from other referrers (cosign, SBOMs, etc.). Use `oras.PackManifest` to build and push it:
  ```go
  desc, err := oras.PackManifest(ctx, store, oras.PackManifestVersion1_1, ownership.ArtifactType, oras.PackManifestOptions{
      Subject:             &resourceDesc,
      ManifestAnnotations: ownershipAnnotations,
  })
  ```
- In `uploadAndUpdateLocalArtifact` (`repository.go`), capture the descriptor returned by `pack.ArtifactBlob` and pass it as `Subject` to `oras.PackManifest`.
- ORAS handles Referrers API vs tag fallback automatically.

**Expected referrer manifest** (discoverable via `oras discover`):

```json
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.oci.image.manifest.v1+json",
  "artifactType": "application/vnd.ocm.ownership.v1+json",
  "config": {
    "mediaType": "application/vnd.oci.empty.v1+json",
    "digest": "sha256:44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a",
    "size": 2
  },
  "layers": [
    {
      "mediaType": "application/vnd.oci.empty.v1+json",
      "digest": "sha256:44136fa355b3678a1146ad16f7e8649e94fb4fc21fe77e8310c060f61caaff8a",
      "size": 2
    }
  ],
  "subject": {
    "mediaType": "application/vnd.oci.image.manifest.v1+json",
    "digest": "sha256:<resource-manifest-digest>",
    "size": 1769
  },
  "annotations": {
    "software.ocm.component.name": "ocm.software/my-component",
    "software.ocm.component.version": "1.0.0",
    "software.ocm.artifact": "{\"identity\":{\"name\":\"my-image\"},\"kind\":\"resource\"}"
  }
}
```

**Done Criteria**
- [ ] Ownership referrer with correct `subject` and annotations is pushed for each by-value resource upload
- [ ] By-reference resources and existing OCI push/pull flows are unaffected
- [ ] Referrer is discoverable via `oras discover --artifact-type application/vnd.ocm.ownership.v1+json`

piotrjanik:

Lets you trace an OCI artifact back to the OCM component version that
shipped it. Today, once a resource lands in a registry it looks like any
other image — there is no standardised reverse lookup.

When `oci.WithOwnershipReferrerPolicy(OwnershipReferrerPolicyEnabled)` is
set (default off), `AddLocalResource` also pushes a minimal OCI 1.1
referrer manifest (`artifactType: application/vnd.ocm.software.ownership.v1+json`)
whose `subject` points at the just-uploaded resource and whose annotations
carry the owning component name, version, and artifact identity (ADR 0016).
The resource manifest itself is untouched, so cosign / Notary signatures
stay valid. The referrer omits `org.opencontainers.image.created`, so
re-runs of `ocm add cv` collapse to a single referrer at the registry.

Fixes: https://github.com/open-component-model/ocm-project/issues/1024


bindings/go/oci/spec/annotations/ownership.go
```diff
@@ -0,0 +1,25 @@
...
+	// name on an ownership referrer or asset manifest.
+	OwnershipComponentName = "software.ocm.component.name"
+
+	// OwnershipComponentVersion is an annotation that records the plain
+	// component version on an ownership referrer or asset manifest.
+	//
+	// Distinct from [OCMComponentVersion] in annotations.go
+	// ("software.ocm.componentversion"), which encodes "<name>:<version>" as a
+	// single legacy string.
+	OwnershipComponentVersion = "software.ocm.component.version"
```
matthiasbruns, line 21: we wont end up with both on the refererer annotations, would we?

bindings/go/constructor/construct.go
```diff
@@ -711,7 +711,11 @@ func addColocatedResourceLocalBlob(
 
 	uploaded, err := repo.AddLocalResource(ctx, component, version, descResource, data)
 	if err != nil {
-		return nil, fmt.Errorf("error adding local resource %q based on input type %q as local resource to component %q : %w", resource.ToIdentity(), resource.Input.GetType(), component, err)
+		inputType := "<none>"
```
matthiasbruns: I have never seen something like this `<none>` in our codebase - cant we just not set it?

bindings/go/oci/ownership_referrer.go
```diff
@@ -0,0 +1,77 @@
...
+// (sorted keys), the struct fields are already alphabetical, and leaves are
+// strings — so encoding/json's defaults match RFC 8785 for this schema.
+func marshalOwnershipArtifact(identity runtime.Identity, kind annotations.ArtifactKind) (string, error) {
+	payload := struct {
+		Identity runtime.Identity         `json:"identity"`
+		Kind     annotations.ArtifactKind `json:"kind"`
+	}{Identity: identity, Kind: kind}
+	data, err := json.Marshal(payload)
+	if err != nil {
+		return "", err
```
matthiasbruns: can we wrap this err?

bindings/go/oci/ownership_referrer.go
```diff
@@ -0,0 +1,77 @@
...
+
+	return nil
+}
+
+// marshalOwnershipArtifact serialises the {identity, kind} value stored under
+// [annotations.ArtifactAnnotationKey] on an ownership referrer (ADR 0015).
+// Output is JCS-canonical by construction: identity is map[string]string
+// (sorted keys), the struct fields are already alphabetical, and leaves are
+// strings — so encoding/json's defaults match RFC 8785 for this schema.
+func marshalOwnershipArtifact(identity runtime.Identity, kind annotations.ArtifactKind) (string, error) {
```
matthiasbruns: name misleading - in the end its used as `ArtifactAnnotationKey`

bindings/go/oci/ownership_referrer.go
```diff
@@ -0,0 +1,77 @@
...
+	payload := struct {
+		Identity runtime.Identity         `json:"identity"`
+		Kind     annotations.ArtifactKind `json:"kind"`
+	}{Identity: identity, Kind: kind}
+	data, err := json.Marshal(payload)
+	if err != nil {
+		return "", err
+	}
+	return string(data), nil
+}
```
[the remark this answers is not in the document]
matthiasbruns: what the rabbit says

bindings/go/oci/spec/annotations/ownership.go
piotrjanik: No. Ownership manifest brings 3 different annotations.

bindings/go/constructor/construct.go
fabianburth: generally seems like a lot of code for an error message

bindings/go/oci/repository.go
```diff
@@ -454,6 +466,12 @@ func (repo *Repository) uploadAndUpdateLocalArtifact(ctx context.Context, compon
 		return fmt.Errorf("failed to pack resource blob: %w", err)
 	}
 
+	if resource, ok := artifact.(*descriptor.Resource); ok {
+		if err := pushOwnershipReferrer(ctx, store, desc, resource, component, version); err != nil {
+			return fmt.Errorf("failed to push ownership referrer: %w", err)
+		}
+	}
```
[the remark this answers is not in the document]
fabianburth: Thinking about this, I'm generally wondering whether ownership artifacts should really be something that is hard coded in our ocm primitives. I feel like this should rather be an opt-in concept. I doubt every ocm user is interested potentially obsolete oci artifacts.

bindings/go/oci/ownership_referrer.go
piotrjanik: @matthiasbruns renamed

bindings/go/oci/ownership_referrer.go
piotrjanik: @matthiasbruns yesss

matthiasbruns (dismissed):
lgtm

bindings/go/oci/spec/annotations/ownership.go
```diff
@@ -0,0 +1,25 @@
...
+// Ownership referrer constants. See docs/adr/0015_ownership_annotations.md.
+//
+// An ownership referrer is a minimal OCI manifest whose subject points to a
+// resource manifest and whose annotations record the owning component version
+// and the artifact identity.
+const (
+	// OwnershipArtifactType is the OCI artifactType used for ownership
+	// referrers. It enables filtering via the Referrers API
+	// (GET /v2/<name>/referrers/<digest>?artifactType=...).
+	OwnershipArtifactType = "application/vnd.ocm.ownership.v1+json"
```
frewilhelm: is there a reason why you omit "software" part?

bindings/go/oci/spec/annotations/ownership.go
```diff
@@ -0,0 +1,25 @@
...
+	// OwnershipComponentName is an annotation that records the plain component
+	// name on an ownership referrer or asset manifest.
+	OwnershipComponentName = "software.ocm.component.name"
+
+	// OwnershipComponentVersion is an annotation that records the plain
+	// component version on an ownership referrer or asset manifest.
+	//
+	// Distinct from [OCMComponentVersion] in annotations.go
+	// ("software.ocm.componentversion"), which encodes "<name>:<version>" as a
+	// single legacy string.
```
frewilhelm: what do you mean by legacy?

bindings/go/oci/ownership_referrer.go
```diff
@@ -0,0 +1,90 @@
...
+		Subject: &subject,
+		ManifestAnnotations: map[string]string{
+			annotations.OwnershipComponentName:    component,
+			annotations.OwnershipComponentVersion: version,
+			annotations.ArtifactAnnotationKey:     artifactValue,
+			// Pin org.opencontainers.image.created so the referrer manifest
+			// digest is a pure function of its ownership data. Without this,
+			// oras.PackManifest defaults the value to time.Now() and a retried
+			// AddLocalResource produces a second, distinct referrer for the
+			// same subject — breaking idempotency on the Referrers API.
```
frewilhelm: Why do we do that? I believe it could be important for people to know when the annotation was created.

bindings/go/oci/ownership_referrer_test.go
```diff
@@ -0,0 +1,262 @@
...
+// a referrer, even though the subject would technically be valid.
+func TestRepository_AddLocalSource_DoesNotPushOwnershipReferrer(t *testing.T) {
+	r := require.New(t)
+	ctx := t.Context()
+
+	fs, err := filesystem.NewFS(t.TempDir(), os.O_RDWR)
+	r.NoError(err)
+	archive := ctf.NewFileSystemCTF(fs)
+	store := ocictf.NewFromCTF(archive)
+	repo := Repository(t, ocictf.WithCTF(store))
```
frewilhelm: In this test the referrer is not pushed because the policy is `None` (default). If you want to test, if the condition works as expected for local sources you need to enable the policy, e.g. 
`repo := Repository(t, ocictf.WithCTF(store), oci.WithOwnershipReferrerPolicy(oci.OwnershipReferrerPolicyEnabled))`.

Actually - what happens if the policy is enabled for a local source? We do not error out, do we? We just silently omit it

frewilhelm (commented):
I believe we should update the `doc.go` as well

bindings/go/oci/ownership_referrer_test.go
piotrjanik: I have refactored this test. Please check.

bindings/go/oci/ownership_referrer.go
piotrjanik: agree, changed

bindings/go/oci/spec/annotations/ownership.go
piotrjanik: It was generated. Good catch, I did not notice legacy.

bindings/go/oci/spec/annotations/ownership.go
piotrjanik: You're right, it's always `software.ocm...`, so I added `soatware` here and in the ADR for consistency.

frewilhelm: What was the reason for moving it to `bindings/go/oci/ownership_referrer.go`?

bindings/go/oci/ownership_referrer.go
frewilhelm: What happens if the manifest already exists? Would it now overwrite the manifest? Can we add a test to make sure it behaves like we want to

bindings/go/oci/ownership_referrer_test.go
frewilhelm: Did you delete the test? Because of the force-pushes on one commit I cannot see the changes :D

frewilhelm: Do we have a test for local source now? I can't find any test

bindings/go/oci/spec/annotations/ownership.go
piotrjanik: It's a type not annotation

bindings/go/oci/ownership_referrer.go
Skarlso: I'm also curious what happens if we keep repackaging it over and over on every push. Shouldn't we first check if it's already owned or not..?

bindings/go/oci/ownership_referrer.go
```diff
@@ -0,0 +1,89 @@
...
+
+	meta := resource.GetElementMeta()
+	artifactValue, err := marshalArtifactAnnotation(meta.ToIdentity(), annotations.ArtifactKindResource)
+	if err != nil {
+		return fmt.Errorf("failed to build ownership artifact annotation: %w", err)
+	}
+
+	opts := oras.PackManifestOptions{
+		Subject: &subject,
+		ManifestAnnotations: map[string]string{
```
Skarlso: Somewhere where we set annotations already there is a thing for this to be set first by oras or something like that.

Because otherwise, if you call PackManifest, that has this:

```go
	annotations, err := ensureAnnotationCreated(opts.ManifestAnnotations, ocispec.AnnotationCreated)
	if err != nil {
		return ocispec.Descriptor{}, err
	}
```

Now if AnnotationCreated is not there, it will literally do a time.Now(). And since we are re-packaging it on every push if I get this right, it will always end up being a different timestamp ending up with a different Digest.

Skarlso: This might even be more of a problem if we push the artifact successfully, then push a referral, that fails. The retry is a no-op on resource push, same digest, but the referral is pushed with a new timestamp and a new digest. I would like to see some tests for this once it's fixed if possible.

Skarlso (commented):
One remark that I think might be problematic.

bindings/go/oci/ownership_referrer_test.go
matthiasbruns: can we please stop force pushing if its not correcting an error that you did a few seconds ago? its hard to follow while reviewing :)

bindings/go/oci/repository.go
```diff
@@ -454,6 +470,12 @@ func (repo *Repository) uploadAndUpdateLocalArtifact(ctx context.Context, compon
 		return fmt.Errorf("failed to pack resource blob: %w", err)
 	}
 
+	if resource, ok := artifact.(*descriptor.Resource); ok && repo.ownershipReferrerPolicy == OwnershipReferrerPolicyEnabled {
+		if err := pushOwnershipReferrer(ctx, store, desc, resource, component, version); err != nil {
+			return fmt.Errorf("failed to push ownership referrer: %w", err)
+		}
+	}
```
frewilhelm: Just from an architecturally point of view, I am unsure if this call belongs in `uploadAndUpdateLocalArtifact` because I believe this should be source/resource agnostic. Would it make sense to put it into `AddLocalResource` only after the call of `uploadAndUpdateLocalArtifact`?

bindings/go/oci/repository.go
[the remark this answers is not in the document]
piotrjanik: @fabianburth 
I made it opt-in. Currently, it is disabled, and integration tests enable it before running.
I would keep it as is and create a follow-up task to add a config option that allows users to enable that feature.

bindings/go/oci/ownership_referrer.go
piotrjanik: The timestamp has been removed because we do not assign a Creation Time for CVs or resources. 
If a timestamp was available on a CV descriptor (or resource), the same issue would occur there. 
Since descriptors are not annotated with `org.opencontainers.image.created`, I believe the referrer manifest should be handled similarly.

The current code pushes the manifest without this annotation, which also means that no matter how many times you push this referrer manifest, there will be only one in the repository, as the digest does not change.

bindings/go/oci/ownership_referrer.go
piotrjanik: replied above

bindings/go/oci/ownership_referrer.go
```diff
@@ -0,0 +1,109 @@
...
+
+// pushOwnershipReferrer pushes an OCI referrer manifest linking subject to
+// the owning component version, per docs/adr/0015_ownership_annotations.md.
+//
+// The manifest is built by hand rather than via [oras.PackManifest] because
+// PackManifest unconditionally stamps org.opencontainers.image.created (caller
+// value or time.Now()), which would make the digest non-deterministic.
+// Omitting that annotation is spec-legal — created is optional in the OCI
+// image-spec — so identical inputs produce identical bytes and re-running
+// `ocm add cv` is a no-op at the registry instead of a new referrer per run.
```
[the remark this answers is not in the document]
piotrjanik: @coderabbitai  Calling store.Push on a manifest goes through ORAS's pushWithIndexing, which is the OCI 1.1 referrers-tag-schema fallback handler.

bindings/go/oci/repository.go
piotrjanik: good idea, just moved that

bindings/go/oci/ownership_referrer_test.go
piotrjanik: @frewilhelm I have moved referrer logic to AddLocalResource

bindings/go/oci/ownership_referrer.go
```diff
@@ -0,0 +1,109 @@
...
+func pushOwnershipReferrer(ctx context.Context, store spec.Store, subject ociImageSpecV1.Descriptor, resource *descriptor.Resource, component, version string) error {
+	if !introspection.IsOCICompliantManifest(subject) {
+		return nil
+	}
+
+	meta := resource.GetElementMeta()
+	artifactValue, err := marshalArtifactAnnotation(meta.ToIdentity(), annotations.ArtifactKindResource)
```
Skarlso: ToIdentity will also include extraIdentity. So platform information and any extra values and informations are also put into the subject. I think we need to canonicalize the JSON from that as well, right? Since ordering matters. Can you add a test for that please?

Other then that this is 👍

piotrjanik: The JSON is already canonical — encoding/json sorts map keys alphabetically, so extraIdentity values get ordered correctly.

Anyway, I added more tests :)

Skarlso (dismissed):

bindings/go/oci/ownership_referrer.go
```diff
@@ -0,0 +1,109 @@
...
+//
+// The manifest is built by hand rather than via [oras.PackManifest] because
+// PackManifest unconditionally stamps org.opencontainers.image.created (caller
+// value or time.Now()), which would make the digest non-deterministic.
+// Omitting that annotation is spec-legal — created is optional in the OCI
+// image-spec — so identical inputs produce identical bytes and re-running
+// `ocm add cv` is a no-op at the registry instead of a new referrer per run.
+func pushOwnershipReferrer(ctx context.Context, store spec.Store, subject ociImageSpecV1.Descriptor, resource *descriptor.Resource, component, version string) error {
+	if !introspection.IsOCICompliantManifest(subject) {
+		return nil
```
frewilhelm: So, if the subject is not an OCI compliant manifest we silently fail, without noticing the user are we? Admittedly, we create the `subject` and it should be compliant. But if we introduce a bug somewhere and the `subject` is not OCI compliant anymore, then we would just return nil and the user would assume that the referrer was created but it was not.

Would it make sense to error out here?

frewilhelm (commented):
Overall this looks good to me. However, I got the impression that we mostly test happy-paths. Would it make sense to add another integration test with some "unhappy" paths?

bindings/go/oci/ownership_referrer.go
piotrjanik: @frewilhelm  I added slog.DebugContext

Skarlso (dismissed):

bindings/go/oci/repository.go
```diff
@@ -227,13 +241,23 @@ func (repo *Repository) AddLocalResource(
...
 	resource = resource.DeepCopy()
 
-	if err := repo.uploadAndUpdateLocalArtifact(ctx, component, version, resource, b); err != nil {
+	desc, store, err := repo.uploadAndUpdateLocalArtifact(ctx, component, version, resource, b)
+	if err != nil {
 		return nil, err
 	}
 
+	if repo.ownershipReferrerPolicy == OwnershipReferrerPolicyEnabled {
+		if err := pushOwnershipReferrer(ctx, store, desc, resource, component, version); err != nil {
```
jakobmoellerdev: I think this might not be the correct way to push for the referrer. I think you should instead modify the existing graph traversal logic inside the upload and have it detect and transfer the referrer. Main reason being that you currently do a sequential upload of the referrer instead of a concurrent one during the upload. This will increase latency unnecessarily

piotrjanik: @jakobmoellerdev yes but this task does not address ocm transfer but ocm add cv

bindings/go/oci/ownership_referrer.go
piotrjanik: @frewilhelm @Skarlso can we resolve this one?

matthiasbruns (dismissed):

bindings/go/oci/repository.go
jakobmoellerdev: add cv is still a transfer of the component version through copy graph so it does not matter which command this is coming from I believe

jakobmoellerdev: also were talking about adding of resources here so we have independent OCI images as graphs

piotrjanik: https://github.com/oras-project/oras-go/blob/763cf5e4c812d6eb295ba9008d1d5ae377711b98/copy.go#L176 that graph?

jakobmoellerdev: yes correct. the copy function by default only considers manifest / index references directly and no referrers.

jakobmoellerdev: This is the function that could be used for this extension I believe: https://github.com/oras-project/oras-go/blob/763cf5e4c812d6eb295ba9008d1d5ae377711b98/copy.go#L120

bindings/go/oci/ownership_referrer.go
```diff
@@ -0,0 +1,119 @@
...
+// than \u-encoded — which RFC 8785 requires but encoding/json does not do by
+// default.
+func marshalArtifactAnnotation(identity runtime.Identity, kind annotations.ArtifactKind) (string, error) {
+	payload := struct {
+		Identity runtime.Identity         `json:"identity"`
+		Kind     annotations.ArtifactKind `json:"kind"`
+	}{Identity: identity, Kind: kind}
+	var buf bytes.Buffer
+	enc := json.NewEncoder(&buf)
+	enc.SetEscapeHTML(false)
```
jakobmoellerdev: Why do you need to do jcs canonicalization manually here? We have libraries for this already in the code so maybe you could use that

bindings/go/oci/tar/blob_io.go
```diff
@@ -146,6 +153,13 @@ func proxyOCIStoreWithTopLevelDescriptor(ctx context.Context, idx int, ociStore
 		return ociImageSpecV1.Descriptor{}, nil, fmt.Errorf("failed to read top level descriptor stream: %w", err)
 	}
 
+	var attachments []ociImageSpecV1.Descriptor
+	var attachmentsStore content.ReadOnlyStorage
+	if opts.Attachments != nil {
+		if attachments, attachmentsStore, err = opts.Attachments(ctx, topLevelDesc); err != nil {
```
jakobmoellerdev: to me it feels like too much of an indirection to have an attachments func even though this gives a lot of flexibility. Would it not be sufficient to simply pass in the attachments a list of subject manifests? I dont think computation of the subject manifest should be expensive so it wouldnt need to be within a meta-function like here

bindings/go/oci/tar/blob_io.go
```diff
@@ -157,9 +171,17 @@ func proxyOCIStoreWithTopLevelDescriptor(ctx context.Context, idx int, ociStore
 		}
 		opts.FindSuccessors = func(ctx context.Context, fetcher content.Fetcher, desc ociImageSpecV1.Descriptor) ([]ociImageSpecV1.Descriptor, error) {
 			if content.Equal(desc, topLevelDesc) {
-				return append([]ociImageSpecV1.Descriptor{manifest.Config}, manifest.Layers...), nil
+				return append(append([]ociImageSpecV1.Descriptor{manifest.Config}, manifest.Layers...), attachments...), nil
```
jakobmoellerdev: I think this is a very complicated syntax and order does not matter so you can probably write `append(manifest.Layers, manifest.Config, attachments...)`

bindings/go/oci/tar/blob_io.go
```diff
@@ -157,9 +171,17 @@ func proxyOCIStoreWithTopLevelDescriptor(ctx context.Context, idx int, ociStore
...
 			}
-			return content.Successors(ctx, ociStore, desc)
+			successors, err := content.Successors(ctx, fetcher, desc)
+			if err != nil {
+				return nil, err
+			}
+			// Drop the attachment's Subject pointing back at root; otherwise
+			// CopyGraph would loop back into root.
+			return slices.DeleteFunc(successors, func(d ociImageSpecV1.Descriptor) bool {
```
jakobmoellerdev: why is this loop a problem. The CAS would be able to simply check that it already exists right? Not sure this should be handled by this function. See the proxy store in copy graph for how the dedup works

bindings/go/oci/tar/blob_io.go
```diff
@@ -171,36 +193,58 @@ func proxyOCIStoreWithTopLevelDescriptor(ctx context.Context, idx int, ociStore
 		}
 		opts.FindSuccessors = func(ctx context.Context, fetcher content.Fetcher, desc ociImageSpecV1.Descriptor) ([]ociImageSpecV1.Descriptor, error) {
 			if content.Equal(desc, topLevelDesc) {
-				return index.Manifests, nil
+				return append(slices.Clone(index.Manifests), attachments...), nil
```
jakobmoellerdev: why do we clone the manifests here?

bindings/go/oci/tar/blob_io.go
```diff
@@ -171,36 +193,58 @@ func proxyOCIStoreWithTopLevelDescriptor(ctx context.Context, idx int, ociStore
...
+			}
+			successors, err := content.Successors(ctx, fetcher, desc)
+			if err != nil {
+				return nil, err
 			}
-			return content.Successors(ctx, ociStore, desc)
+			// Drop the attachment's Subject pointing back at root; otherwise
+			// CopyGraph would loop back into root.
+			return slices.DeleteFunc(successors, func(d ociImageSpecV1.Descriptor) bool {
```
jakobmoellerdev: this is code duplication from the above path

bindings/go/oci/tar/blob_io.go
```diff
@@ -171,36 +193,58 @@ func proxyOCIStoreWithTopLevelDescriptor(ctx context.Context, idx int, ociStore
...
 	}
 	return topLevelDesc, proxy, nil
 }
 
 type descriptorStoreProxy struct {
-	raw  []byte
-	desc ociImageSpecV1.Descriptor
+	raw         []byte
+	desc        ociImageSpecV1.Descriptor
+	attachments content.ReadOnlyStorage // optional, consulted before the base store
```
jakobmoellerdev: why are the attachments coming from a different store than the base? the annotation would be from the same store as source no?

bindings/go/oci/tar/blob_io.go
piotrjanik: imo it looks like that:
Step 1: Main manifest -> Layers, Config, Referrer
Step 2: Layers
Step 3: Config
Step 4: Referrer -> Main manifest through subject

Something would have to prevent Findsuccessors from processing the Main manifest, as it has already been processed.

bindings/go/oci/tar/blob_io.go
piotrjanik: yes, I will extract FindSuccessors once the approach is confirmed

jakobmoellerdev (commented):
nice, much cleaner implementation now. we still have to take a look at the looping behavior but apart from that this looks great already

bindings/go/oci/tar/blob_io.go
piotrjanik: It actually might be a deadlock related to `done, committed := tracker.TryCommit(node)`

https://github.com/oras-project/oras-go/blob/7b1e83d7e4e5007e3a6ea95eee435a1f02d23f7b/copy.go#L217
https://github.com/oras-project/oras-go/blob/7b1e83d7e4e5007e3a6ea95eee435a1f02d23f7b/copy.go#L261

CopyGraph:
1. Starts with the main manifest. Creates a done channel, which is closed at the end (defer)
2. Then goes through successors
3. Eventually, it processes the successors of the referrer (main manifest), and it hangs here

matthiasbruns: can we have a huddle today to debug this together?

piotrjanik: I think the issue occurs here because, for the top level descriptor, we provide 3 successors/descriptors: layout(s), config, and referrer.
The referrer differs significantly from the others because it points upward on the graph, whereas the rest point downward.
`content.Successors` creates a channel(TryCommit) to hold current descriptor till all sucessors are processed.

In this scenario, the top-level descriptor pauses until its children are processed. Meanwhile, one of these children—the referrer—waits for the subject to be processed, which turns out to be the top-level descriptor itself.

Currently, for a descriptor with `artifactType` set, I remove the top-level descriptor from successors.

bindings/go/oci/tar/blob_io.go
```diff
@@ -155,12 +173,8 @@ func proxyOCIStoreWithTopLevelDescriptor(ctx context.Context, idx int, ociStore
 		if err := opts.MutateParentFunc(&topLevelDesc); err != nil {
 			return ociImageSpecV1.Descriptor{}, nil, fmt.Errorf("failed to mutate manifest descriptor before copy: %w", err)
 		}
-		opts.FindSuccessors = func(ctx context.Context, fetcher content.Fetcher, desc ociImageSpecV1.Descriptor) ([]ociImageSpecV1.Descriptor, error) {
-			if content.Equal(desc, topLevelDesc) {
-				return append([]ociImageSpecV1.Descriptor{manifest.Config}, manifest.Layers...), nil
-			}
-			return content.Successors(ctx, ociStore, desc)
-		}
+		rootChildren := slices.Concat([]ociImageSpecV1.Descriptor{manifest.Config}, manifest.Layers, referrerDescriptors)
```
jakobmoellerdev: this definetly needs some comment inline

jakobmoellerdev: also do you really require concat here?

bindings/go/oci/tar/blob_io.go
```diff
@@ -169,38 +183,120 @@ func proxyOCIStoreWithTopLevelDescriptor(ctx context.Context, idx int, ociStore
...
+	}
+}
+
+// mediaTypeOCIArtifactManifest is the deprecated OCI artifact manifest (image-spec v1.1.0-rc1/rc2); the oras-go constant lives in an internal package.
+const mediaTypeOCIArtifactManifest = "application/vnd.oci.artifact.manifest.v1+json"
+
+// successorsWithoutSubject works like [content.Successors] but never returns the Subject field.
+// Manifest types without a Subject pass straight through.
+func successorsWithoutSubject(ctx context.Context, fetcher content.Fetcher, node ociImageSpecV1.Descriptor) ([]ociImageSpecV1.Descriptor, error) {
+	switch node.MediaType {
```
jakobmoellerdev, line 266: this is missing a lot of types related to docker and could cuase issues no?

bindings/go/oci/tar/blob_io.go
```diff
@@ -80,9 +81,21 @@ func copyToOCILayoutInMemoryAsync(ctx context.Context, src content.ReadOnlyStora
...
+// follow by default.
+type ReferrersFunc func(ctx context.Context, top ociImageSpecV1.Descriptor) ([]Referrer, error)
+
 type CopyOCILayoutWithIndexOptions struct {
 	oras.CopyGraphOptions
 	MutateParentFunc func(*ociImageSpecV1.Descriptor) error
+	ReferrersFunc    []ReferrersFunc
 }
 
 func CopyOCILayoutWithIndex(ctx context.Context, dst content.Storage, src blob.ReadOnlyBlob, opts CopyOCILayoutWithIndexOptions) (top ociImageSpecV1.Descriptor, err error) {
```
jakobmoellerdev, line 122: this function now needs much more docs

bindings/go/oci/tar/blob_io.go
piotrjanik: I am providing custom handling for these manifests that may have a subject; the remaining types are handled at the end by calling content.Successors.

bindings/go/oci/tar/blob_io.go
piotrjanik: Otherwise, I would have to append twice, as referrerDescriptors is a slice. I assumed that there might be more referrers in future, so this logic allows adding more.

bindings/go/oci/tar/blob_io.go
```diff
@@ -80,11 +81,44 @@ func copyToOCILayoutInMemoryAsync(ctx context.Context, src content.ReadOnlyStora
...
+//     top-level descriptor before copy. Typical use: inject annotations onto
+//     the root manifest/index. Required.
+//   - [CopyOCILayoutWithIndexOptions.ReferrersFunc] returns extra [Referrer]s
+//     to be pushed alongside the artifact in the same traversal — e.g. OCI
+//     referrer manifests, which [oras.CopyGraph]'s default `FindSuccessors`
+//     does not follow via the `subject` field. Each returned referrer is
+//     served from an in-memory proxy and injected as an additional child of
+//     the root.
+//
+// Returns the descriptor of the root that was copied.
```
jakobmoellerdev, line 121: nice docs!

jakobmoellerdev (dismissed):

matthiasbruns (approved):
great tests, well written and documented - good job

Skarlso (approved):
nice, well done! :)

PR #2387, opened 2026-04-24
feat: automate sprint hygiene for stale Needs Refinement issues

### LINKED ISSUE #1007: Automate the process of updating issues once their sprint setting expires
**Description**

Previously this has been done manually, but we would like to have automation around this. Something that runs every night, check issues that have sprint settings that expired and are in Needs Refinement state.

These are there probably because we just couldn't get to them in time. And since we didn't update the sprint they stayed there and we forgot about them.

Just use a plain graphql query in the github action through the gh cli.

Here are a bunch of examples on gh cli graphql usage: https://cli.github.com/manual/gh_api
and the GraphQL doc: https://docs.github.com/en/graphql

**Done Criteria**
- [ ] Issues no longer have to be manually moved

matthiasbruns:
## What this PR does / why we need it

Adds a nightly GitHub Action that automatically moves stale "Needs Refinement" issues to the current sprint when their assigned sprint has expired.

Previously this was done manually — issues with expired sprints would stay in "Needs Refinement" and get forgotten. This automation runs every night at 3 AM UTC and:

1. Fetches the project board configuration (Status + Sprint fields) via GraphQL
2. Uses server-side filtering (`sprint:<@current status:"..."`) to efficiently find stale items
3. Updates each item's sprint to the current one
4. Leaves a comment on each issue documenting the automatic move

The script is reusable across projects via the `projectNumber` parameter, and all field names (including the "Needs Refinement" status) are looked up dynamically — nothing is hardcoded.

A local runner (`sprint-hygiene.local.js`) is included for testing with `--apply`, `--limit`, `--project`, and `--org` flags. It defaults to dry-run mode for safety.

**Token**: Uses the existing OCMBOT GitHub App (`OCMBOT_APP_ID` / `OCMBOT_PRIV_KEY`). The app needs "Organization projects: Read & write" permission if not already configured.

## Which issue(s) this PR fixes

Fixes open-component-model/ocm-project#1007

### How to test the changes

```bash
cd .github/scripts
npm install
# Dry-run (default) — shows what would be updated
node sprint-hygiene.local.js
# Update a single issue for testing
node sprint-hygiene.local.js --apply --limit 1
# Run unit tests
node sprint-hygiene.test.js
```

### Tested with

- open-component-model/ocm#1173 — sprint updated from "49 - Gerald wants Sprints!" to "Sprint 61"
- open-component-model/ocm#1131 — sprint updated from "52 - API Apocalypse" to "Sprint 61", [comment added automatically](https://github.com/open-component-model/ocm/issues/1131#issuecomment-4311985310)

### Verification

- [x] Tests pass locally (`node sprint-hygiene.test.js`)
- [x] Script tested against live project board with `--apply --limit 1`


.github/scripts/sprint-hygiene.js
```diff
@@ -0,0 +1,270 @@
...
+    return;
+  }
+
+  const needsRefinementOption = statusField.options.find((o) =>
+    o.name.includes("Needs Refinement"),
+  );
+  if (!needsRefinementOption) {
+    core.setFailed('Could not find a status option containing "Needs Refinement"');
+    return;
+  }
```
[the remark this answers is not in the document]
matthiasbruns: Addressed in 9f2be5aab.

.github/scripts/sprint-hygiene.js
```diff
@@ -0,0 +1,270 @@
...
+    core.setFailed("Could not determine current sprint");
+    return;
+  }
+  core.info(`Current sprint: ${currentSprint.title} (${currentSprint.id})`);
+
+  // 3. Fetch stale items using server-side filter
+  //    The `query` parameter on projectV2.items supports the same filter syntax
+  //    as the Projects UI, so we let GitHub do the heavy lifting.
+  const itemsFilter = `sprint:<@current status:"${needsRefinementOption.name}"`;
+  core.info(`Filter: ${itemsFilter}`);
```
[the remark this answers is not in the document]
matthiasbruns: Addressed in 9f2be5aab.

.github/scripts/sprint-hygiene.local.js
```diff
@@ -0,0 +1,64 @@
...
+
+const token = process.env.GH_TOKEN || process.env.GITHUB_TOKEN;
+if (!token) {
+  console.error("Error: GH_TOKEN or GITHUB_TOKEN must be set");
+  process.exit(1);
+}
+
+const org = option("org", "open-component-model");
+const projectNumber = Number(option("project", "10"));
+const limit = option("limit", undefined) ? Number(option("limit")) : undefined;
```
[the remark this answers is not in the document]
matthiasbruns: Addressed in 9f2be5aab.

.github/workflows/sprint-hygiene.yaml
```diff
@@ -0,0 +1,32 @@
...
+    - cron: '0 3 * * *' # Every night at 3 AM UTC
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  update-expired-sprints:
+    name: Move "Needs Refinement" issues with expired sprints to current sprint
+    runs-on: ubuntu-24.04-arm
```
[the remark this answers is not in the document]
matthiasbruns: Addressed in 9f2be5aab.

morri-son:
@coderabbitai review

.github/workflows/sprint-hygiene.yaml
```diff
@@ -0,0 +1,36 @@
...
+        uses: actions/create-github-app-token@f8d387b68d61c58ab83c6c016672934102569859 # v3.0.0
+        with:
+          app-id: ${{ secrets.OCMBOT_APP_ID }}
+          private-key: ${{ secrets.OCMBOT_PRIV_KEY }}
+
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          sparse-checkout: .github/scripts
+
+      - uses: actions/github-script@ed597411d8f924073f98dfc5c65a23a2325f34cd # v8
```
morri-son: @3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9 already available

matthiasbruns: done

.github/workflows/sprint-hygiene.yaml
```diff
@@ -0,0 +1,36 @@
+name: Sprint Hygiene
+
+on:
+  schedule:
+    - cron: '0 3 * * *' # Every night at 3 AM UTC
+  workflow_dispatch:
+
+permissions:
+  contents: read
```
fabianburth, line 9: don't we need more permissions?

morri-son: we need more permissions, but the workflow uses the token from the OCM Bot/App which is almighty. So, the permissions statement here is a bit misleading, as it covers the permissions the workflow assigns to the GITHUB_TOKEN. The actions done with that token indeed only require contents: read. Long story short: The workflow does not require more permissions set at that point.

fabianburth: Ahh, thanks for the explanation. Instead of using the almighty ocm bot token, wouldn't it be more idiomatic to give the workflow the correct permissions?

matthiasbruns: we cannot - the GH_TOKEN cannot access org resources - and our project is on org level

morri-son (dismissed):

.github/workflows/sprint-hygiene.yaml
```diff
@@ -0,0 +1,36 @@
...
+
+concurrency:
+  group: ${{ github.workflow }}
+  cancel-in-progress: true
```
Skarlso, line 14: I don't think this is needed, since it's running at night, right? It would actually leave an upcompleted flow where some itesm are updated, some aren't.

.github/scripts/sprint-hygiene.js
```diff
@@ -0,0 +1,316 @@
...
+    o.name.includes("Needs Refinement"),
+  );
+
+  if (matches.length === 0) {
+    throw new Error('Could not find a status option containing "Needs Refinement"');
+  }
+
+  const statusName = matches[0].name;
+
+  if (statusName.includes('"')) {
```
Skarlso: Might as well add something like `/` or `\`.

.github/scripts/sprint-hygiene.js
```diff
@@ -0,0 +1,316 @@
...
+  const staleItems = await fetchStaleItems(github, core, { projectId, statusName });
+
+  core.info(`Found ${staleItems.length} issue(s) in "Needs Refinement" with expired sprints`);
+
+  if (staleItems.length === 0) {
+    core.info("Nothing to update.");
+    return;
+  }
+
+  const itemsToProcess = limit !== undefined ? staleItems.slice(0, limit) : staleItems;
```
Skarlso: Err, why not just push the limit into the pagination of fetchStaleItems? 🤔

Skarlso (commented):
Just a few notes, otherwise, looks okay.

.github/workflows/sprint-hygiene.yaml
matthiasbruns: it can be run manually as well so I'd keep this

.github/scripts/sprint-hygiene.js
matthiasbruns: welp - yeah ...

matthiasbruns: done

.github/scripts/sprint-hygiene.js
```diff
@@ -0,0 +1,319 @@
+// @ts-check
+
+/**
+ * Automated sprint hygiene for GitHub Projects (v2).
+ *
+ * Finds issues in "Needs Refinement" status whose sprint has expired
```
frewilhelm, line 6: Why do we do this only for issues with "Needs Refinement" status?

frewilhelm: Probably because the issue states only the "Needs refinement" status. But usually all issues from different columns (apart from `Review` and `Done`) are moved to the next sprint manually

.github/workflows/sprint-hygiene.yaml
fabianburth: ahh, got it - thanks! 👌

.github/scripts/sprint-hygiene.js
matthiasbruns: cause its stated in the issue :=

matthiasbruns: so lets go through the options

- things we never looked at are in backlog anyway
- things thats are in next up etc should be moved automatically after each sprint
- refinement just might be the ones we might lose after sprints

frewilhelm: > things thats are in next up etc should be moved automatically after each sprint

how?

I just had another look and we lose items from
- Next up (items that were not picked up in the current sprint)
- ToDo
- Waiting for QA/External
- Review (Spillover)
- In Progress (Spillover)

matthiasbruns: there were some items that are set to "needs refinement" but were never assigned to a sprint

matthiasbruns: so basically, all items that are not in a final state with a sprint assigned that is in the past should be updated? @Skarlso  wdyt?

Skarlso: So are we doing anything about this right now? Or is this okay like it is?

matthiasbruns: We can iterate and see if we need to mutate this - I wouldn't do this now until it becomes a problem

matthiasbruns: talked about this in the daily, we will focus on "needs refinement" for now

Skarlso (approved):

fabianburth (approved):

PR #2399, opened 2026-04-27
feat(runtime): add Scheme.ResolveCanonicalType for alias resolution

### LINKED ISSUE #704: Make Credential Graph returned typed objects instead of plain maps

matthiasbruns:

Adds `Scheme.ResolveCanonicalType(typ Type) Type` to `runtime.Scheme`. This method resolves alias types to their canonical (default) type, or returns the type unchanged if it is already a default or not registered.

This is needed by the typed credentials work (#2343) to validate identity → credential type compatibility through alias resolution. Previously, alias comparison required instantiating objects via `NewObject` and comparing `reflect.TypeOf` — this is simpler and avoids reflection.

**Example:** If `HelmHTTPCredentials/v1` is registered as the default with `HelmHTTPCredentials` as an unversioned alias, then:
- `ResolveCanonicalType("HelmHTTPCredentials")` → `HelmHTTPCredentials/v1`
- `ResolveCanonicalType("HelmHTTPCredentials/v1")` → `HelmHTTPCredentials/v1` (already canonical)
- `ResolveCanonicalType("Unknown/v1")` → `Unknown/v1` (not registered, returned as-is)

Related to https://github.com/open-component-model/ocm-project/issues/704
Part of #2343

##### How to test the changes

Run the runtime tests:
```bash
cd bindings/go/runtime && go test ./... -run TestScheme_ResolveCanonicalType -v
```

##### Verification


bindings/go/runtime/registry.go
```diff
@@ -274,6 +274,23 @@ func (r *Scheme) IsRegistered(typ Type) bool {
...
+	defer r.mu.RUnlock()
+
+	if _, exists := r.defaults.GetLeft(typ); exists {
+		return typ
+	}
+	if def, ok := r.aliases[typ]; ok {
+		return def
+	}
+	return typ
+}
```
fabianburth: How would I distinct between `default` and `unregistered`?

matthiasbruns: maybe instead of this we return an additional ok?

morri-son (dismissed):

bindings/go/runtime/registry.go
fabianburth: i'd rather return an error, it is more explicit

matthiasbruns: I don't know about this. Why do we always have to return errors? In this case, I do not see it tbh, shouldn't ok be enough? Otherwise, the caller must know if we have a typed error or something to handle such cases.

piotrjanik (dismissed):
LGTM

bindings/go/runtime/registry_test.go
```diff
@@ -317,6 +317,89 @@ func TestRegistry_NewObject_Based_On_Alias(t *testing.T) {
...
+			input:      ociAlias,
+			expected:   ociDefault,
+			expectedOk: true,
+		},
+		{
+			name:       "alias of one type does not resolve to another type",
+			input:      helmAlias,
+			expected:   helmDefault,
+			expectedOk: true,
+		},
```
frewilhelm: This is a duplicate of `unversioned alias resolves to default`

matthiasbruns: removed

morri-son (approved):

bindings/go/runtime/registry.go
frewilhelm: I believe errors make sense if we can provide some more context about the "failure", so the user can try to resolve it. The only failure-context I can think of in this case, would be "cannot resolve canonical type because ..."
- type not registered, but XYZ is registered
- type not registered, but no type is registered.

In both cases something went wrong because the passed type is not registered in the scheme and the user needs to fix that. Hence, I believe returning `ok` should be enough.

frewilhelm (approved):

fabianburth (approved):

PR #2411, opened 2026-04-29
chore: add filepath patterns to gitguardian config

morri-son:
add filepath patterns to gitguardian config

### DIFF
--- a/.gitguardian.yaml
+++ b/.gitguardian.yaml
@@ -1,11 +1,31 @@
+version: 2
+
 secret:
   ignored_matches:
-  - match: 4132766f296d24c64452646571d72a1f88e66b72f27838c6749b071b0468939d
-    name: kubernetes/controller/examples public keys
-  - match: c41c71837a9ffbb3eb165eb1fb62ebed35c129c9597ee17460a831a66a372422
-    name: cli/cmd/cmd_test.go
-  - match: 57d4c1be78223b477215a9133e64eed43804b319138ad69e0d4f4d5b0ada836b
-    name: kubernetes/controller/test/e2e/testdata/docker-config-json/.docker-config-json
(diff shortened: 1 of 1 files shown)


.gitguardian.yaml
```diff
@@ -1,11 +1,31 @@
...
+    # --- Test and mock data ---
+    - "**/test/**"
+    - "**/tests/**"
+    - "**/*test*"
+    - "**/mock/**"
+    - "**/*.test.*"
+    
+    # --- Package manager lock files ---
+    - "**/package-lock.json"
+    - "**/go.sum"
```
frewilhelm, line 31: Does this work in a `.gitguardian.yaml` file?

morri-son: in ggshield, the CLI tool, this works. Since we already have a `.gitguardian.yaml` and ignored pattern, the file paths should also work. I don't see gitguardian as workflow, so I expect it to be installed as GH app. And it seems that it checks out the repo, as it uses the existing `.gitguardian.yaml`.

frewilhelm: can you try this out by adding a secret to the PR and check if ggshield is flagging it? Maybe use the file that ggshield flagged in your sigstore PR

frewilhelm:
@coderabbitai review

.gitguardian.yaml
frewilhelm: In general I am unsure if those path-based ignores are really helpful and support the purpose why we use ggshield in the beginning: we do not want to leak secrets. 

path-based ignores will create blind-spots. Imagine while developing, you use your own GH token in a test and then forget about it (this will happen 100%. Not necessarily today or tomorrow, not necessarily to you - but at one point, someone will leak something by accident. This already happened (to the best of us)).

The match-based ignores will make sure that this does not happen because you have to ignore every secret manually. This gives us more safety even if its cumbersome.

PR #2427, opened 2026-05-04
fix(ci): resolve all zizmor security findings in GitHub Actions workflows

jakobmoellerdev:

- Add persist-credentials: false to all actions/checkout steps (artipacked)
- Add cache: false to actions/setup-go in PR-triggered publish workflows (cache-poisoning)
- Move ${{ expression }} interpolations from run: blocks to env: vars (template-injection)
- Replace softprops/action-gh-release with gh release create (superfluous-actions)
- Replace secrets: inherit with explicit secret mappings where possible (secrets-inherit)
- Suppress pull_request_target findings with inline zizmor: ignore comments (dangerous-triggers) where the trigger is intentional and no PR code is executed
- Add secrets declarations to workflow_call blocks in conformance.yml and publish-helminput-plugin-component.yaml

addresses some findings I found when scanning through the repo with [zizmor](https://docs.zizmor.sh/). Based on the review / feedback here I will open a PR that will add Zizmor as a Job

#### Testing

Im testing this all on this branch here.

##### Verification


.github/workflows/conformance.yml
```diff
@@ -38,6 +38,10 @@ on:
         type: string
         required: false
         default: ""
+    secrets:
+      GITHUB_TOKEN:
+        description: "GitHub token for setup-task repo-token"
+        required: false
```
frewilhelm, line 44: Why is this added?

.github/workflows/publish-helminput-plugin-component.yaml
```diff
@@ -21,6 +21,10 @@ on:
         description: "The ref to build on (branch or tag). Defaults to the current ref."
         type: string
         required: false
+    secrets:
+      GITHUB_TOKEN:
+        description: "GitHub token"
+        required: false
```
frewilhelm, line 27: So, instead of inheriting all secrets, we now pass a `GITHUB_TOKEN` with the respective permissions?

.github/workflows/release-go-submodule.yaml
```diff
@@ -37,21 +37,24 @@ jobs:
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
         with:
           fetch-depth: 0
-
-      - name: Fetch all tags
+          persist-credentials: false
```
frewilhelm, line 41: This removes the step `Fetch all tags`. How does the `run:` works afterwards?

frewilhelm (commented):
> Based on the review / feedback here I will open a PR that will add Zizmor as a Job

i think this would make sense

Skarlso (commented):
Looks okay aside from what Frederic already said.

.github/workflows/release-go-submodule.yaml
jakobmoellerdev: The separate Fetch all tags step was removed because actions/checkout with fetch-depth: 0 already should fetch the full history including tags, no?

.github/workflows/conformance.yml
jakobmoellerdev: zizmor flags `secrets: inherit` as a security finding (["secrets-inherit"](https://docs.zizmor.sh/audits/#secrets-inherit)) because it passes **all** repository secrets to the called workflow, even ones it does not need. Adding an explicit `secrets` declaration to the `workflow_call` block lets callers pass only `GITHUB_TOKEN` explicitly instead of `secrets: inherit`. This follows the principle of least privilege — the called workflow declares what it needs, callers pass only that.

.github/workflows/publish-helminput-plugin-component.yaml
jakobmoellerdev: Exactly. Instead of `secrets: inherit` (which passes every secret the caller has access to), the caller now passes only `GITHUB_TOKEN` explicitly. The called workflow declares what it needs via this `secrets:` block. If the workflow is ever compromised, it only has access to the single token that was explicitly granted, not the full secret store.

.github/workflows/release-go-submodule.yaml
jakobmoellerdev: Good catch — this was actually a bug in the original. The `run: git fetch --tags` was a sibling key to `uses:` on the same step, which is invalid YAML for GitHub Actions (a step is either `uses:` OR `run:`, not both). GitHub silently ignored the `run:` field, so tags were never being fetched by that step.

Tags were still available because `fetch-depth: 0` fetches the full history including tags by default. I have now also added `fetch-tags: true` explicitly to make this intent clear and future-proof.

Additionally, I fixed the `persist-credentials: false` on the release job checkout — since `git push` needs auth, the push step now sets the remote URL with the token before pushing (8d0527e0).

.github/workflows/ci.yml
```diff
@@ -307,7 +322,8 @@ jobs:
         env:
           GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
           GITHUB_USER: ${{ github.actor }}
-        run: task ${{ matrix.module }}:test/integration
+          MODULE: ${{ matrix.module }}
+        run: task "${MODULE}:test/integration"
```
matthiasbruns, line 326: why do we need `GH_TOKEN` in integration tests in the first place? I have checked the codebase for `GH_TOKEN` but couldn't find any use.

jakobmoellerdev: GH_TOKEN is used by gh cli and we use it in integration tests for interacting with ghcr.io from what I know.

.github/workflows/conformance.yml
matthiasbruns: In my understanding, `GITHUB_TOKEN ` doesn't need to be explicitly set as part of the secrets - but maybe I am wrong here.

So if the caller does not pass any secrets, `GITHUB_TOKEN` should be available anyway since the workflow runner provides the token based on permission settings. PATs are different though, but thats another story

.github/workflows/markdown.yml
```diff
@@ -18,6 +18,8 @@ jobs:
     runs-on: ubuntu-24.04-arm
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          persist-credentials: false
```
matthiasbruns, line 22: hoping for this to land
https://github.com/actions/checkout/pull/1687

this is madness :D

matthiasbruns (commented):
Overall very good cleanup and security fixes. I am just uncertain of this passing of the `GITHUB_TOKEN`

.github/workflows/ci.yml
matthiasbruns: makes sense

.github/workflows/conformance.yml
jakobmoellerdev: @matthiasbruns You are right that `GITHUB_TOKEN` is automatically available in every workflow run via `github.token`. However, in this case the callers (e.g. `kubernetes-controller.yml:328`) explicitly pass it as:

```yaml
secrets:
  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```

For that explicit pass-through to work, the called workflow must declare the secret in its `workflow_call` block — otherwise actionlint (and GitHub itself) will reject it:

> secret "GITHUB_TOKEN" is not defined in reusable workflow. no secret is defined

The alternative would be to remove the explicit pass from the callers and use `github.token` directly in the called workflow. But since this PR is moving away from `secrets: inherit` toward explicit secret mapping, the declaration is needed to match.

.github/workflows/release-go-submodule.yaml
frewilhelm: what :D I think your current approach works now, but it looked broken before and the AI confused itself a bit :D

.github/workflows/conformance.yml
matthiasbruns: Thanks for the clarification :)

matthiasbruns (dismissed):

frewilhelm (dismissed):
do you plan to introduce zizmor action as a follow up then?

jakobmoellerdev:
yeah exactly ill do a follow up PR.

jakobmoellerdev:
you know what, can also add it now...

.github/workflows/cli-release.yml
```diff
@@ -145,18 +148,17 @@ jobs:
...
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          RC_TAG: ${{ needs.prepare.outputs.new_tag }}
+          RC_VERSION: ${{ needs.prepare.outputs.new_version }}
+        run: |
+          gh release create "${RC_TAG}" \
+            --title "CLI ${RC_VERSION}" \
+            --notes-file "${{ runner.temp }}/CHANGELOG.md" \
+            --prerelease \
+            "${{ runner.temp }}/rc-build-assets/bin/ocm-"* \
+            "${{ runner.temp }}/rc-build-assets/oci/"*
```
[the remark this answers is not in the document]
jakobmoellerdev: I really think this is not required since it automatically points to the repo

.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,26 @@
...
+    branches: ["main"]
+  pull_request:
+    branches: ["**"]
+
+permissions: {}
+
+jobs:
+  zizmor:
+    name: Run zizmor
+    runs-on: ubuntu-latest
```
frewilhelm: I believe zizmor should also run on `arm` (https://github.com/zizmorcore/zizmor/pkgs/container/zizmor)

jakobmoellerdev: good catch!

frewilhelm (dismissed):

.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,26 @@
+name: GitHub Actions Security Analysis with zizmor
+
+on:
+  push:
+    branches: ["main"]
+  pull_request:
+    branches: ["**"]
```
matthiasbruns, line 9: do we maybe want a filter here to only run if something in `.github` changes?

Skarlso (dismissed):

.github/workflows/zizmor.yml
jakobmoellerdev: agreed, good addition!

jakobmoellerdev: should be changed now!

frewilhelm (approved):

matthiasbruns (approved):
lgtm

PR #2446, opened 2026-05-06
chore: clarify credential resolution

matthiasbruns:
Clarification of questions that came up according to how credentials (direct vs indirect) are being resolved.

Contributes: https://github.com/open-component-model/ocm-project/issues/702
Contributes: https://github.com/open-component-model/ocm-project/issues/980


docs/adr/0018_typed_credentials.md
```diff
@@ -126,6 +126,36 @@ sequenceDiagram
...
+
+The credential graph resolves credentials through two paths, determined at ingestion time.
+
+**Direct credentials** apply when a consumer's credential entry is a `Credentials/v1` (plain `properties` map) or a typed
+credential registered in the `CredentialTypeScheme` (e.g., `HelmHTTPCredentials/v1`). The graph stores these as
+`runtime.Typed` directly on the identity node at ingestion time; `ResolveTyped` returns them immediately as a leaf
+lookup with no plugin call.
+
+**Indirect credentials** apply when the credential type requires a plugin (e.g., `AWSSecretsManager`,
+`HashiCorpVault`). The plugin is looked up via the `CredentialPluginProvider` — the plugin registry for credential
```
matthiasbruns: seems to confuse while reading - update

fabianburth:
Why wasn't the maintainer team requested as reviewers here?

matthiasbruns:
> Why wasn't the maintainer team requested as reviewers here?

I have no idea - now I did a request - doesn't that usually happen by repo rules?

docs/adr/0021_typed_credentials.md
```diff
@@ -126,6 +126,36 @@ sequenceDiagram
     C->>C: type-assert fields<br/>(CertFile, KeyFile, Keyring)
 ```
 
+### Credential Resolution: Direct vs. Indirect
+
+The credential graph resolves credentials through two paths, determined at ingestion time.
+
+**Direct credentials** apply when a consumer's credential entry is a `Credentials/v1` (plain `properties` map) or a typed
+credential registered in the `CredentialTypeScheme` (e.g., `HelmHTTPCredentials/v1`). The graph stores these as
```
fabianburth, line 114: Since `Credentials/v1` also have a `type` field, couldn't we implement them as type in `CredentialTypeScheme` as well?

docs/adr/0018_typed_credentials.md
```diff
@@ -126,6 +126,36 @@ sequenceDiagram
     C->>C: type-assert fields<br/>(CertFile, KeyFile, Keyring)
 ```
 
+### Credential Resolution: Direct vs. Indirect
+
+The credential graph resolves credentials through two paths, determined at ingestion time.
```
fabianburth: This is not generally correct. There are also credential repositories (although they are not that relevant here).

docs/adr/0018_typed_credentials.md
```diff
@@ -126,6 +126,36 @@ sequenceDiagram
...
+from its child before it can be called:
+
+```text
+HashiCorpVault:myvault.example.com
+  └─ needs: AWSSecretsManager (secretId: vault-access-creds)
+       └─ has: DirectCredentials {roleid: "my-role-id"}   ← leaf, resolved directly at ingestion
+```
+
+Resolution is bottom-up: the AWS plugin is called first with `{roleid}` and returns a Vault token;
+the HashiCorpVault plugin is then called with that token to return the final credentials.
```
fabianburth: I feel like this graph is rather confusing. Couldnt we reference a section from the original ADR here?

jakobmoellerdev:
I think github is playing tricks on us again, we didnt change any rules.

docs/adr/0021_typed_credentials.md
matthiasbruns: CredentialTypeScheme is for every "external" credential type, that is not in the credential package - I can check if it makes sense to put them also into the CredentialTypeScheme, since it's a well-known type within the credential package.

docs/adr/0018_typed_credentials.md
matthiasbruns: I will add the repo lookup as well

jakobmoellerdev (dismissed):

jakobmoellerdev (dismissed):

morri-son (dismissed):

jakobmoellerdev:
@matthiasbruns needs a rebase, but still good

jakobmoellerdev (dismissed):

docs/adr/0021_typed_credentials.md
```diff
@@ -104,6 +104,50 @@ sequenceDiagram
...
+```go
+// Consumer usage — pass typed identity directly
+identity := &HelmChartRepositoryIdentity{Hostname: "charts.example.com"}
+typed, err := resolver.ResolveTyped(ctx, identity)
+```
+
+The graph internally works with `runtime.Typed` for identities throughout. The only place where `runtime.Identity` is
+still used is at the boundary to `CredentialPlugin` and `RepositoryPlugin` interfaces, which still accept
+`runtime.Identity` in their signatures. This conversion is migration scaffolding that Phase 3 removes when those
+plugin interfaces migrate to `runtime.Typed`.
```
Skarlso: This is no longer true, right? We aren't planning any migration.

docs/adr/0021_typed_credentials.md
```diff
@@ -104,6 +104,50 @@ sequenceDiagram
...
+
+```go
+// Consumer usage — pass typed identity directly
+identity := &HelmChartRepositoryIdentity{Hostname: "charts.example.com"}
+typed, err := resolver.ResolveTyped(ctx, identity)
+```
+
+The graph internally works with `runtime.Typed` for identities throughout. The only place where `runtime.Identity` is
+still used is at the boundary to `CredentialPlugin` and `RepositoryPlugin` interfaces, which still accept
+`runtime.Identity` in their signatures. This conversion is migration scaffolding that Phase 3 removes when those
```
Skarlso: `runtime.Identity` -> this is not true, I believe? If you mean this interface:

```go
// CredentialPlugin defines the interface for plugins that handle custom credential
// resolution logic. These plugins are responsible for:
// - Mapping credentials to consumer identities
// - Resolving credentials for specific identities
type CredentialPlugin interface {
	// GetConsumerIdentity maps a credential to a consumer identity.
	// This identity is used to look up credentials in the credential graph.
	GetConsumerIdentity(ctx context.Context, credential runtime.Typed) (runtime.Identity, error)

	// Resolve attempts to resolve credentials for a given consumer identity.
	// The provided credentials map may contain pre-resolved credentials from
	// the credential graph.
	// TODO(matthiasbruns): Migrate to runtime.Typed https://github.com/open-component-model/ocm-project/issues/980
	Resolve(ctx context.Context, identity runtime.Identity, credentials map[string]string) (map[string]string, error)
}
```

that's `map[string]string`. It is the same as `runtime.Identity` but not the same _type_.

jakobmoellerdev: I think we should unify on runtime.Identity and not a plain map

Skarlso: That sounds good. 👍

docs/adr/0021_typed_credentials.md
matthiasbruns: somewhat reverted - bindings can still provide typed identities as done in https://github.com/open-component-model/open-component-model/pull/2490/changes#diff-a969f48a50029e73abefebfd07d9686e72003323bd6e4bf5864b18b70be5b202 but the API will only rely un runtime.Identity as well as the graph

docs/adr/0021_typed_credentials.md
matthiasbruns: So I will continue with `ResolveTyped` `runtime.Identity` instead of `maps` as in https://github.com/open-component-model/open-component-model/pull/2452/changes#diff-d6026f20583f6d7e0aa73ba162ab73943efc873a1e52f0304a2e8d0c8cb1ef74R31

jakobmoellerdev (approved):

Skarlso (approved):

PR #2457, opened 2026-05-07
chore(docs): add support for .dockerconfigjson format in docstring

reshnm:

Apparently the ocm controllers not only support pull credentials in ocm config format but also in .dockerconfigjson format.
Update the docstring in the API to reflect this.

##### Verification

### DIFF
--- a/kubernetes/controller/api/v1alpha1/common_types.go
+++ b/kubernetes/controller/api/v1alpha1/common_types.go
@@ -41,7 +41,7 @@ const (
 // +kubebuilder:validation:XValidation:rule="((!has(self.apiVersion) || self.apiVersion == \"\" || self.apiVersion == \"v1\") && (self.kind == \"Secret\" || self.kind == \"ConfigMap\")) || (self.apiVersion == \"delivery.ocm.software/v1alpha1\" && (self.kind == \"Repository\" || self.kind == \"Component\" || self.kind == \"Resource\" || self.kind == \"Replication\"))",message="apiVersion must be one of \"v1\" with kind \"Secret\" or \"ConfigMap\" or \"delivery.ocm.software/v1alpha1\" with the kind of an OCM kubernetes object"
 type OCMConfiguration struct {
 	// Ref reference config maps or secrets containing arbitrary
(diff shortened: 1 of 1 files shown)


reshnm:
I tried to update the manifests to reflect the changes. But nothing I do seems to update them. Can someone help me?

```
task generate
task manifests
```

jakobmoellerdev:
@Skarlso I think there might be something off with the generators due to the recent changes on the chart. could you please take a look?

Skarlso:
Will take a look.

Skarlso:
This is an internal struct in common types. It's not a forward facing component in our API. It was never part of our API docs.

Skarlso:
Meaning, you're good here if all you want is to comment on the config struct. If not, you'll have to update the place where it's included at. 

For example:

```go
// DeployerSpec defines the desired state of Deployer.
type DeployerSpec struct {
	// ResourceRef is the k8s resource name of an OCM resource containing the ResourceGroupDefinition.
	// +required
	ResourceRef ObjectKey `json:"resourceRef"`

	// OCMConfig defines references to secrets, config maps or ocm api
	// objects providing configuration data including credentials.
	// +optional
	OCMConfig []OCMConfiguration `json:"ocmConfig,omitempty"`
```

reshnm:
@Skarlso Makes sense. I didn't see that it was inlined.
For me it is fine as is. Since it is included in all other types, I would not update the types where it is include but rather move it to the documentation of the `OCMConfiguration` itself. This would avoid to copy the same text to multiple locations.

Skarlso (approved):

jakobmoellerdev (approved):

frewilhelm:
does it make sense to put an ocm config into a k8s secret of "type" `dockerconfigjson`?

reshnm:
> does it make sense to put an ocm config into a k8s secret of "type" `dockerconfigjson`?

No, it does not. But the ocm controllers support a .dockerconfig secret type as an alternative to an ocm config. I just learned this today. https://ocm.software/docs/how-to/configure-credentials-for-ocm-controllers/#create-a-kubernetes-secret-with-credentials

PR #2468, opened 2026-05-08
feat: unified release flow for a single github release version

### LINKED ISSUE #1058: Fix release process
**Description**
The release process failed b/c the conformance tests used the cli build from the main branch. The cli built a component constructor with
```
input:
  type: file
```

This is translated to
```
access:
  type: LocalBlob/v1 # newly introduced capital case
```
The capital case is only introduced in main and was thus not supported by the controllers. 

**Decision:**
- we need to build and test the controllers together in a workflow, so that they can be tested in the same conformance test run
- we need to build on commits to the release branches, so the cli workflow can use the controllers already built for the release branch (and vice versa) 

**Done Criteria**

Skarlso:

_Update_: We still do create the `cli/v*` tag as well, because things depend on it. 🤷 

Okaaay. This is a larger one. I'll try to sensibly explain what my thinking was here and how I ended up with `pipelines.yaml` and `release.yaml`.

So this pr replaces our per-component release with a single unified release. We drop the `cli/v*` tags, because frankly, they aren't needed anymore, no-one is importing the CLI. We keep the kubernetes/controller/v* tags because those tags are used by people who import all things Kubernetes. If we ever want to create the cli tag again, that's super easy to add. ( we need to add the tag check to the javascript regex, release have to be made aware to push it, and create-tag.js updated ).

Now, if we don't tag, how does this all work? That's the neat part and that's why this unification is so nice! It's built from the same ref and get tested as a pair always!

What changed?

These are mostly structural changes. I tried to limit the number of things I moved but I deleted a bunch of action files, and copied most of their contents verbatim or with only very slight modifications because of the tags. The new things are mostly added for workflow coordination! The other thing we talked about is the conformance testing trigger. I moved that into the pipeline section which has arguably a lot of triggers, but this is MAIN FLOW for all builds, pushes and runs. And I would argue that everything is now IN ONE PLACE instead of several other places and files. And surprisingly, it's not significantly longer.

The tradeoff. CLI changes WILL run conformance and e2e. This is by design to make sure that the new CLI does not break the conformance tests. We can improve here in a followup ( because I wanted to keep this as minimal as possible ) to pass in a flag from CLI saying something like `skipE2E` and then at least that will not run. The conformance does NEED to run though and for that we need the build and chart because those are artifacts used by the conformance test.

The Javascript changes I did by hand first to my limited knowledge, then ran it through an llm a couple of times to see if it could improve on my logic. It did in a few places and added tests as well, which is nice.

Here is a test of this entire flow: https://github.com/Skarlso/open-component-model/actions/runs/25546869852

My test included an ocm-bot ( I created a github app, added GPG signing keys and everything ) so the flow is the same without any stupid modifications locally that would make it work. The bot had registry read/write and content read/write access.

Here is the release and pre release it created: https://github.com/Skarlso/open-component-model/releases/tag/v0.7.0
And: https://github.com/Skarlso/open-component-model/releases/tag/v0.7.0-rc.1
Ignore the changelog it's HUGE. I didn't have anything so it generated the entire commit history. :D

I also forgot to set up reviewers and the gate so it immediately ran through without waiting. :) That shouldn't happen here.

What else... I tried running some failure scenarios as well, to make sure I could restart the entire flow. The entire thing is pretty solid. You can literally re-run the whole thing over and over and it should either fail early with things like, already exists, yada yada, or just run through normally if nothing changed. The only real issue if the tag / release would be somehow pushed incorrectly. Then we'd have to revoke it, but that's an existing problem.

I did not update RELEASE_PROCESS.md yet. I'll do it after we decided that this is good. 😆 

The pipeline:
Screenshot 2026-05-08 at 13 57 49

Fixes https://github.com/open-component-model/ocm-project/issues/1058.

##### Verification


cliff.toml
```diff
@@ -1,6 +1,9 @@
 # git-cliff ~ configuration file
 # https://git-cliff.org/docs/configuration
-
+#
+# A unofied cliff.toml. Replaces the per-component configs at cli/cliff.toml
+# and kubernetes/controller/cliff.toml. I copied over whatever was revelvant
+# since both are now included in the release.
```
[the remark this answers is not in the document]
Skarlso: >Also consider rewording the personal-voice 

I will never consider rewording my personal voice.

.github/workflows/pipeline.yml
```diff
@@ -0,0 +1,348 @@
...
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+env:
+  REGISTRY: ghcr.io
+
+jobs:
+  # === This is new: workflow_call wiring ===
```
Skarlso, line 82: I added these things so you, the reviewer, can easily track where they came from, how, and what changed.

Skarlso:
Okay, it seems dropping cli/v* tag wasn't a good idea apparently. I'll add that back and then revert the fallback behavior of the script.

.github/workflows/release.yml
```diff
@@ -0,0 +1,613 @@
...
+
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          ref: ${{ github.ref_name }}
+          token: ${{ steps.get_token.outputs.token }}
+          fetch-tags: true
+          # persist-credentials kept (intentionally) — create-tag.js does
+          # `git push origin refs/tags/...` which needs the app token persisted
+          # in .git/config to authenticate. Zizmor flags this; accepted.
```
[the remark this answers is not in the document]
Skarlso: Without this, the release flow fails because the scripts need the token to create the tag.

.github/workflows/release.yml
```diff
@@ -0,0 +1,613 @@
...
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          fetch-depth: 0
+          ref: ${{ github.ref_name }}
+          token: ${{ steps.get_token.outputs.token }}
+          fetch-tags: true
+          # persist-credentials kept (intentionally) — create-tag.js does
+          # `git push origin refs/tags/...` which needs the app token persisted
+          # in .git/config to authenticate. Zizmor flags this; accepted.
```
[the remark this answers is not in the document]
Skarlso: Same here.

Skarlso:
Running a full release test here: https://github.com/Skarlso/open-component-model/actions/runs/25560299137

For the creation of the cli tag that I just added back in and the creds persistence change.

Skarlso:
Okay, here is a short description of how the release looks like right now:

Daily CI and releases both run through pipeline.yml, which builds the CLI and controller in parallel from the same ref, runs conformance against the resulting pair of them and only publishes if conformance passes. Because both components come from the same SHA _every time_, conformance always tests a pair that belongs together, including on release branches and backports.

Releases are dispatched via release.yml. This calls pipeline.yml for the build, cuts an RC tag, waits at an environment gate, and on approval promotes the RC to a final v0.X.Y. This did largely not change from the original.

We create three tags on each release. One canonical final vX.Y.X. One, cli/* and one for kubernetes/controller/* like I wrote above. Attestations, chart verification, changelog generation are the same as before. The refactor just unified triggers and ordering and did _not_ change what gets shipped eventually.

.github/workflows/release.yml
```diff
@@ -0,0 +1,613 @@
...
+          set -euo pipefail
+          echo "digest=$(oras resolve $CONTROLLER_CHART:${VERSION})" >> "$GITHUB_OUTPUT"
+
+      - name: Attest final chart
+        uses: actions/attest-build-provenance@a2bbfa25375fe432b6a289bc6b6cd05ecd0c4c32 # v4
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          subject-digest: ${{ steps.chart-digest.outputs.digest }}
+          subject-name: ${{ env.CONTROLLER_CHART }}
+          push-to-registry: true
```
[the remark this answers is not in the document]
Skarlso: Are you sure about this?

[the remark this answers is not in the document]
Skarlso: This is wrong. The chart has `chart` as a name in Chart.yaml. And helm push appends that to the URL.

If I add `chart` it would end up as `chart/chart`.  Please verify.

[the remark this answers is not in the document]
Skarlso: goddamnright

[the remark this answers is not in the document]
matthiasbruns: @coderabbitai you will have to call @Skarlso from now on Heisenberg in this pr

.github/scripts/create-tag.js
```diff
@@ -66,66 +66,118 @@ function defaultExecGit(args) {
...
- * Idempotent: skips if the tag already exists.
- * Sets output `pushed=true` on success or idempotent skip.
- *
- * Expects env vars: TAG
- *
- * @param {object} args
- * @param {object} args.core - GitHub Actions core module.
- * @param {function} [args.execGit] - Git executor (for testing).
+ * Create one annotated tag pointing at HEAD. Idempotent.
+ * Returns true if created or already at HEAD; false if a hard mismatch was reported.
```
matthiasbruns, line 83: https://jsdoc.app/tags-returns

.github/scripts/create-tag.js
```diff
@@ -66,66 +66,118 @@ function defaultExecGit(args) {
...
+    core.info(`Tag ${tag} already exists at HEAD (idempotent)`);
+    return true;
+  }
+  createAndPushTag({ tag, commit: "HEAD", message, execGit });
+  core.info(`Created tag ${tag}`);
+  return true;
+}
+
+/**
+ * Create one annotated tag pointing at a specific commit. Idempotent.
```
matthiasbruns, line 102: jsdocs are missing

.github/scripts/create-tag.js
```diff
@@ -66,66 +66,118 @@ function defaultExecGit(args) {
...
+// --------------------------
+
+/**
+ * Create RC tags for the unified release: the canonical v0.X.Y-rc.N tag and
+ * any number of side tags (cli/v0.X.Y-rc.N, kubernetes/controller/v0.X.Y-rc.N,
+ * etc.) all pointing at HEAD.
+ *
+ * Expects env vars:
+ *   CANONICAL_TAG  Required. The user-facing release tag (e.g. "v0.7.0-rc.1").
+ *   MODULE_TAGS    Optional. Comma-separated list of side tags to emit on the
```
matthiasbruns: This could be confused with the bindings - I first thought you implement binding release tags here, maybe just call it `ADDITIONAL_TAGS`?

.github/scripts/create-tag.js
```diff
@@ -66,66 +66,118 @@ function defaultExecGit(args) {
...
+  const { CANONICAL_TAG: canonicalTag, MODULE_TAGS: moduleTagsRaw } = process.env;
+
+  if (!canonicalTag) {
+    core.setFailed("Missing CANONICAL_TAG environment variable");
     return;
   }
 
-  const message = `Release candidate ${tag}`;
-  createAndPushTag({ tag, commit: "HEAD", message, execGit });
+  const moduleTags = parseModuleTagList(moduleTagsRaw);
```
matthiasbruns: `parseModuleTagList` is a simple string to list util, maybe we should name it that or inline the functionality, since it's basic js

.github/scripts/publish-final-release.js
```diff
@@ -2,6 +2,17 @@
 import fs from "fs";
 import path from "path";
 
+// What is this and why is it here? GitHub has a hard limit on the length of the changelog.
```
matthiasbruns: 👉 maybe rewrite that part

.github/workflows/conformance.yml
```diff
@@ -1,20 +1,25 @@
...
     inputs:
       cli_image:
-        description: "CLI Docker image reference"
+        description: "CLI Docker image reference (artifact mode is preferred)"
         type: string
         required: false
         default: ""
       toolkit_image:
-        description: "OCI reference to the controller Helm chart (with optional version tag)"
+        description: "OCI reference to the controller Helm chart (artifact mode is preferred)"
```
matthiasbruns: what is 

> artifact mode

.github/workflows/kubernetes-controller.yml
```diff
@@ -104,8 +87,12 @@ jobs:
       artifact_name: ${{ env.ARTIFACT_NAME }}
       artifact_id: ${{ steps.upload-artifacts.outputs.artifact-id }}
       version: ${{ steps.version.outputs.version }}
-      should_publish: ${{ steps.branch-check.outputs.should_publish }}
     steps:
+      - name: Compute image refs (lowercase owner)
```
matthiasbruns:
```suggestion
      - name: Compute image refs
```

.github/workflows/conformance.yml
```diff
@@ -73,6 +68,16 @@ jobs:
       matrix:
         arch: [arm64]
     steps:
+      - name: Compute image refs (lowercase owner)
```
matthiasbruns:
```suggestion
      - name: Compute image refs
```

.github/workflows/kubernetes-controller.yml
```diff
@@ -223,6 +191,11 @@ jobs:
       contents: read
       actions: read
     steps:
+      - name: Compute image refs (lowercase owner)
```
matthiasbruns:
```suggestion
      - name: Compute image refs
```

.github/workflows/pipeline.yml
```diff
@@ -0,0 +1,348 @@
+name: Pipeline
```
matthiasbruns: would it be possible to give this a better name? `pipeline.yml` could mean anything

.github/workflows/pipeline.yml
```diff
@@ -0,0 +1,348 @@
...
+        with:
+          script: |
+            const ref = process.env.INPUT_REF || process.env.GITHUB_HEAD_REF || process.env.GITHUB_REF_NAME;
+            core.setOutput('ref', ref);
+            if (context.eventName === 'pull_request') {
+              core.setOutput('publish', 'false');
+              return;
+            }
+            const defaultBranch = context.payload.repository?.default_branch || 'main';
+            const eligible = ref === defaultBranch || /^v\d+\.\d+(\.\d+)?(-.*)?$/.test(ref);
```
matthiasbruns, line 140: maybe add a comment what the regex would allow

.github/workflows/release.yml
```diff
@@ -0,0 +1,613 @@
+name: Release
+
+# Unified release flow. Creates a single v0.X.Y version covering both CLI and
+# controller. I replaced cli-release.yml and controller-release.yml with this trigger.
+#
+# Tag scheme on each release:
+#   v0.X.Y                              This is the main release version that will get released
+#   kubernetes/controller/v0.X.Y        Go-module tag for controller API so we don't break people who import it.
```
matthiasbruns, line 8: since you mentioned bringing back `cli` tags in the comments, would it make sense to reflect this here as well?

matthiasbruns (commented):
this release flow is awesome 🚀
structure and files, call-sites are way easy to understand - mostly minor comments.

.github/scripts/create-tag.js
Skarlso: ah fair.

.github/scripts/publish-final-release.js
Skarlso: which part? :DD

.github/workflows/pipeline.yml
Skarlso: I was thinking about this a lot. I thought pipeline is cool, because it's literally a pipeline. :D Stuff goes through it. But I'm open to suggestions. ;)

.github/scripts/publish-final-release.js
matthiasbruns: the whole :D

.github/workflows/pipeline.yml
matthiasbruns: I will *cook* something up while in the car. 
heisenberg

.github/scripts/create-tag.js
Skarlso: Good point.

jakobmoellerdev (commented):
This PR should not go in without a change to release process md reflecting all of this

.github/scripts/create-tag.js
Skarlso: Did it. Let's see if it's more readable.

Skarlso: Looks like:

```diff
-  const moduleTags = parseModuleTagList(moduleTagsRaw);
+  const moduleTags = (moduleTagsRaw || "").split(",").map(s => s.trim()).filter(Boolean);
```

.github/scripts/publish-final-release.js
Skarlso: BTW, some of the comments are temporary ofc, and I just added them for review purposes. But I will rewrite this one. :)

.github/workflows/conformance.yml
Skarlso: It means it's passin it through artifact mode in Github instead of oci registry. I'll try to rewrite this.

.github/workflows/release.yml
Skarlso: Yep. added.

Skarlso:
>This PR should not go in without a change to release process md reflecting all of this

Yes. I wrote it in the Description that I'll update it as soon as we agree on this at all. :)

Skarlso:
I can also update it right now instead I guess. :D

.github/scripts/release-versioning.js
```diff
@@ -107,39 +118,41 @@ export function run(core, executable, args) {
...
+ * Extract the base version prefix from a release branch name.
+ *
+ * @param {string} branch Release branch (e.g. "releases/v0.7").
+ * @returns {string} The base prefix (e.g. "0.7").
+ * @throws {Error} If the branch does not match the expected pattern.
+ */
 export function parseBranch(branch) {
   const match = /^releases\/v(0\.\d+)/.exec(branch);
   if (!match) throw new Error(`Invalid branch format: ${branch}`);
   return match[1];
```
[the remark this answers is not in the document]
Skarlso: Are you sure?

.github/scripts/publish-final-release.js
```diff
@@ -2,6 +2,14 @@
 import fs from "fs";
 import path from "path";
 
+// GitHub has a hard limit on the length of the changelog.
+// We could use git-cliff's [limit_commits](https://git-cliff.org/docs/configuration/git#limit_commits) setting, however,
+// that is a _hard_ limit. It will basically cut off at a preconfiugred number. Meaning if we have too many commits, the
+// user will not know that it was cut off.
+const GITHUB_RELEASE_BODY_LIMIT = 125000;
+const MAX_RELEASE_BODY_LENGTH = GITHUB_RELEASE_BODY_LIMIT - 5000; // safety buffer
+const TRUNCATION_NOTICE = `\n\n---\n\n*Release notes truncated to fit GitHub's ${GITHUB_RELEASE_BODY_LIMIT}-character body limit. See the source changelog or \`git log\`for the complete history.*`;
```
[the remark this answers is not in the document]
Skarlso: Are you sure?

.github/scripts/create-tag.js
```diff
@@ -66,66 +66,105 @@ function defaultExecGit(args) {
...
     }
-    core.info(`Tag ${tag} already exists at HEAD, skipping (idempotent)`);
-    core.setOutput("pushed", "true");
+    core.setFailed(`Tag ${tag} exists at ${existingSha.substring(0, 7)}, expected ${commit.substring(0, 7)}`);
+    return false;
+  }
+  createAndPushTag({ tag, commit, message, execGit });
+  core.info(`Created tag ${tag} at ${commit.substring(0, 7)}`);
+  return true;
+}
```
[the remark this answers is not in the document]
Skarlso: Are you sure?

.github/workflows/cli.yml
frewilhelm, line 12: I believe the `artifact_name` is unused

.github/workflows/cli.yml
```diff
@@ -138,33 +115,6 @@ jobs:
...
-          script: |
-            const ref = process.env.INPUT_REF || process.env.GITHUB_REF_NAME;
-            core.setOutput(
-              "should_push_oci_image",
-              ref === context.payload.repository.default_branch ||
-              /^cli\/v\d+\.\d+(\.\d+)?(-.*)?$/.test(ref)
-            )
-
       # Upload CLI binaries and OCI layout as build artifacts
       - name: Upload Artifacts
```
frewilhelm, line 120: Is my assumption correct that we only need the artifact-name for other workflows (that are called in the same context) to download the artifact? Because we are only propagating the name and I am wondering how the other workflows know which artifact to download from which workflow

.github/workflows/pipeline.yml
```diff
@@ -0,0 +1,349 @@
...
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8
+        with:
+          name: ${{ needs.build_cli.outputs.artifact_name }}
+
+      - name: Set LAYOUT
+        env:
+          LAYOUT: ${{ github.workspace }}/oci/cli.tar
+        run: |
+          test -e "$LAYOUT" || exit 1
+          echo "LAYOUT=$LAYOUT" >> "$GITHUB_ENV"
```
frewilhelm, line 170: What is the meaning of that? just checking if the tar (that is presumably downloaded in the step before) exist? And if it exists, we propagate the location (not the layout - whatever this might be) to GITHUB_ENV

.github/workflows/website-live-test-install-script.yml
```diff
@@ -52,7 +51,6 @@ jobs:
             }
 
             core.setOutput('version', version);
-            core.setOutput('tag', `${tagPrefix}${version}`);
```
frewilhelm, line 55: I guess you removed it because we did not use it anywhere, right?

.github/workflows/release.yml
```diff
@@ -0,0 +1,613 @@
...
+#
+# We will stop creating cli/v* tags. Since they are not required.
```
frewilhelm: I believe this contradicts your [comment](https://github.com/open-component-model/open-component-model/pull/2468#issuecomment-4406783568)

> Okay, it seems dropping cli/v* tag wasn't a good idea apparently. I'll add that back and then revert the fallback behavior of the script.

Is this up-to-date?

.github/workflows/release.yml
```diff
@@ -0,0 +1,613 @@
...
+  # === controller-release.yml::tag_rc ===
+
+  # === This is completely new. It starts the orchestration process for building and testing the components. ===
+  # It uses the new workflow helper called `pipline.yaml`.
+  # This will execute conformance tests, e2e tests, validations and builds for the respective components.
+  # This pipeline here is a daily pipeline as well.
+  build_and_test:
+    name: Build and Test
+    if: ${{ github.event.inputs.dry_run == 'false' && needs.tag_rc.outputs.pushed == 'true' }}
+    needs: [prepare, tag_rc]
```
frewilhelm, line 213: Is `prepare` needed here? I believe this is already guarded by `tag_rc` which needs `prepare`

.github/workflows/release.yml
```diff
@@ -0,0 +1,613 @@
...
+          script: |
+            const script = await import('${{ github.workspace }}/.github/scripts/create-tag.js');
+            await script.createRcTags({ core });
+  # === controller-release.yml::tag_rc ===
+
+  # === This is completely new. It starts the orchestration process for building and testing the components. ===
+  # It uses the new workflow helper called `pipline.yaml`.
+  # This will execute conformance tests, e2e tests, validations and builds for the respective components.
+  # This pipeline here is a daily pipeline as well.
+  build_and_test:
```
frewilhelm, line 210: what happens with the tags if this fails?

.github/workflows/release.yml
```diff
@@ -0,0 +1,613 @@
...
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          for file in ${{ runner.temp }}/binaries/ocm-*; do
+            echo "Verifying $(basename "$file")..."
+            gh attestation verify "$file" --repo "$GH_REPO"
+          done
+
+      - name: Verify CLI OCI attestation
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
frewilhelm, line 361: `gh` should be able to handle `GITHUB_TOKEN` too

> `GH_TOKEN`, `GITHUB_TOKEN` (in order of precedence): an authentication token that will be used when a command targets either `github.com` or a subdomain of `ghe.com`. Setting this avoids being prompted to authenticate and takes precedence over previously stored credentials.


Is there a reason why you overwrite this?

.github/workflows/release.yml
```diff
@@ -0,0 +1,613 @@
...
+        run: |
+          gh attestation verify "oci://$CONTROLLER_CHART@${CHART_DIGEST}" --repo "$GH_REPO"
+  # === cli-release.yml::verify_attestations + controller-release.yml::verify_attestations ===
+
+  # === Copied from controller-release.yml::promote_and_release_final + cli-release.yml::promote_final + cli-release.yml::release_final and merged ===
+  # These are literally copied verbatim from the original files from the above sentence.
+  # All I changed was applying some of the zizmor stuff.
+  promote_and_release_final:
+    name: Promote and Release Final
+    needs: [prepare, build_and_test, verify_attestations]
```
frewilhelm, line 386: I believe this should only need `verify_attestations`

frewilhelm (commented):
Preliminary review

.github/workflows/release.yml
```diff
@@ -0,0 +1,613 @@
...
+          passphrase: ${{ secrets.GPG_PASSPHRASE }}
+          git_user_signingkey: true
+          git_tag_gpgsign: true
+
+      - name: Create RC tags (canonical + cli + controller side tags)
+        id: tag
+        uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9
+        env:
+          CANONICAL_TAG: ${{ needs.prepare.outputs.new_tag }}
+          MODULE_TAGS: ${{ needs.prepare.outputs.cli_module_rc_tag }},${{ needs.prepare.outputs.controller_module_rc_tag }}
```
piotrjanik: script tskes: CANONICAL_TAG, ADDITIONAL_TAGS (not MODULE_TAGS)

.github/workflows/release.yml
```diff
@@ -0,0 +1,613 @@
...
+        with:
+          github-token: ${{ steps.get_token.outputs.token }}
+          script: |
+            const script = await import('${{ github.workspace }}/.github/scripts/create-tag.js');
+            await script.createRcTags({ core });
+  # === controller-release.yml::tag_rc ===
+
+  # === This is completely new. It starts the orchestration process for building and testing the components. ===
+  # It uses the new workflow helper called `pipline.yaml`.
```
piotrjanik: nit: typo "pipline"

.github/workflows/pipeline.yml
Skarlso: No idea. This is copied verbatim. :) I will not touch it, if it works. :D

.github/workflows/cli.yml
Skarlso: And...? I mean is this a question, or what...? :D

.github/workflows/cli.yml
Skarlso: Not a 100% sure what you are asking here, but all artifacts are part of the same workflow now. All artifacts are scoped to that run. This means that with just `name` we can resolve everything now.

.github/workflows/website-live-test-install-script.yml
Skarlso: Correct. :)

.github/workflows/release.yml
Skarlso: Fuck, I only fixed this in one place. :D thanks.

.github/workflows/release.yml
Skarlso: ah thanks!

.github/workflows/release.yml
Skarlso: the tags will exist, but no release is created. on next run they would be bumped, this would result in orphaned tags if we fail forward.

If we don't want that, since at this point, nothing really happened, it would be safe to delete the tags and rerun with the same version. But since we don't want that normally, we would fail forward.

.github/workflows/release.yml
Skarlso: GH_TOKEN is preferred by Zizmor.

.github/workflows/release.yml
Skarlso: prepare is needed here because build_and_test reads its outputs directly so without it it would resolve to empty string

.github/workflows/release.yml
Skarlso: nope

.github/workflows/cli.yml
frewilhelm: Why is it there if it is not used :D

.github/workflows/cli.yml
frewilhelm: ok perfect. I just wanted to confirm this because i was wondering. Thanks a lot!

.github/workflows/pipeline.yml
frewilhelm: :D oh no

.github/workflows/website-live-test-install-script.yml
frewilhelm: should we actually use it? :D I remember adding it to check against `ocm version`. Nothing we need right now though

.github/workflows/cli.yml
Skarlso: Oh! :D Good call. :D

.github/workflows/release.yml
frewilhelm: argh

.github/workflows/pipeline.yml
matthiasbruns: well, I cooked but forgot to respond

...

....

.....

what about:

```
 ____      _                       ____  _            _ _            
|  _ \ ___| | ___  __ _ ___  ___  |  _ \(_)_ __   ___| (_)_ __   ___ 
| |_) / _ \ |/ _ \/ _` / __|/ _ \ | |_) | | '_ \ / _ \ | | '_ \ / _ \
|  _ <  __/ |  __/ (_| \__ \  __/ |  __/| | |_) |  __/ | | | | |  __/
|_| \_\___|_|\___|\__,_|___/\___| |_|   |_| .__/ \___|_|_|_| |_|\___|
                                          |_|                        
```

Skarlso: The problem is that this isn't _just_ for releasing. :) But good start. :)

jakobmoellerdev: How about "monorepo shared build & publish"

.github/scripts/release-versioning.js
```diff
@@ -4,53 +4,63 @@ import { execFileSync } from "child_process";
...
-    const tagPrefix = `${componentPath}/v`;
 
-    // Get previous base version tag using Git's native version sort (descending)
-    // Filter out RC tags after fetching since git doesn't support negative pattern matching
+    // Stable tags on this minor: v0.X.Y (no component prefix)
     const stableTags = run(core, "git", [
-        "tag", "--list", `${tagPrefix}${basePrefix}.*`,
+        "tag", "--list", `v${basePrefix}.*`,
         "--sort=-version:refname"
     ]);
```
frewilhelm, line 29: Will this work with our previous tags? I ask because the `componentPath` is missing now

.github/scripts/publish-final-release.js
```diff
@@ -2,6 +2,14 @@
 import fs from "fs";
 import path from "path";
 
+// GitHub has a hard limit on the length of the changelog.
+// We could use git-cliff's [limit_commits](https://git-cliff.org/docs/configuration/git#limit_commits) setting, however,
+// that is a _hard_ limit. It will basically cut off at a preconfiugred number. Meaning if we have too many commits, the
+// user will not know that it was cut off.
+const GITHUB_RELEASE_BODY_LIMIT = 125000;
+const MAX_RELEASE_BODY_LENGTH = GITHUB_RELEASE_BODY_LIMIT - 5000; // safety buffer
+const TRUNCATION_NOTICE = `\n\n---\n\n*Release notes truncated to fit GitHub's ${GITHUB_RELEASE_BODY_LIMIT}-character body limit. See the source changelog or \`git log\` for the complete history.*`;
```
frewilhelm, line 11: I can't see it right now, but will the changelog now contain all commits between two tags or only the ones for `cli` and `kubernetes/controller`?

RELEASE_PROCESS.md
```diff
@@ -1,370 +1,104 @@
...
 
-This guide is for OCM release managers.
+For OCM release managers. CLI and Kubernetes Controller ship together in lockstep from a single workflow.
 
-Scope: **CLI and Kubernetes Controller are released together in version lockstep** (RC and final).
+## How it works
 
----
+* Development happens on `main`. Releases are cut from `releases/vX.Y` branches.
+* Minor and patch releases only (no majors).
```
frewilhelm, line 8: like never? or do we plan to have majors at some point?

RELEASE_PROCESS.md
```diff
@@ -1,370 +1,104 @@
...
+  * `v0.X.Y` (canonical, the GitHub release)
+  * `cli/v0.X.Y` (website install scripts)
+  * `kubernetes/controller/v0.X.Y` (Go module side tag)
+* Cadence: one release per sprint (two weeks). RC at sprint start, promote previous RC at the next sprint start.
 
-## General Information
+## Release responsible
 
-### Context and scope
+Rotates each sprint. Decides if a release is needed, runs the workflows, approves the gate, handles patches.
```
frewilhelm, line 88: This is very nit - but the release responsible cannot necessarily approve the gate if he/she started the workflow :D

RELEASE_PROCESS.md
```diff
@@ -1,370 +1,104 @@
...
-| When | Action |
-|------|--------|
-| Sprint Start | Create new release branch and RCs for both components |
-| During sprint | Issue patch releases or increment RCs if required (Security issues, bugs, ..) |
-| Sprint End | Assign next release responsible |
-| Next Sprint Start | Promote previous RCs to Final |
-
----
-
-## Release Workflow Diagram
```
frewilhelm, line 73: what is wrong with that diagram? :(

RELEASE_PROCESS.md
```diff
@@ -1,370 +1,104 @@
...
-2. Run workflow **[Controller Release](https://github.com/open-component-model/open-component-model/actions/workflows/controller-release.yml)** with equivalent inputs.
-3. Verify both pre-releases were created successfully on the GitHub Releases page.
-   - CLI assets: platform binaries (`ocm-*`) and OCI tarballs
-   - Controller assets: Helm chart (`.tgz`)
+Run [Create OCM Release Branch](https://github.com/open-component-model/open-component-model/actions/workflows/release-branch.yml) with `target_branch=releases/vX.Y`.
 
 > [!CAUTION]
-> **Always do a dry-run first** before the actual release.
-> CLI and Controller are released together. Do not release only one of them.
+> Once the branch exists, only bugfixes and docs land on it. Everything else needs release-responsible approval. Fix on `main` first, cherry-pick.
```
frewilhelm: why docs?

> Everything else needs release-responsible approval.

IMO the release-responsible should always be involved when backporting happens

frewilhelm: > Fix on `main` first, cherry-pick.

I believe this sentence needs a bit more love

RELEASE_PROCESS.md
```diff
@@ -1,370 +1,104 @@
...
 
-<details>
-<summary>Understanding the Dry-Run Summary</summary>
+## 2. Cut a release candidate
```
frewilhelm:
```suggestion
## 2. Create a release candidate
```

RELEASE_PROCESS.md
```diff
@@ -1,370 +1,104 @@
...
+2. `dry_run=true` first. Check the prepare summary: next RC version, next final version, set-latest decision.
+3. `dry_run=false` to actually cut the RC.
 
-**Important for Patch Releases:**
+The workflow builds CLI + controller, runs conformance, then publishes the RC as a GitHub pre-release with all assets attached.
 
-When releasing a patch for an older version line (e.g., `v0.16.1` when `v0.17.0` already exists),
-the dry-run summary will show:
+> [!IMPORTANT]
+> Always dry-run first. When patching an older minor (e.g. `v0.6.1` while `v0.7.0` exists), `set_latest=false` is expected. That prevents `latest` from rolling backwards.
```
frewilhelm:
```suggestion
> Always dry-run first. When patching an older minor (e.g. `v0.6.1` while `v0.7.0` exists), `set_latest=false` is required. That prevents `latest` from rolling backwards.
```

RELEASE_PROCESS.md
```diff
@@ -1,370 +1,104 @@
...
-- Final is only valid when both components are promoted in the same cycle.
-
-</details>
-
-### 4) Create a patch release (if needed)
-
-Patch releases address critical fixes for an already-released version.
-The fix must always land on `main` first, then be cherry-picked to the release branch.
+1. Wait for the RC
+2. Open the workflow run, approve the environment gate.
```
frewilhelm: I believe we should add another note or even a warning that the approval should only apply to RCs that were created in the sprint before/were tested/are in sprint cadence..

RELEASE_PROCESS.md
```diff
@@ -1,370 +1,104 @@
...
+3. The workflow verifies all attestations, re-packages the chart with final version strings, pushes everything, and publishes the GitHub release.
 
-> [!IMPORTANT]
-> ⚠️ **Older version lines:** When creating a patch for an older release line
-> (e.g., `v0.16.1` when `v0.17.0` exists), the release will NOT be marked as "latest"
-> on GitHub and the `latest` OCI tags will remain on the newer version.
-> This is intentional — use dry-run first to verify this behavior in the summary.
+If attestation verification fails the run aborts. Cut a fresh RC.
```
frewilhelm: "and do an investigation why the f the attestation verification failed" :D This should not happen

RELEASE_PROCESS.md
```diff
@@ -1,370 +1,104 @@
...
-
-# Create a PR to the release branch
-gh pr create \
-   --title "[releases/vX.Y] cherry-pick: <Original PR or Commit>" \
-   --body "Cherry-pick of <Original PR or Commit> from main to releases/vX.Y" \
-   --base releases/vX.Y \
-   --draft
+git cherry-pick <commit-from-main>
+gh pr create --base releases/vX.Y \
+  --title "[releases/vX.Y] cherry-pick: <original PR>"
```
frewilhelm: i believe we need the whole `chore: ` shenanigans again

RELEASE_PROCESS.md
```diff
@@ -1,370 +1,104 @@
...
 ```
 
-Merge the PR, then follow the RC and Final promotion steps as usual.
-
-</details>
-
----
+After merge, run the Release workflow on that branch. Same RC + promote flow as above.
```
frewilhelm: On a hotfix we usually do not wait for a sprint to promote it.

RELEASE_PROCESS.md
```diff
@@ -1,370 +1,104 @@
...
-**CLI and Controller versions diverged**
-
-- Ensure both workflows were run for the same release branch in the same cycle.
-- Stop promotion and align on a fresh RC cycle in lockstep.
+| Symptom | Action |
+|---------|--------|
+| Workflow failed before tag push | Re-run with `dry_run=true`, then `dry_run=false`. |
+| Promotion fails: attestation verify | Cut a fresh RC (attestations are built per-RC). |
+| CLI and controller diverged | Stop. Start a fresh RC cycle on the release branch. |
+| Conformance failed | Treat as a blocker. Fix on `main`, cherry-pick, new RC. |
```
frewilhelm: If the conformance failed, we cannot just fix something on main. This would be a crucial issue because it should've worked when cut the release branch.

If the attestation verification fails, this should be treated as a crucial security vulnerability until proven otherwise.

I would rather drop the whole table because none of these symptoms should occur and have an easy fix

RELEASE_PROCESS.md
```diff
@@ -1,370 +1,104 @@
...
+1. Mark the GitHub release as pre-release.
+2. Prepend a retraction notice to the body:
    ```markdown
-   > ⚠️ **RETRACTED**: This release has been retracted due to [reason].
-   > Please upgrade to vX.Y.Z instead.
+   > **RETRACTED**: superseded by vX.Y.Z due to <reason>.
    ```
-3. Create a new patch release with the fix as soon as possible.
-4. If security-related, consider filing a security advisory via GitHub's Security tab.
+3. Ship a patch release with the fix. File a security advisory if relevant.
```
frewilhelm: Additionally, the retraction should be made available in the [`go.mod`](https://go.dev/ref/mod#go-mod-file-retract) of the respective module

RELEASE_PROCESS.md
frewilhelm, line 1: this file feels very short now :D I can agree that the previous one was maybe a bit to detailed but I am afraid that you removed to many explanations. But I guess we can always add stuff later when necessary

frewilhelm:
Did you test if `website/static/install-cli.sh` will still work with this release flow?

Skarlso:
> Did you test if website/static/install-cli.sh will still work with this release flow?

The tag is still pushed the same way. So I think it should. I will test it explicitly.

.github/workflows/pipeline.yml
fabianburth: I like that. I think "kubernetes-controller" and "cli" could also be renamed to "controller build & test" and "cli build & test".
Then, this would form an intuitive set of workflows.

Skarlso: Perfect. Love it. Will do that. :)

.github/workflows/conformance.yml
```diff
@@ -139,7 +143,7 @@ jobs:
         if: ${{ inputs.toolkit_image_artifact != '' }}
         env:
           TOOLKIT_VERSION: ${{ inputs.toolkit_version }}
-          IMAGE_NAME: ${{ env.REGISTRY }}/${{ env.CONTROLLER_IMAGE_NAME }}
+          IMAGE_NAME: ${{ env.CONTROLLER_IMAGE_NAME }}
```
fabianburth: I believe we could omit this entirely and just use `env.CONTROLLER_IMAGE_NAME` directly

.github/workflows/release.yml
```diff
@@ -0,0 +1,611 @@
...
+
+permissions:
+  contents: read
+
+env:
+  REGISTRY: ghcr.io
+  CONTROLLER_PATH: kubernetes/controller
+  GH_REPO: ${{ github.repository }}
+  # Image refs (CLI_IMAGE, CONTROLLER_IMAGE, CONTROLLER_CHART) are set per-job
+  # via a "Compute image refs" step that lowercases github.repository_ownerk.
```
fabianburth:
```suggestion
  # via a "Compute image refs" step that lowercases github.repository_owner.
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,611 @@
...
+      - name: Compute next RC version
+        id: compute
+        uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9
+        env:
+          BRANCH: ${{ github.ref_name }}
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            const script = await import('${{ github.workspace }}/.github/scripts/release-versioning.js');
+            await script.default({ core, github, context });
```
fabianburth: nit: I believe only `core` is used by the default export

.github/workflows/release.yml
```diff
@@ -0,0 +1,611 @@
...
+          version: 'v2.12.0'
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          config: cliff.toml
+          args: |
+            --tag-pattern "^v\\d+\\.\\d+\\.\\d+(?:[-\\w\\.]+)?$" \
+            --ignore-tags "^v\\d+\\.\\d+\\.\\d+-rc\\.\\d+$" \
+            --tag "${{ steps.compute.outputs.new_tag }}" \
+            -o "$RUNNER_TEMP/CHANGELOG.md" \
+            --use-branch-tags \
+            ${{ steps.compute.outputs.changelog_range && format('-- "{0}"', steps.compute.outputs.changelog_range) || '--latest' }}
```
fabianburth, line 100: Since we use `steps.compute.outputs.changelog_range` within the `args` , the `env: CHANGELOG_RANGE` is currently obsolete, I believe.

.github/scripts/release-versioning.js
Skarlso: Technically, there won't be any previous tags. This will be a completely fresh, new release, since there are no such tags previously. We could _try_ to make this check the old releases but that would be insane. :) It would require so much fiddling, that it's just not feasible.

.github/scripts/publish-final-release.js
Skarlso: It's one release. So yes, everything. There is no more `cli` or `kubernetes/controller`. It's just `release`. :)

RELEASE_PROCESS.md
Skarlso: I guess not never 🤔

RELEASE_PROCESS.md
Skarlso: True. :D

RELEASE_PROCESS.md
Skarlso: It's not accurate anymore. :D 

I'll see if I can generate a new one. :)

RELEASE_PROCESS.md
Skarlso: >why docs?

This could also be website changes that we talked about yesterday. Because if there is a bugfix, you usually want to update some kind of documentation for it, right?

>IMO the release-responsible should always be involved when backporting happens

That's fair.

cliff.toml
```diff
@@ -1,6 +1,9 @@
 # git-cliff ~ configuration file
 # https://git-cliff.org/docs/configuration
-
+#
+# A unified cliff.toml. Replaces the per-component configs at cli/cliff.toml
+# and kubernetes/controller/cliff.toml. I copied over whatever was relevant
+# since both are now included in the release.
```
fabianburth, line 6: Do we have one unified changelog for both, cli and controllers now? 
If we do dependency upgrades or breaking changes, you now have to figure out by the scope of the conventional commit whether it affected cli or controller?

RELEASE_PROCESS.md
Skarlso: Fuck :D

.github/scripts/release-versioning.js
jakobmoellerdev: this will break our release model because we need to continue with 0.6 no?

RELEASE_PROCESS.md
Skarlso: Fair.

.github/scripts/publish-final-release.js
frewilhelm: This would mean that all commits for example the `bindings/go` modules would be part of the changelog too, right?

.github/workflows/conformance.yml
Skarlso: Oh nice spot.

.github/workflows/release.yml
Skarlso: Oh another nice one!

.github/workflows/release.yml
Skarlso: Goddamn you're on a roll. :D

cliff.toml
Skarlso: Yeah, that's right. We don't enforce `scope` I think, but yes, that's the only thing we can use to figure out what's a breaking change. We can do some enforcement on scope and then that would help a lot in ordering the changelog. Once that's enforced, we can do a `group_by` here as per cliff's doc. And that would do it already. I wouldn't want to extend the parser however, I think that way lie dragons.

Skarlso:
Because we did change a lot of things now, I'll run another release test on my fork to make sure we didn't break anything existing. :)

Skarlso:
https://github.com/Skarlso/open-component-model/actions/runs/25786980284

created the rc https://github.com/Skarlso/open-component-model/releases/tag/v0.9.0-rc.1
and the release https://github.com/Skarlso/open-component-model/releases/tag/v0.9.0

Skarlso:
Also Fixes https://github.com/open-component-model/ocm-project/issues/904.

Skarlso:
The release passed :tada:

.github/scripts/release-versioning.js
Skarlso: Yes. This will be a new release since this will be a new canonical tag.

RELEASE_PROCESS.md
frewilhelm: I believe the previous diagram was about the "theoretical" release process not the technical one. And it is still accurate regarding our RC and hotfix pattern i believe

RELEASE_PROCESS.md
```diff
@@ -1,370 +1,139 @@
...
-- Rerun as dry-run first, then rerun with `dry_run=false`.
-
-**Final promotion fails with "no RC found"**
-
-- Verify that the latest RC tag exists on the release branch.
-- Create a new RC for both components, then promote again.
+Generated by [git-cliff](https://git-cliff.org) from conventional commits. No manual editing for normal runs.
 
-**CLI and Controller versions diverged**
+### Retract a release
```
frewilhelm:
```suggestion
## Retract a release
```

.github/scripts/publish-final-release.js
Skarlso: Yes, it would be. I can actually fix this part though. 🤔  I think that should be possible with adding `--include-path` back into the cliff toml.

frewilhelm: I believe this could make sense. or we even create two changelogs and append them. Otherwise changes from bindings/go that are added to cli and kubernetes/controller would appear doubled in the same list

RELEASE_PROCESS.md
Skarlso: I think this is honestly better. :D The previous one didn't contain too much detail.. I will update it to have something like that, but I would want to keep this information as is. 🤔

Skarlso: @frewilhelm please check the new diagram. :)

.github/scripts/release-versioning.js
Skarlso: @jakobmoellerdev Are you okay with that?

RELEASE_PROCESS.md
frewilhelm: nice - thanks a lot!

frewilhelm:
> > Did you test if website/static/install-cli.sh will still work with this release flow?
> 
> The tag is still pushed the same way. So I think it should. I will test it explicitly.

Did it work?

.github/scripts/release-versioning.js
jakobmoellerdev: this does not address my concern that we cant start with 0.6. im not willing to reset to 0.1

.github/workflows/release.yml
```diff
@@ -0,0 +1,610 @@
...
+          config: cliff.toml
+          args: |
+            --include-path "cli/**" \
+            --include-path "kubernetes/controller/**" \
+            --tag-pattern "^v\\d+\\.\\d+\\.\\d+(?:[-\\w\\.]+)?$" \
+            --ignore-tags "^v\\d+\\.\\d+\\.\\d+-rc\\.\\d+$" \
+            --tag "${{ steps.compute.outputs.new_tag }}" \
+            -o "$RUNNER_TEMP/CHANGELOG.md" \
+            --use-branch-tags \
+            ${{ steps.compute.outputs.changelog_range && format('-- "{0}"', steps.compute.outputs.changelog_range) || '--latest' }}
```
frewilhelm, line 100: I tried this locally with `v0.6` on a fresh open-component-model local repository and got an error because the tag `v0.5.0` does not exist. Should we create the canonical tag to make this work? Or how did you make it run on your fork?

.github/scripts/release-versioning.js
Skarlso: Oh sorry, I misunderstood 🤦 Of course we can start with 0.6.

.github/workflows/release.yml
Skarlso: That would be weird. 🤔 How did you test this? I literally started with 0.7.0. :) https://github.com/Skarlso/open-component-model/releases/tag/v0.7.0

https://github.com/Skarlso/open-component-model/actions/runs/25546869852

```

Previous Release Tag | (none — first release)
-- | --
Changelog Range | (full history)

```

Skarlso: Can you share the exact error?

Skarlso: Maybe you might have `v0.5.0` on your fork somehow by accident?

frewilhelm: I followed the checkout procedure from the checkout action. Those tags are available on upstream (if i am not mistaken):

```
$> git tag --list | grep v0.5
cli/v0.5.0
cli/v0.5.0-rc.1
kubernetes/controller/v0.5.0
kubernetes/controller/v0.5.0-rc.1
```

I believe there is no tag `v0.5.0` that we can run `git-cliff` against

frewilhelm: I ran this command:
```
open-component-model(releases/v0.6): git-cliff --config=cliff.toml --tag-pattern "^v\\d+\\.\\d+\\.\\d+(?:[-\\w\\.]+)?$"   --ignore-tags "^v\\d+\\.\\d+\\.\\d+-rc\\.\\d+$"   --tag "v0.6.0-rc.1"   -o "$CHANGELOG.md"   --use-branch-tags   -- "v0.5.0..HEAD"
```

```
Error: SetCommitRangeError("v0.5.0..HEAD", Error { code: -3, klass: 4, message: "revspec 'v0.5.0' not found" })
```

Skarlso: But... that's not how the job works. 🤔 It would never end up as `v0.5.0..HEAD` if v0.5.0 doesn't exist. I'm not sure what your are trying to test here with this.

frewilhelm: What will be the commit range when we run the release action on upstream?

Skarlso: It would be... a full release. :) Like, I said. It would be like a first release.

That's why I asked if a first release is okay. If we don't want that, we can create a pseudo previous tag that contains everything up until the last release manually and then the new release would only contain the diff until then.

frewilhelm: > It would be... a full release. :) Like, I said. It would be like a first release.

I missed that completely, sorry.

Skarlso:
> > > Did you test if website/static/install-cli.sh will still work with this release flow?
> > 
> > 
> > The tag is still pushed the same way. So I think it should. I will test it explicitly.
> 
> Did it work?

ah, yep 👍

frewilhelm (commented):
I just wondered about the TOCTOU window between verifying the OCI images for CLI, controller, and chart and re-tagging them for the promoted release. I noticed that we actually verify `IMAGE-REFERENCE:DIGEST` and then re-tag `IMAGE-REFERENCE:RC-TAG`. 

Is it possible that someone with malicious intend and write access to the registry, could re-tag our initial image tagged with the RC and we would then promote it to an official release because we only verify the image based on the digest and not the RC tag?

EDIT: Besides that I believe all of this should work!

Skarlso:
> I just wondered about the TOCTOU window between verifying the OCI images for CLI, controller, and chart and re-tagging them for the promoted release.

These things don't happen at the same time. So there shouldn't be any TOCTOU.

.github/scripts/release-versioning.js
```diff
@@ -107,39 +118,41 @@ export function run(core, executable, args) {
   }
```
morri-son, line 118: This issue was there before:  try/catch swallows infra errors and returns "", the same value as a `git tag --list v0.x.*` with no matches (no stdout and RC 0). But in case we have previousBaseVersions (or RC), the version bumps would not happen correctly, or?

.github/workflows/pipeline.yml
```diff
@@ -0,0 +1,348 @@
...
+          {
+            echo "IMAGE_NAME=ghcr.io/$OWNER/kubernetes/controller"
+            echo "CHART_REPO=ghcr.io/$OWNER/kubernetes/controller/chart"
+          } >> "$GITHUB_ENV"
+
+      - name: Checkout
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          repository: ${{ github.event.pull_request.head.repo.full_name || github.repository }}
+          ref: ${{ inputs.ref || github.head_ref || github.ref_name }}
```
morri-son, line 248: I guess triggered from a PR we never hit `github.event.pull_request.head.repo.full_name` and `github.head_ref` as we're in the publish job. publish == 'true' is not emited for PRs. So, we may remove `repository` completely as it defaults to `github.repository` and `github.head_ref`can also be removed.

.github/scripts/create-tag.js
morri-son, line 55: Not introduced by this PR, but maybe worth mentioning it, especially in the light of constant GH issues 😊. Does not block this PR, but maybe worth for a follow-up. 

It may be that there is a delay (second to tens of seconds) in the GH API delivering results of the `git push`, e.g. using a `git fetch`. So, we possibly see `couldn't find remote ref` two jobs downstream from the script action. Now that we push three tags we may even run into this problem more-likely. We can introduce a check if the tag resolves to the expected SHA in `createAndPushTag`. This introduces exactly the delay the GH API has (second to tens of seconds) in every run, but the error is much clearer as if we get an ominous one later in the workflow.

.github/scripts/publish-final-release.js
morri-son, line 137: Again a pre-existing issue, not introduced here, and in the light of GH issues. Does not block this PR, but maybe worth for a follow-up. Since we now also have more release assets, it makes sense to stabilize the upload, not to get stalled on just x of y assets. We can introduce a retry on uploadReleaseAsset and check for 502/503/422 errors. Since we already have all data from the API, we could use res.data.state == "uploaded" and res.data.size ==
  data.length for a verification to avoid marking assets as "Uploaded" which are corrupted.

morri-son:
@Skarlso looks pretty solid 👍🏻, great consolidation / cleanup / streamlining of our former workflows. I'm sure this will help us a lot! I made just some comments, none of them blocks the approval, but maybe worth a follow-up PR.

.github/scripts/release-versioning.js
Skarlso: Sure, will fix in a follow-up.

.github/workflows/pipeline.yml
Skarlso: Sure, will fix in a follow-up.

.github/scripts/create-tag.js
Skarlso: Sure, will fix in a follow-up.

.github/scripts/publish-final-release.js
Skarlso: Sure, will fix in a follow-up.

Skarlso: https://github.com/open-component-model/ocm-project/issues/1076

.github/scripts/create-tag.js
Skarlso: https://github.com/open-component-model/ocm-project/issues/1076

.github/workflows/pipeline.yml
Skarlso: https://github.com/open-component-model/ocm-project/issues/1076

.github/scripts/release-versioning.js
Skarlso: https://github.com/open-component-model/ocm-project/issues/1076

matthiasbruns (approved):

jakobmoellerdev (approved):

Skarlso:
Frederic's concern is valid and will be fixed in the followup -> https://github.com/open-component-model/ocm-project/issues/1076

PR #2469, opened 2026-05-08
feat!: revert typed identy and move back to runtime.Identity

matthiasbruns:

We introduces `runtime.Typed` for both credentials and identities.
While valuable for credentials, our code-base heavily relies on `runtime.Identity`. The problem is that we have to type-assign at several places just to migrate from `runtime.Identity` to `runtime.Typed`. There is no real benefit yet, nor do we have a working concept or even idea how we would provide identity comparison across plugin-boundaries for arbitrary types. We can't even imagine identities that would need such a change right now.

~I kept `CredentialTypeSchemeProvider` for now - no idea if this adds much to our system anymore.~

Rolls back: https://github.com/open-component-model/ocm-project/issues/800


matthiasbruns:
@coderabbitai review

bindings/go/credentials/graph.go
```diff
@@ -59,11 +59,11 @@ type Graph struct {
 	repositoryPluginProvider     RepositoryPluginProvider     // injection for resolving custom repository types
 	credentialPluginProvider     CredentialPluginProvider     // injection for resolving custom credential types
 	identityTypeRegistry         *IdentityTypeRegistry        // validates consumer identity types and accepted credentials
-	credentialTypeSchemeProvider CredentialTypeSchemeProvider // validates credential types from config
+	credentialTypeSchemeProvider CredentialTypeSchemeProvider // optional: enables typed credential ingestion
```
jakobmoellerdev, line 57: hmm, feels to me like this is dead code now, but will defer that decision to @fabianburth

matthiasbruns: Yeah. We will discuss if we need this at all on Monday. I don't think so just to log a warning. Doesn't make sense to me.

bindings/go/credentials/synced_dag.go
```diff
@@ -38,12 +38,12 @@ func (g *syncedDag) getVertex(id string) (v *dag.Vertex[string], ok bool) {
...
 
-func (g *syncedDag) getIdentity(id string) (runtime.Typed, bool) {
+func (g *syncedDag) getIdentity(id string) (runtime.Identity, bool) {
 	v, ok := g.getVertex(id)
 	if !ok {
 		return nil, false
 	}
-	identity, ok := v.Attributes[attributeIdentity].(runtime.Typed)
+	identity, ok := v.Attributes[attributeIdentity].(runtime.Identity)
 	return identity, ok
```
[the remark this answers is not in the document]
matthiasbruns: did you check how it is stored? does the graph store the original or a shallow-copy of the identity?

bindings/go/credentials/type_registry.go
```diff
@@ -1,88 +1,100 @@
...
 		return err
 	}
 
-	if len(acceptedCredentialTypes) > 0 && len(types) > 0 {
+	if len(acceptedCredentialTypes) > 0 {
 		r.mu.Lock()
 		defer r.mu.Unlock()
-		// Store under the default type (first in the list).
 		r.acceptedCreds[types[0]] = acceptedCredentialTypes
 	}
```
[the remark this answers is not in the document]
matthiasbruns: we will probably drop the identity->credential mapping completely

bindings/go/credentials/graph.go
matthiasbruns: removed

jakobmoellerdev (approved):

Skarlso (approved):

PR #2484, opened 2026-05-11
fix(website): add manual verification instructions to install script

### LINKED ISSUE #2348: Verify release binaries without requiring gh CLI authentication
## Summary

The install script currently relies on `gh attestation verify` to check release artifact integrity, which requires the user to be authenticated with the GitHub CLI. If the user is not authenticated, a warning is printed but installation proceeds with an **unverified** binary.

We should provide a verification method that does not require authentication or a GitHub-specific tool.

## Motivation

- **Security:** An unauthenticated user (or a script calling the installer non-interactively) can silently end up with an unverified binary if the warning is missed.
- **OpenSSF Scorecard:** Direct signature verification on release artifacts is likely required to improve the project's OpenSSF score.
- **Usability:** Users should not need a GitHub token just to verify a download.

### Option 1: Checksums only

Generate a `checksums.txt` with SHA256 hashes per binary. The install script downloads it and verifies the binary hash.

- **Pros:** Zero extra dependencies (`sha256sum`/`shasum` are ubiquitous). Simple to implement. This is what **kubectl** and **Go** do.
- **Cons:** Verifies integrity (was the file corrupted?) but not authenticity (did it come from us?). An attacker who can tamper with the binary on GitHub Releases can also tamper with the checksums file.
- **Reference:** [kubernetes/kubectl install docs](https://kubernetes.io/docs/tasks/tools/install-kubectl-linux/)

### Option 2: GPG detached signatures

Sign the checksums file (or each binary) with a project GPG key. Publish the public key. The install script verifies using `gpg --verify`.

- **Pros:** `gpg` is pre-installed on most Linux systems. Verifies both integrity and authenticity. Well-understood, battle-tested. This is what **Helm** does.
- **Cons:** Requires managing a GPG keypair (generation, rotation, revocation). macOS does not ship `gpg` by default. Helm makes signature verification opt-in for this reason (`VERIFY_SIGNATURES=true`).
- **Reference:** [helm/helm install script](https://github.com/helm/helm/blob/main/scripts/get-helm-3), [Helm release assets](https://github.com/helm/helm/releases)

### Option 3: Cosign keyless signing (Sigstore)

Sign the checksums file with `cosign sign-blob` in the release workflow using GitHub Actions OIDC identity. No private key to manage.

- **Pros:** No key management. Identity tied to the GitHub Actions workflow. Transparency log provides an audit trail. This is what **yq** does.
- **Cons:** Requires `cosign` binary to verify - not pre-installed anywhere. Creates a bootstrapping problem for an install script (who verifies cosign?). Effectively replaces the `gh` dependency with a `cosign` dependency.
- **Reference:** [mikefarah/yq release workflow](https://github.com/mikefarah/yq/blob/master/.github/workflows/release.yml)

### Option 4: Keep `gh attestation` but add checksums as fallback

Keep the current `gh attestation verify` path for authenticated users. Additionally generate and publish a checksums file so unauthenticated users get at least integrity verification.

- **Pros:** No breaking change. Authenticated users still get full attestation. Unauthenticated users get checksums instead of nothing.
- **Cons:** Does not fully solve the authenticity problem for unauthenticated users. Two code paths to maintain.

### Option 5 (chosen): Just log a warning message
 
We will try gh attestation verify, but if that fails, the user will see a detailed message explaining how to verify binaries manually. 

## Comparison

| Approach | Extra dependency | Integrity | Authenticity | Key management |
|---|---|---|---|---|
| Checksums only | None | Yes | No | None |
| GPG signatures | `gpg` (usually pre-installed) | Yes | Yes | Manual (keypair) |
| Cosign keyless | `cosign` (not pre-installed) | Yes | Yes | None (OIDC) |
| `gh attestation` (current) | `gh` + auth token | Yes | Yes | None (GitHub-managed) |
| Checksums + `gh attestation` fallback | None / `gh` | Yes | Only if authenticated | None |

## Implementation notes

- The release workflow is in `.github/workflows/cli-release.yml` and uses a custom build step (no GoReleaser).
- A checksums file would need to be generated during the build step (e.g. `sha256sum ocm-* > checksums.txt`) and uploaded as a release asset.
- The install script lives at `website/static/install.sh`.

## References

- https://github.com/open-component-model/open-component-model/pull/2322#issuecomment-4273077142

jakobmoellerdev:
## Summary

When `gh` is unavailable or unauthenticated, show detailed manual verification instructions instead of a terse warning. Three options: `gh attestation verify`, `cosign verify-blob-attestation`, and manual SHA-256 hash check via the public GitHub attestation API.

Closes #2348

## Test plan

- [x] `bash install-cli.sh` with `gh` authenticated → attestation verified
- [x] `bash install-cli.sh` without `gh` → detailed warning printed
- [x] `OCM_SKIP_VERIFY=true` → no warning
- [x] Option B (cosign) verified against cli/v0.5.0 and cli/v0.6.0-rc.1
- [x] CI test updated to match new warning format

##### Verification


website/static/install-cli.sh
```diff
@@ -186,28 +186,72 @@ download_binary() {
...
+    warn "══════════════════════════════════════════════════════════════════════"
+    warn ""
+    warn "  Reason: ${reason}"
+    warn ""
+    warn "  After installation completes, verify the binary at:"
+    warn "    ${BIN_DIR}/ocm"
+    warn ""
+    warn "  Option A — Verify with GitHub CLI (recommended):"
+    warn "    1. Install gh: https://cli.github.com/"
+    warn "    2. Authenticate: gh auth login"
```
frewilhelm:
```suggestion
    warn "    2. Authenticate and chose GitHub.com: gh auth login"
```

website/static/install-cli.sh
```diff
@@ -186,28 +186,72 @@ download_binary() {
...
+    warn "    3. Verify the installed binary:"
+    warn "       gh attestation verify ${BIN_DIR}/ocm --repo ${GITHUB_REPO}"
+    warn ""
+    warn "  Option B — Verify with cosign (no GitHub auth needed):"
+    warn "    DIGEST=\"sha256:\$(${hash_cmd} ${BIN_DIR}/ocm | cut -d' ' -f1)\""
+    warn "    curl -sfL \\"
```
frewilhelm: Having the `[WARN]` in the output makes it pretty hard to just copy the commands. 

```text
[WARN]      curl -sfL \
[WARN]        "https://api.github.com/repos/open-component-model/open-component-model/attestations/${DIGEST}" \
[WARN]        | jq -r '.attestations[0].bundle' > attestation.jsonl
[WARN]      cosign verify-blob-attestation \
[WARN]        --bundle attestation.jsonl \
[WARN]        --new-bundle-format \
[WARN]        --certificate-oidc-issuer https://token.actions.githubusercontent.com \
[WARN]        --certificate-identity-regexp \
[WARN]          '^https://github.com/open-component-model/open-component-model/' \
[WARN]        <path>/.local/bin/ocm
```


Can we put all this in just `warn` call, so its easier to copy it?

website/static/install-cli.sh
```diff
@@ -186,28 +186,72 @@ download_binary() {
...
+    warn "      \"https://api.github.com/repos/${GITHUB_REPO}/attestations/\${DIGEST}\" \\"
+    warn "      | jq -r '.attestations[0].bundle' > attestation.jsonl"
+    warn "    cosign verify-blob-attestation \\"
```
frewilhelm: This command failed:

```bash
$> cosign verify-blob-attestation --bundle attestation.jsonl --new-bundle-format --certificate-oidc-issuer https://token.actions.githubusercontent.com  --certificate-identity-regexp '^https://github.com/open-component-model/open-component-model/' <path>/.local/bin/ocm
Error: invalid predicate type, expected custom got https://slsa.dev/provenance/v1
error during command execution: invalid predicate type, expected custom got https://slsa.dev/provenance/v1
```

Any idea what the problem could be?

website/content/docs/getting-started/ocm-cli-installation.md
```diff
@@ -100,6 +100,76 @@ Expected output:
...
+If automatic verification is unavailable, you can verify manually using one of the methods below.
+
+{{< tabs "verification-methods" >}}
+
+{{< tab "GitHub CLI" >}}
+
+The simplest method — requires the [GitHub CLI](https://cli.github.com/) with authentication.
+
+```shell
+gh auth login
```
frewilhelm: I believe it should prompt you against which GitHub instance you want to authenticate to.

website/static/install-cli.sh
```diff
@@ -186,28 +186,72 @@ download_binary() {
...
+    warn "    ${BIN_DIR}/ocm"
+    warn ""
+    warn "  Option A — Verify with GitHub CLI (recommended):"
+    warn "    1. Install gh: https://cli.github.com/"
+    warn "    2. Authenticate: gh auth login"
+    warn "    3. Verify the installed binary:"
+    warn "       gh attestation verify ${BIN_DIR}/ocm --repo ${GITHUB_REPO}"
+    warn ""
+    warn "  Option B — Verify with cosign (no GitHub auth needed):"
+    warn "    DIGEST=\"sha256:\$(${hash_cmd} ${BIN_DIR}/ocm | cut -d' ' -f1)\""
```
matthiasbruns: we might have to make this os specific - as you stated in https://github.com/open-component-model/open-component-model/pull/2484/changes#diff-bf5afbfa91d7722f67d8faf69a991f849ac7777072c324628c133326350ed525R128

website/static/install-cli.sh
```diff
@@ -186,28 +186,72 @@ download_binary() {
...
+    warn "    cosign verify-blob-attestation \\"
+    warn "      --bundle attestation.jsonl \\"
+    warn "      --new-bundle-format \\"
+    warn "      --certificate-oidc-issuer https://token.actions.githubusercontent.com \\"
+    warn "      --certificate-identity-regexp \\"
+    warn "        '^https://github.com/${GITHUB_REPO}/' \\"
+    warn "      ${BIN_DIR}/ocm"
+    warn ""
+    warn "  Option C — Manual SHA-256 hash check (integrity only):"
+    warn "    DIGEST=\"sha256:\$(${hash_cmd} ${BIN_DIR}/ocm | cut -d' ' -f1)\""
```
matthiasbruns: same here

website/content/docs/getting-started/ocm-cli-installation.md
jakobmoellerdev: I think we can hardcode it to github.com since we dont host anywhere else

website/static/install-cli.sh
jakobmoellerdev: I think the attestation type default in older cosign versions didnt work, i tested with cosign 3.x. Anyhow i added the correct type in there now, please double check

website/static/install-cli.sh
jakobmoellerdev: Is the change I pushed what you thought of?

website/static/install-cli.sh
jakobmoellerdev: please see my new version where I hardcoded github.com

website/static/install-cli.sh
jakobmoellerdev: confused, isn't hash_cmd os specific already?

website/static/install-cli.sh
jakobmoellerdev: lets discuss above

website/static/install-cli.sh
matthiasbruns: confused me - you are right

frewilhelm (approved):
Very cool - thanks a lot!

website/static/install-cli.sh
frewilhelm: Worked like a charm now

website/static/install-cli.sh
frewilhelm: yes. now i can copy, paste, and run the command. Thank you!

website/static/install-cli.sh
frewilhelm: yes, that works!

morri-son (approved):

PR #2489, opened 2026-05-11
chore(docs): ADR 0022 — streaming oci resource transfer via oras storage

### LINKED ISSUE #2485: Direct OCI-to-OCI streaming transfer transformation (TransferOCIArtifact/v1alpha1)

```
```

Problems:

## Proposed Design

```
```

### Activation

### Interface opt-in

```go
    repository.ResourceRepository
}
```

## Components

| What | Where |
|------|-------|

## Prototype

## Acceptance Criteria

## Out of Scope

jakobmoellerdev:
## Summary

- Adds ADR 0022 proposing `ResourceStream` interface (OCI-specific) that wraps `content.ReadOnlyGraphStorage` from oras-go
- Eliminates tar materialization for OCI→OCI and OCI→CTF transfers by exposing the oras store already inside our repository implementations
- Generic `ResourceRepository` interface stays unchanged; streaming is opt-in via `StreamingResourceRepository` in the OCI package

## Design

`DownloadResourceStream` returns a lazy handle (zero I/O). `UploadResourceStream` passes it to `oras.CopyGraph` which streams blob-by-blob with HEAD-check dedup. Tar only created if `Materialize()` is explicitly called (legacy transformer compat).

## Motivation

Resolves #2485 — replaces the fused `TransferOCIArtifact` transformer approach with a cleaner abstraction at the repository level that benefits all target types (OCI, CTF) without graph builder special-casing.

## Test plan

- [ ] Review ADR for technical accuracy
- [ ] Validate oras interface compatibility with current `oras-go/v2` dependency
- [ ] Discuss with team: naming, package placement, migration phases


fabianburth (approved):

fabianburth:
Not sure if `Materialize` sounds a bit too magic.

matthiasbruns (approved):

jakobmoellerdev:
i can think of an alternative name for that method while engineering on this, no problem :+1:

PR #2491, opened 2026-05-11
feat(cli): add version update notification

### LINKED ISSUE #1044: Add version update notification to OCM CLI
**Description**
- Add a version check to the OCM CLI that warns when a newer release is available
- Check runs on command invocation, comparing the current binary version against the latest published GitHub release
- Warning is printed to stderr (non-blocking) with upgrade guidance
- Results are cached locally (~24h) to avoid network calls on every invocation
- Warning is rate-limited to show at most once per time interval (~24h), even if the cached check indicates a newer version
- Fails silently on network errors and in offline/air-gapped environments
- Opt-out via flag or ocm config

**Done Criteria**
- [x] Version check implemented with local caching and silent failure on network errors
- [x] Opt-out mechanism via flag and environment variable

jakobmoellerdev:
## Summary

Adds a non-blocking version update notification to the OCM CLI that warns users when a newer stable release is available.

- Async version check via goroutine + `cobra.OnFinalize` — no CLI slowdown
- Compares running binary version against latest `cli/v*` GitHub release
- 24h cache in `$XDG_CACHE_HOME/ocm/version-check.json`
- Rate-limited warnings via `slog.Warn` (max once per 24h)
- Silent failure on network errors / offline / air-gapped environments
- Only considers stable releases (excludes drafts, pre-releases, non-CLI tags)

### Opt-out mechanisms (priority order)

1. `OCM_DISABLE_VERSION_CHECK=1` environment variable
2. ~`--version-check=disable` persistent flag~ this was cut during review
3. OCM config file: `versioncheck.cli.config.ocm.software/v1alpha1` with `disabled: true`

### Skip conditions

- Dev builds (`BuildVersion == "n/a"`)
- The `ocm version` command itself
- Any of the above opt-out mechanisms active

Resolves: open-component-model/ocm-project#1044

## Test plan

- [x] Unit tests for versioncheck package (83.6% coverage)
- [x] Integration tests hitting real GitHub API (6 tests)
- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] All existing unit tests pass (no regressions)
- [ ] Manual test with `BuildVersion=0.0.1` shows slog warning
- [ ] Manual test with `OCM_DISABLE_VERSION_CHECK=1` suppresses warning
- [ ] Manual test offline — no error, no delay


cli/docs/reference/ocm.md
```diff
@@ -60,6 +60,8 @@ ocm [sub-command] [flags]
       --plugin-directory string            default directory path for ocm plugins. (default "$HOME/.config/ocm/plugins")
       --plugin-shutdown-timeout duration   Timeout for plugin shutdown. If a plugin does not shut down within this time, it is forcefully killed (default 10s)
       --temp-folder string                 Specify a custom temporary folder path for filesystem operations.
+      --version-check enum                 control automatic version update checks
```
matthiasbruns: i think this deserves its own config command - not a flag that every sub cmd now needs to document

jakobmoellerdev: I have an ocm config struct that exposes this already, can simply drop the flag if thats desired

morri-son: if the check is the default and can be disabled in the config, I would drop the flag

jakobmoellerdev: now dropped the flag

cli/cmd/setup/versioncheck.go
```diff
@@ -0,0 +1,136 @@
...
+		slog.Debug("version check disabled via config policy")
+		return true
+	}
+
+	return false
+}
+
+// envDisabled checks the OCM_DISABLE_VERSION_CHECK environment variable.
+func envDisabled() bool {
+	val := os.Getenv(VersionCheckEnvVar)
```
matthiasbruns, line 103: What about parse bool here?

https://pkg.go.dev/strconv#ParseBool

cli/cmd/setup/versioncheck.go
```diff
@@ -0,0 +1,136 @@
...
+	vcCfg, err := versioncheck.LookupConfig(cfg)
+	if err != nil {
+		slog.Debug("version check config lookup failed", slog.String("error", err.Error()))
+		return false
+	}
+	return vcCfg.Policy == versioncheck.PolicyDisable
+}
+
+// printWarning logs the upgrade notification using the structured logging stack.
+func printWarning(result *versioncheck.Result) {
```
matthiasbruns: Sounds like a general warning until. Maybe printUpgradeWarning?

cli/internal/versioncheck/check.go
```diff
@@ -0,0 +1,225 @@
...
+type githubRelease struct {
+	TagName    string `json:"tag_name"`
+	Draft      bool   `json:"draft"`
+	Prerelease bool   `json:"prerelease"`
+}
+
+// fetchLatestVersion queries the GitHub Releases API and returns the highest stable CLI version.
+// It filters out drafts, pre-releases, non-CLI tags, and semver pre-release versions.
+func fetchLatestVersion(ctx context.Context, opts Options) (string, error) {
+	url := fmt.Sprintf("%s/repos/%s/%s/releases?per_page=%d",
```
matthiasbruns, line 178: We. About just filter for releases and not drafts can we?

matthiasbruns (commented):
Nits. Good work.

cli/internal/versioncheck/check.go
jakobmoellerdev: i think this is doable, checking

cli/internal/versioncheck/config.go
```diff
@@ -0,0 +1,82 @@
...
+// --version-check CLI flag. This allows administrators to disable version checks
+// for an entire environment via config, while individual users can still re-enable
+// them via --version-check=auto on the command line.
+//
+// Example config:
+//
+//	type: generic.config.ocm.software/v1
+//	configurations:
+//	- type: versioncheck.cli.config.ocm.software/v1alpha1
+//	  policy: disable
```
piotrjanik, line 37: so that we dont have same problem as with access types, should policies be CamelCase, snake_case or lowercase? I think it should be consistent. For referrer I added Auto|Disabled as types are also uppercase, but I dont have here any strong opinion.

cli/internal/versioncheck/check.go
```diff
@@ -0,0 +1,228 @@
...
+		slog.Debug("version check: cache read failed", slog.String("error", err.Error()))
+	}
+	if cache != nil && cache.IsFresh(now) {
+		return compareVersions(current, cache.LatestVersion)
+	}
+
+	// Cache is stale or missing — fetch from GitHub.
+	latest, err := fetchLatestVersion(ctx, opts)
+	if err != nil {
+		return nil
```
piotrjanik, line 123: fetchLatestVersion can return error in many cases, here it is swallowed. Maybe slog.Debug?

cli/internal/versioncheck/check.go
```diff
@@ -0,0 +1,228 @@
...
+
+	// Persist the new result, preserving the previous WarnedAt timestamp.
+	entry := &CacheEntry{
+		LatestVersion: latest,
+		CheckedAt:     now,
+	}
+	if cache != nil {
+		entry.WarnedAt = cache.WarnedAt
+	}
+	if err := WriteCache(cacheDir, entry); err != nil {
```
piotrjanik, line 134: What if a user is behind a firewall or offline and can't retrieve versions? In that case they'd face a few seconds of delay on every invocation, right? Maybe we should store extra info in the cache — e.g. retry in 1h.  After several failures we could also print a hint: "Sorry, we couldn't fetch the latest version. If you're behind a firewall, you can disable the version check."

Tha could be the case in air-gapped envs, right?

cli/internal/versioncheck/config.go
jakobmoellerdev: @fabianburth please give your opinion on this, will standardize around that

cli/internal/versioncheck/check.go
jakobmoellerdev: hmm im wondering if the 24 hour interval should also count for failure

cli/cmd/setup/versioncheck.go
jakobmoellerdev: added

cli/cmd/setup/versioncheck.go
jakobmoellerdev: done

cli/internal/versioncheck/check.go
jakobmoellerdev: unfortunately i cant pull in this filter on server side via query string via the rest api, maybe I could with the graphql api but didnt check too deeply yet

cli/internal/versioncheck/check.go
jakobmoellerdev: changed it so that a fetch failure also accounts for a check grace period

cli/internal/versioncheck/check.go
jakobmoellerdev: added debug handling

cli/internal/versioncheck/check.go
jakobmoellerdev: no filter in gql as well :/

cli/internal/versioncheck/config.go
```diff
@@ -0,0 +1,82 @@
...
+
+func init() {
+	configScheme.MustRegisterWithAlias(&Config{}, runtime.NewVersionedType(ConfigType, ConfigVersion))
+}
+
+// Config configures the default policy for version checking.
+// The policy field sets the baseline behavior which can be overridden by the
+// --version-check CLI flag. This allows administrators to disable version checks
+// for an entire environment via config, while individual users can still re-enable
+// them via --version-check=auto on the command line.
```
frewilhelm: I believe the `--version-check` CLI flag was removed, right?

jakobmoellerdev: good callout

jakobmoellerdev: removed

cli/cmd/setup/hooks/pre_run.go
```diff
@@ -51,5 +52,7 @@ func PreRunEWithConfig(cmd *cobra.Command, cfg Config) error {
 		cmd.SetErr(parent.ErrOrStderr())
 	}
 
+	setup.VersionCheck(cmd, version.BuildVersion)
```
piotrjanik, line 55: I don't believe this hook functions with `--help`. 
At least, I wasn't able to receive a version notification when I used `--help`.

frewilhelm (approved):

cli/cmd/setup/hooks/pre_run.go
jakobmoellerdev: do we really want the hook to function with help?

matthiasbruns (approved):

PR #2495, opened 2026-05-12
feat(oci): add ResourceStream for tar-free OCI content streaming

### LINKED ISSUE #2485: Direct OCI-to-OCI streaming transfer transformation (TransferOCIArtifact/v1alpha1)

jakobmoellerdev:
Adds a streaming abstraction for OCI resource transfers that avoids creating tar files. Instead of downloading all blobs into a tar archive before uploading, this returns a lazy handle that streams blobs one at a time directly to the destination.

Design documented in ADR 0022: #2489

Why? Transferring large OCI artifacts (30+ GiB AI model images) currently causes OOM kills because the entire artifact is buffered as a tar file in `/tmp` before upload begins. With streaming, only one blob is in flight at a time and existing blobs at the destination are skipped entirely.

```go
stream, _ := repo.DownloadResourceStream(ctx, res)
// Nothing downloaded yet — just a handle

oras.CopyGraph(ctx, stream, dstStore, stream.Root(), opts)
// Streams blobs one by one, skips what destination already has
```

The old `download()` method now delegates to `downloadStream().Materialize()`, meaning that all codepaths still go down to the same code for streaming behind the scenes. streaming just omits the autoserializing to tar.

What's included

- `oci/stream` package: `ResourceStream` interface + `OCIResourceStream` implementation
- `Repository.DownloadResourceStream` / `UploadResourceStream` methods
- `resource.ResourceRepository` credential-aware wrapper
- `TransferOCIArtifact` spec type + transformer for the transfer graph
- 5 unit tests covering Fetch, Exists, Root, Materialize, and CopyGraph
- Refactored `download()` to use streaming internally

Follow-up: Transfer module graph builder wiring lives on [`feat/streaming-resource-transfer`](https://github.com/jakobmoellerdev/open-component-model/tree/feat/streaming-resource-transfer) and will be a separate PR after this lands.

## Test plan

- [x] `go test ./stream/...` passes
- [x] `golangci-lint` clean
- [x] `go build ./...` passes (pre-existing `credentials.go` issue unrelated)
- [ ] Integration test with real registry (follow-up with transfer wiring)

Resolves #2485


jakobmoellerdev:
Im actually not sure yet on the contract of the transformer. I think it would be sufficient to have a source resource and then a target image

matthiasbruns:
> Im actually not sure yet on the contract of the transformer. I think it would be sufficient to have a source resource and then a target image

you mean instead of a full target resource?

matthiasbruns:
> > Im actually not sure yet on the contract of the transformer. I think it would be sufficient to have a source resource and then a target image
> 
> you mean instead of a full target resource?

are you going to try that out? otherwise, I'd want to try that one out and probably approve it :) 
All of our `oci` transfers currently follow the same pattern, we could, if this works, probably change the original `AddOCI` as well, if desired, right?

bindings/go/oci/stream/stream.go
```diff
@@ -0,0 +1,41 @@
...
+
+// ResourceStream is a lazy handle to OCI content.
+// It implements content.ReadOnlyStorage so it can be passed
+// directly to oras.CopyGraph or consumed blob-by-blob.
+// No data is fetched until Fetch or Materialize is called.
+type ResourceStream interface {
+	content.ReadOnlyStorage
+
+	// Root returns the top-level descriptor (manifest or index).
+	Root() ocispec.Descriptor
```
matthiasbruns, line 22: Could we in theory support this streaming interface for any resource?
So having a generic interface in some central place later on that we can opt-in to support streaming for any kind of resource

jakobmoellerdev: I do not think so. A streaming interface requires knowledge of the underlying data structure to stay efficient, otherwise you would have to serialize it. Actually this is what we already have. 

Note that we already support "streaming" in a traditional sense, since our ReadOnlyBlobs are unbuffered datastreams, but most implementations have to buffer to read them (i.e. tars have to be buffered for their filesystem to be readable).

I am viewing this as a more specialised interface that allows access to streams per blob and descriptor and not just a huge blob for the state of the world

jakobmoellerdev:
What is there to try out? I linked a second branch where I have the full integration running with the transformers I believe, is that what you mean?

bindings/go/oci/stream/oci.go
```diff
@@ -0,0 +1,56 @@
...
+	"oras.land/oras-go/v2/content"
+
+	"ocm.software/open-component-model/bindings/go/blob"
+	"ocm.software/open-component-model/bindings/go/oci/tar"
+)
+
+// OCIResourceStream wraps a content.ReadOnlyStorage (typically a remote.Repository)
+// and a resolved root descriptor. No network I/O occurs at construction time.
+type OCIResourceStream struct {
+	store    content.ReadOnlyStorage
```
fabianburth: nit: since we expose the whole API surface of `content.ReadOnlyStorage` anyway, we could embed it and omit the method re-implementations

bindings/go/oci/stream/oci.go
```diff
@@ -0,0 +1,56 @@
...
+// New creates a ResourceStream from an oras ReadOnlyStorage and a resolved root descriptor.
+func New(store content.ReadOnlyStorage, root ocispec.Descriptor, copyOpts oras.CopyGraphOptions, tempDir string, tags []string) *OCIResourceStream {
+	return &OCIResourceStream{
+		store:    store,
+		root:     root,
+		copyOpts: copyOpts,
+		tempDir:  tempDir,
+		tags:     tags,
+	}
+}
```
fabianburth: nit: I don't really know what I would want or need to pass as `tags`, a short comment might be useful

bindings/go/oci/transformer/transfer_oci_artifact.go
```diff
@@ -0,0 +1,89 @@
...
+)
+
+// TransferOCIArtifact is a fused transformer that streams OCI artifacts directly
+// between registries using ResourceStream, bypassing tar materialization.
+// Falls back to DownloadResource + UploadResource if streaming is not supported.
+type TransferOCIArtifact struct {
+	Scheme             *runtime.Scheme
+	Repository         repository.ResourceRepository
+	CredentialProvider credentials.Resolver
+}
```
fabianburth, line 22: I did not look at your other PR yet. But intuitively, I feel like we might just want to use `streaming.Repository` and omit the fallback as a code path.
If the fallback is serialised in the transformation graph, it is much more obvious that we did in fact fallback.

bindings/go/oci/stream/stream.go
```diff
@@ -0,0 +1,41 @@
...
+
+	// Materialize produces a ReadOnlyBlob (OCI layout tar) for legacy consumers.
+	// This is the only code path that creates a tar file.
+	Materialize(ctx context.Context) (blob.ReadOnlyBlob, error)
+}
+
+// Repository extends the generic ResourceRepository with
+// OCI-native streaming. Only implemented by OCI-backed repositories.
+type Repository interface {
```
fabianburth:
```suggestion
// Repository extends the generic ResourceRepository with
// OCI-native streaming. Only implemented by OCI-backed repositories.
type ResourceRepository interface {
```

nit: for consistency, we might still want to call this a `ResourceRepository`

bindings/go/oci/transformer/transfer_oci_artifact.go
fabianburth: Should have checked the implementation first (: In that case, the comment is just a bit misleading

bindings/go/oci/repository/resource/resource_repository.go
```diff
@@ -269,3 +270,73 @@ func createRepository(
 	repo, err := oci.NewRepository(options...)
 	return repo, err
 }
+
+var _ ocistream.Repository = (*ResourceRepository)(nil)
+
+func (p *ResourceRepository) DownloadResourceStream(ctx context.Context, resource *descriptor.Resource, credentials map[string]string) (ocistream.ResourceStream, error) {
```
fabianburth: nit: we might want to deduplicate / share code between `DownloadResourceStream`, `UploadResourceStream`, `DownloadResource`, and `UploadResource`

bindings/go/oci/repository.go
```diff
@@ -747,63 +748,13 @@ func (repo *Repository) DownloadSource(ctx context.Context, src *descriptor.Sour
 }
 
 // download downloads an artifact specified by an access from the repository into a blob.ReadOnlyBlob.
-// It is expected that the access is
-//   - a valid [accessv1.OCIImage], or
-//   - a [v2.LocalBlob] access that has a [v2.LocalBlob.GlobalAccess] set that can be interpreted as [accessv1.OCIImage].
+// It delegates to downloadStream and materializes the result into a tar-based OCI layout blob.
 func (repo *Repository) download(ctx context.Context, access runtime.Typed) (data blob.ReadOnlyBlob, err error) {
-	typed, err := repo.scheme.NewObject(access.GetType())
+	stream, err := repo.downloadStream(ctx, access)
```
fabianburth, line 757: The way we re-use the stream implementation seems a bit convoluted, a lot of code (especially type conversion and assertion) is being called twice now. I did not dive deep now, but on first glance, I'd have expected that `DownloadResource` could call `DownloadResourceStream` and `Materialize`.

bindings/go/oci/repository.go
```diff
@@ -923,3 +874,97 @@ func (repo *Repository) AddComponentVersionAlias(ctx context.Context, component,
...
+		return nil, fmt.Errorf("failed to stream resource via copy: %w", err)
+	}
+
+	ref, err := looseref.ParseReference(access.ImageReference)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse target access image reference %q: %w", access.ImageReference, err)
+	}
+	if err := ref.ValidateReferenceAsTag(); err != nil {
+		return nil, fmt.Errorf("can only copy %q if it is tagged: %w", access.ImageReference, err)
+	}
```
fabianburth: Can we really just copy if it is tagged?

bindings/go/oci/repository.go
```diff
@@ -923,3 +874,97 @@ func (repo *Repository) AddComponentVersionAlias(ctx context.Context, component,
...
+
+	if err := store.Tag(ctx, rs.Root(), ref.Tag); err != nil {
+		return nil, fmt.Errorf("failed to tag artifact with tag %q: %w", ref.Tag, err)
+	}
+
+	res = res.DeepCopy()
+	access.ImageReference = ref.String()
+	res.Access = &access
```
fabianburth, line 987: In the "normal" upload, we also (re-)populate the resource digest every time. Are we sure it is okay to skip this here? What if the digest of the original resource is empty (or isn't this a valid case?)

bindings/go/oci/transformer/transfer_oci_artifact.go
jakobmoellerdev: how is it misleading?

bindings/go/oci/repository.go
jakobmoellerdev: i have tried to join even more of the logic together

bindings/go/oci/repository.go
jakobmoellerdev: I do think that I now also populate the digest, please check again!

bindings/go/oci/repository.go
jakobmoellerdev: adjusted

bindings/go/oci/repository/resource/resource_repository.go
jakobmoellerdev: Tried to unify after your feedback

bindings/go/oci/stream/stream.go
jakobmoellerdev: agreed, modified

bindings/go/oci/stream/oci.go
jakobmoellerdev: agree, changed

bindings/go/oci/transformer/transfer_oci_artifact.go
fabianburth: Both me and my AI independently expected the fallback behaviour to be implemented by the transformer itself after reading the comment. But if I understand correctly, the fallback behaviour would be facilitated through the TGD generation?

jakobmoellerdev: understood, thanks. will remove that sentence

bindings/go/oci/repository/resource/resource_repository.go
fabianburth: perfection

bindings/go/oci/repository.go
fabianburth: ah, yes you do! 👌

fabianburth (approved):

matthiasbruns (approved):
lgtm

PR #2502, opened 2026-05-12
feat(website): versioned website

### LINKED ISSUE #1053: versioned releases for ocm website
**Description**

This is a follow up to [the floating main version for the ocm website](https://github.com/open-component-model/ocm-project/issues/1013).

The website currently serves two versions of the documentation:
- `main` — content in `website/content/`, including authored Markdown and committed generated reference (e.g. JSON schemas from Go types)
- `latest` — generated content from the most recent release tag, imported via Hugo module BUT the manually written content is the same as in main which makes it still impossible to prepare guides without immediately publishing them on the website.

Users on older, still-supported OCM releases have no documentation matching their version. We want to support multiple pinned release versions alongside `main` and `latest`, stored as snapshots in this monorepo.

CLI, controller, and website release in lockstep, so each release produces a single doc version — no per-component selectors needed.

**Goals**

- Store all served doc versions (`main`, `latest`, pinned releases) as content within `website/`.
- Eliminate the Hugo module import for `latest`, replacing it with a snapshot.
- Snapshot pinned versions as part of the lockstep release process by copying the docs tree at the tagged commit into a versioned path.
- Define a support policy for which versions stay published, and how they are retired.

**Layout/Shortcode Compatibility**

Layouts under `website/` are shared across all stored versions, so a single set of layouts must render every snapshot correctly. Rather than vendoring layouts per snapshot (which freezes bugs and blocks site-wide improvements), treat `website/layouts/shortcodes/` as a public API:

- Parameters may be added (with defaults), never removed or renamed.
- Shortcodes may be deprecated but not deleted while any live snapshot still references them.
- For genuinely breaking changes, introduce a new shortcode name (`foo-v2`) alongside the old one and migrate `main` content; older snapshots continue using the old version until retired.
- A CI check scans all snapshot directories for shortcode usage and fails if a PR removes or renames a still-referenced shortcode.
- Document each shortcode's parameters and behavior in `website/layouts/shortcodes/SHORTCODES.md`.

**`latest` handling**

`/docs/latest/*` is served as a Netlify rewrite to the newest pinned snapshot's path (e.g. `/docs/v2.1/:splat`), not a copy. This keeps a single source of content per version, so patches to the current release only need to be made in one place. The rewrite target is updated as part of the release process. Layouts emit a canonical link tag to avoid duplicate-content issues in search.

**SEO**

Multiple versions create duplicate-content URLs that need explicit handling:
- Pinned versions emit a `<link rel="canonical">` pointing to the corresponding page under `/docs/latest/`. This consolidates search ranking on the latest version while keeping older versions indexable for users who explicitly need them.
- The `main` (dev) version emits `<meta name="robots" content="noindex">` to keep unstable docs out of search results entirely.
- Implemented in the Hugo base template based on the page's path prefix.

A stricter alternative would be to `noindex` all non-latest pinned versions as well, removing them from search entirely. This avoids users landing on stale docs via search but forces them to use the version selector to find docs matching their installed version. We default to canonical-only because users on supported older releases have a legitimate need to find their docs via search.

**Open Questions / Decisions Needed**

- **Snapshot location**: `website/content/docs/<version>/` (Hugo-native, simplest) vs. a sibling tree mounted via Hugo config.
- **Snapshot trigger**: step in the existing release workflow vs. a separate workflow on release tag. Should integrate with `RELEASE_PROCESS.md` either way.
- **Repo size**: each pinned version adds the full doc tree. Estimate growth over ~2 years of releases.
- **Retirement**: delete out-of-support snapshot directories or keep them with a deprecation banner.

**Done Criteria**

- [ ] Snapshot location and naming convention defined
- [ ] Snapshot mechanism implemented and integrated with `RELEASE_PROCESS.md`
- [ ] `latest` migrated from module import to Netlify rewrite of newest pinned snapshot
- [ ] At least one pinned version snapshotted and reachable at its own path
- [ ] Hugo module imports for documentation content removed
- [ ] Version selector lists `main`, `latest`, and pinned versions
- [ ] Version retirement policy documented
- [ ] Shortcode contract documented in `SHORTCODES.md` and enforced via CI check
- [ ] Canonical and `noindex` tags emitted correctly per version
- [ ] RSS feed and sitemap behave correctly across versions

fabianburth:
A versioned website setup enables:
- Merging docs on main without showing them in the default documentation (so, the documentation for the latest release). This is important, so we can add the documentation alongside the actual features without confusing users reading the website documentation.
- Providing users still using older versions access to accurate documentation for that particular version.

This website setup:
- Creates tags for the website (which is a go module) to leverage hugo's go module import system to set up documentation for each version.
- The website tag is created on release (together with `cli` and `controller`) although we DO NOT create `rc` for the website. Instead, the website tag is created AFTER the `cli` and `controller` `rc`'s are promoted and their actual release tags (vX.Y.Z) exist.
- After the creation of all the tags, a release step opens a PR against main to add the corresponding hugo module imports to the main branches website hugo configs.

So, in conclusion, it adds a manual step (merging the created PR) to the release process.

fix https://github.com/open-component-model/ocm-project/issues/1053

##### Verification


website/scripts/register-docs-version.js
```diff
@@ -0,0 +1,414 @@
...
+// Serialize a parsed config back to YAML
+function dumpYaml(parsed) {
+    return yaml.dump(parsed, { lineWidth: -1, noRefs: true });
+}
+
+// Maximum number of minor versions (excluding special versions like main/legacy)
+const MAX_MINOR_VERSIONS = 10;
+
+// Compare two SemVer strings (X.Y or X.Y.Z). Returns <0 if a<b, >0 if a>b, 0 if equal.
+function compareSemver(a, b) {
```
Skarlso, line 49: I'm pretty sure we have _several_ functions by now doing this comparison :D

.github/workflows/release.yml
```diff
@@ -608,3 +611,74 @@ jobs:
...
   # === controller-release.yml::promote_and_release_final + cli-release.yml::promote_final + cli-release.yml::release_final ===
+
+  # === Website version release ===
+  # Replaces the old standalone website-version-release.yml. Runs after the
+  # final promotion succeeded, when the website/v0.X.Y tag exists. Opens a PR
+  # to main pinning Hugo module imports to the just-created website tag, with
+  # binding versions resolved from the released CLI module's go.mod.
+  create_website_update_pr:
+    name: Open website update PR
+    needs: [prepare, promote_and_release_final]
```
Skarlso, line 622: `promote_and_release_final` already has prepare. I don't think you need prepare here again?

.github/workflows/release.yml
```diff
@@ -608,3 +611,74 @@ jobs:
...
+        run: |
+          set -euo pipefail
+          git fetch origin "refs/tags/cli/v${VERSION}:refs/tags/cli/v${VERSION}" --no-tags
+          git show "cli/v${VERSION}:cli/go.mod" > "${{ runner.temp }}/cli-go.mod"
+
+      - name: Release docs version
+        working-directory: website
+        env:
+          VERSION: ${{ needs.prepare.outputs.base_version }}
+        run: node scripts/register-docs-version.js "$VERSION" --cli-gomod "${{ runner.temp }}/cli-go.mod"
```
Skarlso, line 664: why aren't you using `actions/github-script`? 🤔 you wouldn't need the npm ci thing in the begin I believe.

Skarlso, line 664: This thing calls `go` without `setup-go` action. It happened to work because the runner has some kind of go but it's not going to work if it isn't compatible with what we have.

website/README.md
```diff
@@ -141,9 +141,9 @@ npm run cutoff -- 0.0.2
 What the script does (brief):
 
 - Copies `content/` to `content_versioned/version-x.y.z`.
-- Appends a `[versions."x.y.z"]` stanza (without weight) to `config/_default/hugo.toml` and optionally keeps
```
Skarlso, line 144: I think you can remove the things about the cut off being mentioned, right? Since you removed that.

agents.md
```diff
@@ -146,9 +146,9 @@ Hugo-based documentation site published at <https://ocm.software>. Read `website
 - **Frontmatter**: Every page needs `title`, `description`, optional `logo`, and `weight` (lower weight ranks higher).
 - **Internal links**: Always use `{{< relref "filename.md" >}}`. Use the bare filename when unique across `content/`, otherwise the full path relative to `content/`.
 - **CLI reference** (`content/docs/reference/`) is mounted via Hugo modules from source repos; do not edit those files in this repo.
-- **Versioning**: `npm run cutoff -- x.y.z` snapshots `content/` into `content_versioned/version-x.y.z` and updates `config/_default/hugo.toml` and `module.toml`. Add `--keepDefault` to keep the current `defaultContentVersion`. Never edit the version configs by hand.
```
Skarlso, line 149: There are more mentions of `cutoff`. :) 

```md
8. **Editing versioned website content** — Files under `website/content_versioned/` are snapshots. Edit `website/content/` and cut a new version via `npm run cutoff`.
9. **Hand-editing Hugo version configs** — `config/_default/hugo.yaml` and `module.yaml` version stanzas are managed by the cutoff script.
```

RELEASE_PROCESS.md
```diff
@@ -41,7 +42,16 @@ gitGraph TB:
     commit tag: "v0.8.0-rc.1" type: REVERSE
 ```
 
-Two things to notice. First, **promotion does not create a new commit.** The canonical final tag (`v0.7.0`) lands on the same commit as the most recent RC tag (`v0.7.0-rc.2`). The workflow just stamps additional tags on the RC commit. Second, **each tagged release commit carries three tags, not one.** The diagram shows only the canonical tag for legibility; the same commit also receives `cli/v0.7.0-rc.2` + `cli/v0.7.0` and `kubernetes/controller/v0.7.0-rc.2` + `kubernetes/controller/v0.7.0`. The side tags exist so the website install script and Go module consumers can address each component directly.
+Two things to notice. First, **promotion does not create a new commit.** The
```
Skarlso, line 45: Should also add a warning that if we forget to merge this BEFORE the next release there are going to be problems. :D

RELEASE_PROCESS.md
```diff
@@ -41,7 +42,16 @@ gitGraph TB:
     commit tag: "v0.8.0-rc.1" type: REVERSE
 ```
 
-Two things to notice. First, **promotion does not create a new commit.** The canonical final tag (`v0.7.0`) lands on the same commit as the most recent RC tag (`v0.7.0-rc.2`). The workflow just stamps additional tags on the RC commit. Second, **each tagged release commit carries three tags, not one.** The diagram shows only the canonical tag for legibility; the same commit also receives `cli/v0.7.0-rc.2` + `cli/v0.7.0` and `kubernetes/controller/v0.7.0-rc.2` + `kubernetes/controller/v0.7.0`. The side tags exist so the website install script and Go module consumers can address each component directly.
+Two things to notice. First, **promotion does not create a new commit.** The 
+canonical final tag (`v0.7.0`) lands on the same commit as the most recent 
+RC tag (`v0.7.0-rc.2`). The workflow just stamps additional tags on the RC 
+commit. Second, **each tagged release commit carries multiple tags, not one.
+** The same commit also receives `cli/v0.7.0-rc.2` + `cli/v0.7.0`,
```
Skarlso: this seems.. off.. the `**` is on the next line :D

Skarlso (commented):
Couple of things, honestly, great work. This looks fantastic!

.github/workflows/release.yml
fabianburth:
```
      - name: Fetch CLI go.mod at release tag
        env:
          VERSION: ${{ needs.prepare.outputs.base_version }}
        run: |
          set -euo pipefail
          git fetch origin "refs/tags/cli/v${VERSION}:refs/tags/cli/v${VERSION}" --no-tags
          git show "cli/v${VERSION}:cli/go.mod" > "${{ runner.temp }}/cli-go.mod"
```
Since this step "needs" the `base_version` from prepare, I think I do need it.

.github/workflows/release.yml
fabianburth: The script uses `js-yaml` to serialise the hugo config files. Still good point, I tried to slim it down!

fabianburth: Since we just do a `go mod edit`, I think the risk would actually be neglectable. But honestly, I just did not notice, and adding it is definitely more idiomatic! Good catch!

RELEASE_PROCESS.md
fabianburth: Added a warning as you should definitely not forget it (for other reasons). I believe there wouldn't really be any problems besides merge conflicts.

.github/workflows/release.yml
fabianburth: actually thought about it a bit and went back to `npm ci` because it provides more resilient version management through the package-lock.json

website/scripts/register-docs-version.js
fabianburth: yeah, at some point it might be worth unifying them! do you want me to take care of it now? you wanted to do some further clean ups on the whole release related workflows anyway, maybe we should rather do it there

website/README.md
fabianburth: I hope I caught most of it now!

RELEASE_PROCESS.md
fabianburth: Actually, removed the warning again. I'm not sure what should happen, so this would just give a false sense of danger.

.github/workflows/release.yml
fabianburth: Actually, if it bothers us and we don't mind reading json, we could also switch to json instead of yaml. That'd be in the `actions/github-script` environment by default.

website/scripts/register-docs-version.js
Skarlso: No, let's not touch it. :)

.github/workflows/release.yml
Skarlso: Ah, yes. Dang. I missed that. In my refactor ( the optimization ) I'm pulling together certain actions so we don't cross-share this much. :) But that's later.

.github/workflows/release.yml
Skarlso: Ugh. :D Let's not do that in here :D But I'll push this as a note to the optimization issue, WDYT? https://github.com/open-component-model/ocm-project/issues/1076

Skarlso: Ah, that's true as well. :D thanks for fixing it anyways. :)

RELEASE_PROCESS.md
Skarlso: If we don't merge that PR wouldn't that mean that the website isn't working right since it wouldn't point to the right version? 🤔

Skarlso (dismissed):
LET'S GOOOOO.

RELEASE_PROCESS.md
```diff
@@ -69,20 +79,35 @@ flowchart TD
...
 ```
 
 Phase 1 (RC, blue) runs end-to-end without human intervention once you trigger the workflow with `dry_run=false`. The run then parks at the `release` environment gate. Approving it kicks off Phase 2 (final, green): attestation verification, tag promotion, chart repack with byte-equivalence check against the RC, and the final GitHub release. Anything failing in Phase 1 aborts before any user-facing tags or releases are visible; anything failing in Phase 2 leaves the RC intact so you can re-approve after fixing the underlying issue.
 
+### Website
+
+The website integrates into the same workflow run, after `promote_and_release_final` succeeds:
+
+* The `website/v0.X.Y` tag is created at the same commit as the canonical tag — same `ADDITIONAL_TAGS` step that emits `cli/` and `kubernetes/controller/` tags.
+* A separate `create_website_update_pr` job then opens a PR to `main` updating `website/config/_default/{hugo.yaml,module.yaml}` to pin the new minor's Hugo module imports to the just-created `website/v0.X.Y` tag. The PR uses the OCMBot app token, signed commits, and `add-paths: website/config/`.
```
frewilhelm, line 102: Can we rerun the job to recreate the PR?

Because IIRC we usually have problems with such generated PRs if they need rebasing for a linear commit history. Apparently, we cannot rebase with a signed commit, so we just recreate it (see e.g. https://github.com/open-component-model/open-component-model/pull/2566)

fabianburth: Currently, you cannot individually. I did not extract it into its own workflow for simplicity. So I guess I'd need to check how idempotent the rest of the release workflow is.

matthiasbruns (dismissed):

RELEASE_PROCESS.md
fabianburth: never mind, you can re-run it, although it will create a new PR because I currently add a timestamp to the branch suffix - let me change that
(see https://github.com/fabianburth/open-component-model/pulls)

RELEASE_PROCESS.md
fabianburth: Sure, the version would be missing. We should not forget to merge it. But that's the whole danger, you should be able to merge it whenever you want (even if you don't and the next release comes along, the missing version would be the whole danger).

RELEASE_PROCESS.md
fabianburth: alright, adjusted the branch suffix to work w/ the same branch on re-runs (didn't test w/ this now, hope it will be fine!)

Skarlso (approved):

jakobmoellerdev (approved):

PR #2504, opened 2026-05-12
feat(config): add oci.config.ocm.software with ownershipReferrerPolicy

### LINKED ISSUE #1043: Ownership Referrer Toggle in: ownership.config.ocm.software/v1alpha1
**Description**

Expose the existing `OwnershipReferrerPolicy` toggle through the public OCM config system so users can opt in to ownership-referrer creation via a config file or CLI flag.

**Implementation notes**

- Sample YAML:
```yaml
configurations:
  - type: ownership.config.ocm.software/v1alpha1
    policy: AddIfSupported
    repositories:
      - repository:
          type: OCIRepository/v1
        policy: Never
      - repository:
          type: OCIRepository/v1
          baseUrl: ghcr.io
          subPath: my-org/components
        policy: AddIfSupported
      - repository:
          type: S3/v1
          baseUrl: s3.aws
        policy: Never
```

**Done Criteria**
- [ ] ...

piotrjanik:

Lets users opt in to the asset-to-owner OCI referrer from `.ocmconfig` — the existing `ocm.config.ocm.software/v1` now accepts `ownershipReferrerPolicy: auto|disabled`, off by default.

```yaml
type: generic.config.ocm.software
configurations:
  - type: ocm.config.ocm.software/v1
    ownershipReferrerPolicy: auto
```

This is the `bindings/go/configuration`part.

Fixes: https://github.com/open-component-model/ocm-project/issues/1043


bindings/go/configuration/ocm/v1/spec/config.go
```diff
@@ -107,10 +128,6 @@ type Resolver struct {
 }
 
 // Lookup creates a new Config from a central V1 config.
-//
-// Deprecated: Resolvers are deprecated and are only added for backwards
```
jakobmoellerdev, line 111: ehm..... why was this removed?

matthiasbruns:
looks good but what happened to the comment?

bindings/go/configuration/ocm/v1/spec/config.go
piotrjanik: Oops, wrong line. I wanted to remove the deprecated comment above Config, as it gives the impression that Config is deprecated. I moved that comment above Referrers within Config.

matthiasbruns (approved):

bindings/go/configuration/ocm/v1/spec/config.go
```diff
@@ -41,10 +54,6 @@ func init() {
 //	    filePath: ./ocm/primary-transport-archive
 //	    priority: 10
 //
-// Deprecated: Resolvers are deprecated and are only added for backwards
```
jakobmoellerdev, line 44: i think youre missing the addition of this in the new resolvers no? And i do believe the old resolvers are deprecated so this needs a replacement right?

piotrjanik: I moved that comment below. What am I missing for new resolvers?

jakobmoellerdev: we have a new resolver concept in place that is well documented on the website as well! does that help? https://ocm.software/docs/how-to/migrate-from-fallback-to-deterministic-repository-resolvers/

jakobmoellerdev: Im assuming from quick glance on the code that you just added the policy to one of the resolvers, not both, maybe im wrong?

piotrjanik: Actually, my intention was to add this field to the`ocm.config.ocm.software/v1` config like this:

```yaml
configurations:
  - type: ocm.config.ocm.software/v1
    ownershipReferrerPolicy: auto
    resolvers: [] # deprecated
```

So ownershipReferrerPolicy is not set under resolvers.

piotrjanik: I think the problem is that I used `ocm.config.ocm.software/v1` config type which is deprecated as it includes deprecated `resolvers`.

With my change I remove deprecation comment from config type, but not from `resolvers`.

Might be we need another place to add this toggle, if `ocm.config.ocm.software/v1` must remain deprecated.

I can create nw config type like: `referrers.ocm.config.ocm.software/v1`

jakobmoellerdev: I dont think this is correct. ocm config is only containing resolvers right now and should stay deprecated. you need to configure referrer policy by repository I believe, not globally

piotrjanik: @jakobmoellerdev  Something like that?
```yaml
configurations:
  - type: resolvers.config.ocm.software/v1alpha1
    resolvers:
      - repository:
          type: OCIRepository/v1
          baseUrl: ghcr.io
          subPath: my-org/components
        componentNamePattern: "example.com/services/*"
        versionConstraint: ">=1.0.0 <2.0.0"
        ownershipReferrerPolicy: auto # HERE
```

jakobmoellerdev: I believe so. please check the configs with @matthiasbruns and make sure it works for both resolvers. I also hope you chatted with @fabianburth if this even all makes sense and is as we want :D

matthiasbruns: @piotrjanik let's discuss this in the wr today

matthiasbruns: I would not want to add this to `resolvers`, BUT we should think about a repo/endpoint-based policy config thingy that might work like resolvers..
1. resolvers only work on componentReferences
2. resolvers are not the place for this tbh
We would have to add a similar config logic for sure, but mixing those would be bad.
The only question is, how would a repo get these new informations?

piotrjanik:
```yaml
configurations:
  - type: referrers.config.ocm.software/v1alpha1
    ownershipReferrerPolicy: disabled # by default unless it is...
    repositories:
      - repository:
          type: OCIRepository/v1
          baseUrl: ghcr.io
          subPath: my-org/components
        ownershipReferrerPolicy: auto # ...enabled for particular repo
```


@jakobmoellerdev @fabianburth ? what do you think ?

PR #2452, opened 2026-05-06
feat!: add typed methods to credential plugin interfaces

matthiasbruns:
#### What this PR does / why we need it
Phase 3 of the credential migration - this is a breaking change since implementors must adapt.

This PR adds typed support to both CredentialPlugin and RepositoryPlugin as defined in the [ADR](https://github.com/open-component-model/open-component-model/blob/main/docs/adr/0018_typed_credentials.md#phase-3-plugin-interfaces)

#### Which issue(s) this PR fixes
Contributes: https://github.com/open-component-model/ocm-project/issues/1047

The full impact can be seen in https://github.com/open-component-model/open-component-model/pull/2508

##### How to test the changes

Added unit tests for the new method in `credentials`.
With `go.work` enabled, you will see that this breaks consumers -> `ocm.software/open-component-model/bindings/go/plugin`

##### Verification

- [x] I have added/updated tests for my changes (see [Test Requirements](../CONTRIBUTING.md#test-requirements))
- [x] Tests pass locally (`task test` and `task test/integration` if applicable)
- [x] My changes do not decrease test coverage

PR #2508, opened 2026-05-13
feat!: resolve typed plugins bindings impl

matthiasbruns:
Phase 3 of the credential migration - this is a breaking change since implementors must adapt.

This PR adds typed support to both CredentialPlugin and RepositoryPlugin as defined in the [ADR](https://github.com/open-component-model/open-component-model/blob/main/docs/adr/0018_typed_credentials.md#phase-3-plugin-interfaces)

This PR implements the new methods in all bindings and makes it work - this will probably needs to be split up.
Follow up of https://github.com/open-component-model/open-component-model/pull/2452

Contributes: https://github.com/open-component-model/ocm-project/issues/1047


matthiasbruns:
Closed in favor of https://github.com/open-component-model/open-component-model/pull/2519

PR #2513, opened 2026-05-13
feat(oidcflow): add OIDC token caching to avoid repeated browser prompts

morri-son:
## Files changed (this PR only)

builds on #2346, so has many files that will disappear after rebase once #2346 was merged.

- `cli/internal/oidcflow/cache.go` — token cache persist/load/refresh logic
- `cli/internal/oidcflow/cache_test.go` — unit tests for cache layer
- `cli/internal/oidcflow/oidcflow.go` — integrate cache into GetIDToken flow

## Summary

- Cache OIDC refresh token and id_token to `~/.ocm/cache/oidc/` after the first interactive browser flow
- On subsequent sign operations, silently refresh the token instead of opening the browser again
- Handle providers like Dex that omit `id_token` in refresh responses by falling back to the cached one

## Depends on

- #2346 (introduces the OIDC flow this builds upon)

## How it works

1. First `ocm sign cv` → browser opens, user authenticates, tokens cached
2. Subsequent signs within the id_token validity window (~10 min) → silent, no browser
3. After id_token expires → graceful fallback to interactive flow

## Test plan

- [x] Unit tests for cache persist/load/refresh/remove (`cache_test.go`)
- [x] E2e test signing 3 components in sequence — browser opens only once
- [x] `go test ./internal/oidcflow/` passes


cli/internal/oidcflow/cache.go
```diff
@@ -0,0 +1,141 @@
...
+)
+
+type cachedToken struct {
+	AccessToken  string `json:"access_token"`
+	RefreshToken string `json:"refresh_token"`
+	TokenType    string `json:"token_type"`
+	IDToken      string `json:"id_token"`
+}
+
+var userHomeDir = os.UserHomeDir
```
jakobmoellerdev, line 24: shouldnt a cache use user cache dir? also shouldnt the lookup of the dir be done only once and not on every call?

jakobmoellerdev:
while the feature is appreciated this probably deserves some love and rework after we revisited the credential system

PR #2518, opened 2026-05-15
feat!: credentials phase 3 - credential bindings

matthiasbruns:
Phase 3 of the credential migration

This PR updated the `credential` bindings.

This PR adds typed support to both CredentialPlugin and RepositoryPlugin as defined in the [ADR](https://github.com/open-component-model/open-component-model/blob/main/docs/adr/0018_typed_credentials.md#phase-3-plugin-interfaces)

Additionally, we also replace the deprecated `Resolve` methods from the interfaces, as well as from the credentials.

You can see the full impact in https://github.com/open-component-model/open-component-model/pull/2517

Contributes: https://github.com/open-component-model/ocm-project/issues/1047


bindings/go/credentials/resolve_direct.go
```diff
@@ -50,70 +43,17 @@ func (g *Graph) resolveFromGraph(ctx context.Context, identity runtime.Identity)
...
-
-		// Let the plugin resolve the child's credentials.
-		credentials, err := plugin.Resolve(ctx, identity, childMap)
+		credentials, err := plugin.ResolveTyped(ctx, identity, childCredentials)
 		if err != nil {
 			return nil, fmt.Errorf("no credentials for node %q resolved from plugin: %w", edgeID, err)
 		}
 
-		// Merge the resolved credentials into the result
-		maps.Copy(result, credentials)
```
matthiasbruns, line 63: fixing this one

matthiasbruns: fixed :D

bindings/go/credentials/resolve_direct.go
```diff
@@ -26,16 +24,14 @@ func (g *Graph) resolveFromGraph(ctx context.Context, identity runtime.Identity)
 		return nil, err
 	}
 
-	// Leaf node: return the credentials directly.
 	creds, cached := g.getCredentials(vertex.ID)
 	if cached {
 		return creds, nil
 	}
 
 	node := identity.String()
```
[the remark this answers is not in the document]
matthiasbruns: this shouldn't have changed compared to main - can you double check?

bindings/go/credentials/plugins.go
```diff
@@ -20,11 +20,11 @@ type RepositoryPlugin interface {
...
-	// Resolve attempts to resolve credentials for a given repository configuration
-	// and consumer identity. The provided credentials map may contain pre-resolved
-	// credentials from the credential graph.
-	// TODO(matthiasbruns): Migrate to runtime.Typed https://github.com/open-component-model/ocm-project/issues/980
-	Resolve(ctx context.Context, cfg runtime.Typed, identity runtime.Identity, credentials map[string]string) (map[string]string, error)
+	// ResolveTyped resolves credentials for a given repository configuration and consumer
+	// identity, returning a runtime.Typed credential.
+	// The credentials parameter contains any pre-resolved credentials from the credential
+	// graph; it may be nil when no graph credentials are available for this repository.
+	ResolveTyped(ctx context.Context, cfg runtime.Typed, identity runtime.Identity, credentials runtime.Typed) (runtime.Typed, error)
```
jakobmoellerdev: looking at this here, why do we rename the Method signature if we anyhow drop the old resolve? 

Im wondering why you dont leave it on `Resolve` or alternatively name it `ResolveTyped` and keep the signature for resolve?

bindings/go/credentials/resolve_direct.go
```diff
@@ -50,70 +49,75 @@ func (g *Graph) resolveFromGraph(ctx context.Context, identity runtime.Identity)
...
-	}
-	if dc, ok := cred.(*v1.DirectCredentials); ok {
-		return maps.Clone(dc.Properties)
+// mergeTyped combines multiple resolved typed credentials into a single value,
+// preserving the original map-merge semantic (later entries override earlier
+// ones per field).
+//
+// The scheme must know every input's runtime type so scheme.Convert can
+// serialize it to runtime.Raw.
+func mergeTyped(creds []runtime.Typed, scheme *runtime.Scheme) (runtime.Typed, error) {
```
jakobmoellerdev, line 79: I am almost certain that this is "ok" to merge credentials, because it is ordering dependent on creds in the slice right? That said, this should be documented not just in the internal method, but in the behavior documentation for the credential config. 

It is also important for me to note that you should probably do an unstructured conversion here, and not a json conversion. Is there a reason why you did it this way?

bindings/go/credentials/plugins.go
matthiasbruns: The old `Resolve` should be dropped, but I can simply change the credentials param to runtime.Typed and get rid of the `ResolveTyped`. 

The idea was to have everything non-breaking, which did not really work since we moved spec/identity, that's why there was a `ResolveTyped` in the first place. :sadface:

bindings/go/credentials/resolve_direct.go
matthiasbruns: I took over the old implementation strategy there - I did not really question it - I will update to Unstructured :)

bindings/go/credentials/plugins.go
jakobmoellerdev: understood. yeah considering our strategy change this should probably be adapted

matthiasbruns: Renamed to Resolve

bindings/go/credentials/resolve_direct.go
matthiasbruns: ok, the problem is

unstructured allows `map[string]interface{} `but DirectCredentials are `map[string]string`
thats why I used the json marshal way

jakobmoellerdev (dismissed):

jakobmoellerdev (approved):

piotrjanik (approved):

PR #2556, opened 2026-05-18
chore(deps): Renovate/patch golang versions 1.26.3

matthiasbruns:
Followup of https://github.com/open-component-model/open-component-model/pull/2522

Sets Golang to ~v1.26.2~ v.1.26.3


jakobmoellerdev:
a few things here: 1. why exactly did you need to close the pr? 2. I think it makes sense to not have the minimum window for go core because we already have 1.26.3 right?

matthiasbruns:
> a few things here: 1. why exactly did you need to close the pr? 2. I think it makes sense to not have the minimum window for go core because we already have 1.26.3 right?

1. The build failed on the original PR - should I have kept it open and wait until it closes itself after merging this one?
2. Yeah I'll push to 1.26.3

jakobmoellerdev:
I think  the build should not have failed because the grouping of renovate should have included the bumps you did as well. If it didnt do that the grouping rules are probably off.

matthiasbruns:
> I think the build should not have failed because the grouping of renovate should have included the bumps you did as well. If it didnt do that the grouping rules are probably off.

yeah - the docker stuff didnt work https://github.com/open-component-model/open-component-model/actions/runs/26017346081/job/76470456696

jakobmoellerdev (approved):

morri-son (approved):

PR #2573, opened 2026-05-19
fix(release): make release auth explicit

matthiasbruns:

https://github.com/open-component-model/open-component-model/actions/runs/26096370263
Permissions are not set up correctly after updating permissions in all workflows.

https://github.com/open-component-model/open-component-model/actions/runs/26096370263

This aligns CLI release tag creation with the existing controller release pattern.

Controller already uses the OCM bot app token for:
- RC tag creation: https://github.com/open-component-model/open-component-model/blob/main/.github/workflows/controller-release.yml#L53-L98
- Final release tag creation: https://github.com/open-component-model/open-component-model/blob/main/.github/workflows/controller-release.yml#L241-L287

CLI release previously used the default `GITHUB_TOKEN` for the equivalent tag pushes, so this PR makes the two release workflows consistent.


.github/workflows/cli-release.yml
```diff
@@ -94,8 +103,9 @@ jobs:
           TAG: ${{ needs.prepare.outputs.new_tag }}
           # create-tag.js uses GITHUB_TOKEN to authenticate git push because
           # the checkout above sets persist-credentials: false (zizmor requirement).
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GITHUB_TOKEN: ${{ steps.get_token.outputs.token }}
         with:
+          github-token: ${{ steps.get_token.outputs.token }}
```
frewilhelm: Why do we need `GITHUB_TOKEN` and `github-token`?

.github/workflows/cli-release.yml
```diff
@@ -155,6 +166,7 @@ jobs:
           RC_VERSION: ${{ needs.prepare.outputs.new_version }}
         run: |
           gh release create "${RC_TAG}" \
+            --repo "${{ github.repository }}" \
```
frewilhelm, line 170: Do we need the `--repo` flag because we did not check out the repository?

matthiasbruns: correct

.github/workflows/cli-release.yml
matthiasbruns: fixed

frewilhelm (dismissed):

.github/workflows/cli-release.yml
```diff
@@ -94,7 +103,7 @@ jobs:
           TAG: ${{ needs.prepare.outputs.new_tag }}
           # create-tag.js uses GITHUB_TOKEN to authenticate git push because
           # the checkout above sets persist-credentials: false (zizmor requirement).
```
morri-son:
```suggestion
  # Use the OCM bot app token (not GITHUB_TOKEN) because:
  # 1. persist-credentials: false (zizmor) → git push needs explicit token.
  # 2. Pushes via secrets.GITHUB_TOKEN do NOT trigger other workflows;
  #    cli.yml listens on `push: tags: ['cli/v**']` to publish the OCI image.
```

.github/workflows/cli-release.yml
```diff
@@ -285,7 +305,7 @@ jobs:
           NEW_RELEASE_TAG: ${{ needs.prepare.outputs.promotion_tag }}
           # create-tag.js uses GITHUB_TOKEN to authenticate git push because
           # the checkout above sets persist-credentials: false (zizmor requirement).
```
morri-son:
```suggestion
  # Use the OCM bot app token (not GITHUB_TOKEN) because:
  # 1. persist-credentials: false (zizmor) → git push needs explicit token.
  # 2. Pushes via secrets.GITHUB_TOKEN do NOT trigger other workflows;
  #    cli.yml listens on `push: tags: ['cli/v**']` to publish the OCI image.
```

jakobmoellerdev (approved):

frewilhelm (approved):

PR #2519, opened 2026-05-15
feat: 1047 typed credentials - full migration path

matthiasbruns:
#### What this PR does / why we need it
This PR basically migrates the whole code-base to the final state of credentials including cli and controllers.

- updates all interfaces that used credentials map[string]string
- updates plugin boundary auth parsing
- updates all consumers and cross references
- updates cli & controllers
- updates all tests and integration tests
- cleans up deprecated consts and helper functions
- adds a util to convert `runtime.Typed` into typed credentials in the bindings where needed
- introduces typed credentials in sigstore

#### Which issue(s) this PR fixes
Contributes: 
- https://github.com/open-component-model/ocm-project/issues/1047
- https://github.com/open-component-model/ocm-project/issues/1055
- https://github.com/open-component-model/ocm-project/issues/1056
- https://github.com/open-component-model/ocm-project/issues/1057

#### Binding release order 

 Dependency graph (changed → changed only):
  
  blob          →  (none)
  signing       →  (none)
  rsa           →  (none)
  sigstore      →  signing
  repository    →  blob, credentials
  oci           →  blob, credentials, repository
  constructor   →  blob, credentials, oci, repository
  plugin        →  blob, constructor, credentials, repository, signing
  input/dir     →  blob, constructor, credentials, repository
  input/file    →  blob, constructor, credentials, repository
  input/utf8    →  blob, constructor, credentials, repository
  helm          →  blob, constructor, credentials, oci, plugin, repository
  transfer      →  blob, credentials, helm, oci, repository, signing

PR plan:

  
| Gate | PRs                                                                           | Modules                                   | Can review in parallel |
|------|-------------------------------------------------------------------------------|-------------------------------------------|------------------------|
| 1    | ✅  PR 1 (https://github.com/open-component-model/open-component-model/pull/2580) | blob, signing, rsa                        | ✅ (3 modules)          |
| 2    | ✅ PR 2 (https://github.com/open-component-model/open-component-model/pull/2586)                                                                         | repository, sigstore                      | ✅ (2 modules)          |
| 3    | PR 3 (https://github.com/open-component-model/open-component-model/pull/2594)                                                                          | oci                                       | —                      |
| 4    | PR 4                                                                          | constructor                               | —                      |
| 5    | PR 5                                                                          | plugin, input/dir, input/file, input/utf8 | ✅ (4 modules)          |
| 6    | PR 6                                                                          | helm                                      | —                      |
[...]

PR #2580, opened 2026-05-20
feat!: update rsa, blob and signing to use runtime.Typed

matthiasbruns:
This PR migrates the following bindings to runtime.Typed credentials
- blob 
- rsa
- signing

Since we touch central interfaces, this is a breaking change.

The migration path can be observed here: https://github.com/open-component-model/open-component-model/pull/2519

Contributes: 
- https://github.com/open-component-model/ocm-project/issues/1055

#### Testing
Fully tested locally with `go.work` enabled in https://github.com/open-component-model/ocm-project/issues/1047


bindings/go/rsa/spec/credentials/v1/rsa_credentials.go
```diff
@@ -46,14 +93,56 @@ func MustRegisterCredentialType(scheme *runtime.Scheme) {
...
+// FromTyped converts [runtime.Typed] into RSACredentials.
+// Direct conversation as well as converting from [v1.DirectCredentials] is supported.
+// Other supported [runtime.Typed] implementations are [runtime.Raw] and [runtime.Unstructured].
+// For unsupported [runtime.Typed] implementations, an error will be returned.
+func FromTyped(creds runtime.Typed) (*RSACredentials, error) {
+	if creds == nil {
+		return nil, nil
+	}
+	switch t := creds.(type) {
+	case *RSACredentials:
```
jakobmoellerdev: what is happening here :D this is most certainly not how I expect this to look like here

bindings/go/rsa/spec/credentials/v1/rsa_credentials.go
```diff
@@ -46,14 +93,56 @@ func MustRegisterCredentialType(scheme *runtime.Scheme) {
...
+		}
+		props := map[string]string{}
+		if err := json.Unmarshal(data, &props); err != nil {
+			return nil, fmt.Errorf("error converting unstructured credentials to RSACredentials: %w", err)
+		}
+		return FromDirectCredentials(props), nil
 	}
+
+	slog.Error("unexpected credential type, expected RSACredentials or DirectCredentials", "type", creds.GetType())
+	return nil, fmt.Errorf("unexpected credential type: %T", creds)
```
frewilhelm: this seems redundant. why do we have this twice?

bindings/go/rsa/spec/credentials/v1/rsa_credentials.go
matthiasbruns: Well, this pattern evolved due to the fact that we get `runtime.Typed` and want to work with our internal typed in the bindings.
So I build a util func - also since we can get `Raw` and `Unstructured` typed from the plugin boundaries.

This came up during the plugin changes, I am open for other ideas though

bindings/go/rsa/spec/credentials/v1/rsa_credentials.go
matthiasbruns: changed error handling a bit

bindings/go/rsa/spec/credentials/v1/rsa_credentials.go
jakobmoellerdev: I believe this is not the valid approach I think. There should have been a scheme of credential types that is setup by the transformation handler / cli and that can use convert

matthiasbruns: The idea in the ADR was that credentials are being validated at the consumer level, not at the resolve stage
We have a scheme here, maybe I can use that to make the function less convoluted

> setup by the transformation handler / cli and that can use convert

how would this work in the first place. sigstore for example allows creds to be either OIDC or TrustedRoot in the same function call - nobody, but sigstore would know how to handle that, right?

jakobmoellerdev: Youre in the RSA bindings here so you know about the known credential types and you have a scheme to decode with

matthiasbruns: So what I could change

first I check for `DirectCredentials`, in all other cases I use `scheme` to convert to `RSACredentials`

bindings/go/rsa/signing/handler/internal/credentials/credentials.go
```diff
@@ -8,15 +8,12 @@ import (
...
 
-// IdentityTypeRSA is the consumer identity type for RSA signing.
-//
-// Deprecated: Use identityv1.VersionedType or identityv1.V1Alpha1Type instead.
-var IdentityTypeRSA = identityv1.V1Alpha1Type
-
 func PrivateKeyFromCredentials(creds *rsacredentialsv1.RSACredentials) (*rsa.PrivateKey, error) {
+	if creds == nil {
+		return nil, nil
+	}
```
fabianburth, line 16: I don't think we usually program that defensively, do we?

bindings/go/rsa/spec/credentials/v1/rsa_credentials.go
```diff
@@ -11,17 +14,53 @@ const (
...
-	CredentialKeyPrivateKeyPEM     = "private_key_pem"
-	CredentialKeyPrivateKeyPEMFile = CredentialKeyPrivateKeyPEM + "_file"
+	// Deprecated: Use CredentialKeyPublicKeyPEM instead.
+	DeprecatedCredentialKeyPublicKeyPEM = "public_key_pem"
+	// Deprecated: Use CredentialKeyPublicKeyPEMFile instead.
+	DeprecatedCredentialKeyPublicKeyPEMFile = "public_key_pem_file"
+	// Deprecated: Use CredentialKeyPrivateKeyPEM instead.
+	DeprecatedCredentialKeyPrivateKeyPEM = "private_key_pem"
+	// Deprecated: Use CredentialKeyPrivateKeyPEMFile instead.
+	DeprecatedCredentialKeyPrivateKeyPEMFile = "private_key_pem_file"
```
fabianburth: do we want a follow up issue here to remove this?

bindings/go/rsa/spec/credentials/v1/rsa_credentials.go
```diff
@@ -30,30 +67,61 @@ const (
...
+	// PublicKeyPEMFile is a path to a PEM file containing an RSA public key or X.509 certificate chain.
+	// Same semantics as PublicKeyPEM, but loaded from disk. Ignored when PublicKeyPEM is also set.
+	PublicKeyPEMFile string `json:"publicKeyPEMFile,omitempty"`
+	// PrivateKeyPEM is an inline PEM-encoded RSA private key (PKCS#1 or PKCS#8).
+	// Required for signing; not used during verification.
+	// Takes precedence over PrivateKeyPEMFile when both are set.
+	PrivateKeyPEM string `json:"privateKeyPEM,omitempty"`
+	// PrivateKeyPEMFile is a path to a PEM file containing an RSA private key (PKCS#1 or PKCS#8).
+	// Same semantics as PrivateKeyPEM, but loaded from disk. Ignored when PrivateKeyPEM is also set.
+	PrivateKeyPEMFile string `json:"privateKeyPEMFile,omitempty"`
```
fabianburth, line 53: This entire documentation is duplicated w/ the constants above now. I think this should serve as source of truth and be sufficient (if you feel the need, make a comment that refers to this).

bindings/go/rsa/spec/credentials/v1/rsa_credentials.go
```diff
@@ -30,30 +67,61 @@ const (
...
+
+	if dc, ok := creds.(*v1.DirectCredentials); ok {
+		return fromDirectCredentials(dc.Properties), nil
+	}
+
+	rsaCreds := RSACredentials{}
+	if err := Scheme.Convert(creds, &rsaCreds); err != nil {
+		return nil, err
+	}
+	return &rsaCreds, nil
```
fabianburth: I can't remember from your ADR. How long are we planning to support `*DirectCredentials`?

bindings/go/rsa/spec/credentials/v1/rsa_credentials_test.go
```diff
@@ -1,14 +1,110 @@
...
+func (f *fakeTyped) SetType(_ runtime.Type)       {}
+func (f *fakeTyped) DeepCopyTyped() runtime.Typed { return &fakeTyped{} }
+
+func TestFromTyped(t *testing.T) {
+	typ := runtime.NewVersionedType(RSACredentialsType, Version)
+
+	raw := &runtime.Raw{}
+	require.NoError(t, json.Unmarshal([]byte(
+		`{"type":"RSACredentials/v1","privateKeyPEM":"my-key","publicKeyPEMFile":"/path/pub.pem"}`),
+		raw))
```
fabianburth: nit: you should very easily be able to just instantiate `runtime.Raw{}` w/o the `json.Unmarshal` call.

bindings/go/rsa/spec/credentials/v1/rsa_credentials_test.go
```diff
@@ -1,14 +1,110 @@
...
+		name    string
+		input   runtime.Typed
+		want    *RSACredentials
+		wantErr bool
+	}{
+		{
+			name:  "nil",
+			input: nil,
+			want:  nil,
+		},
```
fabianburth: This goes towards the defensive programming direction. Not sure if this should be valid.

bindings/go/rsa/spec/credentials/v1/rsa_credentials_test.go
```diff
@@ -1,14 +1,110 @@
...
+		{
+			name: "Raw with deprecated snake_case keys",
+			input: func() *runtime.Raw {
+				r := &runtime.Raw{}
+				require.NoError(t, json.Unmarshal([]byte(
+					`{"type":"RSACredentials/v1","private_key_pem":"my-key","public_key_pem_file":"/path/pub.pem"}`),
+					r))
+				return r
+			}(),
```
fabianburth: same here, could be much simpler by just instantiating the raw

bindings/go/rsa/signing/handler/handler_test.go
```diff
@@ -222,10 +221,8 @@ func Test_RSA_Handler(t *testing.T) {
 									},
 								}
 							},
-							creds: func(t *testing.T) map[string]string {
-								return map[string]string{
-									rsacredentialsv1.CredentialKeyPublicKeyPEMFile: rootPEM,
-								}
+							creds: func(t *testing.T) runtime.Typed {
+								return &rsacredentialsv1.RSACredentials{PublicKeyPEMFile: rootPEM}
 							},
```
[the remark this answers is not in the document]
matthiasbruns: done

bindings/go/rsa/spec/credentials/v1/rsa_credentials_test.go
```diff
@@ -1,14 +1,110 @@
...
+				require.Error(t, err)
+				return
+			}
+			require.NoError(t, err)
+			assert.Equal(t, tt.want, got)
+		})
+	}
+}
+
 func TestFromDirectCredentials(t *testing.T) {
```
fabianburth: since `FromTyped` calls `fromDirectCredentials`, do we even require all those tests?

bindings/go/rsa/spec/credentials/v1/scheme_test.go
```diff
@@ -0,0 +1,68 @@
...
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"ocm.software/open-component-model/bindings/go/runtime"
+)
+
+func TestMustRegisterCredentialType(t *testing.T) {
```
fabianburth: do we typically include such tests in our modules? feels like they are rather testing the `scheme` than any implementation here

bindings/go/rsa/signing/handler/handler.go
```diff
@@ -145,10 +136,13 @@ func (h *Handler) Verify(
 	signed descruntime.Signature,
 	// we use hints from the signature to determine the correct settings, so no additional config is needed
 	_ runtime.Typed,
-	creds map[string]string,
+	creds runtime.Typed,
 ) error {
-	typedCreds := rsacredentialsv1.FromDirectCredentials(creds)
-	pubFromCreds, err := rsacredentials.PublicKeyFromCredentials(typedCreds)
+	rsaCreds, err := rsacredentialsv1.FromTyped(creds)
```
fabianburth: nit: maybe a me problem, but `FromTyped` somehow tells my brain that whatever I'm getting back is not typed anymore.

matthiasbruns:
So now I am happy :)

bindings/go/rsa/signing/handler/handler_test.go
```diff
@@ -173,61 +170,62 @@ func Test_RSA_Handler(t *testing.T) {
...
+											Value:     si.Value,
+											Issuer:    c.interm.Subject.String(),
+										},
+									}
+								},
+								creds: func(t *testing.T) runtime.Typed {
+									return &rsacredentialsv1.RSACredentials{PublicKeyPEMFile: rootPEM}
+								},
+							}
+						}(),
```
fabianburth: why do several whole test cases need to be constructed by a function now?

bindings/go/rsa/signing/handler/internal/credentials/credentials.go
matthiasbruns: Well either here or in the callers - in the end credentials are somewhat optional still, right?

bindings/go/rsa/spec/credentials/v1/rsa_credentials.go
matthiasbruns: yes -> https://github.com/open-component-model/ocm-project/issues/1072

bindings/go/rsa/spec/credentials/v1/rsa_credentials.go
matthiasbruns: already cleaned up in one of the 100 commits in the meantime :D

bindings/go/rsa/spec/credentials/v1/rsa_credentials.go
matthiasbruns: As long as we support legacy .ocmconfigs - also I don't really know how we support hybrid credentials like the one in sigstore

bindings/go/rsa/spec/credentials/v1/rsa_credentials_test.go
matthiasbruns: I would keep it as mentioned above

bindings/go/rsa/spec/credentials/v1/rsa_credentials_test.go
matthiasbruns: not anymore, it was public before

bindings/go/rsa/spec/credentials/v1/scheme_test.go
matthiasbruns: removed

bindings/go/rsa/signing/handler/handler.go
matthiasbruns: maybe - something else in mind?

bindings/go/rsa/signing/handler/handler_test.go
matthiasbruns: rolled back

bindings/go/rsa/signing/handler/internal/credentials/credentials.go
fabianburth: Yeah, I guess you've got a point in general. I still feel like this is not the correct level to handle this.
Besides, for the signing handler, I don't think those credentials are actually optional.

bindings/go/rsa/spec/credentials/v1/rsa_credentials.go
fabianburth: what do you mean by hybrid credentials?

bindings/go/rsa/signing/handler/handler.go
fabianburth: hmm... since those functions are module / package specific now anyway, maybe something like `ToRSA(creds)` or `AsRSA()`? or maybe we stick to our runtime terminology and call it `Convert` or `ConvertToRSA`?

also not really convinced by any of those yet

bindings/go/rsa/spec/credentials/v1/rsa_credentials_test.go
fabianburth: I actually meant by doing
```
r := &runtime.Raw{
  Type: "RSACredentials/v1", // or even runtime.NewVersionedType(Type, Version)
  Data: `{"type":"RSACredentials/v1","private_key_pem":"my-key","public_key_pem_file":"/path/pub.pem"}`
}
```
then you can omit the whole inline function here

bindings/go/rsa/signing/handler/internal/credentials/credentials.go
matthiasbruns: well I can move the nil guard one layer up

bindings/go/rsa/spec/credentials/v1/rsa_credentials.go
matthiasbruns: this
https://github.com/open-component-model/open-component-model/pull/2519/changes#diff-7cc53b30ceb0858a2ebd35fb99a6026489b8d2dce810938ed3a300d6e6107886R95

bindings/go/rsa/signing/handler/internal/credentials/credentials.go
fabianburth: I'd appreciate if you try to do so 🫶 I think, in an upper layer, it should also be more obvious what is optional and what is required

bindings/go/rsa/spec/credentials/v1/rsa_credentials.go
fabianburth: just by looking at that function, I don't even really comprehend yet how the use case looks like and what you are trying to achieve there

bindings/go/rsa/spec/credentials/v1/rsa_credentials_test.go
matthiasbruns: well, we might have a problem with our design and the keys

so when I do
```go
	raw := &runtime.Raw{}
	err := raw.UnmarshalJSON([]byte(`{"type":"RSACredentials/v1","privateKeyPEM":"my-key","publicKeyPEMFile":"/path/pub.pem"}`))
	raw.SetType(VersionedType)
```

raw internally converts from

```json
{"type":"RSACredentials/v1","privateKeyPEM":"my-key","publicKeyPEMFile":"/path/pub.pem"}
```

into

```json
{"private_key_pem":"my-key","public_key_pem_file":"/path/pub.pem"}
```

with:

```go
   u.Data, err = jsoncanonicalizer.Transform(u.Data)
```

So using `Scheme.Convert` does not work with `Raw` and `camelCase` json tags in `RSACredentials` @jakobmoellerdev  also important for you. I would add `Raw` handling back into `FromTyped` since we need to handle `snake_case` there again

bindings/go/rsa/signing/handler/internal/credentials/credentials.go
matthiasbruns: done

bindings/go/rsa/spec/credentials/v1/rsa_credentials.go
matthiasbruns: I'll explain in the WR - that's simpler

bindings/go/rsa/spec/credentials/v1/rsa_credentials_test.go
matthiasbruns: yeah updated - ty

bindings/go/rsa/signing/handler/handler.go
matthiasbruns: Renamed to `ConvertToRSACredentials`

bindings/go/rsa/spec/credentials/v1/rsa_credentials_test.go
matthiasbruns: resolved thanks to @jakobmoellerdev  <3

fabianburth (approved):

jakobmoellerdev (approved):

Issue #2373, opened 2026-04-23
feat: Implement Replication Controller for component version transfer


frewilhelm:
## Summary

Implement the Replication Controller as described in [ADR-0017 (PR)](https://github.com/open-component-model/open-component-model/pull/2316)/[ADR-0017](https://github.com/open-component-model/open-component-model/blob/b74283882e6f31e0970fcc87579a947939fa015a/docs/adr/0017_replication_controller.md). The controller transfers component versions from a source to a target OCI repository using the existing transfer library (`bindings/go/transfer/`).

## Scope

**In Scope:** `Replication` CRD, dedicated transfer worker pool, two-phase reconciliation (Plan -> Execute), finalizer with bounded drain, unit/integration/e2e tests.

**Out of Scope:** multi-version replication, CRD split into Replication + Transfer, transfer policies, `ReplicationSet` fan-out, drift detection at target.

### CRD (`delivery.ocm.software/v1alpha1/Replication`)

**Spec** references a source `Component` CR, a target `Repository` CR, a `transferConfig` (either a ConfigMap reference or inlined `generic.config.ocm.software/v1` configuration containing `GlobalTransferOptions/v1alpha1` with `recursive` and `copyMode` settings), optional `ocmConfig` for credentials, an `interval`, and a `suspend` flag.

**Status** tracks `lastTransferredVersion`, `lastTransferredDigest` (`*v2.Digest`), `componentInfo` (mirrored from source), `effectiveOCMConfig`, and standard conditions:
- `Ready` — with reasons: `TransferComplete`, `TransferFailed`, `BuildGraphDefinitionFailed`
- `TransferInProgress` - `True` while a transfer is in-flight

A mismatch between `componentInfo.digest` and `lastTransferredDigest` signals a pending/in-flight/failed transfer; conditions disambiguate which.

### Transfer Worker Pool

New package at `internal/transfer/workerpool/`, separate from the resolution worker pool. Key differences from resolution:

- No result caching (TGDs regenerate cheaply)
- Longer-running tasks with per-item cancellation (keyed by CR UID)
- Transient error retry with exponential backoff inside the worker
- Sentinel `ErrTransferInProgress` for duplicate submissions

Exposes: `Submit`, `Cancel`, `HasInFlight`, and an event source for completion-triggered reconciliation. Registered with the manager via `mgr.Add()`. Worker count, queue length, and subscriber buffer size are configurable via CLI flags.

### Controller Reconciliation

[...]

PR #2587, opened 2026-05-20
feat(transfer)!: add transfer config

frewilhelm:

Adds a typed wire-format configuration (`TransferConfiguration/v1alpha1`) for transfer operations under `bindings/go/transfer/v1alpha1`, and reshapes `transfer.Options` to embed it. Consumers (CLI, the upcoming Replication controller per [ADR-0020](docs/adr/0020_replication_controller.md)) can now load transfer knobs (`recursive`, `copyMode`, `uploadType`) from YAML/JSON via `transfer.FromConfig` and apply them alongside the runtime-only mappings.

This is the binding-side prerequisite for the Replication CRD reconciler that lands in a follow-up PR.

> [!WARNING]
> Breaking change: the `transfer.CopyMode` / `transfer.UploadType` int constants previously in `bindings/go/transfer/internal/options.go` are removed. Callers now use the string-typed enums in `transferv1alpha1`.

Related to #2373

##### How to test the changes

Required files to test the changes:

A transfer config
```yaml
type: TransferConfiguration/v1alpha1
recursive: true
copyMode: allResources
uploadType: ociArtifact
```

- Check out [frewilhelm/feat-add-transfer-config-cli](https://github.com/frewilhelm/open-component-model/tree/feat-add-transfer-config-cli)
- Init go work (remove breaking module if necessary but keep `cli` and `transfer`
- Build the binary

```bash
ocmv2 transfer cv --transfer-config transfer.config ...
```

##### Verification


Skarlso:
I think this looks okay. It's obviously going to break things... But that's expected. And we are going to use this during the replication controller update.

Skarlso (dismissed):

bindings/go/transfer/options.go
```diff
@@ -1,99 +1,40 @@
...
 	return func(o *Options) {
 		o.CopyMode = mode
 	}
 }
 
-// WithRecursive enables or disables recursive transfer of referenced component versions.
-// When enabled, the transfer discovers all component references in the source descriptor
-// and transfers them (and their transitive references) to the target repository.
-func WithRecursive(recursive bool) Option {
+func WithRecursive(recursive int) Option {
```
fabianburth: theoretically breaking change on public API (i did not know it already exists when I told you to align)

bindings/go/transfer/v1alpha1/spec/config.go
```diff
@@ -0,0 +1,97 @@
...
+		return nil
+	}
+
+	if !cfg.Type.IsEmpty() {
+		if cfg.Type.Name != ConfigType || (cfg.Type.Version != "" && cfg.Type.Version != Version) {
+			return fmt.Errorf("invalid type %q (must be %q or %q)",
+				cfg.Type, ConfigType, runtime.NewVersionedType(ConfigType, Version))
+		}
+	}
+
```
frewilhelm, line 79: I believe now we could also validate for valid integers for `recursive`

bindings/go/transfer/doc.go
```diff
@@ -75,4 +75,37 @@
...
+// # Driving the Transfer from a Wire-Format Config
+//
+// The transfer knobs (recursive, copy mode, upload type) are also expressed by
+// [transferv1alpha1.Config] - a canonical wire format that downstream consumers
+// (CLI, controllers, other tooling) can load from YAML/JSON and feed in directly.
+// Use [FromConfig] to convert a loaded config into [Option]s alongside the
+// runtime-only mappings:
+//
+//	cfg := &transferv1alpha1.Config{
+//	    Recursive:  ptr.To(true), // *bool, see GetRecursive() for the default
```
frewilhelm: This is outdated

bindings/go/transfer/v1alpha1/spec/config.go
```diff
@@ -0,0 +1,97 @@
...
+	Type runtime.Type `json:"type"`
+
+	// Recursive configures transferring component references with the parent component.
+	// -1 means infinite recursion, 0 means no recursion.
+	Recursive int `json:"recursive,omitempty"`
+
+	CopyMode CopyMode `json:"copyMode,omitempty"`
+
+	UploadType UploadType `json:"uploadType,omitempty"`
+}
```
fabianburth, line 61: The diff between the config and the flags will very likely cause confusion. Are we aware? Do we want to accept this? (I admit, I prefer your names as well)

```
Flags:
      --copy-resources         copy all resources in the component version
      --dry-run                build and validate the graph but do not execute
  -h, --help                   help for component-version
  -o, --output enum            output format of the component descriptors
                               (must be one of [json ndjson yaml]) (default yaml)
  -r, --recursive              recursively discover and transfer component versions
      --transfer-spec string   path to a transfer specification file (use "-" for stdin)
  -u, --upload-as enum         Define whether copied resources should be uploaded as OCI artifacts (instead of local blob resources).
```

bindings/go/transfer/v1alpha1/config.go
```diff
@@ -0,0 +1,97 @@
...
+// so any new transfer setting belongs here first.
+//
+// +k8s:deepcopy-gen:interfaces=ocm.software/open-component-model/bindings/go/runtime.Typed
+// +k8s:deepcopy-gen=true
+// +ocm:typegen=true
+// +ocm:jsonschema-gen=true
+type Config struct {
+	// +ocm:jsonschema-gen:enum=TransferConfiguration/v1alpha1
+	// +ocm:jsonschema-gen:enum:deprecated=TransferConfiguration
+	Type runtime.Type `json:"type"`
```
fabianburth: I think transfer config will need to be extensible w/ additional types in the future. I think in my transfer config ADR, we said we want purpose scoped config types, so we can iterate. Also your validate will become huge eventually.

So, it is fine if this is just ONE type (like the type representing the cli flags as config options), but not the entire "top level" type (which TransferConfiguration sounds like).

bindings/go/transfer/v1alpha1/spec/config.go
Skarlso: There is no concept yet what _valid_ means for the recursion since it's either 0 or -1. 🤔

bindings/go/transfer/options.go
Skarlso: Yes. That's why the PR is marked as breaking. :) ( amongst other things :D )

bindings/go/transfer/v1alpha1/spec/config.go
Skarlso: Errr, good catch. 🤔 Humm hummm. I would leave it as is here and then we'll adjust in code, and perhaps update the CLI to the new name if we prefer this? :)

frewilhelm: I don't think we need to adjust the cli option names (right now). They are parsed in the command and then passed as transfer options currently

bindings/go/transfer/v1alpha1/config.go
Skarlso: Sooooooo.... What is it exactly you would want here? Instead of runtime.Type we should put specific types here...? Pinged in slack to discuss. :)

Skarlso: For now, we are going to ignore this, because it actually requires a deeper thinking. This TransferConfig should be the umbrella options config and not the actual config. And then, there should be overrides with specific configs.

bindings/go/transfer/internal/discovery.go
```diff
@@ -159,7 +159,7 @@ type discoverer struct {
 //     another parent with a different resolver, an error is returned — the ambiguity must
 //     be resolved by the caller via an explicit WithTransfer mapping for that component.
 func (d *discoverer) Discover(ctx context.Context, parent *discoveryValue) ([]string, error) {
-	if !d.recursive {
+	if d.recursive == 0 {
```
matthiasbruns: this would also fix https://github.com/open-component-model/ocm-project/issues/907

bindings/go/transfer/v1alpha1/spec/config.go
```diff
@@ -0,0 +1,97 @@
...
+type Config struct {
+	// +ocm:jsonschema-gen:enum=TransferConfiguration/v1alpha1
+	// +ocm:jsonschema-gen:enum:deprecated=TransferConfiguration
+	Type runtime.Type `json:"type"`
+
+	// Recursive configures transferring component references with the parent component.
+	// -1 means infinite recursion, 0 means no recursion.
+	Recursive int `json:"recursive,omitempty"`
+
+	CopyMode CopyMode `json:"copyMode,omitempty"`
```
matthiasbruns, line 52: can we get some docs on the other properties as well?

matthiasbruns (commented):
we should definitely align `--recursive` in the cli, since it's still a boolean flag afaik

docs/adr/0020_replication_controller.md
```diff
@@ -68,9 +68,9 @@ spec:
     inlined:
       type: generic.config.ocm.software/v1
       configurations:
-        - type: GlobalTransferOptions/v1alpha1 # Current not existing configuration type.
+        - type: TransferConfiguration/v1alpha1 # Current not existing configuration type.
```
jakobmoellerdev: nice, much better name

docs/adr/0020_replication_controller.md
```diff
@@ -94,9 +94,9 @@ data:
   transferConfig: |
     type: generic.config.ocm.software/v1
     configurations:
-      - type: GlobalTransferOptions/v1alpha1 # Current not existing configuration type.
+      - type: TransferConfiguration/v1alpha1 # Current not existing configuration type.
         spec:
-          recursive: false
+          recursive: -1
```
jakobmoellerdev: I think this is not the correct way to put this, it probably should allow true as well. this is because its now user facing

Skarlso: I don't get it. :D The user-facing one is the CLI, right? And that's already an int I believe. 🤔 Or by userfacing do mean parts of the API contract? ( I.e.: Public function arguments ).

matthiasbruns: in tranfer, it's a `bool` - there was a `int` in other commands, but we never were able to actually limit the depth and I honestly feel that `int` is really counter intuitive for this

Skarlso: The int makes sense if we will really integrate it with things like the depth tracking and restriction. I think it will be a lot harder to change later. I'm all for leaving it as an int. I'm just not sure which public facing int Jakob is referring to in this case. :D /cc @jakobmoellerdev

bindings/go/transfer/v1alpha1/spec/config.go
Skarlso: added

docs/adr/0020_replication_controller.md
jakobmoellerdev: TransferConfiguration is a public API so technically anyone can use it from bindings.

Skarlso: But we literally are just introducing it right now. 🤔

Skarlso: I completely misread this 🤦 Jakob is saying that this needs better UX. I read it like it's been used already. :D Sorry folks for the extra rounds.

Skarlso: Done 👍

bindings/go/transfer/v1alpha1/config.go
```diff
@@ -0,0 +1,110 @@
...
+var Scheme = runtime.NewScheme()
+
+func init() {
+	Scheme.MustRegisterWithAlias(&Config{},
+		runtime.NewVersionedType(ConfigType, Version),
+		runtime.NewUnversionedType(ConfigType),
+	)
+}
+
+// Config is the canonical wire format for transfer knobs. Downstream consumers
```
fabianburth: "knobs"?

bindings/go/transfer/v1alpha1/config.go
```diff
@@ -0,0 +1,110 @@
...
+// (CLI, controllers) load it from YAML/JSON and pass it into [transfer.FromConfig],
+// so any new transfer setting belongs here first.
```
fabianburth: Isn't this the opposite of what I said in a previous comment? 
I believe we likely want individual purpose built transfer config types. This type is purpose built to mirror the CLI flags (or maybe more general, options that affect all components / resources, so kind of "global" as @frewilhelm phrased it originally). 
IMHO, if you wanted to introduce an option that only affects a specific set of resource (e.g. specified through a selector), you should introduce a new config type.

bindings/go/transfer/v1alpha1/spec/config.go
```diff
@@ -0,0 +1,110 @@
...
+	// with their own repository references.
+	UploadType UploadType `json:"uploadType,omitempty"`
+}
+
+func (cfg *Config) GetCopyMode() CopyMode {
+	if cfg == nil || cfg.CopyMode == "" {
+		return CopyModeLocalBlobResources
+	}
+	return cfg.CopyMode
+}
```
fabianburth: Is our idiomatic pattern really to perform `nil` checks everywhere in order to return defaults? Checking our existing config types, I believe we never implemented getters at all before.

bindings/go/transfer/v1alpha1/config.go
fabianburth: If you prepare for this properly, I think you need a 
```
Lookup(cfg *genericv1.Config) (*Config, error)
```

bindings/go/transfer/v1alpha1/spec/config.go
fabianburth, line 1: The established pattern would be to store the config under ‎*bindings/go/transfer/v1alpha1/**spec**/config.go*‎. For this type, you don't need a corresponding **runtime** package, because your config type does not contain other `runtime.Typed` objects (compared to e.g. resolvers)

bindings/go/transfer/v1alpha1/spec/config.go
```diff
@@ -0,0 +1,110 @@
...
+
+// Validate rejects a non-matching [Config.Type] and unknown enum values.
+// An empty Type is allowed so callers constructing a Config programmatically
+// (without going through [Scheme.Decode]) do not need to set it explicitly.
+// Empty enum fields are allowed; consumers must call [Config.GetCopyMode] /
+// [Config.GetUploadType] to resolve them.
+func (cfg *Config) Validate() error {
+	if cfg == nil {
+		return nil
+	}
```
fabianburth, line 71: Also weird, but connected to the way the methods above are implemented.

bindings/go/transfer/v1alpha1/spec/config.go
```diff
@@ -0,0 +1,110 @@
...
+	return cfg.UploadType
+}
+
+// Validate rejects a non-matching [Config.Type] and unknown enum values.
+// An empty Type is allowed so callers constructing a Config programmatically
+// (without going through [Scheme.Decode]) do not need to set it explicitly.
+// Empty enum fields are allowed; consumers must call [Config.GetCopyMode] /
+// [Config.GetUploadType] to resolve them.
+func (cfg *Config) Validate() error {
```
fabianburth, line 68: Would it make sense to try to handle this through JSON schema validation? Kind of similar to what we do [here](https://github.com/open-component-model/open-component-model/blob/ad110c16e03862828e70a1ae733b3f3e257392eb/bindings/go/descriptor/v2/validate.go#L51)? 

As we are under a bit of time pressure right now, that might be something to tackle later.

bindings/go/transfer/v1alpha1/spec/recursive.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+// Recursive controls whether component references are transferred along with
+// their parent component, and how deep that recursion goes.
+//
+// In the wire format it accepts either an integer or a boolean:
+//   - an integer sets an explicit depth: -1 means infinite recursion, 0 means
+//     no recursion, and n > 0 limits recursion to n levels;
+//   - a boolean is shorthand: true maps to infinite recursion (-1), false to
+//     no recursion (0).
```
fabianburth: I know @jakobmoellerdev prompted this, but this feels like over-engineering that might come back haunting us later. Feels like the same situation as when I allowed port to be passed as string or number in ocm v1 and @jakobmoellerdev hated it.

bindings/go/transfer/doc.go
```diff
@@ -75,4 +75,37 @@
 //	        transfer.FromResolver(repoResolver),
 //	    ),
 //	)
+//
+// # Driving the Transfer from a Wire-Format Config
+//
+// The transfer knobs (recursive, copy mode, upload type) are also expressed by
```
fabianburth: again, "knobs"??

fabianburth (changes requested):

bindings/go/transfer/v1alpha1/config.go
Skarlso: Errr... which previous comment pls? :) I don't think we changed the config since your comments on the multiple types thing. 🤔 And we agreed that we leave it as is for now, because we can ither provider an umbrella type and then wait for overrides, or we don't. :D

bindings/go/transfer/v1alpha1/spec/recursive.go
jakobmoellerdev: I hate to admit it but youre probably right...

bindings/go/transfer/v1alpha1/spec/config.go
Skarlso: Don't think so, no. But it does give me some help in determining recursion given the two types. I'll figure out something better here.

bindings/go/transfer/doc.go
Skarlso: all "knobs" are gone :D :D

bindings/go/transfer/internal/graph.go
```diff
@@ -67,9 +68,9 @@ type TransferRoot struct {
 func BuildGraphDefinition(
 	ctx context.Context,
 	roots map[string]TransferRoot,
-	recursive bool,
-	copyMode int,
-	uploadType int,
+	recursive int,
+	copyMode transferv1alpha1.CopyMode,
+	uploadType transferv1alpha1.UploadType,
```
fabianburth: do we eventually need to pass the entire config here? (asking because we are already in the business of breaking things anyway)

bindings/go/transfer/options.go
```diff
@@ -1,99 +1,40 @@
...
-// components via [Component], and a target via [ToRepositorySpec].
+// It embeds the wire-format [transferv1alpha1.Config] (so the declarative settings
+// are accessed as promoted fields) and adds [Mappings], which carries the live
+// runtime objects (resolvers, repository specs) that the wire format cannot
+// serialize.
+//
+// The embedded [transferv1alpha1.Config.Type] is wire-only and not consulted
+// by the runtime; it is exposed only because Go's struct-embedding promotes
+// every field. Callers should not set it on Options.
 type Options struct {
```
fabianburth: Wondering whether this is how we want to do it. I think at other places, we started just passing around the plain config right? 
(see e.g. https://github.com/open-component-model/open-component-model/blob/c5ed93a358c34a8c40efccd4ccb616f2be6d5c00/bindings/go/helm/repository/resource/resource_repository.go#L51)

I even think we had a previous PR where you introduced options to wrap a config and we had you remove it.

bindings/go/transfer/integration/integration_test.go
```diff
@@ -1296,3 +1296,128 @@ func Test_Integration_TransferDockerManifestLocalBlob_CTFToOCI(t *testing.T) {
...
+// of [Test_Integration_TransferOCIImageResource_CopyModeAllResources]: same component
+// shape and target outcome, but the transfer is configured via a [transferv1alpha1.Config]
+// fed through [transfer.FromConfig] - the same path the CLI's --transfer-config flag and
+// the replication controller's CRD spec take.
+//
+// This pins the SSOT property at runtime: a Config struct must drive a transfer
+// identically to the equivalent functional-options call. If FromConfig drifts from
+// the With* setters (or if the v1alpha1 enum values fall out of sync with what the
+// internal layer expects), this test fails.
+func Test_Integration_TransferOCIImageResource_FromConfig(t *testing.T) {
```
fabianburth: I think even requiring this test shows why we might want to get rid of that translation

bindings/go/transfer/v1alpha1/spec/config_test.go
```diff
@@ -0,0 +1,205 @@
...
+	}{
+		{"valid empty", Config{}, ""},
+		{"valid all fields", Config{Recursive: -1, CopyMode: CopyModeAllResources, UploadType: UploadAsOciArtifact}, ""},
+		{"valid recursive false", Config{Recursive: 0}, ""},
+		{"valid copyMode localBlob", Config{CopyMode: CopyModeLocalBlobResources}, ""},
+		{"valid uploadType default", Config{UploadType: UploadAsDefault}, ""},
+		{"valid uploadType localBlob", Config{UploadType: UploadAsLocalBlob}, ""},
+		{"invalid copyMode", Config{CopyMode: "garbage"}, "invalid copyMode"},
+		{"invalid uploadType", Config{UploadType: "garbage"}, "invalid uploadType"},
+		{"valid recursive depth", Config{Recursive: 3}, ""},
```
fabianburth: Maybe this should currently throw some kind of "Not Implemented" error?

bindings/go/transfer/v1alpha1/spec/config_test.go
```diff
@@ -0,0 +1,205 @@
...
+	r := require.New(t)
+
+	in := &Config{Recursive: -1, CopyMode: CopyModeAllResources, UploadType: UploadAsOciArtifact}
+	// DefaultType stamps the canonical wire-format type onto a struct that has
+	// none. Asserting the result pins the registration contract from init():
+	// the first alias passed to MustRegisterWithAlias (the versioned form) is
+	// what new writes serialise with. A reorder there would flip this assertion.
+	_, err := Scheme.DefaultType(in)
+	r.NoError(err)
+	r.Equal(runtime.NewVersionedType(ConfigType, Version), in.Type)
```
fabianburth: Do we really need this test?

fabianburth: Generally with the whole tests here not sure whether this is rather testing scheme behaviour (especially testing it for yaml and json).

bindings/go/transfer/doc.go
```diff
@@ -75,4 +75,39 @@
...
+//	        ),
+//	    )...,
+//	)
+//
+// [FromConfig] skips empty enum fields - so callers can overlay explicit
+// overrides on top of a partial config without the zero values clobbering
+// them. The resulting [Options] therefore still carries empty enum strings
+// until [BuildGraphDefinition] runs; that's where the empties are resolved to
+// their canonical defaults ([transferv1alpha1.CopyModeLocalBlobResources],
+// [transferv1alpha1.UploadAsDefault]).
```
fabianburth: thats a lot of docs required to explain how to use a config

Skarlso: True. :D

bindings/go/transfer/internal/graph.go
Skarlso: Might as well, yes. 👍

bindings/go/transfer/v1alpha1/spec/config_test.go
Skarlso: You mean if someone would define a recursion and expect to actually work but nothing is there yet so we should warn them? Yeah, that sounds good.

bindings/go/transfer/v1alpha1/spec/config_test.go
Skarlso: Yeah, I think I agree. This is mostly covered by marshalling anyways.

bindings/go/transfer/options.go
Skarlso: That's right. 🤔

bindings/go/transfer/v1alpha1/spec/config.go
Skarlso: Removed all the getters. :)

Skarlso:
I might have gone a bit overboard with the cleanup... 🤔 But not many other configs do anything that is being done here so I was trying to align things.

bindings/go/transfer/v1alpha1/spec/config.go
```diff
@@ -0,0 +1,158 @@
...
+	return nil
+}
+
+// LookupConfig extracts the transfer configuration from a central generic
+// config. All entries of type [ConfigType] are decoded, validated, and merged
+// via [Merge]. Returns nil if cfg is nil or contains no transfer entries.
+func LookupConfig(cfg *genericv1.Config) (*Config, error) {
+	if cfg == nil {
+		return nil, nil
+	}
```
fabianburth: all the other configs are not doing this kind of defensive programming

fabianburth (dismissed):

bindings/go/transfer/v1alpha1/spec/config.go
Skarlso: fuck, thought I got that one. sorry.

Skarlso: >all the other configs are not doing this

BTW, not true... :D bindings/go/configuration/resolvers/v1alpha1/spec/config.go is also doing it. ;)

Skarlso: Removed. :)

fabianburth (approved):

jakobmoellerdev (approved):

PR #2594, opened 2026-05-21
feat: migrate oci to typed credentials

matthiasbruns:
This PR migrates the following bindings to runtime.Typed credentials
- oci

Since we cascade the repository breaking changes, set public consts to private and change the spec, this is a breaking change.

The migration path can be observed here: https://github.com/open-component-model/open-component-model/pull/2519

Contributes: 
- https://github.com/open-component-model/ocm-project/issues/1055

#### Testing
- breaking change only tested with binding tests
- go.work disabled, task test is green


matthiasbruns:
@coderabbitai review

bindings/go/oci/spec/identity/v1/convert.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"log/slog"
+
+	"ocm.software/open-component-model/bindings/go/runtime"
+)
+
+// ToIdentity converts an [OCIRegistryIdentity] into a [runtime.Identity].
+// Empty fields are omitted from the resulting map. If the type field is unset,
+// the canonical [VersionedType] is used.
+func ToIdentity(identity *OCIRegistryIdentity) runtime.Identity {
```
matthiasbruns, line 12: currently unused, but would want to keep it if we decide to change identity handling

bindings/go/oci/spec/identity/v1/convert.go
```diff
@@ -0,0 +1,60 @@
...
+	if identity.Path != "" {
+		id[runtime.IdentityAttributePath] = identity.Path
+	}
+	return id
+}
+
+// FromIdentity converts a [runtime.Identity] into an [OCIRegistryIdentity].
+// Attributes outside the OCI registry schema are ignored. If the type
+// attribute is missing or empty, the canonical [VersionedType] is used.
+func FromIdentity(id runtime.Identity) *OCIRegistryIdentity {
```
matthiasbruns, line 40: used in  https://github.com/matthiasbruns/open-component-model/blob/66d9100e212a5fe1c25067f1ddc97968bd934ffb/bindings/go/oci/credentials/docker_config.go#L103
and
https://github.com/matthiasbruns/open-component-model/blob/91417511873b9fccd4ac9b60e735fa89fbe6beac/bindings/go/oci/spec/identity/v1/type.go#L25

matthiasbruns:
I am not 100% sure about the identity parsing here - I'd like your feedback on this

jakobmoellerdev (approved):

bindings/go/oci/spec/identity/v1/convert.go
jakobmoellerdev: while i would prefer a method, this will work just fine

frewilhelm (approved):
nice

PR #2596, opened 2026-05-21
feat(transfer): wire TransferOCIArtifact into transfer graph builder

### LINKED ISSUE #2485: Direct OCI-to-OCI streaming transfer transformation (TransferOCIArtifact/v1alpha1)

jakobmoellerdev:
## Summary

Follow-up to #2495. Wires the `TransferOCIArtifact` fused transformer (added to the OCI module in #2495) into the transfer graph builder.

- When `UploadAsOciArtifact` is set, `processOCIArtifact` now takes the streaming fast path via `processOCIArtifactStreaming`, emitting a single `TransferOCIArtifact` node instead of the two-node `GetOCIArtifact` + `AddOCIArtifact` pair
- `NewDefaultBuilder` registers the `TransferOCIArtifact` transformer alongside the existing Get/Add transformers
- OCI dependency bumped to post-#2495 pseudo-version (`v0.0.43-0.20260520144124-c7e5118f86a6`) since streaming support is not yet tagged

The `TransferOCIArtifact` transformer internally calls `DownloadResourceStream` + `UploadResourceStream` via the `ocistream.Repository` interface, which uses `oras.CopyGraph` to stream blobs one-at-a-time with HEAD-checked deduplication at the destination. No tar file is ever created.

The CTF and `UploadAsDefault` paths are unchanged — they still use the existing `GetOCIArtifact` + `AddLocalResource` two-node path.

Closes #2485

## Test plan

- [ ] `go build ./...` in `bindings/go/transfer` passes
- [ ] `go vet ./...` passes
- [ ] `go test ./...` passes — updated `TestBuildGraphDefinition_OCIImageUploadAsOCIArtifact` to assert 3 nodes (single `TransferOCIArtifact`) instead of old 4-node Get+Add structure
- [ ] Existing CTF target test (`TestBuildGraphDefinition_CTFTarget`) unchanged — confirms no regression on non-OCI paths
- [ ] Existing `TestBuildGraphDefinition_OCIImageWithCopyAllResources` (`UploadAsDefault`) unchanged — confirms two-node path still works


bindings/go/transfer/integration/go.mod
```diff
@@ -10,11 +10,11 @@ require (
 	github.com/testcontainers/testcontainers-go/modules/registry v0.42.0
 	golang.org/x/crypto v0.50.0
 	ocm.software/open-component-model/bindings/go/blob v0.0.12
-	ocm.software/open-component-model/bindings/go/credentials v0.0.10
+	ocm.software/open-component-model/bindings/go/credentials v0.0.11
 	ocm.software/open-component-model/bindings/go/ctf v0.4.0
 	ocm.software/open-component-model/bindings/go/descriptor/runtime v0.0.0-20260505072254-1c17fcd5c971
 	ocm.software/open-component-model/bindings/go/descriptor/v2 v2.0.3-alpha3
-	ocm.software/open-component-model/bindings/go/oci v0.0.40
+	ocm.software/open-component-model/bindings/go/oci v0.0.43-0.20260520144124-c7e5118f86a6
```
fabianburth: is this due to dependency issues w/ creds?

bindings/go/transfer/internal/oci.go
```diff
@@ -141,8 +139,64 @@ func imageReferenceFromAccess(id string) referenceNameOption {
...
+
+	unstructured, err := runtime.UnstructuredFromMixedData(map[string]any{
+		"resource": resource,
+		"targetResource": map[string]any{
+			"name":     resource.Name,
+			"version":  resource.Version,
+			"type":     resource.Type,
+			"relation": resource.Relation,
+			"access": map[string]any{
+				"type":           runtime.NewVersionedType(ociv1.LegacyType, ociv1.LegacyTypeVersion).String(),
```
fabianburth: is it desired to use the `ociv1.LegacyType` here?

bindings/go/transfer/internal/oci.go
```diff
@@ -141,8 +139,64 @@ func imageReferenceFromAccess(id string) referenceNameOption {
...
+				"imageReference": targetImageReference,
```
fabianburth: In `ociUploadAsArtifact`, you used `referenceNameOption` with 

```
// staticReferenceName returns a referenceNameOption that constructs the reference name by combining the target repository base URL and the given reference name.
// If the target repository base URL is empty, it returns the given reference name as is.
func staticReferenceName(referenceName string) referenceNameOption 
```

I have to admit, I don't know why we needed that - but if we needed it there, do we need it here too?

bindings/go/transfer/integration/go.mod
jakobmoellerdev: Yes — credentials v0.0.11 introduced the `runtime.Typed` credential interface required by `TransferOCIArtifact`. The PR predated the v0.0.12 tag. After rebasing onto main this is moot; all dep versions are now at main's newer levels (credentials v0.0.12, oci v0.0.45).

bindings/go/transfer/internal/oci.go
jakobmoellerdev: Yes, intentional — matches the pattern in `ociUploadAsArtifact` which uses the same `LegacyType` for the target access type. This keeps backward compatibility with OCM descriptors written by older tooling; the runtime scheme resolves both legacy and modern type strings to the same `OCIImage` struct.

bindings/go/transfer/internal/oci.go
jakobmoellerdev: Good catch — fixed. The inline `targetRepoBaseURL + "/" + referenceName` produced a leading slash when `BaseUrl` is empty. Changed to `staticReferenceName(referenceName)(targetRepoBaseURL)` to match the guard logic in `ociUploadAsArtifact`.

matthiasbruns (dismissed):

bindings/go/transfer/internal/builder.go
```diff
@@ -61,6 +61,13 @@ func NewDefaultBuilder(
 		CredentialProvider: credentialProvider,
 	}
 
+	// Streaming OCI-to-OCI transfer transformer
+	ociTransferOCIArtifact := &ocitransformer.TransferOCIArtifact{
+		Scheme:             transformerScheme,
+		Repository:         resourceRepo,
+		CredentialProvider: credentialProvider,
+	}
```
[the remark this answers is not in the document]
jakobmoellerdev: this is acceptable because its anyhow only for OCI

fabianburth (approved):

Skarlso (approved):

matthiasbruns (approved):

PR #2611, opened 2026-05-22
feat: add ocm skill pull/push commands for AI skill catalogue distribution

jakobmoellerdev:
## Summary

- Adds `ocm skill push <dir>` — scans a skills directory for `SKILL.md` files, emits a component constructor YAML (`ai.skill/v1` resources) ready for `ocm add component-version`
- Adds `ocm skill pull <ref>` — resolves an OCM component version, filters `ai.skill/v1` resources, writes each skill to `~/.claude/skills/<name>/SKILL.md`
- Path traversal protection (`validateSkillName`, prefix check, symlink rejection)
- Idempotent installs (`O_TRUNC` instead of `O_APPEND`)

## Motivation

Claude Code skills are Markdown files that teach the AI how to work with a codebase. This PR enables teams to package, version, sign, and distribute skills via OCI registries exactly like container images — one OCM component = one skill catalogue release.

## Files changed

| Path | Description |
|------|-------------|
| `cli/cmd/cmd.go` | Registers `skill` subcommand |
| `cli/cmd/skill/cmd.go` | Root `skill` command |
| `cli/cmd/skill/pull/cmd.go` | Pull implementation |
| `cli/cmd/skill/push/cmd.go` | Push implementation |
| `cli/cmd/skill/pull/cmd_test.go` | 8 unit tests |
| `cli/cmd/skill/push/cmd_test.go` | 9 unit tests (incl. round-trip) |
| `cli/integration/skill_integration_test.go` | 3 integration tests over real OCI registry |
| `docs/blog/shipping-ai-skills-with-ocm.md` | Tutorial blog post |

## Test plan

- [ ] `go test ./cli/cmd/skill/...` — all 17 unit tests pass
- [ ] Integration tests in `cli/integration/skill_integration_test.go` pass (requires Docker for OCI registry)
- [ ] `TestPushRoundTrip` validates full CTF round-trip (push → add cv → pull)
- [ ] Path traversal test `TestPullRejectsPathTraversalInResourceName` rejects `../evil` resource names
- [ ] Repeated pulls produce identical file content (idempotency)

### DIFF
--- a/cli/cmd/skill/pull/cmd.go
+++ b/cli/cmd/skill/pull/cmd.go
@@ -1,6 +1,7 @@
 package pull
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"io"
@@ -21,9 +22,18 @@ import (
 
 const (
 	skillResourceType = "ai.skill/v1"
-	defaultSkillDir   = ".claude/skills"
-	flagSkillName     = "skill"
-	flagOutput        = "output"
+
+	// Default install directories per coding agent.
+	claudeSkillDir = ".claude/skills"
+	codexSkillDir  = ".agents/skills"
+
+	flagSkillName = "skill"
+	flagOutput    = "output"
+	flagTarget    = "target"
+
+	targetClaude = "claude"
+	targetCodex  = "codex"
+	targetAll    = "all"
 )
 
 func New() *cobra.Command {
@@ -33,12 +43,22 @@ func New() *cobra.Command {
 		Long: `Pull one or all AI skills from an OCM component version that packages skills as resources with type ai.skill/v1.
 
 When --skill is given, only that resource is downloaded. Without --skill, all ai.skill/v1 resources are downloaded.
-Skills are written to ~/.claude/skills/<skill-name>/SKILL.md by default.`,
-		Example: `  # Pull a single skill
+
+By default skills are installed for Claude Code (--target claude):
+  ~/.claude/skills/<skill-name>/SKILL.md
+
+Use --target codex to install for OpenAI Codex CLI instead:
+  ~/.agents/skills/<skill-name>/SKILL.md
+
+Use --target all to install for both agents simultaneously.`,
+		Example: `  # Pull a single skill into Claude Code (default)
   ocm skill pull ./catalogue//jakob.io/ai-skill-catalogue:1.0.0 --skill ocm-guide
 
-  # Pull all skills from catalogue to default location
-  ocm skill pull ./catalogue//jakob.io/ai-skill-catalogue:1.0.0
+  # Pull all skills into OpenAI Codex CLI
+  ocm skill pull ./catalogue//jakob.io/ai-skill-catalogue:1.0.0 --target codex
+
+  # Pull all skills into both Claude Code and Codex
+  ocm skill pull ./catalogue//jakob.io/ai-skill-catalogue:1.0.0 --target all
 
   # Pull a skill to a custom path
   ocm skill pull ./catalogue//jakob.io/ai-skill-catalogue:1.0.0 --skill ocm-guide --output /tmp/ocm-guide.md`,
@@ -49,6 +69,7 @@ Skills are written to ~/.claude/skills/<skill-name>/SKILL.md by default.`,
 
 	cmd.Flags().String(flagSkillName, "", "name of skill resource to pull (pulls all ai.skill/v1 resources when omitted)")
 	cmd.Flags().String(flagOutput, "", "output path for the skill file (only valid with --skill)")
+	cmd.Flags().String(flagTarget, targetClaude, `coding agent to install skills for: "claude" (~/.claude/skills/), "codex" (~/.agents/skills/), or "all"`)
 
 	return cmd
 }
@@ -69,8 +90,22 @@ func pullSkill(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("getting output flag failed: %w", err)
 	}
 
+	target, err := cmd.Flags().GetString(flagTarget)
+	if err != nil {
+		return fmt.Errorf("getting target flag failed: %w", err)
+	}
+
 	if output != "" && skillName == "" {
 		return fmt.Errorf("--output requires --skill to be set")
+	}
+	if output != "" && target != targetClaude {
+		return fmt.Errorf("--output cannot be combined with --target")
+	}
+
+	switch target {
+	case targetClaude, targetCodex, targetAll:
+	default:
+		return fmt.Errorf("unknown --target %q: must be %q, %q, or %q", target, targetClaude, targetCodex, targetAll)
 	}
 
 	ref, err := compref.Parse(args[0])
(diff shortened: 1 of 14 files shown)


morri-son:
skills are not only used in Claude. Would we like to offer additional tools like codex, cursor, gemini, aider and use their specific paths to pull/push skills, using Claude Code as default?

jakobmoellerdev:
did you read the blog post? I think I also added support for codex, open for others.

morri-son:
@jakobmoellerdev 
>did you read the blog post? I think I also added support for codex, open for others.

ah, not when I wrote the comment. I just used what was in the PR description. Went through all files in detail now. Code has Claude and codex, blog mentions cursor and continue on top.

cli/cmd/skill/push/cmd_test.go
```diff
@@ -0,0 +1,283 @@
...
+	var spec map[string]any
+	require.NoError(t, yaml.Unmarshal(out.Bytes(), &spec))
+
+	resources := spec["components"].([]any)[0].(map[string]any)["resources"].([]any)
+	input := resources[0].(map[string]any)["input"].(map[string]any)
+	path := input["path"].(string)
+	require.True(t, strings.HasPrefix(path, "/"), "resource path must be absolute, got %q", path)
+}
+
+func TestPushRoundTrip(t *testing.T) {
```
morri-son, line 215: This fails for me on MacOS, most-likely because how MacOS treats /tmp. It's a symlink to /private/tmp, and t.TempDir() / mktemp -d returns paths under var/folders/... which symlinks to /private/var/folders/... Why doesn't it simply do it like Linux ?! 😂. The error is not wrapped, so you have to scroll too the right, to see the interesting part

```
...
failed to open path "../../../../../../../private/var/folders/s1/9ybhw_m510zgf1679xh4y9pw0000gn/T/TestPushRoundTrip3859444100/001/skills/test-skill/SKILL.md" in root "/var/folders/s1/9ybhw_m510zgf1679xh4y9pw0000gn/T/TestPushRoundTrip3859444100/001": openat ../../../../../../../private/var/folders/s1/9ybhw_m510zgf1679xh4y9pw0000gn/T/TestPushRoundTrip3859444100/001/skills/test-skill/SKILL.md: path escapes from parent
--- FAIL: TestPushRoundTrip (0.01s)
```

matthiasbruns:
wouldn't this be nice once we have cobra plugins up and running to be part of the community repo instead?

jakobmoellerdev:
will we actually ship cobra plugins? its currently permanently deprioritized due to lack of demand

matthiasbruns:
> will we actually ship cobra plugins? its currently permanently deprioritized due to lack of demand

I was looking for this for my tui community pr tbh, but I get it

cli/cmd/skill/pull/cmd.go
```diff
@@ -0,0 +1,297 @@
...
+	"ocm.software/open-component-model/bindings/go/runtime"
+	"ocm.software/open-component-model/cli/cmd/download/shared"
+	"ocm.software/open-component-model/cli/internal/repository/ocm"
+)
+
+const (
+	skillResourceType = "ai.skill/v1"
+
+	// Default install directories per coding agent.
+	claudeSkillDir = ".claude/skills"
```
matthiasbruns, line 27: I dont know about hardcoding this - if new agents land, we would have to update this everytime

maybe we need something like this

```yml
- type: skill.config.ocm.software/v1
  agents:
    - name: claude
      dir: ~/.claude/skills
    - name: cursor
      dir: ~/.config/cursor/rules
```

cli/cmd/skill/push/cmd.go
```diff
@@ -0,0 +1,214 @@
...
+	flagOutputFile = "output"
+
+	skillResourceType = "ai.skill/v1"
+	skillFileName     = "SKILL.md"
+)
+
+func New() *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "push <skills-dir> --component <name> --version <v>",
+		Short: "Generate an OCM component-constructor for a directory of AI skills",
```
matthiasbruns, line 28: this is not a "push" command is it? when I saw push, I though we would do a construct + add cv transparently

matthiasbruns (commented):
generally needs some love an refactor - cmd.go are painfully long - would move things into `cmd/internal/skills` to reuse logic and clean up of the cmd files.

matthiasbruns:
Well - I still do not see that this is a first citizen command - as discussed yesterday, @jneisener had an alternative idea to use what we already have in ocm to distribute skills instead of adding a new command

jneisener:
> Well - I still do not see that this is a first citizen command - as discussed yesterday, @jneisener had an alternative idea to use what we already have in ocm to distribute skills instead of adding a new command

Right, so my pitch was to ship a skill that essentially implements `ocm skill push` on the client side by instrumenting the ocm commands that we already have. 

Main question on that approach would be: How is this bootstrapped? Because if you need a skill to pull your skills, where do you get the initial skill from? 

My idea was that we instead implement something like `ocm init`, which could serve multiple functions, e.g. installing plugins, one of which could be to download the skill to work with skills. This way we could support skills almost out-of-the-box, but without having to build a feature that is only used/useful for AI. WDYT?

matthiasbruns:
> My idea was that we instead implement something like `ocm init`, which could serve multiple functions, e.g. installing plugins, one of which could be to download the skill to work with skills. This way we could support skills almost out-of-the-box, but without having to build a feature that is only used/useful for AI. WDYT?

what about how its done in other projects? there are skill marketplaces, and skills can be shipped in the repo itself and you tell your ai to install it?

jneisener:
> > My idea was that we instead implement something like `ocm init`, which could serve multiple functions, e.g. installing plugins, one of which could be to download the skill to work with skills. This way we could support skills almost out-of-the-box, but without having to build a feature that is only used/useful for AI. WDYT?
> 
> what about how its done in other projects? there are skill marketplaces, and skills can be shipped in the repo itself and you tell your ai to install it?

Well sure, but transferring things is kind of our thing. If they need a third party marketplace to begin with, whats the pitch to use ocm for skill transport over that marketplace?

matthiasbruns:
> 
> Well sure, but transferring things is kind of our thing. If they need a third party marketplace to begin with, whats the pitch to use ocm for skill transport over that marketplace?

My bad - I was only talking about an ocm skill that explains ocm including how to transfer skills with localBlob  for example - the interesting part that we need to figure out though is how to "comsume" such skills and how to put them into the right locations on the users machine after they have been downloaded again

jakobmoellerdev:
Guys whats the status here?

matthiasbruns:
> Guys whats the status here?

to not do it like in this PR - maybe we should create an issue to actually think about this more or refactor this pr

Skarlso:
Agreed. This is not the way to do this right now.

However, we do agree that something needs to be done. 

To this, I created the following EPIC: https://github.com/open-component-model/ocm-project/issues/1270

And I created a sub-issue for THIS specific pr and purpose: https://github.com/open-component-model/ocm-project/issues/1271

To be refined in our next refinement. :) Thanks Jakob 🙇‍♀️

PR #2616, opened 2026-05-22
feat!: migrate transfer and controller bindings to runtime.Typed credentials (gate 7)

jakobmoellerdev:
#### What this PR does / why we need it

Gate 7 of the typed credentials migration (ocm-project#1047).

Migrates `bindings/go/transfer` and `kubernetes/controller` to use `runtime.Typed` credentials instead of `map[string]string`.

**bindings/go/transfer:**
- Bump `blob→v0.0.13`, `credentials→v0.0.12`, `oci→v0.0.43`, `repository→v0.0.9`
- Update OCI identity import path: `oci/spec/credentials/identity/v1` → `oci/spec/identity/v1`

**kubernetes/controller:**
- `resolveResourceCredentials` and `VerifyResource` return `runtime.Typed` instead of `map[string]string`
- Replace `map[string]string` RSA credential literals with `rsacredentialsv1.RSACredentials` structs
- Update import aliases for `oci/spec/credentials` and `oci/spec/identity/v1`
- `GetComponentVersionRepository` call sites updated to pass `runtime.Typed` credentials

#### Which issue(s) this PR fixes

Contributes:
- https://github.com/open-component-model/ocm-project/issues/1047
- https://github.com/open-component-model/ocm-project/issues/1055
- https://github.com/open-component-model/ocm-project/issues/1056
- https://github.com/open-component-model/ocm-project/issues/1057

#### Binding release order

| Gate | PRs | Modules | Status |
|------|-----|---------|--------|
| 1 | ✅ #2580 | blob, signing, rsa | merged |
| 2 | ✅ #2586 | repository, sigstore | merged |
| 3 | ✅ #2594 | oci | merged |
| 4 | ✅ #2598 | constructor | merged |
| 5 | ✅ #2602 | plugin, input/dir, input/file, input/utf8 | merged |
| 6 | ✅ #2612 | helm | merged |
| 7 | **this PR** | transfer, controller | 👈 |
| 8 | PR 8 | cli | — |

PR #2618, opened 2026-05-22
fix(oci): handle *runtime.Raw access type in ProcessResourceDigest

jakobmoellerdev:
## Problem

\`Repository.ProcessResourceDigest\` type-switches on \`res.Access\` and only handles \`*v2.LocalBlob\` and \`*accessv1.OCIImage\`. When the access field arrives as \`*runtime.Raw\` (not yet decoded from JSON — as is the case when resources come from descriptors loaded off the wire), it hits the \`default\` case and returns:

\`\`\`
unsupported resource access type: *runtime.Raw
\`\`\`

## Fix

Add a \`*runtime.Raw\` case that converts via the scheme before re-dispatching — the same pattern already used for \`*v2.LocalBlob.GlobalAccess\`:

\`\`\`go
case *runtime.Raw:
    converted, err := repo.scheme.NewObject(typed.GetType())
    ...
    res.Access = converted
    return repo.ProcessResourceDigest(ctx, res)
\`\`\`

This makes the inner \`Repository\` transparently handle any registered type regardless of whether it has been decoded yet, so callers don't need to pre-convert before calling.

## Context

Discovered during gate 7 of the typed-credentials migration (#2616). The outer \`resource_repository.ProcessResourceDigest\` wrapper had to paper over this by converting \`resource.Access\` before delegating — this PR fixes the root cause in the right place.


jakobmoellerdev:
still unsure if this is the proper fix for the credential unblock, need to investigate further

matthiasbruns (approved):

piotrjanik (approved):

fabianburth (changes requested):
that's not the correct design

matthiasbruns:
https://github.com/open-component-model/open-component-model/pull/2616/changes/f2e06d1ee808445d33728005d26dc95b23387325#diff-4cc5d5a0a0751c6864122e7edfc77b534b0a43f2ca6eedf9c8bd74e6d22bac79
https://github.com/open-component-model/open-component-model/pull/2616/changes/fdb6af920206197d411bde6333e5d2b8b604e5cc

this might be the root of the problem

`epo.ProcessResourceDigest` assumes that we have anything but `Raw`

```go
func (repo *Repository) ProcessResourceDigest(ctx context.Context, res *descriptor.Resource) (_ *descriptor.Resource, err error) {
	ctx = slogcontext.NewCtx(ctx, repo.logger)
	done := log.Operation(ctx, "process resource digest",
		log.IdentityLogAttr("resource", res.ToIdentity()))
	defer func() {
		done(err)
	}()
	res = res.DeepCopy()
	switch typed := res.Access.(type) {
	case *v2.LocalBlob:
		if typed.GlobalAccess == nil {
			return nil, fmt.Errorf("local blob access does not have a global access and cannot be used")
		}
		globalAccess, err := repo.scheme.NewObject(typed.GlobalAccess.GetType())
		if err != nil {
			return nil, fmt.Errorf("error creating typed global blob access with help of scheme: %w", err)
		}
		if err := repo.scheme.Convert(typed.GlobalAccess, globalAccess); err != nil {
			return nil, fmt.Errorf("error converting global blob access: %w", err)
		}
		res.Access = globalAccess
		return repo.ProcessResourceDigest(ctx, res)
	case *accessv1.OCIImage:
		return repo.processOCIImageDigest(ctx, res, typed)
	default:
		return nil, fmt.Errorf("unsupported resource access type: %T", typed)
	}
}
```

matthiasbruns:
https://github.com/open-component-model/open-component-model/pull/2640

PR #2638, opened 2026-05-25
Http config + client + ADR (PoC)

piotrjanik:

Extends the HTTP client configuration type (`http.config.ocm.software/v1alpha1`) with more granular timeout, transport-layer, logging, tracing, and proxy configurations. Adds per-host overrides for HTTP settings and facilitates merging configurations with priority rules.

Key additions include:
- **TimeoutConfig:** granular control of HTTP request, TLS handshake, and connection timeouts.
- **TransportConfig:** retry policies, TLS settings, custom DNS overrides, and proxy support.
- **LoggingConfig:** request/response logging with header/body inclusion options.
- **TracingConfig:** OpenTelemetry distributed tracing support.
- **HostConfig:** per-host override capabilities.

Updates unit tests to cover all new configuration fields and scenarios.

Fixes: https://github.com/open-component-model/ocm-project/issues/1731

##### Verification


bindings/go/httpclient/doc.go
jakobmoellerdev, line 1: any specific reason this is not part of runtime?

piotrjanik: @jakobmoellerdev That way we have cleaner dependency: 
```
httpclient 
               | -> configuration
               | -> runtime
```

When httpclient is in runtime, the runtime would depend on configuration. 

That was my thinking, but I don't have any strong opinion here.

jakobmoellerdev: understood and agreed. can we rename the package to `http` instead?

PR #2647, opened 2026-05-27
fix(cli): eliminate flicker in transfer progress renderer

jakobmoellerdev:
## Problem

`ocm transfer component-version` flickers visibly during transfer. Two independent causes:

**1. Dual render paths racing each other**

`HandleEvent()` called `renderLocked()` synchronously on every node state change, while the animation goroutine independently fires `renderLocked()` every 80ms. Rapid node completions (e.g. 3 nodes finishing within 5ms) triggered 3 full `clearLines() + redraw` cycles in quick succession, colliding with the background ticker and producing visible cursor jumping.

**2. Partial-frame writes**

Each render frame was written to the terminal as multiple separate `fmt.Fprint` calls — clear lines, then header, then each event line, then the progress bar. The terminal could paint partial frames between syscalls, causing a visible flash.

## Fix

**Event ingestion decoupled from rendering** (`bar_visualizer.go`): `HandleEvent()` now only mutates state. The animation ticker is the sole render driver — at most 10 redraws/sec regardless of event rate. The 100ms lag between a node completing and it appearing on screen is imperceptible.

**Single-write frames** (`bar_visualizer.go`): The full frame (cursor movements, header, events, bar) is assembled into a `strings.Builder` and flushed with a single `io.WriteString` call. The terminal sees a complete frame or nothing.

**Animation interval** (`animation.go`): 80ms → 100ms. 12.5 redraws/sec was imperceptible over 10 and added unnecessary syscall pressure.

## Test plan

- [ ] `go test ./cli/internal/render/progress/bar/...` passes
- [ ] Run `ocm transfer component-version <source> <target>` against a registry with several components — no flicker during node completions, smooth spinner animation


piotrjanik (approved):
Oh, it really helped. Looks much nicer :D GJ!

matthiasbruns (approved):
lgtm

Skarlso (approved):

PR #2648, opened 2026-05-27
docs(adr): ADR-0024 — optional component-descriptors path prefix

jakobmoellerdev:
## Summary

- Documents the friction caused by the mandatory `component-descriptors` path segment in OCM component version references
- Proposes three options; recommends **Option A** (configurable, opt-in prefix — no default change)
- Surfaces four open angles for community discussion before implementation begins

## Motivation

Every OCM component version is stored at a path like:

```
ghcr.io/my-org/platform/component-descriptors/ocm.software/myapp:v1.2.0
```

The `component-descriptors` segment was introduced as a namespace guardrail. It has become a source of friction: teams expect to address a component at its natural OCI path, the extra segment surprises new adopters, and native OCI tooling has to be explicitly told about it.

The Go bindings parser already handles the no-prefix form today (`ghcr.io/my-org//ocm.software/myapp:v1.2.0`). The restriction is entirely on the write path and CTF lister.

## Decision (proposed)

**Option A — opt-in configurable prefix.** Add `WithComponentDescriptorPath("")` option to the OCI URL resolver and CTF store. Make the CTF lister prefix-agnostic. Default behaviour unchanged.

No parser, validator, or annotation changes needed — all three already strip the prefix defensively.

## Open angles (seeking input)

1. **`Ref.String()` for empty prefix** — `//` double-slash (unambiguous, zero parser change) vs single `/` (cleaner but needs parser awareness). What does the community prefer?

2. **Spec wording** — should the spec PR relax to `SHOULD`, introduce a named configuration option, or set a deprecation sunset? This ADR does not prescribe.

3. **Migration tooling** — should `ocm transfer` gain a flag to rewrite references to no-prefix form during transfer? Out of scope here; tracking issue welcome.

4. **CTF compatibility fixtures** — new no-prefix fixture files should be added alongside existing ones (not replacing them).

## Prerequisite

A spec PR on `open-component-model/ocm-spec` must relax the `component-descriptors` requirement from MUST to SHOULD (or equivalent) before the code change merges.

## Test plan

- [ ] Review ADR text and open angles
- [ ] Agree on spec PR scope with OCM TSC
- [ ] Confirm Option A is the right first step or redirect to an alternative
- [ ] Follow-up issue for `ocm transfer` migration flag if desired


docs/adr/0024_optional_component_descriptors_prefix.md
```diff
@@ -0,0 +1,317 @@
...
+`WithComponentDescriptorPath("")` option when constructing a resolver or CTF store.
+The CTF lister is made prefix-agnostic so it can read archives written with either
+convention. This is a stepping stone: once the spec has been updated and ecosystem
+tooling has adapted, a later ADR can flip the default (Option B).
+
+### Open angles requiring community discussion
+
+The following questions are deliberately left open in this ADR:
+
+1. **`Ref.String()` serialisation for empty prefix.** When `Prefix == ""`, the current
```
matthiasbruns, line 102: I always found the `//` pretty weird. One thing here - does external tooling work with the `//` at all? I'd opt for teaching the parser to handle single `/`

docs/adr/0024_optional_component_descriptors_prefix.md
```diff
@@ -0,0 +1,317 @@
...
+```go
+// Existing — unchanged behaviour:
+resolver, _ := url.New(url.WithBaseURL("ghcr.io"), url.WithSubPath("my-org"))
+// resolver.BasePath() == "ghcr.io/my-org/component-descriptors"
+
+// New opt-in:
+resolver, _ := url.New(
+    url.WithBaseURL("ghcr.io"),
+    url.WithSubPath("my-org"),
+    url.WithComponentDescriptorPath(""),
```
matthiasbruns, line 172: how are we configuring this in the constructor? I guess this should be a per-cv setting right?

docs/adr/0024_optional_component_descriptors_prefix.md
jakobmoellerdev: // is a replacement for `/component-descriptors/`. if we allow `/` it would not be the same. External tooling cannot deal with // as it is just a shortcut that OCM CLI knows

matthiasbruns: I know but isn't the goal to make this work with external tooling, or did I get this wrong?
If so, what options do we actually have here?

jakobmoellerdev: There are 2 main options imho, either we follow URL specification or not. If we do follow it, then // is only valid as part of the scheme demoninator I believe

docs/adr/0024_optional_component_descriptors_prefix.md
jakobmoellerdev: I think if this should be configurable at all, it would be per repository and not per component version, as the prefix doesnt change across many CVs in the same repo. I think realistically the empty path should probably be the default

docs/adr/0024_optional_component_descriptors_prefix.md
matthiasbruns: If we want to support external tooling, we should aim for following the url spec

morri-son: I also vote for teaching the parser how to deal with /

docs/adr/0024_optional_component_descriptors_prefix.md
```diff
@@ -0,0 +1,317 @@
...
+The following questions are deliberately left open in this ADR:
+
+1. **`Ref.String()` serialisation for empty prefix.** When `Prefix == ""`, the current
+   serialiser emits `//` (double-slash). This is unambiguous and already parsed
+   correctly, but aesthetically awkward. The alternative is to emit a single `/` and
+   teach the parser to treat a component name appearing immediately after the repository
+   path (without any known prefix) as the no-prefix case. The double-slash form is
+   proposed as the interim canonical representation, pending a UX review.
+
+2. **Spec relaxation scope.** The spec PR should decide whether `component-descriptors`
```
morri-son, line 109: I opt for "deprecated with a sunset timeline". Cut the old stuff :)

docs/adr/0024_optional_component_descriptors_prefix.md
```diff
@@ -0,0 +1,317 @@
...
+   becomes a `SHOULD` (recommended but not required), a named configuration option, or
+   is deprecated with a sunset timeline. This ADR does not prescribe which.
+
+3. **Migration tooling.** Should the `ocm transfer` command grow a flag to rewrite
```
morri-son, line 113: yes, something like --no-prefix would help users to migrate their stuff. Using Oras or skopeo would work too if scripted and explained, but from and end user perspective I prefer to use the same tool that introduced the prefix.

morri-son (approved):

docs/adr/0024_optional_component_descriptors_prefix.md
jakobmoellerdev: we cannot make external tools teach the // since they are not under our maintenance and will not accept this "special case"

Skarlso (approved):
Nice! This will be interesting for tools like ocm-kit. :)

PR #2657, opened 2026-05-27
fix(transfer): support Docker manifest media types in OCI-compliant manifest detection

jakobmoellerdev:

Docker manifest v2 (`application/vnd.docker.distribution.manifest.v2+json`) and Docker manifest list (`application/vnd.docker.distribution.manifest.list.v2+json`) were excluded from `isOCICompliantManifest` / `IsOCICompliantMediaType` in both `bindings/go/oci/internal/introspection` and `bindings/go/transfer/internal`. This caused LocalBlob resources with Docker manifest media types to silently bypass the OCI artifact upload path when using `UploadAsOciArtifact` mode during transfer — they were stored as plain local blobs instead.

The root cause was a TODO in both functions noting Docker manifests were not yet supported. This PR resolves that TODO by adding the two Docker distribution manifest media type constants (mirroring `oras-go/v2`'s internal definitions, which are not importable) and extending the switch statements to accept them.

Internal report: standard `docker manifest` transfer broken in v2 transfer path (`UploadAsOciArtifact` mode).

##### How to test the changes

```bash
# Unit tests (fast)
cd bindings/go/transfer && go test ./internal/... -run "TestIsOCICompliant|TestBuildGraphDefinition_Docker" -v
cd bindings/go/oci && go test ./internal/introspection/... -v

# Integration test (requires Docker)
cd bindings/go/transfer/integration && go test -v -run Test_Integration_TransferDockerManifestLocalBlob_CTFToOCI
```

##### Verification

### DIFF
--- a/bindings/go/oci/internal/introspection/manifest.go
+++ b/bindings/go/oci/internal/introspection/manifest.go
@@ -4,6 +4,14 @@ import (
 	ociImageSpecV1 "github.com/opencontainers/image-spec/specs-go/v1"
 )
 
+// Docker manifest media types as defined by the Docker distribution spec.
+// oras-go/v2 defines these in internal/docker/mediatype.go (not importable),
+// so we redeclare them here.
+const (
+	MediaTypeDockerManifest     = "application/vnd.docker.distribution.manifest.v2+json"
+	MediaTypeDockerManifestList = "application/vnd.docker.distribution.manifest.list.v2+json"
+)
+
 // IsOCICompliantManifest checks if a descriptor describes a manifest that is recognizable by OCI.
 func IsOCICompliantManifest(desc ociImageSpecV1.Descriptor) bool {
 	return IsOCICompliantMediaType(desc.MediaType)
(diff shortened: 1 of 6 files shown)


matthiasbruns (commented):
conflicts - it works  - good job

bindings/go/oci/internal/introspection/manifest.go
```diff
@@ -4,6 +4,14 @@ import (
...
 )
 
+// Docker manifest media types as defined by the Docker distribution spec.
+// oras-go/v2 defines these in internal/docker/mediatype.go (not importable),
+// so we redeclare them here.
+const (
+	MediaTypeDockerManifest     = "application/vnd.docker.distribution.manifest.v2+json"
+	MediaTypeDockerManifestList = "application/vnd.docker.distribution.manifest.list.v2+json"
+)
+
```
fabianburth, line 14: we were quite sure, we do not want to offer this as public API?

fabianburth (approved):

Skarlso (approved):

bindings/go/oci/internal/introspection/manifest.go
jakobmoellerdev: I really dont want people to depend on this here but instead use moby for this. Its just that we ourselves dont want to depend on moby. I think if someone should export it, its probably oras

PR #2661, opened 2026-05-27
docs: Add `Task` to prerequisites for building

jneisener:
#### What this PR does / why we need it
`Task` is missing in the list of perquisites to build from source. Its listed in the overall perquisites (https://ocm.software/community/onboarding/) but so is `Go`, so it would make sense to repeat it as well.

#### Which issue(s) this PR fixes
List of prerequisites is missing `Task`

##### How to test the changes

Its a one line change in the documentation 

##### Verification

- [ ] ~I have added/updated tests for my changes (see [Test Requirements](../CONTRIBUTING.md#test-requirements))~
- [ ] ~Tests pass locally (`task test` and `task test/integration` if applicable)~
- [ ] I~f touching multiple modules, `go work` is enabled (see `go.work`)~
- [x] My changes do not decrease test coverage
- [ ] ~I have tested the changes locally by running `ocm`~

PR #2662, opened 2026-05-27
chore: split PR title validation and drop blackduck

matthiasbruns:

This PR splits `pull-request.yaml` into two workflows:

| Workflow | Trigger | Permissions | What it does |
|---|---|---|---|
| `validate-pr-title.yaml` | `pull_request` | `contents: read` | Validates PR title against Conventional Commit regex, calls `core.setFailed()` on mismatch |
| `pull-request.yaml` (renamed "Label Pull Request") | `pull_request_target` | `pull-requests: write` | Applies labels only — retains `pull_request_target` because `addLabels()` requires write access unavailable on fork PRs under the plain `pull_request` trigger |

This PR also drops blackduck.


.github/workflows/validate-pr-title.yaml
```diff
@@ -0,0 +1,49 @@
...
+on:
+  pull_request:
+    types:
+      - opened
+      - edited
+      - synchronize
+      - reopened
+
+permissions:
+  contents: read
```
jakobmoellerdev, line 14: why do we even need contents read here?

.github/workflows/pull-request.yaml
jakobmoellerdev, line 14: same here, do we really need content read?

frewilhelm: I believe this is to set a baseline of least privilege permissions. This actually does not make much sense because we set permissions for every job later but GitHub Security will flag it because

> Risk: High (vulnerable to malicious code additions)
This check determines whether the project's automated workflows tokens follow the
principle of least privilege. This is important because attackers may use a
compromised token with write access to, for example, push malicious code into the
project.
It is currently limited to repositories hosted on GitHub, and does not support
other source hosting repositories (i.e., Forges).
The highest score is awarded when the permissions definitions in each workflow's
yaml file are set as read-only at the
[top level](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#permissions)
and the required write permissions are declared at the
[run-level](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#jobsjob_idpermissions).

jakobmoellerdev: wouldnt it make sense to explicitly set the baseline as no permissions?

matthiasbruns: I removed the permission - no need for them

frewilhelm, line 38: This list does not meet the requirements for the job `verify-labels` I believe. For example this https://github.com/open-component-model/open-component-model/pull/2661 was automatically labeled `area/documentation` but failed later on `verify-labels` https://github.com/open-component-model/open-component-model/actions/runs/26522146327/job/78115964846. 

Can you either adjust `TYPE_TO_LABEL` or `any_of` (when the `verify-labels` workflow is called), so this works

frewilhelm, line 64: What is this about? This looks like ocm v1

frewilhelm: > wouldnt it make sense to explicitly set the baseline as no permissions?

Yes, you can do this too. It is either `permissions: {}` or `contents: read` i believe

matthiasbruns: I moved out the inline script - it made me crazy, will ping you once the pr is ready again

.github/workflows/validate-pr-title.yaml
```diff
@@ -0,0 +1,38 @@
...
+jobs:
+  validate-title:
+    name: Validate PR Title (Conventional Commit)
+    runs-on: ubuntu-24.04-arm
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
+        with:
+          ref: ${{ github.base_ref }}
+          sparse-checkout: |
+            .github/scripts/label-pr.js
```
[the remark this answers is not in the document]
matthiasbruns: removed

.github/workflows/pull-request.yaml
matthiasbruns: thats the original script - https://github.com/open-component-model/open-component-model/blob/main/.github/workflows/pull-request.yaml#L125

matthiasbruns:
@coderabbitai fill review

.github/workflows/pull-request.yaml
frewilhelm: original from where? :D

The script points to directories from ocm v1 if I am not mistaken. For example `components/` or `resources/`. I don't think we need that job and we should delete it.

WDYT?

frewilhelm (commented):
I saw you requested another review after coderabbitai posted some review comments, e.g. 

> With all permissions denied, actions/checkout cannot authenticate against the GitHub API to clone the repository. The job needs at minimum contents: read for sparse-checkout to work, especially in private repositories. The current setting may happen to work in public repos (anonymous HTTPS clone), but it is fragile and undocumented behaviour.

which I think is valid.

What about them?

matthiasbruns:
> I saw you requested another review after coderabbitai posted some review comments, e.g.
> 
> > With all permissions denied, actions/checkout cannot authenticate against the GitHub API to clone the repository. The job needs at minimum contents: read for sparse-checkout to work, especially in private repositories. The current setting may happen to work in public repos (anonymous HTTPS clone), but it is fragile and undocumented behaviour.
> 
> which I think is valid.
> 
> What about them?

wow didn't push my changes...

.github/workflows/validate-pr-title.yaml
jakobmoellerdev: this wasnt really addressed so im reopening here

PR #2665, opened 2026-05-28
feat(transfer): add multi-version discovery to transfer cv

jakobmoellerdev:
## Summary

- `ocm transfer cv` now supports omitting the version from the source reference, discovering and transferring all matching versions — mirroring `get cv` behaviour
- New `--semver-constraint` flag filters which versions are transferred (e.g. `">= 1.0.0, < 2.0.0"`)
- New `--latest` flag transfers only the newest matching version
- Both flags are no-ops (with a warning) when an explicit version is present in the reference
- Reference docs updated to document the new flags and examples

## Test plan

- [ ] Unit tests: `go test ./cli/cmd/transfer/component-version/...` — covers all-versions, semver-constraint, latest-only, exact-version-unchanged, warn-on-ignored-flags
- [ ] Integration tests: `go test ./cli/integration/... -run Test_Integration_TransferComponentVersion` — AllVersions, SemverConstraint, LatestOnly against a real OCI registry container
- [ ] Build: `go build ./cli/...`


cli/integration/transfer_component_version_multi_integration_test.go
```diff
@@ -0,0 +1,244 @@
...
+		urlresolver.WithPlainHTTP(true),
+		urlresolver.WithBaseClient(client),
+	)
+	r.NoError(err)
+	targetRepo, err := oci.NewRepository(oci.WithResolver(resolver), oci.WithTempDir(t.TempDir()))
+	r.NoError(err)
+
+	for _, v := range versions {
+		verifyVersionPresent(t, ctx, targetRepo, componentName, v)
+	}
```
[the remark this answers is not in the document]
jakobmoellerdev: this is just a test so not really a big issue.

matthiasbruns (commented):
Great addition, just one thought:

If one transfer fails, how would the system react? Would it continue with the others in parallel or stop at this point? How can a user recover from this?

jakobmoellerdev:
> Great addition, just one thought:
> 
> If one transfer fails, how would the system react? Would it continue with the others in parallel or stop at this point? How can a user recover from this?

Good question. If the policy is set to override it should be fine with that. Also it should cancel operations independently but finish each node

cli/cmd/transfer/component-version/cmd.go
```diff
@@ -336,13 +354,52 @@ func buildGraphDefinitionFromArgs(
...
+		repo, err := repoProvider.GetComponentVersionRepositoryForComponent(ctx, fromSpec.Component, "")
+		if err != nil {
+			return nil, fmt.Errorf("could not access ocm repository: %w", err)
+		}
+		versions, err = ocm.VersionsWithFiltering(ctx, fromSpec.Component, repo, ocm.VersionOptions{
+			SemverConstraint: constraint,
+			LatestOnly:       latestOnly,
+		})
+		if err != nil {
+			return nil, fmt.Errorf("listing component versions failed: %w", err)
```
frewilhelm: Thats exactly the same error that is returned by `ocm.VersionsWithFiltering`:

```
	versions, err := repo.ListComponentVersions(ctx, component)
	if err != nil {
		return nil, fmt.Errorf("listing component versions failed: %w", err)
	}
```

As this function does listing and filtering, I would adjust the error message

cli/cmd/transfer/component-version/cmd.go
```diff
@@ -336,13 +354,52 @@ func buildGraphDefinitionFromArgs(
...
+		}
+		if len(versions) == 0 {
+			return nil, fmt.Errorf("no versions found for component %q matching constraint %q", fromSpec.Component, constraint)
```
frewilhelm: What if the user specified `--latest` and no semver constraint but the cv list returned 0 versions. This error would look odd.

frewilhelm (commented):
overall lgtm. I wonder if we should add some failure-tests as well.

cli/cmd/transfer/component-version/cmd.go
```diff
@@ -30,12 +30,14 @@ import (
...
-	FlagUploadAs      = "upload-as"
-	FlagTransferSpec  = "transfer-spec"
+	FlagDryRun           = "dry-run"
+	FlagOutput           = "output"
+	FlagRecursive        = "recursive"
+	FlagCopyResources    = "copy-resources"
+	FlagUploadAs         = "upload-as"
+	FlagTransferSpec     = "transfer-spec"
+	FlagSemverConstraint = "semver-constraint"
+	FlagLatest           = "latest"
```
matthiasbruns, line 41: do we need both flags? cant the default behaviour be latest if no version is specified instead?

matthiasbruns: or instead of this, just call it "version" which can be anything - a semver, a constraint or a floating tag or latest?

jakobmoellerdev: the default would be to transfer all versions right now, we can also change it but then get would need to change as well if we wanna keep behavior parallel

cli/cmd/transfer/component-version/cmd.go
jakobmoellerdev: why is that? isnt the error correct?

cli/cmd/transfer/component-version/cmd.go
jakobmoellerdev: adjusted

cli/cmd/transfer/component-version/cmd.go
matthiasbruns: ok, then besides the default being no constraint at all, we can still merge the semver constraints & floating tags/latest into the same flag, couldn't we?

cli/cmd/transfer/component-version/cmd.go
frewilhelm: now it looks good. I believe in a previous version we did not differentiate between constraint and latest

cli/cmd/transfer/component-version/cmd.go
jakobmoellerdev: Not if we keep behavior the same across both get and transfer right? 
Im just wondering how you would merge them, happy to hear suggestions!

matthiasbruns: If you can wait a bit (latest EoD) I will dig into that before answering directly

matthiasbruns: Ok so now I get what you meant

Those flags are already present in `get cv` - which I didn't even know existed 😅
Lets not break with the current patterns

cli/cmd/transfer/component-version/cmd.go
```diff
@@ -336,13 +354,52 @@ func buildGraphDefinitionFromArgs(
...
+		versions, err = ocm.VersionsWithFiltering(ctx, fromSpec.Component, repo, ocm.VersionOptions{
+			SemverConstraint: constraint,
+			LatestOnly:       latestOnly,
+		})
+		if err != nil {
+			return nil, fmt.Errorf("listing component versions failed: %w", err)
+		}
+		if len(versions) == 0 {
+			return nil, fmt.Errorf("no versions found for component %q matching constraint %q", fromSpec.Component, constraint)
+		}
```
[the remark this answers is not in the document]
matthiasbruns: still valid?

matthiasbruns (approved):

frewilhelm (approved):

PR #2670, opened 2026-05-28
chore(fix): prevent renovate from using the canonical version for bindings go module versions

Skarlso:

So this was a bit more involved actually... :D

Renovate created this pr: https://github.com/open-component-model/open-component-model/pull/2664

The solution in the end is to tell renovate to NOT use the canonical tag as a tag for the submodules.

##### Verification


jakobmoellerdev:
This will effectively break all monorepo updates. How would you expect that no bumps are the correct expected behavior?

Skarlso:
> This will effectively break all monorepo updates. How would you expect that no bumps are the correct expected behavior?

I don't think it will. 🤔

Skarlso:
Ah, I see. We did have proper sha updates.

Skarlso:
Ahhh I see. I think my first instinct will be right after all. Let me update this.

piotrjanik (dismissed):

Skarlso:
Mmmm.... my test still returns the canonical tag... hold on here.

.github/renovate.json5
```diff
@@ -172,6 +172,21 @@
...
+      ],
+      matchPackageNames: [
+        'ocm.software/open-component-model/**',
+      ],
+      matchCurrentValue: '/^v0\\.0\\.0-/',
+      matchUpdateTypes: [
+        'major',
+        'minor',
+        'patch',
+      ],
```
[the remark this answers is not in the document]
Skarlso: >In .github/renovate.json5, the gomod rule for ocm.software/open-component-model/** with matchCurrentValue: '/^v0\\.0\\.0-/' sets enabled: false only for matchUpdateTypes: ['major','minor','patch']. If the goal is “don’t touch v0.0.0-* pseudo-versions”, this is too narrow—remove matchUpdateTypes (or replace it with the full set you intend to exclude).


AH! I see. Yes, that was the intent. :D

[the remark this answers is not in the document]
Skarlso: No wait. the rule is not to "don't touch pseudo versions" the rule should be don't use incoming version matching vx.y.z as something to use as an update. Which would mean that our bindings modules with versions like bindings/go/whatever/v0... should still update.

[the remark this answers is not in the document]
Skarlso: Nope, I still think this is wrong. I need to investigate the rule checking more.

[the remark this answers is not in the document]
Skarlso: I think what I actually want is allowedVersions match.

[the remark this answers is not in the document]
Skarlso: I think it's actually the opposite of negation.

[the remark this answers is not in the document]
Skarlso: Exactly that, yes. I believe so. And my CLI testing shows it's working. :)

.github/renovate.json5
```diff
@@ -172,6 +172,16 @@
...
     },
+    {
+      matchManagers: [
+        'gomod',
+      ],
+      matchPackageNames: [
+        'ocm.software/open-component-model/**',
+      ],
+      matchCurrentValue: '/^v0\\.0\\.0-/',
+      allowedVersions: '/^v0\\.0\\.0-/',
```
jakobmoellerdev: shouldnt this also allow matching for suffixed releases?

Skarlso: oh crap

Skarlso: I wonder if I do `/^v0\\.0\\./` would that be enough... 🤔  the problem is that renovate will not make the minor / major jump. But I think that we do that by hand anyways?

Skarlso: FWIW if all modules would have proper tags, we wouldn't have this problem because then I could scope it to a proper module tag. :P

.github/renovate.json5
```diff
@@ -172,6 +172,16 @@
...
       minimumReleaseAge: null,
     },
+    {
+      matchManagers: [
+        'gomod',
+      ],
+      matchPackageNames: [
+        'ocm.software/open-component-model/**',
+      ],
+      matchCurrentValue: '/^v0\\.0\\.0-/',
```
jakobmoellerdev, line 182: isnt this a regex? if so then this doesnt make sense to me X)

Skarlso: Why not? :D It is a regex, yes. Which part doesn't make sense?

jakobmoellerdev: because you match for `0-` only, and not for anything after, so it wouldnt match

Skarlso: Yes, correct. We want the pesudo versions to match but the real tagged values don't. And then the pseudo versions match, and THEN we filter out the canonical version scheme.

At least that's how I understand this config after reading 
https://docs.renovatebot.com/string-pattern-matching/
https://docs.renovatebot.com/configuration-options/#matchcurrentvalue
and
https://docs.renovatebot.com/configuration-options/#allowedversions

Skarlso: And I ran the renovate CLI to make sure the versions don't appear but the other fixed versions do appear.

jakobmoellerdev (approved):

frewilhelm (approved):
lgtm, but i would really love a comment explaining what is happening, so we do not have to wonder in the future

Skarlso:
@frewilhelm I'll add one in my next pr dealing with the release flow. :) Don't want to dismiss the 👍 s :D

PR #2676, opened 2026-05-28
feat: add and transfer ownership referrers (ADR 0016)

### LINKED ISSUE #1025: Transfer Ownership Referrers
**Description**

Ownership referrers must survive `ocm transfer component-version` across all transfer scenarios: OCI-to-OCI, OCI-to-CTF, and CTF-to-OCI. After transfer, consumers should be able to discover which component version owns each resource.

**Implementation notes**

- Add a new transformation type (e.g. `AddOwnershipReferrer`) that runs after the resource upload step and creates the ownership referrer using the uploaded resource's descriptor.

**Done Criteria**
- [ ] Ownership referrers present on target after `ocm transfer` for OCI-to-OCI, OCI-to-CTF, and CTF-to-OCI
- [ ] Referrer has correct `subject.digest` and annotations on target
- [ ] For OCI registry targets, referrer is discoverable via `oras discover`

piotrjanik:

Given an OCI artifact in a registry, there is today no way to find the OCM component
version that shipped it. This PR adds that **asset-to-owner** link (ADR 0016): an opt-in
*ownership referrer* on a resource's manifest, discoverable via the OCI Referrers API
(`oras discover`), that also survives `ocm transfer`. Scope is the **OCI implementation**
only.

Opt-in is **per resource** via `options.ownershipPolicy`, an enum of `Always` / `Never`
(default `Never`). It is a construction-time directive — not persisted to the component
descriptor — so only resources explicitly set to `Always` get a referrer:

```yaml
resources:
  - name: my-image
    type: ociImage
    version: 1.0.0
    options:
      ownershipPolicy: Always   # opt this resource in (default: Never)
    access:
      type: ociArtifact
      imageReference: ghcr.io/acme/my-app:1.0.0
```

The capability is a new optional `OwnershipAwareRepository` interface — implemented by the
OCI repository; a repository that cannot host referrers is a hard error when a resource
opts in.

**How / where it's called**:

- **Creation (`ocm add cv`)** — `constructor.processResource` gates on the opt-in
  (`construct.go:438`/`:750`) → `OwnershipAwareRepository.AddOwnership` →
  `oci.Repository.AddOwnership` (`repository.go:789`), which builds and pushes a
  deterministic, content-addressed referrer (idempotent on re-add).

- **Transfer (`ocm transfer`)** — a referrer that already exists on the source is carried
  to the target during upload, never re-created. The OCI-native by-reference path streams
  directly between registries (`TransferOCIArtifact` transformer, no tar materialization):
  - `DownloadResourceStream` builds an `OCIResourceStream` whose lazy `DiscoverReferrers`
    lists the resource's ownership referrers from the source registry.
  - `UploadResourceStream` runs `oras.ExtendedCopyGraph` (Depth 1); the stream exposes
    those referrers as the root's `Predecessors`, so the copy walks them up from the root
    and uploads `root` + its ownership referrers to the target in a single pass. Signatures
    or other referrers do not ride along.
  - For materialized / by-value uploads (e.g. `AddLocalResource`, layout blobs), the
    existing referrers are instead replayed into the target layout via
    `tar.CopyReferrerRoots` / `tar.InjectReferrersAsSuccessors`.

  Transfer is **copy-only** — it never creates a fresh referrer; creation is the separate
  `Repository.AddOwnership` path, driven on `ocm add` by the constructor through its
  `OwnershipAwareRepository` capability, **not** transfer.

Fixes: https://github.com/open-component-model/ocm-project/issues/1025


bindings/go/oci/repository.go
```diff
@@ -640,7 +654,7 @@ func (repo *Repository) UploadResource(ctx context.Context, res *descriptor.Reso
 
 	res = res.DeepCopy()
 
-	desc, access, err := repo.uploadOCIImage(ctx, res.Access, b)
+	desc, access, err := repo.uploadOCIImage(ctx, res.Access, b, res.Relation == descriptor.LocalRelation)
```
matthiasbruns: so, if we do not own the resource, we will also not copy referrers? doest that make sense?

bindings/go/oci/repository.go
```diff
@@ -716,9 +739,47 @@ func (repo *Repository) uploadOCIImage(ctx context.Context, newAccess runtime.Ty
...
+}
+
+// uploadOwnershipReferrers copies each ownership referrer (ADR 0016) from src to
+// dst. The referrer's subject (main) and any shared empty blobs were already
+// pushed by the main copy, so CopyGraph only adds the referrer manifests. A
+// locally-owned artifact that carries no referrer is logged as a warning and is
+// otherwise a no-op.
+func (repo *Repository) uploadOwnershipReferrers(ctx context.Context, src content.ReadOnlyStorage, dst content.Storage, main ociImageSpecV1.Descriptor, referrers []ociImageSpecV1.Descriptor) error {
+	if len(referrers) == 0 {
+		slogcontext.Log(ctx, slog.LevelWarn, "resource has local relation but no ownership referrer was found in the artifact; nothing to copy", log.DescriptorLogAttr(main))
```
matthiasbruns: if this worth a warning?

bindings/go/oci/repository.go
piotrjanik: That line looks ugly but yes relation local would indicate it. I will probably extract that logic into function so that decision whether ownership is created is in one place.

I am also wondering whether we should update spec..

bindings/go/oci/repository.go
piotrjanik: Well I think so provided that ownership should be created whenever relation is local. In such case missing ownership would mean we're transferring cv which was not created with it.

bindings/go/oci/repository.go
matthiasbruns: don't we want to know who owns this even if we are not the creators?

bindings/go/constructor/runtime/constructor.go
```diff
@@ -72,6 +85,9 @@ type Resource struct {
 	// Relation describes the relation of the resource to the component.
 	// Can be a local or external resource.
 	Relation ResourceRelation `json:"-"`
+	// OwnershipPolicy controls whether an asset-to-owner ownership referrer
+	// (ADR 0016) is created for this resource during construction.
+	OwnershipPolicy OwnershipPolicy `json:"-"`
 	// AccessOrInput defines the access or input information of the resource.
 	// In a component constructor, there is only one access or input information.
 	AccessOrInput `json:"-"`
```
fabianburth: You should settle either for inlined or for the options (currently, there is a diff between spec and runtime, right?)

bindings/go/constructor/runtime/convert_v1.go
```diff
@@ -81,6 +81,25 @@ func ConvertToV1ElementMeta(meta ElementMeta) v1.ElementMeta {
...
+
+// v1OptionsFromOwnershipPolicy builds the optional v1 resource options block from the
+// runtime ownership policy. The default (Never) yields nil so the options block is
+// omitted from the spec rather than emitting an empty object on every resource.
+func v1OptionsFromOwnershipPolicy(p OwnershipPolicy) *v1.ResourceOptions {
+	if p == OwnershipPolicyAlways {
+		return &v1.ResourceOptions{OwnershipPolicy: v1.OwnershipPolicyAlways}
+	}
+	return nil
+}
```
fabianburth: Maybe stick to the naming pattern of the existing convert functions (especially the "v1O" is exhausting to read).

bindings/go/constructor/runtime/convert_v1.go
```diff
@@ -302,9 +323,10 @@ func ConvertToRuntimeComponent(component *v1.Component) Component {
 // for use in a ComponentConstructor.
 func ConvertToRuntimeConstructorResource(resource v1.Resource) Resource {
 	target := Resource{
-		ElementMeta: ConvertFromV1ElementMeta(resource.ElementMeta),
-		Type:        resource.Type,
-		Relation:    ResourceRelation(resource.Relation),
+		ElementMeta:     ConvertFromV1ElementMeta(resource.ElementMeta),
+		Type:            resource.Type,
+		Relation:        ResourceRelation(resource.Relation),
+		OwnershipPolicy: ownershipPolicyFromV1Options(resource.Options),
```
fabianburth: this only looks so weird because the types aren't aligned right?

bindings/go/constructor/construct.go
```diff
@@ -711,7 +779,7 @@ func addColocatedResourceLocalBlob(
 
 	uploaded, err := repo.AddLocalResource(ctx, component, version, descResource, data)
 	if err != nil {
-		return nil, fmt.Errorf("error adding local resource %q based on input type %q as local resource to component %q : %w", resource.ToIdentity(), resource.Input.GetType(), component, err)
+		return nil, fmt.Errorf("error adding colocated local resource %q to component %q: %w", resource.ToIdentity(), component, err)
```
fabianburth: why not mention the input type here?

bindings/go/descriptor/runtime/descriptor.go
```diff
@@ -107,6 +120,10 @@ type Resource struct {
 	// Relation describes the relation of the resource to the component.
 	// Can be a local or external resource.
 	Relation ResourceRelation `json:"-"`
+	// OwnershipPolicy controls whether an asset-to-owner ownership referrer
+	// (ADR 0016) is created for this resource during construction. It is a
+	// construction-time directive and is not persisted to the component descriptor.
+	OwnershipPolicy OwnershipPolicy `json:"-"`
```
fabianburth: We definitely do not want the ownership policy on the descriptor!

bindings/go/oci/repository/resource/resource_repository.go
```diff
@@ -146,6 +150,43 @@ func (p *ResourceRepository) DownloadResource(ctx context.Context, resource *des
...
+	obj, err := p.GetResourceRepositoryScheme().NewObject(t)
+	if err != nil {
+		return nil, fmt.Errorf("error creating new object for type %s: %w", t, err)
+	}
+	if err := p.GetResourceRepositoryScheme().Convert(resource.Access, obj); err != nil {
+		return nil, fmt.Errorf("error converting access to object of type %s: %w", t, err)
+	}
+	resource.Access = obj
+	return resource, nil
+}
```
fabianburth: We don't do this anywhere, and it has only one usage, just inline

cli/cmd/add/component-version/cmd.go
```diff
@@ -460,7 +460,17 @@ func (prov *constructorProvider) GetResourceRepository(ctx context.Context, reso
...
+	// Forward the optional asset-to-owner ownership-referrer capability (ADR 0016)
+	// only when the resolved plugin actually implements it: the in-process OCI
+	// resource repository does, external plugin bridges do not. Wrapping
+	// conditionally keeps the constructor's OwnershipReferrerAttacher type-assert an
+	// accurate gate, so it still warns-and-skips for plugins that cannot host
+	// referrers instead of silently no-opping.
+	if attacher, ok := plugin.(repository.OwnershipReferrerRepository); ok {
+		return &constructorPluginWithOwnershipReferrer{constructorPlugin: base, attacher: attacher}, nil
+	}
+	return base, nil
```
fabianburth: dont you need to do the exact same thing in the constructors code again?

bindings/go/descriptor/runtime/descriptor.go
piotrjanik: done

bindings/go/repository/interface.go
```diff
@@ -73,7 +73,9 @@ type LocalResourceRepository interface {
 	// Resources for non-existent component versions may be stored but may be removed during garbage collection.
 	// The Resource given is identified later on by its own Identity ([descriptor.Resource.ToIdentity]) and a collection of a set of reserved identity values
 	// that can have a special meaning.
-	AddLocalResource(ctx context.Context, component, version string, res *descriptor.Resource, content blob.ReadOnlyBlob) (*descriptor.Resource, error)
+	// Optional [AddLocalResourceOption]s carry construction-time directives (e.g.
+	// ADR-0016 ownership-referrer creation) that are not part of the descriptor.
+	AddLocalResource(ctx context.Context, component, version string, res *descriptor.Resource, content blob.ReadOnlyBlob, opts ...AddLocalResourceOption) (*descriptor.Resource, error)
```
fabianburth: I thought we were clear on the no options decision by now ...?

bindings/go/constructor/spec/v1/constructor.go
```diff
@@ -113,12 +124,26 @@ type Resource struct {
 	// Relation describes the relation of the resource to the component.
 	// Can be a local or external resource.
 	Relation ResourceRelation `json:"relation,omitempty"`
+	// Options bundles optional, construction-time directives for the resource
+	// It is omitted entirely when no options are set.
+	Options *ResourceOptions `json:"options,omitempty"`
 
 	AccessOrInput `json:",inline"`
 
 	ConstructorAttributes `json:",inline"`
```
fabianburth: That's how I expected it to be! But having `Options` and `ConstructorAttributes` alongside each other is very ugly. We should probably add a `TODO`.

bindings/go/constructor/construct.go
```diff
@@ -472,6 +482,64 @@ func (c *DefaultConstructor) processResourceByValue(ctx context.Context, targetR
 	return addColocatedResourceLocalBlob(ctx, targetRepo, component, version, resource, data)
 }
 
+// attachOwnershipReferrer attaches an asset-to-owner ownership referrer (ADR
+// 0016) for a relation=local resource that is kept by reference. It is a no-op
+// when the resource does not opt in via OwnershipPolicyAlways, when no resource
+// repository provider is configured, or when the resolved repository does not
+// implement the optional [OwnershipReferrerAwareRepository] capability (e.g. the
+// out-of-process plugin bridge, which cannot host referrers).
+func (c *DefaultConstructor) attachOwnershipReferrer(ctx context.Context, resource *constructor.Resource, res *descriptor.Resource, component, version string) error {
```
fabianburth: nit (since this is method is not exported): we wanted to avoid `Referrer` in the general APIs as the concept is not supposed to be OCI exclusive

bindings/go/constructor/construct.go
```diff
@@ -472,6 +482,64 @@ func (c *DefaultConstructor) processResourceByValue(ctx context.Context, targetR
...
+	// credential resolution below.
+	if resource.Options.OwnershipPolicy != constructor.OwnershipPolicyAlways {
+		logger.Debug("resource does not opt into ownership referrers, skipping")
+		return nil
+	}
+
+	if c.opts.ResourceRepositoryProvider == nil {
+		logger.Debug("no resource repository provider configured, skipping ownership referrer")
+		return nil
+	}
```
fabianburth: Again, `referrer`. Besides, I doubt, we would just silently want to skip adding ownership if the policy was set to `always`.

bindings/go/repository/interface.go
matthiasbruns: We did not want `opts` right? That' at least what my memory serves me

matthiasbruns (commented):
overall good direction - just the comments from fabian should be taken care of

I would also trim down the generated docs a bit, and make sure oci wording is only used in the oci bindings

bindings/go/repository/interface.go
fabianburth: Yes, we definitely did not want opts

bindings/go/constructor/construct.go
piotrjanik: please check, I changed referrer into ownership

bindings/go/constructor/construct.go
piotrjanik: I removed all occurrences of 'referrer' from the constructor binding.

bindings/go/constructor/spec/v1/constructor.go
piotrjanik: TODO is there

bindings/go/constructor/runtime/convert_v1.go
piotrjanik: fixed

bindings/go/constructor/runtime/convert_v1.go
piotrjanik: done

bindings/go/constructor/runtime/constructor.go
piotrjanik: done

bindings/go/oci/repository/resource/resource_repository.go
piotrjanik: done

cli/cmd/add/component-version/cmd.go
piotrjanik: this part has been changed, so please check again

bindings/go/repository/interface.go
piotrjanik: No opts

bindings/go/constructor/runtime/convert_v1.go
```diff
@@ -81,6 +81,27 @@ func ConvertToV1ElementMeta(meta ElementMeta) v1.ElementMeta {
...
+// ConvertToV1ResourceOptions builds the optional v1 resource options block from
+// the runtime resource options. The zero value (OwnershipPolicyNever) yields nil
+// so the options block is omitted from the spec rather than emitting an empty
+// object on every resource.
+func ConvertToV1ResourceOptions(o ResourceOptions) *v1.ResourceOptions {
+	if o.OwnershipPolicy == OwnershipPolicyAlways {
+		return &v1.ResourceOptions{OwnershipPolicy: v1.OwnershipPolicyAlways}
+	}
+	return nil
+}
```
fabianburth, line 101: This asymmetry is a bit weird - primarily because marshaling would always ignore "Never" now. Effectively, currently no one marshals descriptors, so might not be that relevant. Not sure if we should change it.

bindings/go/constructor/spec/v1/constructor.go
```diff
@@ -91,6 +91,16 @@ const (
...
+// OwnershipPolicy controls whether ownership (ADR 0016) is recorded for a resource during construction (ocm add/create).
+type OwnershipPolicy string
+
+const (
+	// OwnershipPolicyNever is the default: no ownership is recorded.
+	OwnershipPolicyNever OwnershipPolicy = "Never"
+	// OwnershipPolicyAlways records ownership for the resource.
+	OwnershipPolicyAlways OwnershipPolicy = "Always"
+)
+
```
fabianburth, line 103: Is there a particular reason having it as a iota in runtime and a constant in spec? Serialization?

piotrjanik:
@coderabbitai review

bindings/go/constructor/construct.go
piotrjanik: done

bindings/go/oci/stream/oci.go
```diff
@@ -30,10 +34,22 @@ func (s *OCIResourceStream) Root() ocispec.Descriptor {
...
+// content.ReadOnlyGraphStorage, as oras.ExtendedCopyGraph requires of its source.
+// It returns node's real parents (none if the storage can't report them).
+func (s *OCIResourceStream) Predecessors(ctx context.Context, node ocispec.Descriptor) ([]ocispec.Descriptor, error) {
+	finder, ok := s.ReadOnlyStorage.(content.PredecessorFinder)
+	if !ok {
+		return nil, nil
+	}
+	return finder.Predecessors(ctx, node)
+}
+
```
fabianburth: The PR looks pretty solid now! I really like the `oras.ExtendedCopyGraph` solution you found.
But this `ReferrerDescriptors` here and the `Predecessors` on `OCIResourceStream` look a bit sus.

Since the `Predecessor` method just uses `OCIResourceStream.ReadOnlyStorage` (type asserted), couldnt we calculate the `ReferrerDescriptors` on the fly as part of `tar.CopyToOCILayoutInMemory`?

piotrjanik: I changed this implementation.
 By introducing lazy `DiscoverReferrers` I could materialize it for non-stream operation but also use it as Predecessors in `oras.ExtendedCopyGraph`.
 
 Check last commit.

bindings/go/constructor/spec/v1/constructor.go
piotrjanik: valid, point, updated

bindings/go/constructor/runtime/convert_v1.go
piotrjanik: updated :)

bindings/go/constructor/construct.go
```diff
@@ -714,6 +741,22 @@ func addColocatedResourceLocalBlob(
 		return nil, fmt.Errorf("error adding local resource %q based on input type %q as local resource to component %q : %w", resource.ToIdentity(), resource.Input.GetType(), component, err)
 	}
 
+	// A by-value (or input-method) resource that opts in via OwnershipPolicyAlways
+	// gets ownership (ADR 0016) pointing the uploaded
+	// artifact back at the owning component version. Recording it requires a target
+	// repository that implements [OwnershipAwareRepository]; one that cannot (e.g. the
+	// out-of-process plugin bridge, or the deprecated fallback) makes construction fail
+	// rather than silently dropping an explicitly requested ownership link.
+	if resource.Options.OwnershipPolicy == constructor.OwnershipPolicyAlways {
```
matthiasbruns, line 740: nit: this could be inverted and an early return to reduce the nesting

matthiasbruns (commented):
I still feel the amount and content of the godoc screams ai - can you consolidate them a bit?

bindings/go/constructor/construct.go
piotrjanik: I reduced that a bit

jakobmoellerdev:
moved this back to draft as this is not an actual PR for review/merge

PR #2681, opened 2026-05-29
chore(fix): optimize e2e and conformance test to run in parallel

Skarlso:

Parallelize the e2e run with the conformance run. This extract the e2e from the kubernetes-controller flow and puts it as a re-usable workflow so we are able to trigger it at the same time in the same run.

Part of https://github.com/open-component-model/ocm-project/issues/1076

##### Verification


.github/workflows/pipeline.yml
```diff
@@ -144,7 +161,7 @@ jobs:
   # === This is from cli.yml::publish verbatim a date on conformance ===
   publish_cli:
     name: Publish CLI
-    needs: [build_cli, conformance, decide_publish]
+    needs: [build_cli, conformance, e2e, decide_publish]
```
matthiasbruns: why not just `build_cli` & `decide_publish`?

Skarlso: I'm looking into trimming the needs section :)

Skarlso: It's mostly because of the usages from the jobs and the dependencies. Like, publish_cli needs e2e because if it fails, we must not publish the CLI.

matthiasbruns: yes, but decide_publish also needed e2e doesnt it?

Skarlso: yep, that's why I'm looking into trimming it. :) I agree with you.

Skarlso: Done :)

matthiasbruns (approved):

jakobmoellerdev (approved):

PR #2683, opened 2026-05-29
fix(ci): load controller image via docker-daemon to avoid ctr-import digest mismatch

matthiasbruns:

The `E2E Tests (arm64)` job was failing at `kind load image-archive` with:

```
ctr: failed to extract layer (application/vnd.docker.image.rootfs.diff.tar.gzip
  sha256:c172…) to overlayfs:
  failed to get reader from content store: content digest sha256:c172…: not found
```

TBD @matthiasbruns  summarize the problem


matthiasbruns:
will bring this back once some prs are closed

PR #2689, opened 2026-05-30
fix(transfer): use ResourceRepository with empty filesystem config for streaming transfers

jakobmoellerdev:

This is a magic fix to make sure CLI can actually call a streaming transfer until we have fixed our resource repository delegation

follow up https://github.com/open-component-model/open-component-model/issues/2485

##### Verification


bindings/go/transfer/internal/builder.go
```diff
@@ -63,8 +65,20 @@ func NewDefaultBuilder(
 
 	// Streaming OCI-to-OCI transfer transformer
 	ociTransferOCIArtifact := &ocitransformer.TransferOCIArtifact{
-		Scheme:             transformerScheme,
-		Repository:         resourceRepo,
+		Scheme: transformerScheme,
+		// TODO(jakobmoellerdev): This is an ultra-super-duper hack.
```
frewilhelm, line 69: do we already have a follow up issue for that?

frewilhelm (approved):

bindings/go/transfer/internal/builder.go
jakobmoellerdev: I think the issue is linked is it not?

matthiasbruns: // This means that I now have to initialize a raw repository here, until either the builder and/or the
		// ResourceRepository plugin is refactored (see https://github.com/open-component-model/ocm-project/issues/774).

Usually I'd expect the url also closer to the TODO but not worth a change

matthiasbruns (approved):

bindings/go/transfer/internal/builder.go
jakobmoellerdev: Ack, thought about this some time. feel like the TODO is okay TBH and its a nit so resolving

PR #2734, opened 2026-06-02
chore(deps): Fix #2632

frewilhelm:
Supersedes #2632


jakobmoellerdev (approved):

kubernetes/controller/chart/templates/crd/components.delivery.ocm.software.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.20.1
+    controller-gen.kubebuilder.io/version: v0.21.0
```
jneisener, line 6: Seems like these `.yaml` files are the only diff to #2632.

I think adding an annotation like `# renovate: datasource=go depName=sigs.k8s.io/controller-tools` would enable renovate to handle the complete update in the future

jneisener (approved):

kubernetes/controller/chart/templates/crd/components.delivery.ocm.software.yaml
frewilhelm: Those files are generated and renovate does not run any post commands to re-generate files (maybe we could add this)

jneisener: Ah, makes sense. [`postUpgradeTasks`](https://docs.renovatebot.com/configuration-options/#postupgradetasks) might be a good fit

frewilhelm: Yes. I am just a bit concerned because it would run quite often I believe. And considering the GHA minutes that we already spend, we might just stick to the manual fixes

PR #2741, opened 2026-06-03
feat!: floating tag deletion in ctf/oci [2/2]

### LINKED ISSUE #959: Alias (Floating Tag) Delete Support
**Description**
Follow up for https://github.com/open-component-model/ocm-project/issues/720.

We decided to limit the scope of https://github.com/open-component-model/ocm-project/issues/720 to oci add support. We assume this should be sufficient for a lot of use cases already.

Still, we'd eventually like to also offer support for deleting tags as well (for details, check the linked issue).

**Consideration with regards to general `alias` support:**
- Would deleting tags be supported by all potential storage backends (oci, oci layout/ctf, s3)?

**Done Criteria**

matthiasbruns:
This PR is the full implementation of the floating tag deletion support for ctf and remote oci repositories.
Sinces ORAS does not provide a public API for this case, this PR implements the endpoint call itself.

OCM supports mutable aliases on component versions (e.g.latest, edge) that float independently of semver tags.Removing an alias requires deleting only the tag pointer and not the underlying manifest.

ORAS does not provide a public API for this.remote.Repository.Delete only operates by digest, which would wipe the manifest and all tags pointing to it. [The OCI Distribution Spec](https://github.com/opencontainers/distribution-spec/blob/v1.1.1/spec.md#deleting-manifests) defines 

> DELETE `/v2/<name>/manifests/<tag>` 

for tag-only deletion, but neither oras-go nor any community library exposes this path — see oras-go #647
(https: //github.com/oras-project/oras-go/pull/647), oras#1594 (https://github.com/oras-project/oras/discussions/1594), distribution-spec#102
(https: //github.com/opencontainers/distribution-spec/issues/102), distribution#3234 (https://github.com/distribution/distribution/issues/3234).

This PR implements content.Untagger for both the CTF store (via index removal) and the remote OCI store (via a direct `DELETE /v2/<name>/manifests/<tag>` call), and wires
both into RemoveComponentVersionAlias.

This PR needs to be split into two, the ctf and oci bindings.

Contributes: https://github.com/open-component-model/ocm-project/issues/959

##### Verification


bindings/go/oci/resolver/url/remote_store.go
```diff
@@ -0,0 +1,91 @@
...
+// (e.g. REGISTRY_STORAGE_DELETE_ENABLED is not set).
+var ErrTagDeletionDisabled = fmt.Errorf("registry does not support tag deletion (405 Method Not Allowed)")
+
+// remoteStore wraps *remote.Repository and adds content.Untagger support.
+//
+// oras-go implements content.Untagger only for its local OCI layout store
+// (content/oci.Store). The remote registry client (registry/remote.Repository)
+// intentionally omits it: the OCI Distribution Spec treats
+// DELETE /v2/<name>/manifests/<tag> as optional, and not all registries honor it.
+type remoteStore struct {
```
jakobmoellerdev: I believe we have a hardcoded switch case on remote.Repository in pack somewhere. can you double check this still works?

jakobmoellerdev: also Im wondering, shouldnt remoteStore be a type alias to remote.Repository?

bindings/go/oci/repository.go
```diff
@@ -879,6 +880,50 @@ func (repo *Repository) AddComponentVersionAlias(ctx context.Context, component,
...
 
+func (repo *Repository) RemoveComponentVersionAlias(ctx context.Context, component, alias string) (err error) {
+	ctx = slogcontext.NewCtx(ctx, repo.logger)
+	done := log.Operation(ctx, "remove component version alias",
+		slog.String("component", component),
+		slog.String("alias", alias))
+	defer func() { done(err) }()
+
+	if versionRegex.MatchString(alias) {
+		return fmt.Errorf("alias %q uses semantic version format and cannot be removed as an alias (use non-semver names like 'edge' or 'latest' instead)", alias)
```
jakobmoellerdev: nit: the error doesnt really make sense to me. wouldnt this mean that this is not a valid alias but a component version and that it cannot be removed at all (not just as an alias) without delete support?

bindings/go/oci/repository.go
```diff
@@ -879,6 +880,50 @@ func (repo *Repository) AddComponentVersionAlias(ctx context.Context, component,
...
+		return fmt.Errorf("reference %q is not a valid OCM component version: %w", reference, err)
+	}
+
+	untagger, ok := store.(content.Untagger)
+	if !ok {
+		return fmt.Errorf("store does not support alias deletion for component %q", component)
+	}
+
+	if err := untagger.Untag(ctx, alias); err != nil {
+		if errors.Is(err, errdef.ErrNotFound) {
```
jakobmoellerdev: nit: move errors.Is outside of nested err != nil check to reduce nesting

jakobmoellerdev (commented):
can go in mostly as is, just smaller adjustments from my side.

bindings/go/oci/resolver/url/remote_store.go
matthiasbruns: > also Im wondering, shouldnt remoteStore be a type alias to remote.Repository?

```go
type remoteStore struct {
	*remote.Repository
}
```
should be enough but I can double down on it

```go
var (
	_ spec.Store       = (*remoteStore)(nil) // general store spec
	_ content.Untagger = (*remoteStore)(nil) // content.Untagger opt-in
)
```

while I am not sure, if we should also update 

```go
type Store interface {
	content.ReadOnlyStorage
	content.Pusher
	content.TagResolver
}
```

or add a new interface with Untag support.

bindings/go/oci/repository.go
matthiasbruns: I aligned this with the floating-tag-creation logic

https://github.com/open-component-model/open-component-model/blob/6a817abaebd4287cfc02873f89bc1e64c1bb6213/bindings/go/oci/repository.go#L854

Good question though, we can either drop the safeguard and let users delete what every they want or need to rework the logic on both call-sites

bindings/go/oci/resolver/url/remote_store.go
matthiasbruns: also fixed the assertion in the switch-case 
for this I made the wrapper repo public and moved it into internal
I did not want to introduce another interface/abstraction/helper without thinking about it 

```go
// TODO(jakobmoellerdev): Eventually we should find a smarter solution to determine if a store is global
```

bindings/go/oci/repository.go
matthiasbruns: updated the error message - hope that helps

bindings/go/oci/internal/pack/pack.go
```diff
@@ -309,9 +310,10 @@ func setGlobalAccess(baseReference string, desc ociImageSpecV1.Descriptor, local
 // TODO(jakobmoellerdev): Eventually we should find a smarter solution to determine if a store is global.
 func backedByGlobalStore(storage content.Storage) bool {
 	switch storage.(type) {
-	// for ORAS repositories, we know they are global if they are remote repositories.
 	case *remote.Repository:
 		return true
+	case *remotestore.RemoteStore:
```
jakobmoellerdev, line 315: nit: inline with a comma

jakobmoellerdev: I have a smarter solution proposal for this in my cache draft PR

jakobmoellerdev (dismissed):

bindings/go/oci/internal/pack/pack.go
matthiasbruns: yeah that's a nit - I was thorn between another case, falltrough or comma

matthiasbruns: I will keep it as is, since you are proposing something else sooner or later

bindings/go/oci/internal/remotestore/remote_store.go
```diff
@@ -0,0 +1,98 @@
...
+}
+
+var (
+	_ spec.Store       = (*RemoteStore)(nil) // general store spec
+	_ content.Untagger = (*RemoteStore)(nil) // content.Untagger opt-in
+)
+
+// Untag removes the given tag from the remote registry without deleting the underlying manifest.
+// The registry must have tag deletion enabled; a 405 response means it is disabled server-side.
+func (r *RemoteStore) Untag(ctx context.Context, reference string) error {
```
Skarlso, line 43: Okay, so... going through all the trace calls here, I _think_ that this is not called with a digest, but honestly, I would be more at ease if we would add a specific guard statement here making sure that the reference is NOT a digest. Otherwise, we might accidentally delete a manifest.

matthiasbruns: it's public so it could - let me think about how

bindings/go/oci/internal/remotestore/remote_store.go
```diff
@@ -0,0 +1,98 @@
...
+	ref := r.Reference
```
Skarlso, line 44: Why not just do the same as the CTF implementation and do a LooseParseref here as well?

Skarlso:
>Neither oras-go nor any community library exposes this path

Technically, they expose it for local index.json layouts and in-memory blobs: https://github.com/oras-project/oras-go/blob/b11ec27a09e5c379c68f0c761924ea7008a58bb7/content/oci/oci.go#L310, https://github.com/oras-project/oras-go/blob/b11ec27a09e5c379c68f0c761924ea7008a58bb7/internal/resolver/memory.go#L73.

Just being pedantic. :)

matthiasbruns:
> > Neither oras-go nor any community library exposes this path
> 
> Technically, they expose it for local index.json layouts and in-memory blobs: https://github.com/oras-project/oras-go/blob/b11ec27a09e5c379c68f0c761924ea7008a58bb7/content/oci/oci.go#L310, https://github.com/oras-project/oras-go/blob/b11ec27a09e5c379c68f0c761924ea7008a58bb7/internal/resolver/memory.go#L73.
> 
> Just being pedantic. :)

You know what this is about :D 

straight-face

bindings/go/oci/internal/remotestore/remote_store.go
matthiasbruns: This is already done in https://github.com/open-component-model/open-component-model/blob/2c165a96dcd0f08c426d35002a8f64b65ae1d535/bindings/go/oci/resolver/url/resolver.go#L120

bindings/go/oci/internal/remotestore/remote_store.go
matthiasbruns: technically, we should also do a `semver` check here, cause we could also delete by version in that codepath :-/

matthiasbruns: so doing both now - hope thats fine

Skarlso: hm, this is a generic untagger, right? Why would it care for the semver? I think the tag check is enough. Also, why can't you untag a semver? 🤔 I'm not sure this belongs here.

bindings/go/oci/internal/remotestore/remote_store.go
Skarlso: Yes, but in the CTF its done in the Untag function. Here, it's not in the untag function. :)

bindings/go/oci/internal/remotestore/remote_store.go
Skarlso: If this is a restriction for `RemoveComponentVersionAlias` then it should be put in there and not the Untag. :)

bindings/go/oci/internal/remotestore/remote_store.go
Skarlso: Anyways, if it's already parsed than that's fine.

bindings/go/oci/internal/remotestore/remote_store.go
matthiasbruns: weakened up the guard again

jakobmoellerdev (dismissed):

Skarlso (approved):

jakobmoellerdev (approved):

matthiasbruns:
Bringing this to upstream -> https://github.com/oras-project/oras-go/pull/1217

PR #2745, opened 2026-06-03
chore(GHA): replace release GHA

frewilhelm:
Replaces `softprops/action-gh-release` with native `gh release` in the RC pre-release step

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -268,17 +268,23 @@ jobs:
           set -euo pipefail
 
           shopt -s nullglob
-          assets=(
-            "${{ runner.temp }}"/cli-assets/bin/*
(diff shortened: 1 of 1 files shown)


jakobmoellerdev (commented):
is this really even a valid finding? for me the code looks not problematic at all. Happy to stand corrected

frewilhelm:
This was the reason for the flagging

> Detects actions that are known to be "superfluous," i.e. perform an operation already provided by GitHub's own runner images.
Usage of these actions is not itself a security concern. However, many superfluous actions have significant runtime dependencies that unnecessarily increase the attack surface and/or supply chain risk of a workflow.

Hence, there is no immediate action necessary because the action did not show any vulnerability or the like. Will close this PR for now.

PR #2750, opened 2026-06-04
feat(replication): create the workerpool for the transfer resolution service

Skarlso:

We are adding the workpool implementation of the ADR 0020. This PR creates the code for the workerpool and the necessary logic to call transfer. It decides what and how to do things early on and tries not the take any chances once the API is solidified. Meaning it should work regardless of what changes once the API is created.

That said, it might be adjusted later on when the time has come to create the controller loop.

Also added `kubernetes/controller/integration/transfer_workerpool_integration_test.go`. This test is an example of what the CLI does for testing the transfer. It tests that the transfer and the workerpool and the even logic all work nicely. It runs really fast too. The reason this is not an e2e test in the e2e folder is simply because it doesn't require an entire cluster and there is no Replication CR yet to test with. It's calling the library directly.

##### Verification


kubernetes/controller/integration/transfer_workerpool_integration_test.go
```diff
@@ -0,0 +1,274 @@
...
+	)
+	graphBuilder := workerpool.NewGraphBuilder(b)
+
+	// Phase 2: async transfer
+	logger := testr.New(t)
+	pool := workerpool.NewWorkerPool(workerpool.PoolOptions{
+		WorkerCount: 2,
+		QueueSize:   8,
+		Logger:      new(logger),
+	})
```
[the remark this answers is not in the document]
Skarlso: Update your Go version. ;)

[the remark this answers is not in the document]
Skarlso: Do you see any compiler errors in this repository?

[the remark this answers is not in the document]
Skarlso:
```
➜  integration git:(add-replication-controller-workerpool) go test ./...
ok  	ocm.software/open-component-model/kubernetes/controller/integration	3.248s
```

🪄

kubernetes/controller/internal/replication/workerpool/workerpool.go
```diff
@@ -0,0 +1,456 @@
...
+
+	// Cancel every per-key context so running transfers stop at the next safe point.
+	wp.baseCancel()
+
+	done := make(chan struct{})
+	go func() {
+		wp.workersDone.Wait()
+
+		close(wp.workQueue)
+
```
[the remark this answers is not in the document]
Skarlso: Hm, are you sure?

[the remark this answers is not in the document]
Skarlso: Fair point.

kubernetes/controller/internal/replication/workerpool/workerpool.go
```diff
@@ -0,0 +1,493 @@
...
+		stamp:   opts.Stamp,
+		tgd:     opts.TGD,
+		builder: opts.Builder,
+		ctx:     itemCtx,
+	}
+
+	select {
+	case wp.workQueue <- item:
+		wp.inProgress[opts.Key] = &inflight{
+			requesters: []RequesterInfo{opts.Requester},
```
frewilhelm: We don't check if `opts.Requester` could be `nil`. But do we expect any scenario where a transfer occurs without `Requester`?

kubernetes/controller/internal/replication/workerpool/workerpool.go
```diff
@@ -0,0 +1,493 @@
...
+
+// Result consumes the terminal result for a key, returning it only when its
+// Stamp matches desiredStamp.
+//
+// The caller passes the content identity it currently wants (derived from the
+// live source/target/config this reconcile). A result whose Stamp differs is
+// stale: the desired state moved on while the transfer was in flight. It is
+// dropped and (nil, false) is returned, exactly as if no result were present,
+// so the caller re-plans for the current state. A matching result is delivered
+// exactly once.
```
frewilhelm, line 283: I am having a hard time understanding this.

What happens if I started two identical transfers and then update one of them? Is it possible that the transfer deduplication would omit the first transfer and then only process the updated one?

kubernetes/controller/internal/replication/workerpool/builder.go
frewilhelm, line 1: Is my assumption correct that this file only exists to mock the functions for the tests?

Taskfile.yml
```diff
@@ -136,6 +136,10 @@ includes:
     optional: true
     taskfile: ./kubernetes/controller/Taskfile.yml
     dir: ./kubernetes/controller
+  kubernetes/controller/integration:
+    optional: true
+    taskfile: ./kubernetes/controller/integration/Taskfile.yml
+    dir: ./kubernetes/controller/integration
```
frewilhelm, line 150: Will this be run by our CI for the controllers?

kubernetes/controller/internal/replication/workerpool/source_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"k8s.io/client-go/util/workqueue"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+
+	"ocm.software/open-component-model/kubernetes/controller/internal/replication/workerpool"
+)
+
+func TestEventSourceEnqueuesRequestersOnCompletion(t *testing.T) {
```
frewilhelm: A little doc would've saved me a bit of time understanding what is going on here :D

frewilhelm (commented):
I am having a bit of a hard time understanding the purpose of the `stamp`. Lets discuss this please :)

kubernetes/controller/internal/replication/workerpool/workerpool.go
Skarlso: opts.Requester is not a pointer so it's never nil. But we don't check if it's empty, that's true. I don't really see the reason here, but we might as well add a guard. :)

Skarlso: >But do we expect any scenario where a transfer occurs without Requester?

No, there should never be a transfer without a requester.

kubernetes/controller/internal/replication/workerpool/workerpool.go
Skarlso: Dedup is keyed on the CR UID, not the stamp. A burst of submits for the same object collapses onto one  transfer regardless of content. The stamp is separate: it's an identity of the work. It's only checked at result-consume time. For example, if you submit A, then change your mind to B while A is in flight, A is what actually runs (stamp remains). When A finishes, the reconciler asks Result for B, the stamp won't match, so the A result is dropped, and the reconciler re-submits for B. You never finalize stale work, and B still runs once the key frees up.

Does this make better sense? :D

Taskfile.yml
Skarlso: It already is running it 👍 ✅

kubernetes/controller/internal/replication/workerpool/builder.go
Skarlso: Correct. This helps me with the unit tests. Once the full machinery exists I might throw this out and change it to something. But right now, it's quite convenient.

kubernetes/controller/internal/replication/workerpool/source_test.go
Skarlso: Got it. :D I'll add one. :)

kubernetes/controller/internal/replication/workerpool/doc.go
```diff
@@ -0,0 +1,33 @@
+// Package workerpool implements the asynchronous transfer worker pool for the
+// replication controller (ADR 0020).
+//
+// The pool mirrors the resolution worker pool but specializes for transfers:
+//
+//   - Work is keyed by the Replication CR UID, so burst reconciles for an
+//     already-submitted transfer collapse onto a single in-flight item and
+//     return [ErrTransferInProgress] instead of re-submitting.
+//   - Results are one-shot and consumed exactly once via [WorkerPool.Result];
```
jakobmoellerdev, line 14: I think this is not a "strict" correct design. Instead the result cache should live in the controller and then the result should be reflected into the CR status. WDYT?

kubernetes/controller/internal/replication/workerpool/doc.go
```diff
@@ -0,0 +1,33 @@
...
+//     content lives in the CR status (lastTransferredDigest) and in the
+//     content-addressed target registry, not in the pool.
+//   - A result carries the opaque Stamp it was submitted with. Because the key
+//     is the CR UID and not the unit of work, a result can outlive the source
+//     state it was produced for (request coalescing, a dropped event, a resync
+//     landing after the digest moved). [WorkerPool.Result] takes the currently
+//     desired Stamp and returns a result only on a match, dropping a superseded
+//     one so the caller cannot accidentally finalize stale work. The pool is
+//     keyed by UID rather than content on purpose: content keying would share a
+//     transfer across objects with different credentials, letting one object
```
jakobmoellerdev, line 25: Not if we would use the existing OCM cache keying concept right?

kubernetes/controller/internal/replication/workerpool/workerpool.go
```diff
@@ -0,0 +1,493 @@
...
+// via the completion event when the transfer finishes.
+type RequesterInfo struct {
+	NamespacedName types.NamespacedName
+}
+
+// SubmitOptions describes a transfer to submit to the pool.
+type SubmitOptions struct {
+	// Key uniquely identifies the transfer. The Replication CR UID is used so
+	// burst reconciles for the same object map onto a single in-flight item.
+	Key string
```
jakobmoellerdev, line 38: nit: name to UID

kubernetes/controller/internal/replication/workerpool/workerpool.go
```diff
@@ -0,0 +1,493 @@
...
+	// Stamp is the opaque content identity of this unit of work. The pool does
+	// not interpret it; it only matches it, in [WorkerPool.Result], against the
+	// stamp the caller currently wants, dropping a stale outcome. The reconciler
+	// derives it from what defines the transfer (source digest, target, transfer
+	// config), so a result produced for an older source is discarded once the
+	// source has moved on.
+	Stamp string
```
jakobmoellerdev, line 45: what exactly does content identity mean?

kubernetes/controller/internal/replication/workerpool/workerpool.go
```diff
@@ -0,0 +1,493 @@
...
+}
+
+// workItem is the internal unit processed by a worker.
+type workItem struct {
+	key     string
+	stamp   string
+	tgd     *transformv1alpha1.TransformationGraphDefinition
+	builder GraphBuilder
+	// ctx is the per-key context, derived from the pool base context.
+	ctx context.Context
```
jakobmoellerdev, line 98: why do we have a per key context?

kubernetes/controller/internal/replication/workerpool/workerpool.go
```diff
@@ -0,0 +1,493 @@
...
+		baseCtx:     baseCtx,
+		baseCancel:  baseCancel,
+	}
+}
+
+// Subscribe creates a new event subscription channel. Each subscriber gets its
+// own buffered channel so events are not stolen between controllers. The channel
+// is closed when the pool shuts down. Subscribing after shutdown returns an
+// already-closed channel so the consumer stops immediately.
+func (wp *WorkerPool) Subscribe() <-chan []RequesterInfo {
```
jakobmoellerdev: how can a subscription be cancelled?

kubernetes/controller/internal/replication/workerpool/workerpool.go
```diff
@@ -0,0 +1,493 @@
+package workerpool
```
jakobmoellerdev, line 1: it feels too complicated imho but maybe thats just me.

kubernetes/controller/internal/replication/workerpool/doc.go
Skarlso: Yeah, the workpool is technically the cache. The result can be put into the status, and will be I think according to the API in terms of

```go
  lastTransferredVersion: "1.2.3"
  lastTransferredDigest: "sha256:abc123..."
```

kubernetes/controller/internal/replication/workerpool/workerpool.go
Skarlso: >defines the transfer (source digest, target, transfer config) etc.

A unit of work basically.

kubernetes/controller/internal/replication/workerpool/workerpool.go
Skarlso: Errr.... can't be cancelled right now. :D I didn't plan for that...

kubernetes/controller/internal/replication/workerpool/doc.go
Skarlso: This is actually relevant for your other question as well, so I'll try to answer all of them here. :) 

The reason I didn't chose the content as a key rather than the UID and a Stamp is because in case of the resolution service, it's a read in this case it's a write action. A stamp is basically a unit of work that needs to be performed. There is nothing really that needs caching if I understand this right. It basically just needs to be done. ( Also, I wanted to avoid a possible reuse of credentials by the cache acidentally ).

So I thought that a stamp might a combination of everything, like source component + version + digest, target repository spec, transfer config... whatever defines a unit of work. With that, you'll get your stamp. And then when you want to your result by calling `pool.Result....` with your stamp, you'll know that what is coming back for your resource / object is the correct current stamp. If not, you need to submit again and requeue because there was a rapid update inbetween your previous submit and the this request.

This is to prevent the service offerring a stale result as the result of the transfer. Saying something like, hey object A, sure I got your result right here and object A says, thanks all good.

Now, if you decide to keey based on object + digest + config + repo + whatever to make it super unique. That adds a couple of other problems. While the stamping goes away, we get staleness and LRU and needing to make sure that everything is garbage collected, yadayada.

Here, since we only have this as the object, as long as the object is there, there is it's data, and once it's deleted no more data. :) Honestly, I thought this is pretty elegant. :D 

ALSO... If we do use unique keys for everything, we won't get the nice deduplication. Meaning if you submit another request it will start pushing that content simultaniously with the otherone while the otherone is still running and would need to be cancelled. This is something the ADR had some trouble with designing what to do about in-flight stuff. So, this, kinda made sense that we allow ONE for the same object. That's it. If another comes in immedately for the _same_ object it makes sense for it to wait for the result since it already started one. So this is kinda elegant way of saving to have to fiddle with cancel since it will naturally evolve from there. It's fine if we want that, but it's a bigger machine a bigger chunk of work. I thought this as a start would be very much sufficient. :) Yet also allows for improvement by hiding this machinery.

kubernetes/controller/internal/replication/workerpool/workerpool.go
Skarlso: Hmmm. I thought it's actually quite elegant compared to the resolution service. 🤔 

What did you expect to see? Did you have something in your mind? Like just an LRU cache and done? :) 

That also works. I'm doing this because we decided this in the ADR:
```md
### Reconciliation Flow

Similar to the resolution service, this is the two-phase process for transfer.
Following the existing `ErrResolutionInProgress` pattern, phase 2 introduces a
transfer-specific in-progress sentinel error named `ErrTransferInProgress`.
```

kubernetes/controller/internal/replication/workerpool/workerpool.go
Skarlso: I was thinking about how to handle the cancel for an ongoing transfer. So the idea here is that this context is given to the item here:

```go
	itemCtx, cancel := context.WithCancel(wp.baseCtx)
	item := &workItem{
		key:     opts.Key,
		stamp:   opts.Stamp,
		tgd:     opts.TGD,
		builder: opts.Builder,
		ctx:     itemCtx,
	}
...
```

And then, we give the `cancel` function of that to the worker:

```go
		wp.inProgress[opts.Key] = &inflight{
			requesters: []RequesterInfo{opts.Requester},
			cancel:     cancel,
		}
```

And then this cancel can be used in one of two ways.
If the workerQueue is full, it calls cancel on the context. And second, the `inflight` has the cancel:

```go
// inflight tracks the state of a queued or running transfer.
type inflight struct {
	// requesters are all objects waiting on this key, notified on completion.
	requesters []RequesterInfo
	// cancel cancels the per-key context, used for the deletion drain and on
	// pool shutdown.
	cancel context.CancelFunc
}
```

So, here, we have three options to call this cancel. One is the deletion drain and the other is the pool shutdown.

But! I also added a Pool cancel for a key! 

```go

// Cancel cancels the per-key context of a queued or running transfer. It is a
// no-op if no transfer for the key is in flight. The canceled transfer still
// produces a terminal result and a completion event.
func (wp *WorkerPool) Cancel(key string) {
	wp.mu.Lock()
	defer wp.mu.Unlock()

	if inf, ok := wp.inProgress[key]; ok {
		wp.Logger.V(1).Info("canceling in-flight transfer", "key", key)
		inf.cancel()
	}
}
```

Now, this... this can be used to shutdown the actual transfer from the outside for a key. I can imagine something like this in the above described serialized transfer:

We compute the stamp and store it in the Status as -> `InFlightStamp` or whatever. We do a Submit. Transfer begins, inProgress happens. Suddenly, the object is updated because there is an ew version or whatever. Calculate the stamp. Do a Submit... Did we get back an InProgress? Compare it with saved stamp in status. Oh-oh... it differs. That's not good, something is still running. Save stamp. Call `Cancel` and return! Cancel will still send out a done signal and remove the stale result! So the item requeues... Calculate the stamp. Submit... in progress, all good, yay.

Okay, this is a bit rambling, let me try to write the code I would expect to call here (pseudo code):

```go
  desired := computeStamp(digest, spec, ...more)

  switch {
  case !pool.IsInProgress(uid):
      // Nothing running. Did a finished result land for what we want?
      res, ok := pool.Result(uid, desired)
      if ok { return res }
       
      // New round
      errr := pool.Submit(SubmitOptions{Key: uid, Stamp: desired, ...})
      // do things with error
      status.InFlightStamp = desired // save the stamp
      return

  case status.InFlightStamp == desired:
      // The right transfer is already running. Wait for its event. Maybe we got requeued by something else.
      return ctrl.Result{}, nil

  default:
      // Something is running but the stamp didn't match so we cancel it.
      pool.Cancel(uid) // this takes a while to propegate through the Graph
      status.InFlightStamp = "" // clear it because we no longer trust the current run
      return ctrl.Result{}, nil // once cancel is done it will retrigger via the even queue. Maybe to be on the safe side, we would add a requeue time as well?
  }
```

How does that sound? :) I'll be honest, I really like this approach. It makes it so easy to deal with in-progress transfer cancellation. :D But maybe I'm biased.

kubernetes/controller/internal/replication/workerpool/workerpool.go
Skarlso: I had a bit of thinking here last night, trying to sketch out some ideas on paper and realised I can simplify things here. I do want to keep the pool and the map because I believe that it gives us a lot of flexibility and a very good, sane, flow.

What I can do is simplify the internal workings. For example, I don't really _need_ the subscriber model since it's just the replication controller. And I don't  need the eventqueue implementaion since controller runtime has one. And there are a few other things like the workerQueue can be replaced... so I'll simplify that and then you can see if it reads better. :)

kubernetes/controller/internal/replication/workerpool/workerpool.go
```diff
@@ -0,0 +1,439 @@
...
+		opts.EventBufferSize = defaultEventBufferSize
+	}
+
+	if opts.ShutdownTimeout <= 0 {
+		opts.ShutdownTimeout = defaultShutdownTimeout
+	}
+
+	if opts.Logger == nil {
+		opts.Logger = new(logr.Discard())
+	}
```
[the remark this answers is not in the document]
Skarlso: Jesus Christ man, we had this same discussion before. Update your Go version.

kubernetes/controller/internal/replication/workerpool/builder.go
Skarlso: Actually, you know what, I'm gonna remove this thing.

Skarlso: Done 🎉

kubernetes/controller/internal/replication/workerpool/workerpool.go
Skarlso: Not a problem anymore. :)

kubernetes/controller/internal/replication/workerpool/workerpool.go
Skarlso: Technically, I'm calling it a key here to "leave the option open" to have this as anything other. I might even remove the comment specifics. WDYT?

kubernetes/controller/internal/replication/workerpool/doc.go
jakobmoellerdev: We already have a cache key concept for ocm config and from what I understand you can just append all of the relevant data for a transfer, take that hash and then get the resulting ID. Im not saying do not introduce a key, im more leaning towards "lets make key construction easy and understandable and document how it should behave"

kubernetes/controller/internal/replication/workerpool/workerpool.go
jakobmoellerdev: if you make it even more generic it becomes less understandable. the pool will not be used outside this specific purpose so it should have purpose built commentary imho

kubernetes/controller/internal/replication/workerpool/workerpool.go
jakobmoellerdev: see above, this needs a revisit to be more understandable and relatable when looking at it after 2 weeks again.

kubernetes/controller/internal/replication/workerpool/workerpool.go
jakobmoellerdev: To me this is a lot complexity for propagation of 2 cancel funcs so probably worth simplifying to your idea here by just passing that. im not sure you need context for anything else

Skarlso: You mean the workerpool cancel and the context cancel func?

But the context is specific to a single Transfer operation. 🤔 And it's inside the workerpool. I'm not sure how it could be a better or easier separation for cancelling. The transfer is independent of the reconcile.

jakobmoellerdev:
closing for now as we decided to follow a simpler approach and iterate on that

PR #2752, opened 2026-06-04
feat: Add command to print effective config

jneisener:
Implements https://github.com/open-component-model/ocm-project/issues/949

Prints the effectively merged config from all discovered sources (`configuration/extract` in unmerged form for technical reasons).

Output is returned to stdout in either `yaml`, `json`, or `ndjson`.

https://github.com/open-component-model/ocm-project/issues/949

##### How to test the changes
Added unit tests

##### Verification


cli/cmd/get/config/cmd.go
```diff
@@ -0,0 +1,186 @@
...
+	"ocm.software/open-component-model/cli/internal/flags/enum"
+	pluginsv2alpha1 "ocm.software/open-component-model/cli/internal/plugin/spec/config/v2alpha1"
+	"ocm.software/open-component-model/cli/internal/render"
+)
+
+const (
+	FlagOutput = "output"
+)
+
+type EffectiveConfig struct {
```
jakobmoellerdev: This is most likely not quite correct, because config types such as ownership configuration are part of the ocm config, but here they are handled separately from the OCM config. Most likely you want to revisit this logic so that everything becomes part of one big OCM config here.

cli/cmd/get/config/cmd.go
```diff
@@ -0,0 +1,176 @@
...
+	return s
+}()
+
+var credentialsScheme = func() *runtime.Scheme {
+	s := runtime.NewScheme()
+	credentialsv1.MustRegister(s)
+	return s
+}()
+
+func appendEffective(result *genericv1.Config, cfg *genericv1.Config, scheme *runtime.Scheme, lookup func() (runtime.Typed, error)) error {
```
jakobmoellerdev: this func is a little too much meta programming for my taste but if you keep it as is it needs behavior comment on how and why it does things the way it does

cli/cmd/cmd_test.go
```diff
@@ -2030,3 +2030,222 @@ resources:
 	// Verify referenced component is still accessible
 	_ = referencedDesc
 }
+
+func Test_Get_Config(t *testing.T) {
```
jakobmoellerdev, line 2034: here I am looking for test cases for 

- merging multiple configs of the same type
- merging multiple configs from multiple files of the same type

do you have these in here?

cli/cmd/get/config/cmd.go
```diff
@@ -0,0 +1,176 @@
...
+		return nil
+	}
+	typed, err := lookup()
+	if err != nil {
+		return fmt.Errorf("%s: %w", types[0], err)
+	}
+	if typed == nil {
+		return nil
+	}
+	raw := &runtime.Raw{}
```
matthiasbruns: Do we need this at all? our `runtime.Typed` are serializable by default, right? 
So in the cli serialization, this should be done automatically, afaik,

cli/cmd/get/config/cmd.go
```diff
@@ -0,0 +1,176 @@
...
+	return s
+}()
+
+func appendEffective(result *genericv1.Config, cfg *genericv1.Config, scheme *runtime.Scheme, lookup func() (runtime.Typed, error)) error {
+	var types []runtime.Type
+	for typ, aliases := range scheme.GetTypes() {
+		types = append(types, typ)
+		types = append(types, aliases...)
+	}
+	if !hasEntries(cfg, types) {
```
matthiasbruns: we are double looping here - once on this line and then as a fallback when

`if lookup == nil {`

cli/cmd/get/config/cmd.go
jneisener: So what I'm trying to implement here is a consistent usage of a semi consistent interface between the different configuration types. 

Lookup functions have different names and some types don't implement them at all. Some schemas are exported, others are not. I want to serialize all of them and I was going for brevity. If in that pursuit it got too meta I can also make it more verbose again, e.g. similar to how it was here: https://github.com/jneisener/open-component-model/blob/7340900b1817ff9c26b5dc432d1c2bc870fc4c97/cli/cmd/get/config/cmd.go#L100-L168

Would you find something along those lines more preferable?

cli/cmd/cmd_test.go
jneisener: No, all the current tests are with single files. Different output formats and one test case with a 'complete' config file. 

Arguably testing the outcome of the merge tests what is behind the `Lookup`/`LookupConfig` interface, which mainly should be tested elsewhere. But since I'm manually calling all the methods it makes sense to have at least one test to validate the output including a merge.

cli/cmd/get/config/cmd.go
```diff
@@ -93,81 +93,114 @@ func GetConfig(cmd *cobra.Command, args []string) error {
...
-		{ocmv1.Scheme, func() (runtime.Typed, error) { return ocmv1.Lookup(cfg) }}, //nolint:staticcheck // displaying deprecated config for user visibility
-		{resolversv1alpha1.Scheme, func() (runtime.Typed, error) { return resolversv1alpha1.Lookup(cfg) }},
-		{ownershipv1alpha1.Scheme, func() (runtime.Typed, error) { return ownershipv1alpha1.Lookup(cfg) }},
-		{extractv1alpha1.Scheme, nil}, // TODO: bindings/go/configuration/extract/v1alpha1/spec/config.go merge is no-op
-		{pluginsScheme, func() (runtime.Typed, error) { return pluginsv2alpha1.LookupConfig(cfg) }},
-		{credentialsScheme, nil},
+	var typed runtime.Typed
+	var err error
+
+	if hasEntries(cfg, filesystemv1alpha1.ConfigType, filesystemv1alpha1.Version) {
```
jakobmoellerdev: why do we need the explicit hasEntries check, shouldnt this be part of LookupConfig?

cli/cmd/get/config/cmd.go
jakobmoellerdev: tbh I was just looking for docs on behavior in the code :D didnt want you to revert to the old version

cli/cmd/get/config/cmd.go
jneisener: http and plugin configs don't return nil on lookup https://github.com/jneisener/open-component-model/blob/84923963cd20f4687953748c056212cee49fe8ac/cli/internal/plugin/spec/config/v2alpha1/config.go#L68-L97

But the other functions do, so I don't need the `hasEntries` check in most of the places

cli/cmd/get/config/cmd.go
jneisener: Refactored away

cli/cmd/get/config/cmd.go
jneisener: Refactored away

cli/cmd/cmd_test.go
jneisener: Does https://github.com/jneisener/open-component-model/blob/7df33a74d8f83c4c39c43b2749b12c2f63b181a2/cli/cmd/cmd_test.go#L2067-L2107 sufficiently cover both cases?

cli/cmd/get/config/cmd.go
jneisener: No, I think you were right. It wasn't idiomatic go code, it was too fancy. I think the current solution - while more verbose - is easier to understand

cli/cmd/get/config/cmd.go
jneisener: Replaced with `!= nil` and type checks

cli/cmd/get/config/cmd.go
```diff
@@ -0,0 +1,190 @@
...
+		var config extractv1alpha1.Config
+		if err := extractv1alpha1.Scheme.Convert(entry, &config); err != nil {
+			return nil, fmt.Errorf("failed to decode extract config: %w", err)
+		}
+		result.Configurations = append(result.Configurations, &config)
+	}
+
+	// Runtime types of credentials serialize as `json:"-"` (bindings/go/credentials/spec/config/runtime/config.go)
+	// Serialize the raw entry instead (caveat: output is not merged)
+	filtered, err = genericv1.Filter(cfg, &genericv1.FilterOptions{
```
jakobmoellerdev: instead of this, you can convert the runtime back to a serializable form right?

jakobmoellerdev (commented):
just a comment left on credentials serializing, rest looks great!

cli/cmd/get/config/cmd.go
jneisener: Sound plausible, but I'm not sure how.

There is no v1 scheme, only a runtime scheme. I can add my own scheme, but conversion only works on raw types, not typed ones. So I'm not sure there is a builtin way to do it. 

Did you mean a manual conversion like this: https://github.com/open-component-model/open-component-model/pull/2752/changes/78e28a27bc7270b4196579e81c9f8f6b4ed987a2#diff-0386a795ae1c3c97a163ab9dcee29c4a7159db7442901e2ef030bb2221802dccR187-R210?

jakobmoellerdev: https://github.com/open-component-model/open-component-model/tree/main/bindings/go/credentials/spec/config/v1

is that not it?

jneisener: Backwards conversion to be added via #2813

cli/cmd/get/config/cmd.go
```diff
@@ -0,0 +1,194 @@
...
+		result.Configurations = append(result.Configurations, fsCfg)
+	}
+
+	httpCfg, err := httpv1alpha1.LookupConfig(cfg)
+	if err != nil {
+		return nil, fmt.Errorf("config lookup failed for http: %w", err)
+	}
+	if httpCfg != nil && httpCfg.Type != (runtime.Type{}) {
+		result.Configurations = append(result.Configurations, httpCfg)
+	}
```
matthiasbruns: When you are not really using the variables anywhere else but just in the else branch if the if statement, you can also write it like this
```suggestion
	if httpCfg, err := httpv1alpha1.LookupConfig(cfg); err != nil {
		return nil, fmt.Errorf("config lookup failed for http: %w", err)
	} else if httpCfg != nil && httpCfg.Type != (runtime.Type{}) {
		result.Configurations = append(result.Configurations, httpCfg)
	}
```

so `httpCfg` and `err` will only be scoped and available within the if-statement

matthiasbruns (commented):
just a tiny nit

jakobmoellerdev (approved):

matthiasbruns (approved):

PR #2753, opened 2026-06-04
chore: Add `yaml` struct tags to all configuration types

jneisener:
Found while working on https://github.com/open-component-model/ocm-project/issues/949
`enc.Encode` expects yaml types to be deeply tagged, i.e. fixes:
https://github.com/open-component-model/open-component-model/pull/2752/changes#diff-0386a795ae1c3c97a163ab9dcee29c4a7159db7442901e2ef030bb2221802dccR86

##### How to test the changes

Only added equivalent `yaml` tags to the already existing `json` tags, which I assume does not need additional testing

##### Verification

### DIFF
--- a/bindings/go/configuration/extract/v1alpha1/spec/config.go
+++ b/bindings/go/configuration/extract/v1alpha1/spec/config.go
@@ -30,9 +30,9 @@ func init() {
 type Config struct {
 	// +ocm:jsonschema-gen:enum=extract.oci.artifact.ocm.software/v1alpha1
 	// +ocm:jsonschema-gen:enum:deprecated=extract.oci.artifact.ocm.software
-	Type runtime.Type `json:"type"`
+	Type runtime.Type `json:"type" yaml:"type"`
 	// Rules defines rules for extracting layers to specific files.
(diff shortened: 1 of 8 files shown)


frewilhelm:
I assume you added this to print out the effective configuration directly as yaml. However, I am wondering if we want to do this because this implies maintaining the `json` and `yaml` tags which should be the same/redundant.

I am wondering because, usually, an `.ocmconfig` is written in yaml and the current `json` tags work fine. Can you please check how we read the `.ocmconfig` and if the `yaml` tag is really necessary.

jneisener:
> I assume you added this to print out the effective configuration directly as yaml. However, I am wondering if we want to do this because this implies maintaining the `json` and `yaml` tags which should be the same/redundant.
> 
> I am wondering because, usually, an `.ocmconfig` is written in yaml and the current `json` tags work fine. Can you please check how we read the `.ocmconfig` and if the `yaml` tag is really necessary.

Good point, I was using `gopkg.in/yaml.v3` when I should have been using `sigs.k8s.io/yaml`, which relies on the [JSON tags](https://github.com/kubernetes-sigs/yaml#introduction)

PR #2757, opened 2026-06-05
chore: fix build running and building all architectures on basic ci runs

Skarlso:

Attempt to speed up the CLI build so it doesn't take over 6 minutes to run. :D 

These are the high level changes:

Speed up CI builds

CLI
I split the single build job into a 6-way GOOS/GOARCH matrix. Cross-compiles run in parallel on separate runners instead of serially on one. The new version job computes the version once and then shares it. The assemble job downloads the prebuilt binaries, packages the OCI image and CTF _without_ recompiling, then attests and uploads. 

Controller
PR builds produce a linux/arm64 image _only_ since PR gates run only on arm64 runners there is no need for the others because we can build them during push/release ON DEMAND. Meaning they are never built otherwise.

Dockerfile uses BuildKit now to cache mounts for the module and build caches and dropped -a so the stdlib isn't rebuilt every time.

And I updated the path to add triggers for when changing the e2e file.

##### Verification


piotrjanik (dismissed):

cli/Taskfile.yml
```diff
@@ -134,6 +145,18 @@ tasks:
...
         --repository {{ .CTF_LOCATION }}
 
+  assemble/ctf:
+    desc: "Assemble the CTF from binaries already present in tmp/bin without recompiling (CI join entrypoint)"
+    cmds:
+      - task: package/image
+      - task: assemble/component-constructor
+      - ln -sf {{ .TASKFILE_DIR }}/tmp/bin/ocm-{{ .GOOS }}-{{ .GOARCH }} {{ .TASKFILE_DIR }}/tmp/bin/ocm
+      - |
+        rm -rf {{ .CTF_LOCATION }}
```
piotrjanik, line 155: nit: you could create internal task jufr for building ctf and use it here and in `generate/ctf`

kubernetes/controller/Dockerfile
```diff
@@ -9,7 +10,8 @@ COPY go.mod go.mod
 COPY go.sum go.sum
 # cache deps before building and copying source so that we don't need to re-download as much
 # and so that source changes don't invalidate our downloaded layer
-RUN go mod download
+RUN --mount=type=cache,target=/go/pkg/mod,sharing=locked \
```
jakobmoellerdev, line 13: the target directory is now opinionated on the go install on the host to have a running mod cache under /go/pkg/mod which is not canonical and also mutable. Im not sure that this is really desired to be honest, so im cautious to accept it. While I do understand it makes things faster in CI, this can and will break on local installs that dont look exactly like the runner

jakobmoellerdev:
another thing. The job names are getting a bit crazy

```
[monorepo shared build & publish / build_cli / Build CLI (darwin/amd64)
```
we should probably change this to 

```
monorepo shared build & publish / build cli / darwin/amd64
```

Skarlso:
All very good notes, thanks folks! I will address them. :)

kubernetes/controller/Dockerfile
Skarlso: Hm, what I could do is create a specific dockerfile for the CI to use while keeping the main file untouched. 🤔  WDYT about that?

Skarlso: So, read about this and looked at some examples.

We are using buildx that uses BuildKit. That mount is not opinionated on the host, it's actually in the target container base. `target=/go/pkg/mod` is a target in the container which we have. https://docs.docker.com/reference/dockerfile/#run---mounttypecache. So that works.

And since our task files are using buildx anyways this should also work on local but since Docker now is also using BuildKit by default if you use the plain docker build environment even `docker build` works out of the box.

Also to be on the safe side, I checked podman and Buildah supports buildkit since 2022. :D 

TL;DR: I think, we're good.

jakobmoellerdev:
@Skarlso ping on this pr

Skarlso:
Ja ja. :D There were just so many higher priority things I wasn't able to continue / finish this one yet. It's on my todo list. :D

Skarlso:
Adding renaming the builds to this in a sec. :)

Skarlso:
Screenshot 2026-07-17 at 12 51 29

Renamed for better readability. :)

morri-son (dismissed):

morri-son (approved):

matthiasbruns (approved):
Merge it already

Skarlso:
Why has this not merged already?? I set it to auto merge AGES ago

Skarlso:
There are just constantly so many bump prs being merged, it's impossible to merge this. :D

Skarlso:
Okay, I'm going to just abandon this until tomorrow or something.

PR #2762, opened 2026-06-05
fix: Check for length instead of `nil`

### LINKED ISSUE #2742: Transfer of components with no labels fails
**Context**

The command `ocm transfer cv --copy-resources` fails for components whose resources contain no labels. 
This is valid per the OCM Component Descriptor v2 (https://ocm.software/docs/reference/component-descriptor/), as `component.labels` has no `required` tag, but it is rejected:

**Version**

Both OCM CLI 0.8.0 on macOS and 0.8.0-rc.1 installed via https://github.com/open-component-model/open-component-model/issues/2585#issuecomment-4561953872.

`{"major":"0","minor":"8","patch":"0","gitVersion":"0.8.0","goVersion":"go1.26.3","compiler":"gc","platform":"darwin/arm64"}`

**To Reproduce**

OS: macOS Tahoe 26.5 (25F71) (darwin/arm64), shell: zsh

```bash
ocm transfer componentversion \
  --recursive \
  --copy-resources \
  europe-docker.pkg.dev/gardener-project/releases//github.com/gardener/gardener:v1.143.0 \
  ctf::/Users/aoe/Desktop/test.ctf
```
The affected external dependenciy is github.com/gardener/machine-controller-manager:v0.61.3 which contains no label:

```yaml
  transformGithubComGardenerMachineControllerManagerV0613:
    component:
      componentReferences: []
      creationTime: "2026-03-30T13:21:58Z"
      name: github.com/gardener/machine-controller-manager
      provider: SAP SE
      repositoryContexts:
        ...
      resources:
      - access:
        ...
      - access:
        ...
      - access:
        ...
      - access:
        ...
      sources:
      - access:
        ...
      version: v0.61.3
    meta:
```
but is being referenced inside `transformGithubComGardenerMachineControllerManagerV0613Upload`'s `spec.descriptor.component.labels`:
```yaml
- id: transformGithubComGardenerMachineControllerManagerV0613Upload
  spec:
    descriptor:
      component:
        componentReferences: ${environment.transformGithubComGardenerMachineControllerManagerV0613.component.componentReferences}
        labels: ${environment.transformGithubComGardenerMachineControllerManagerV0613.component.labels} # offending line
        ...
      ...
  type: CTFAddComponentVersion/v1alpha1
```

**Actual behavior**
```
Error: failed to process value for node transformGithubComGardenerMachineControllerManagerV0613Upload in graph: cannot compile expression "environment.transformGithubComGardenerMachineControllerManagerV0613.component.labels": ERROR: <input>:1:78: undefined field 'labels'
 | environment.transformGithubComGardenerMachineControllerManagerV0613.component.labels
 | .............................................................................^
environment:
```

**Expected behavior**

 The transfer should succeed. The OCM Component Descriptor v2 spec defines `component.label`
 **not** as a required key. A missing `component.label` is valid per spec and should be accepted.

**Additional Comments**

n/a

frewilhelm:

When transferring a component version a transfer spec is built that goes through the component descriptor. Usually, when a field is `nil` it is not added to the transfer spec. However, non-nil but empty fields (e.g. `labels: []`) are added, although, they cannot be resolved later. This PR fixes this by checking for the length of the object which covers nil and non-nil but empty fields.

Fixes #2742


frewilhelm:
@coderabbitai review

jneisener (dismissed):

bindings/go/transfer/internal/graph_test.go
```diff
@@ -716,6 +716,81 @@ func TestBuildDescriptorSpec_NoLabelsOmitted(t *testing.T) {
 	}
 }
 
+// Sibling of TestBuildDescriptorSpec_LabelsIncluded covering the
+// descriptor-level Signatures field (gated separately at graph.go:405). When
```
piotrjanik: nit: I would avoid pointing to line numbers, as they can change; I think mentioning the `buildDescriptorSpec` function would be safer

jakobmoellerdev:
```
However, non-nil but empty fields (e.g. labels: []) are added, although, they cannot be resolved later.
```

I think arguably this is a bug in the parser as they should resolve to an empty slice.

bindings/go/transfer/internal/graph_test.go
frewilhelm: Yeah, absolutely

jneisener (approved):

jakobmoellerdev (approved):

piotrjanik (approved):

PR #2781, opened 2026-06-08
fix(transfer): check len instead of nil for optional descriptor fields in buildDescriptorSpec

### LINKED ISSUE #2742: Transfer of components with no labels fails

jakobmoellerdev:

When a descriptor from an external registry has \`"labels":[]\` (or similar) in its v2 JSON, \`json.Unmarshal\` produces a non-nil empty slice. \`ConvertToV2\` preserves this, but \`json:"labels,omitempty"\` then omits it from the marshaled environment JSON — so the CEL environment has no \`labels\` field at all.

The old \`!= nil\` check in \`buildDescriptorSpec\` emitted \`\${environment.X.component.labels}\` for any non-nil slice, including empty ones. The CEL compiler then failed with:

\`\`\`
cannot compile expression "environment.X.component.labels": ERROR: undefined field 'labels'
\`\`\`

**Fix:** use \`len() != 0\` instead of \`!= nil\` for \`labels\`, \`repositoryContexts\`, \`sources\`, \`componentReferences\`, and \`signatures\` — so CEL placeholders are only emitted when the field actually has content and will therefore be present in the environment.

Fixes #2742

#### Test

Added \`TestBuildAndCheck_Reproduce2742_EmptyNonNilLabels\` in \`bindings/go/transfer/internal/graph_test.go\` which injects a descriptor with \`Labels = []Label{}\` through the mock resolver (bypassing CTF normalisation) and reproduces the exact CEL compile error from the issue.

### DIFF
--- a/bindings/go/transfer/internal/graph.go
+++ b/bindings/go/transfer/internal/graph.go
@@ -392,17 +392,17 @@ func buildDescriptorSpec(v2desc *descriptorv2.Descriptor, id string, resourceTra
 		"resources": resourcesArray,
 	}
 
-	setOptionalField(componentMap, "labels", id, v2desc.Component.Labels != nil)
-	setOptionalField(componentMap, "repositoryContexts", id, v2desc.Component.RepositoryContexts != nil)
-	setOptionalField(componentMap, "sources", id, v2desc.Component.Sources != nil)
-	setOptionalField(componentMap, "componentReferences", id, v2desc.Component.References != nil)
+	setOptionalField(componentMap, "labels", id, len(v2desc.Component.Labels) != 0)
+	setOptionalField(componentMap, "repositoryContexts", id, len(v2desc.Component.RepositoryContexts) != 0)
+	setOptionalField(componentMap, "sources", id, len(v2desc.Component.Sources) != 0)
+	setOptionalField(componentMap, "componentReferences", id, len(v2desc.Component.References) != 0)
 
 	descSpecMap := map[string]any{
 		"meta":      fmt.Sprintf("${environment.%s.meta}", id),
 		"component": componentMap,
 	}
 
-	if v2desc.Signatures != nil {
+	if len(v2desc.Signatures) != 0 {
 		descSpecMap["signatures"] = fmt.Sprintf("${environment.%s.signatures}", id)
 	}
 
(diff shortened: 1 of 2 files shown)


Skarlso:
Nice, looking good!

PR #2794, opened 2026-06-09
feat(http): pass HTTPConfig to OCI provider and Helm downloader

jakobmoellerdev:
## Summary

Replaces `WithHTTPClient(*http.Client)` with `WithHTTPConfig(*httpv1alpha1.Config)` across `bindings/go/oci` and `bindings/go/helm`. Each layer now builds its own `http.Client` from the serialisable config type internally, so external plugins can round-trip the config over the wire and reconstruct an equivalent client without taking a dependency on `bindings/go/http` themselves.

Prerequisite for the CLI integration PR (#2795).

### bindings/go/oci
- `provider.Options`: replace `HTTPClient *http.Client` with `HTTPConfig *httpv1alpha1.Config`
- Replace `WithHTTPClient` with `WithHTTPConfig(cfg *httpv1alpha1.Config)`
- `NewComponentVersionRepositoryProvider` builds the client internally via `ocmhttp.New(WithConfig, WithUserAgent)`
- Tests: verify HTTP traffic actually reaches the test server; verify timeout causes an error on a real network operation

### bindings/go/helm
- `download.option`: replace `HTTPClient` with `HTTPConfig`; build OCI registry client via `ocmhttp.New`, HTTP/S transport via `ocmhttp.NewTransport` using the global timeout config; apply same transport to `resolveHTTPChartURL` so index fetches also use the configured timeouts
- `helminput.InputMethod`: replace `HTTPClient` with `HTTPConfig`; thread to `GetV1HelmBlob`
- `blob.go` / `resource_repository.go`: propagate `*httpv1alpha1.Config` through the download chain

### go.mod (oci, helm)
- Add `bindings/go/http` as a direct dependency (published pseudo-version, no replace directives)
- `oci/integration`: add `bindings/go/http` as indirect dep for go.sum completeness

## Test plan

- [x] `go build ./...` + `go vet ./...` in `bindings/go/oci` passes
- [x] `go test ./repository/provider/...` in `bindings/go/oci` passes — custom config reaches test server, timeout surfaces on real network call
- [x] `go build ./...` + `go vet ./...` in `bindings/go/helm` passes
- [x] `go test ./internal/download/... ./repository/...` in `bindings/go/helm` passes
- [x] CI green (unit tests, integration tests, golangci-lint, code generation)


bindings/go/helm/internal/download/download.go
```diff
@@ -92,7 +93,18 @@ func NewReadOnlyChartFromRemote(ctx context.Context, helmRepo, targetDir string,
...
 
-	regClient, err := registry.NewClient()
+	var regClientOpts []registry.ClientOption
+	if opt.HTTPConfig != nil {
+		// OCI registry client: full retry+timeout client from config.
+		httpClient := ocmhttp.New(ocmhttp.WithConfig(opt.HTTPConfig))
+		regClientOpts = append(regClientOpts, registry.ClientOptHTTPClient(httpClient))
+		// HTTP/S getter needs a bare *http.Transport (helm API constraint).
+		// Use the global timeout config; per-host overrides are not supported here.
+		getterOpts = append(getterOpts, getter.WithTransport(
```
matthiasbruns: Why is it not supported? Can't we add another getter option that checks for the host and applies the config for that host there?

matthiasbruns (commented):
Lgtm. Just one question. I am currently on my phone so I couldn't check the code myself.

bindings/go/helm/internal/download/download.go
jakobmoellerdev: Good question. The constraint is Helm's `getter.WithTransport` API — it takes a `*http.Transport` specifically, not an `http.RoundTripper`. Our `hostRouter` (which does per-host dispatch in `ocmhttp`) is a `RoundTripper`, so it can't be passed there directly.

The OCI registry path already gets full per-host support because `ocmhttp.New` (which wraps `hostRouter` + retry) is passed directly as `registry.ClientOptHTTPClient` — that path goes through our whole transport chain.

For the HTTP/S getter path, we would need to either implement a custom `getter.Getter` wrapping an `http.Client` with our `hostRouter`, or accept the limitation and document it. I'd lean toward the latter to keep this PR's scope contained — per-host overrides on Helm HTTP/S chart repos is an edge case. Happy to track it as a follow-up. Updating the comment to be more precise.

jakobmoellerdev: implemented it so you can take a look. let me know what you think

bindings/go/helm/internal/download/download.go
```diff
@@ -188,14 +207,27 @@ func NewReadOnlyChartFromRemote(ctx context.Context, helmRepo, targetDir string,
 
 // GetterProviders returns the available getter providers.
 // This replaces the need for cli.New() and avoids the explosion of the dependency tree.
-func GetterProviders() getter.Providers {
+// When client is non-nil, the HTTP/S provider uses a custom getter backed by
+// that client so per-host routing and full timeout config apply on the HTTP/S
+// path as well as the OCI path. The opts struct carries the credential and
+// header values that httpConfigGetter needs, bypassing the opaque
+// getter.Option type.
+func GetterProviders(client *http.Client, opts HTTPConfigGetterOpts) getter.Providers {
```
Skarlso, line 220: I find something like this more readable:
```go
func GetterProviders(client *http.Client, opts HTTPConfigGetterOpts) getter.Providers {
	providers := getter.Providers{
		{
			Schemes: []string{"http", "https"},
			New: func(options ...getter.Option) (getter.Getter, error) {
				options = append(options, defaultOptions...)
				return getter.NewHTTPGetter(options...)
			},
		},
		{
			Schemes: []string{registry.OCIScheme},
			New:     getter.NewOCIGetter,
		},
	}

	if client != nil {
		providers[0].New = func(_ ...getter.Option) (getter.Getter, error) {
			return NewHTTPConfigGetter(client, opts)
		}
	}

	return providers
}

```

But totally up to you. :)

Skarlso (dismissed):
Looks okay, with a tiny tiny nit.

matthiasbruns (dismissed):
Good improvement :) thanks for trying the getter options thingy

Skarlso (approved):

matthiasbruns (approved):

PR #2808, opened 2026-06-10
feat: credential transformers (#847)

matthiasbruns:
This PR adds support for credential aware transformations by adding an optional interface `TransformerWithCredentials`.
This is being type-checked in the runtime.
If a transformation implements TransformerWithCredentials, the runtime will use the `TransformWithCredentials` injecting all credentials that are gathered by the identities provided through GetCredentialConsumerIdentities.
If during a credential resolution an error that is not `credentials.ErrNotFound` occurs, the runtime will error out.

Transformers implementing the new interface will have to implement two methods:

```go
type TransformerWithCredentials interface {
	GetCredentialConsumerIdentities(ctx context.Context, step runtime.Typed) (map[string]runtime.Identity, error)
	TransformWithCredentials(ctx context.Context, step runtime.Typed, credentials map[string]runtime.Typed) (runtime.Typed, error)
}
```

`GetCredentialConsumerIdentities` tells the runtime what identities the transformer handles and gets the credentials for said identities.

`TransformWithCredentials` should implement the transformation logic itself - comparable to `Transform` in the `Transformer` interface. The difference here is, that the runtime takes care of the credential resolution. The current transformers implementing `Transform` without credentials have to take care of this by themselves.

Credentials and identities share the same map key - called `slot` in the issue.
A transformer can get the resolved credentials via this key that it defined in the identity mapping.

Contributes: https://github.com/open-component-model/ocm-project/issues/847


matthiasbruns:
@coderabbitai review

matthiasbruns:
closed due to completely misunderstanding the scope 😅

PR #2822, opened 2026-06-11
feat(http): add InsecureSkipVerify support to HTTP config (global + per-host)

jakobmoellerdev:
## Summary

Implements [ocm-project#1065](https://github.com/open-component-model/ocm-project/issues/1065) — opt-in TLS skip-verify for development/testing with self-signed certificates.

- Adds `TLSConfig` struct with `InsecureSkipVerify *bool`, embedded inline in `Config` and `HostConfig` (flat YAML surface, consistent with `TimeoutConfig` pattern)
- Global: `insecureSkipVerify: true` disables TLS verification for all hosts
- Per-host: `hosts.<hostname>.insecureSkipVerify: true/false` overrides global; explicit `false` re-enables verification even when global is `true`
- `MergeTLSConfig`: last-non-nil-wins, mirrors `MergeTimeoutConfig`
- `NewTransportWithTLS`: allocates fresh `*tls.Config` (never mutates `DefaultTransport`), emits `slog.Warn` at build time
- `insecureWarnTransport`: thin `RoundTripper` decorator; emits `slog.WarnContext` on first connection per host (`sync.Once` prevents log floods)
- Security docs and ADR 0024 updated to reflect opt-in semantics

**Example config:**
```yaml
type: http.config.ocm.software/v1alpha1
insecureSkipVerify: true          # global
hosts:
  registry.example.com:
    insecureSkipVerify: true      # per-host
  trusted.registry.io:
    insecureSkipVerify: false     # re-enable even if global is true
```

## Test plan

- [ ] `TestMergeTLSConfig` — nil/dst/src/override/false-overrides-true
- [ ] `TestConfig_ParseYAML_TLS` — round-trips at global and per-host scope
- [ ] `TestMerge_TLS` — multi-layer merge
- [ ] `TestNewTransportWithTLS` — skip=true sets `TLSClientConfig.InsecureSkipVerify`; `DefaultTransport` not mutated
- [ ] `TestNewTransportWithTLS_ConstructionWarning` — slog.Warn fires at build time
- [ ] `TestInsecureWarnTransport` — first request logs, second to same host does not
- [ ] `TestNewClient_InsecureSkipVerify` — global wraps with warn transport; per-host scoped
- [ ] Integration test: `httptest.NewTLSServer` + default client fails; skip=true client succeeds + warning emitted
- [ ] Security reviewer check before merge


matthiasbruns:
@coderabbitai can you recheck your comments if they are fixed?

bindings/go/http/integration/tls_integration_test.go
```diff
@@ -0,0 +1,76 @@
...
+			TLSConfig: httpv1alpha1.TLSConfig{InsecureSkipVerify: &tr},
+		}
+		c := ocmhttp.NewClient(cfg)
+		resp, err := c.Get(srv.URL)
+		require.NoError(t, err, "client with InsecureSkipVerify=true must succeed against self-signed cert")
+		resp.Body.Close()
+		assert.Equal(t, nethttp.StatusOK, resp.StatusCode)
+
+		msgs := capture.warnMessages()
+		assert.NotEmpty(t, msgs, "warning must be emitted when InsecureSkipVerify=true")
```
matthiasbruns, line 74: can we check for the warning itself? this could be any warning in the call-chain otherwise

bindings/go/http/spec/config/v1alpha1/config.go
```diff
@@ -155,12 +155,142 @@ func MergeTimeoutConfig(dst, src *TimeoutConfig) TimeoutConfig {
...
+//	retry:
+//	  maxRetries: 3      # 0 = infinite; -1 = disable; nil uses default (5)
+//	  minWait: 100ms
+//	  maxWait: 5s
+//
+// +k8s:deepcopy-gen=true
+type RetryConfig struct {
+	// MaxRetries is the maximum number of retry attempts after the initial
+	// request. Zero means infinite retries. -1 disables retry entirely.
+	// Nil uses the library default (5).
```
matthiasbruns, line 208: could this number become stale if the lib defaults change? maybe just pointing to it would be better instead

bindings/go/http/spec/config/v1alpha1/config.go
```diff
@@ -155,12 +155,142 @@ func MergeTimeoutConfig(dst, src *TimeoutConfig) TimeoutConfig {
...
+
+	// Retry overrides the global retry policy for this host.
+	// Fields set here override the corresponding top-level retry value.
+	Retry *RetryConfig `json:"retry,omitempty"`
+}
+
+// Validate checks the per-host timeout and retry config for valid values.
+func (h *HostConfig) Validate() error {
+	if err := h.TimeoutConfig.Validate(); err != nil {
+		return err
```
matthiasbruns: we should wrap this error, shouldn't we?

bindings/go/http/spec/config/v1alpha1/config.go
```diff
@@ -155,12 +155,142 @@ func MergeTimeoutConfig(dst, src *TimeoutConfig) TimeoutConfig {
...
+	}
+	if h.Retry != nil {
+		if err := h.Retry.Validate(); err != nil {
+			return fmt.Errorf("retry: %w", err)
```
matthiasbruns: this could use some more infos

bindings/go/http/insecure_warn_transport.go
```diff
@@ -0,0 +1,26 @@
...
+// sync.Once per host prevents log floods in long-running services.
+type insecureWarnTransport struct {
+	base nethttp.RoundTripper
+	once sync.Map // host (string) -> *sync.Once
+}
+
+func (t *insecureWarnTransport) RoundTrip(req *nethttp.Request) (*nethttp.Response, error) {
+	host := req.URL.Host
+	v, _ := t.once.LoadOrStore(host, &sync.Once{})
+	v.(*sync.Once).Do(func() {
```
matthiasbruns: could this be used on controllers? if so, wouldn't this grow during the lifetime of controllers?

**if** this will be used in controllers as well, we might want to have a way to opt-out of this

bindings/go/http/spec/config/v1alpha1/config.go
jakobmoellerdev: it could but I want us to own our behavior here. I think its important that we fully control this. if upstream changes, we probably dont want to automatically follow.

bindings/go/http/insecure_warn_transport.go
jakobmoellerdev: I will think about this

jakobmoellerdev: Each `NewClient` call constructs a fresh `insecureWarnTransport` instance scoped to a single host chain — the `sync.Once` lives on the instance, not in a global map. A controller that creates a new client per reconcile gets a new transport with a fresh `sync.Once`; one that reuses a single client instance fires the warning once total. Neither case causes unbounded growth. If we later introduce a long-lived shared client across reconcile loops we would revisit, but the current construction model makes it a non-issue.

bindings/go/helm/transformation/spec/v1alpha1/schemas/GetHelmChart.schema.json
```diff
@@ -111,8 +92,9 @@
       "description": "ElementMeta defines an object with name and version containing labels.",
       "properties": {
         "extraIdentity": {
-          "$ref": "#/$defs/ocm.software.open-component-model.bindings.go.runtime.Identity",
-          "description": "ExtraIdentity is the identity of an object.\nAn additional label with key \"name\" is not allowed"
+          "type": "object",
+          "description": "ExtraIdentity is the identity of an object.\nAn additional label with key \"name\" is not allowed",
+          "additionalProperties": true
         },
```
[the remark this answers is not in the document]
jakobmoellerdev: This file was not modified by this PR — the `propertyNames` / reserved-key guard concern is pre-existing and not introduced here. Will open a separate issue to track it so this feature PR is not blocked.

cli/internal/plugin/spec/config/v2alpha1/schemas/Config.schema.json
```diff
@@ -18,7 +18,8 @@
       }
     },
     "type": {
-      "$ref": "#/$defs/ocm.software.open-component-model.bindings.go.runtime.Type"
+      "type": "object",
+      "additionalProperties": true
     }
```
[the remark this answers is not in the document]
jakobmoellerdev: This file was not modified by this PR either — the discriminator erosion concern is pre-existing. Will track separately.

matthiasbruns (dismissed):

matthiasbruns (dismissed):

bindings/go/http/spec/config/v1alpha1/config.go
```diff
@@ -246,11 +283,11 @@ type HostConfig struct {
 // Validate checks the per-host timeout and retry config for valid values.
 func (h *HostConfig) Validate() error {
 	if err := h.TimeoutConfig.Validate(); err != nil {
-		return err
+		return fmt.Errorf("timeout: %w", err)
```
piotrjanik: nit: "invalid timeout config: ...", to match retry config error msg?

piotrjanik (dismissed):
LGTM, one nit

piotrjanik (approved):

matthiasbruns (approved):

PR #2811, opened 2026-06-10
chore(docs): typed credential system

matthiasbruns:
#### What this PR does / why we need it
Updates existing credential docs and adds new docs for the typed credential migration

#### Which issue(s) this PR fixes
Fixes: https://github.com/open-component-model/ocm-project/issues/1122

Preview Link: https://deploy-preview-2811--ocm-website.netlify.app/

PR #2795, opened 2026-06-09
feat(http): wire HTTP config from OCM config into CLI builtin plugins

jakobmoellerdev:
## Summary

Threads `http.config.ocm.software/v1alpha1` from the user's OCM config file through the CLI so that timeout, per-host override, and User-Agent settings apply to all outbound HTTP traffic (OCI registry and Helm downloads).

- `cli/cmd/setup`: resolve HTTP config via `httpv1alpha1.ResolveHTTPConfig`, pass to `builtin.Register`
- `cli/internal/plugin/builtin`: thread `httpConfig` to OCI provider (`WithHTTPConfig`) and Helm input/resource plugins
- `bindings/go/examples`: step 8 — HTTP config tour (defaults, global timeout, per-host overrides, OCI wiring)
- `website/content/docs/how-to`: new guide for constrained environments (corporate proxy, air-gap, high-latency registries)

Depends on #2794 (merged).

fix https://github.com/open-component-model/ocm-project/issues/1052

PR #2824, opened 2026-06-11
fix(ci): poll netlify deploy-preview commit status instead of check suite

matthiasbruns:

The `Verify links (website)` job failed on practically every website PR (#2795, #2811), looping on `No Netlify suite found. Waiting...` until its 10 minute timeout. Two causes:

- The step polled Netlify's **check suite**, which appears minutes *after* the `netlify/ocm-website/deploy-preview` **commit status** (and sometimes never). The step now polls the commit status directly — the earliest signal, which already carries the preview URL — and fails fast on `failure`/`error`.
- A transient HTTP 401 in the `gh pr diff` based changes check silently set `has_website_changes=false`, skipping the whole link check while reporting success. The check now uses `git diff HEAD^1 HEAD` on the local merge commit, with no API dependency.

The timeout is raised to 30 minutes: a single deploy takes ~6 minutes, but Netlify builds one deploy at a time, so push bursts queue up longer delays. In extreme bursts the queue can exceed even that — re-running the job once the queue drains covers those cases.

Fixes the flaky Netlify deploy-preview wait in the Markdown workflow.

##### How to test the changes

Dry-run the new poll loop against PR #2811's head commit (`94c03e2`) — it resolves the preview URL immediately, where the old loop still finds no check suite. Re-run `Verify links (website)` on a website PR after merge for end-to-end validation.

##### Verification

- [x] Workflow passes `actionlint` and `zizmor`
- [x] New poll loop and changes check dry-run verified against PR #2811 data
- [x] Skip path (no website changes) validated by this PR's own `Verify links (website)` run

🤖 Generated with [Claude Code](https://claude.com/claude-code)


jakobmoellerdev (dismissed):

matthiasbruns:
fasted review ever :D @jakobmoellerdev

jakobmoellerdev (approved):

jakobmoellerdev:
:sonicgoesfast:

jneisener (approved):

PR #2833, opened 2026-06-11
feat(oci): blob and resolution cache

jakobmoellerdev:

```go
// Package cache provides two disk-backed, TTL+LRU caches that layer
// on top of an oras [*remote.Repository] via [Repository]:
//
//   - [BlobCache] keys cached blobs by digest and stores each blob as
//     a file under `<Options.Dir>/blobs/<algo>/<hex>`. Its primary
//     consumer is [BlobCache.Fetch], which mirrors the oras-go
//     internal/cas/proxy.go pattern: the first Fetch of a descriptor
//     reads from upstream and tees the bytes to disk; subsequent
//     Fetches for the same digest serve the on-disk file. Push, Tag,
//     Resolve, FetchReference, and Exists on the wrapping
//     [Repository] are pure passthroughs.
//
//   - [ReferenceCache] keys cached resolves by (namespace, reference)
//     so two repositories that happen to share a short reference
//     (e.g. the tag "v1") cannot collide. Each namespace's entries
//     are persisted to its own file under
//     `<Options.Dir>/refs/<fnv1a(namespace)>.json`; the FNV-1a hash
//     is used purely to derive a compact filename — the canonical
//     namespace string is stored inside the snapshot body, so any
//     unlikely hash collision is recoverable.
```

This fixes various performance issues around recurring fetching of component versions, listing component versions and latency around fetching those.

#### Testing

I ran this against remote repositories with my local cli build

##### How to test the changes

Required files to test the changes:

cli/internal/plugin/builtin/oci/register.go
```go
cacheOpts := cache.Defaults()
	CachingComponentVersionRepositoryProvider := provider.NewComponentVersionRepositoryProvider(
		provider.WithTempDir(filesystemConfig.TempFolder),
		provider.WithUserAgent(creator),
		provider.WithHTTPConfig(httpConfig),
		provider.WithBlobCacheOptions(cacheOpts),
		provider.WithReferenceCacheOptions(cacheOpts),
	)
```

```bash
ocm get cv xxx

ocm transfer xxx
```
-->

##### Verification

### DIFF
--- a/bindings/go/oci/cache/blob_cache.go
+++ b/bindings/go/oci/cache/blob_cache.go
@@ -1,0 +1,298 @@
+package cache
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"log/slog"
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/hashicorp/golang-lru/v2/expirable"
+	"github.com/opencontainers/go-digest"
+	ociImageSpecV1 "github.com/opencontainers/image-spec/specs-go/v1"
+	"golang.org/x/sync/singleflight"
+	"oras.land/oras-go/v2/content"
+)
+
+// blobEntry is the LRU value for [BlobCache]. The on-disk path is
+// derived deterministically from the digest key via [pathFor]; storing
+// it on the entry keeps the eviction callback self-contained even if
+// the layout function changes.
+type blobEntry struct {
+	path string
+	size int64
+}
+
+// BlobCache is a digest-keyed, disk-backed LRU+TTL cache for small
+// immutable blobs (manifests and OCM component descriptors). All
+// methods are safe for concurrent use.
+//
+// On startup, [NewBlobCache] reseeds the LRU from any well-formed
+// digest-keyed files already present under [Options.Dir], so cached
+// entries survive a process restart with the same Dir. Stray files
+// (unparseable names, leftover .incoming-* temp files) are removed
+// during the seed pass.
+//
+// See package documentation for the design rationale and the link to
+// oras-go's proxy.go pattern that this cache is built to support.
+type BlobCache struct {
+	opts   Options
+	logger *slog.Logger
+	lru    *expirable.LRU[digest.Digest, blobEntry]
+
+	// sf collapses concurrent upstream fetches for the same digest into
+	// a single round-trip. Because cached blobs are small (manifests
+	// and descriptors, bounded by [Options.MaxBlobSize]), the shared
+	// result is buffered in memory and each waiter is served its own
+	// reader over those bytes.
+	sf singleflight.Group
+}
+
+// NewBlobCache constructs a [BlobCache]. [Options.Dir] is required;
+// other zero-valued fields fall back to [Defaults].
+//
+// The cache stores its files under `<Options.Dir>/blobs/<algo>/<hex>`
+// so the same Dir can also host a [ReferenceCache] (which writes
+// `<Dir>/refs/<sha256(namespace)>.json`) without collisions.
+func NewBlobCache(opts Options) (*BlobCache, error) {
+	opts, err := opts.applyDefaults()
+	if err != nil {
+		return nil, err
+	}
+	// Scope to a dedicated subdirectory so a shared Options.Dir does
+	// not entangle blob layout with anything else the caller stores
+	// at that path.
+	opts.Dir = filepath.Join(opts.Dir, "blobs")
+	if err := ensureDir(opts.Dir); err != nil {
+		return nil, fmt.Errorf("blobcache: %w", err)
+	}
+
+	c := &BlobCache{opts: opts, logger: slog.Default().With(slog.String("dir", opts.Dir))}
+	c.lru = expirable.NewLRU(opts.MaxEntries, c.onEvict, opts.TTL)
+
+	startScan := time.Now()
+	existing, err := scanExisting(opts.Dir)
+	if err != nil {
+		return nil, fmt.Errorf("blobcache: scan existing entries: %w", err)
+	}
+	totalSize := int64(0)
+	for _, e := range existing {
+		c.lru.Add(e.digest, blobEntry{path: e.path, size: e.size})
+		totalSize += e.size
+	}
+	if len(existing) > 0 {
+		c.logger.Debug("blobcache: seeded from disk",
+			slog.Int("count", len(existing)),
+			slog.Duration("duration", time.Since(startScan)),
+			slog.Int64("size", totalSize))
+	}
+	return c, nil
+}
+
+// Accept reports whether desc should be cached, per [Options.Accept].
+// The full descriptor is fed to the admission filter (not just the
+// media type) so future extensions can key on annotations, size, or
+// artifact type without a breaking signature change.
+func (c *BlobCache) Accept(desc ociImageSpecV1.Descriptor) bool {
+	return c.opts.Accept(desc)
+}
+
+// MaxBlobSize returns the configured per-blob size cap (0 means no cap).
+func (c *BlobCache) MaxBlobSize() int64 {
+	return c.opts.MaxBlobSize
+}
+
+// Has reports whether dgst is currently cached without affecting LRU
+// recency.
+func (c *BlobCache) Has(dgst digest.Digest) bool {
+	_, ok := c.lru.Peek(dgst)
+	return ok
+}
+
+// Get returns an open file positioned at offset 0 if dgst is cached.
+// The caller is responsible for closing the returned file. A miss
+// returns (nil, false, nil).
+//
+// If the LRU has the entry but the on-disk file is missing (e.g. the
+// user removed the cache directory), the entry is evicted and the
+// call returns a miss.
+func (c *BlobCache) Get(dgst digest.Digest) (*os.File, bool, error) {
+	e, ok := c.lru.Get(dgst)
+	if !ok {
+		return nil, false, nil
+	}
+	f, err := os.Open(e.path)
+	if err != nil {
+		c.lru.Remove(dgst)
+		if errors.Is(err, os.ErrNotExist) {
+			return nil, false, nil
+		}
+		return nil, false, fmt.Errorf("blobcache: open cached file: %w", err)
+	}
+	return f, true, nil
+}
+
+// populate writes r to disk under dgst and inserts the entry into the
+// LRU. It is a no-op (returns false, nil) when:
+//
+//   - [Options.MaxBlobSize] > 0 and size exceeds it.
+//   - The blob copied does not match size (when size > 0). This guards
+//     against truncated reads from early-Close on the upstream reader.
+//
+// Concurrent calls for the same digest race on the final [os.Rename];
+// the last winner wins. Both yield byte-identical content (the digest
+// is the key) so the LRU bookkeeping is harmless. Callers that want
+// to avoid duplicate upstream fetches must dedupe at a higher layer
+// (see [BlobCache.Fetch], which uses singleflight for exactly this).
+//
+// The boolean return value reports whether the blob was actually
+// inserted into the LRU (true) or skipped (false).
+func (c *BlobCache) populate(ctx context.Context, dgst digest.Digest, size int64, r io.Reader) (bool, error) {
+	if c.opts.MaxBlobSize > 0 && size > c.opts.MaxBlobSize {
+		c.logger.DebugContext(ctx, "blobcache: skipping oversized blob",
+			slog.String("digest", dgst.String()),
+			slog.Int64("size", size),
+			slog.Int64("max", c.opts.MaxBlobSize))
+		// Drain r so the upstream pipe writer doesn't block forever
+		// waiting for someone to read it.
+		_, _ = io.Copy(io.Discard, r)
+		return false, nil
+	}
+
+	path, n, err := writeAtomic(c.opts.Dir, dgst, c.opts.MaxBlobSize, r)
+	if err != nil {
+		return false, err
+	}
+	if size > 0 && n != size {
+		if rerr := removeQuiet(path); rerr != nil {
+			c.logger.WarnContext(ctx, "blobcache: remove truncated entry failed",
+				slog.String("digest", dgst.String()), slog.String("err", rerr.Error()))
+		}
+		return false, fmt.Errorf("blobcache: short read: got %d, want %d", n, size)
+	}
+	c.lru.Add(dgst, blobEntry{path: path, size: n})
+	c.logger.DebugContext(ctx, "blobcache: populated",
+		slog.String("digest", dgst.String()), slog.Int64("size", n))
+	return true, nil
+}
+
+// Fetch is the cache-aware Fetch primitive used by store
+// implementations that want to layer the cache in front of an
+// upstream [content.ReadOnlyStorage]. A cache hit returns the on-disk
+// file directly; a miss reads the (small) blob from upstream
+// verifying its digest, records it to disk, and serves the caller
+// from the in-memory copy.
+//
+// Fetch honors [Options.Accept] and [Options.MaxBlobSize]: if the
+// descriptor fails either filter (or c is nil), the call falls
+// straight through to upstream.Fetch and nothing is buffered. In the
+// oversize case a debug log is emitted so operators can see why the
+// cache did not engage.
+//
+// Concurrent Fetches for the same digest are collapsed via
+// singleflight into a single upstream.Fetch; every caller receives an
+// independent reader over identical bytes.
+func (c *BlobCache) Fetch(ctx context.Context, upstream content.ReadOnlyStorage, target ociImageSpecV1.Descriptor) (io.ReadCloser, error) {
+	if c == nil || !c.Accept(target) {
+		return upstream.Fetch(ctx, target)
+	}
+	if c.MaxBlobSize() > 0 && target.Size > c.MaxBlobSize() {
+		c.logger.DebugContext(ctx, "blobcache: skipping oversized descriptor, falling through",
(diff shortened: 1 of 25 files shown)


bindings/go/oci/cache/blob_cache.go
```diff
@@ -0,0 +1,258 @@
...
+	logger *slog.Logger
+	lru    *expirable.LRU[digest.Digest, blobEntry]
+}
+
+// NewBlobCache constructs a [BlobCache]. [Options.Dir] is required;
+// other zero-valued fields fall back to [Defaults].
+//
+// The cache stores its files under `<Options.Dir>/blobs/<algo>/<hex>`
+// so the same Dir can also host a [ReferenceCache] (which writes
+// `<Dir>/references.json`) without collisions.
```
frewilhelm: Is it also `OptionsDir` or is it another repository?

bindings/go/oci/cache/blob_cache.go
```diff
@@ -0,0 +1,258 @@
...
+//
+// On startup, [NewBlobCache] reseeds the LRU from any well-formed
+// digest-keyed files already present under [Options.Dir], so cached
+// entries survive a process restart with the same Dir. Stray files
+// (unparseable names, leftover .incoming-* temp files) are removed
+// during the seed pass.
+//
+// See package documentation for the design rationale and the link to
+// oras-go's proxy.go pattern that this cache is built to support.
+type BlobCache struct {
```
matthiasbruns, line 42: couldn't this be something in the blob bindings that could be used ocm-wide?

bindings/go/oci/cache/blob_cache.go
```diff
@@ -0,0 +1,258 @@
...
+// is the key) so the LRU bookkeeping is harmless. Callers that want
+// to avoid duplicate upstream fetches must dedupe at a higher layer.
+//
+// The boolean return value reports whether the blob was actually
+// inserted into the LRU (true) or skipped (false).
+func (c *BlobCache) Populate(ctx context.Context, dgst digest.Digest, size int64, r io.Reader) (bool, error) {
+	if c.opts.MaxBlobSize > 0 && size > c.opts.MaxBlobSize {
+		// Drain r so the upstream pipe writer doesn't block forever
+		// waiting for someone to read it.
+		_, _ = io.Copy(io.Discard, r)
```
frewilhelm, line 162: I think a debug log would be beneficial here, so users can be aware that the cache is not used because the size was too large.

bindings/go/oci/cache/blob_cache.go
jakobmoellerdev: where would we have another use for this? Im wondering because its a very specific implementation that is targeted for a specific oci use case

matthiasbruns: yeah thinking about this you might be right - it felt that we could have a generic caching layer somewhere central instead of a specialized one

bindings/go/oci/repository/provider/provider_test.go
```diff
@@ -237,3 +241,34 @@ func TestWithHTTPConfig_ShortTimeoutCausesError(t *testing.T) {
 	_, err = repo.ListComponentVersions(t.Context(), "example.org/component")
 	require.Error(t, err)
 }
+
+// TestCacheIsolation_ReferenceScopeKeys verifies the ScopeKey invariants used
+// by ReferenceCache partitioning:
+//   - different credentials → different keys (isolated ref caches)
+//   - same credentials → same key (single ref cache reused)
+//   - different registries → different keys
+func TestCacheIsolation_ReferenceScopeKeys(t *testing.T) {
```
frewilhelm: This looks like a test for `ScopeKey` only which should be in `cache/scope.go`

bindings/go/oci/cache/reference_cache.go
```diff
@@ -0,0 +1,407 @@
...
+}
+
+// referenceKey is the in-memory LRU key. Splitting namespace and
+// reference into struct fields means we don't have to choose a
+// separator that escapes safely — Go's map equality covers it.
+type referenceKey struct {
+	namespace string
+	reference string
+}
+
```
fabianburth, line 100: This key contains redundant information by design, doesn't it? `namespace` constructed completely based on reference (in fact, it is a subset of the information in reference). So, it does not add any value to the uniqueness of the key.

bindings/go/oci/cache/doc.go
```diff
@@ -0,0 +1,64 @@
...
+//     reads from upstream and tees the bytes to disk; subsequent
+//     Fetches for the same digest serve the on-disk file. Push, Tag,
+//     Resolve, FetchReference, and Exists on the wrapping
+//     [Repository] are pure passthroughs.
+//
+//   - [ReferenceCache] keys cached resolves by (namespace, reference)
+//     so two repositories that happen to share a short reference
+//     (e.g. the tag "v1") cannot collide. Each namespace's entries
+//     are persisted to its own file under
+//     `<Options.Dir>/refs/<fnv1a(namespace)>.json`; the FNV-1a hash
```
fabianburth: I think the scope key for oci credentials should be mentioned here. Besides, I think you ended up using sha256 instead of fnv-1a.

bindings/go/oci/cache/blob_cache.go
```diff
@@ -0,0 +1,250 @@
...
+	// reader over those bytes.
+	sf singleflight.Group
+}
+
+// NewBlobCache constructs a [BlobCache]. [Options.Dir] is required;
+// other zero-valued fields fall back to [Defaults].
+//
+// The cache stores its files under `<Options.Dir>/blobs/<algo>/<hex>`
+// so the same Dir can also host a [ReferenceCache] (which writes
+// `<Dir>/refs/<fnv1a(namespace)>.json`) without collisions.
```
fabianburth: same here, fnv-1a is not actually used

bindings/go/oci/cache/reference_cache.go
```diff
@@ -0,0 +1,407 @@
...
+// referenceEntry is the on-disk shape of a single cached resolve.
+// SavedAt records when the entry was first resolved so a reseed can
+// honour [Options.TTL] instead of resurrecting arbitrarily stale
+// mappings with a fresh in-memory expiry.
+type referenceEntry struct {
+	Descriptor ociImageSpecV1.Descriptor `json:"descriptor"`
+	SavedAt    time.Time                 `json:"savedAt"`
+}
+
```
fabianburth, line 51: `referenceEntry` and `referenceValue` are the exact same data structure. Is the semantic difference of one being persisted while the other is stored in memory worth keeping both?

bindings/go/oci/repository/provider/provider.go
```diff
@@ -191,6 +237,67 @@ func (b *CachingComponentVersionRepositoryProvider) GetComponentVersionRepositor
...
+		if err != nil {
+			slog.Warn("provider: failed to initialise shared blob cache, continuing without caching",
+				slog.String("err", err.Error()))
+			return
+		}
+		b.sharedBlobCache = c
+	})
+	return b.sharedBlobCache
+}
+
```
fabianburth, line 293: I'm wondering, if we would also want to use this cache for resources eventually, shouldn't this also have a credential scope? Especially in the controllers, I could imagine a situation where we have 2 Resource CRs, 1 has sufficient credentials to fetch the resource while 2 does not. Now 2 might be able to get it from the cache anyway.

bindings/go/oci/cache/blob_cache.go
```diff
@@ -0,0 +1,250 @@
...
+// Accept reports whether mediaType should be cached, per [Options.Accept].
+func (c *BlobCache) Accept(mediaType string) bool {
+	return c.opts.Accept(mediaType)
+}
+
+// MaxBlobSize returns the configured per-blob size cap (0 means no cap).
+func (c *BlobCache) MaxBlobSize() int64 {
+	return c.opts.MaxBlobSize
+}
+
```
fabianburth: We could also just have a single `Accept(descriptor ociImageSpecV1.Descriptor)` function. Would be safe for future extension.

bindings/go/oci/cache/blob_cache.go
```diff
@@ -0,0 +1,250 @@
...
+	// sf collapses concurrent upstream fetches for the same digest into
+	// a single round-trip. Because cached blobs are small (manifests
+	// and descriptors, bounded by [Options.MaxBlobSize]), the shared
+	// result is buffered in memory and each waiter is served its own
+	// reader over those bytes.
+	sf singleflight.Group
+}
+
+// NewBlobCache constructs a [BlobCache]. [Options.Dir] is required;
```
fabianburth, line 55: > [Options.Dir] is required

Is it still an option then?

bindings/go/oci/cache/blob_cache.go
```diff
@@ -0,0 +1,250 @@
...
+	return true, nil
+}
+
+// Fetch is the cache-aware Fetch primitive used by store
+// implementations that want to layer the cache in front of an
+// upstream [content.ReadOnlyStorage]. A cache hit returns the on-disk
+// file directly; a miss reads the (small) blob from upstream, tees it
+// to disk, and serves the caller from the in-memory copy.
+//
+// Fetch honours [Options.Accept] and [Options.MaxBlobSize]: if the
```
fabianburth: *honors

bindings/go/oci/cache/blob_cache.go
```diff
@@ -0,0 +1,250 @@
...
+	c.logger.DebugContext(ctx, "blobcache: populated",
+		slog.String("digest", dgst.String()), slog.Int64("size", n))
+	return true, nil
+}
+
+// Fetch is the cache-aware Fetch primitive used by store
+// implementations that want to layer the cache in front of an
+// upstream [content.ReadOnlyStorage]. A cache hit returns the on-disk
+// file directly; a miss reads the (small) blob from upstream, tees it
+// to disk, and serves the caller from the in-memory copy.
```
fabianburth: > tees it to disk

That is not what is currently happening in the function. Currently, it buffers the whole blob (which is fine as long as we never use it for layers)!
```go
		data, err := io.ReadAll(upstreamRC)
		if err != nil {
			return nil, fmt.Errorf("blobcache: read upstream: %w", err)
		}
```

Also, should we consider using a limit reader or would that be too defensive?

bindings/go/oci/cache/blob_cache.go
```diff
@@ -0,0 +1,250 @@
...
+		c.logger.DebugContext(ctx, "blobcache: skipping oversized blob",
+			slog.String("digest", dgst.String()),
+			slog.Int64("size", size),
+			slog.Int64("max", c.opts.MaxBlobSize))
+		// Drain r so the upstream pipe writer doesn't block forever
+		// waiting for someone to read it.
+		_, _ = io.Copy(io.Discard, r)
+		return false, nil
+	}
+
```
fabianburth: This function should probably not be exported. Currently it is only used within `Fetch` anyway, so we have no immediate reason to export it.

It bypasses the media type check and it has to perform a kind of redundant `MaxBlobSize` check.

bindings/go/oci/cache/blob_cache.go
```diff
@@ -0,0 +1,250 @@
...
+		upstreamRC, err := upstream.Fetch(ctx, target)
+		if err != nil {
+			return nil, err
+		}
+		defer func() { _ = upstreamRC.Close() }()
+
+		data, err := io.ReadAll(upstreamRC)
+		if err != nil {
+			return nil, fmt.Errorf("blobcache: read upstream: %w", err)
+		}
```
fabianburth: Our cache never verifies that the bytes actually hash to the claimed digest. If we `ReadAll` anyway, we can collapse this code to `oras.FetchAll` which also noodles the content through a `VerifyReader`.

bindings/go/oci/cache/disk.go
```diff
@@ -0,0 +1,167 @@
...
+		_ = tmp.Close()
+		_ = os.Remove(tmpName)
+	}
+
+	copyR := r
+	if maxSize > 0 {
+		// LimitReader stops at max+1 so we can detect oversize without
+		// reading unbounded amounts.
+		copyR = io.LimitReader(r, maxSize+1)
+	}
```
fabianburth: In the current implementation, this `LimitReader` is too late. We are already buffering the whole content in memory.
(Also, I don't see the purpose of `copyR := r`)

bindings/go/oci/cache/disk.go
```diff
@@ -0,0 +1,167 @@
...
+// and the number of bytes copied.
+//
+// If max > 0 and the copied byte count exceeds max, the temp file is
+// removed and an error is returned. The destination file is never
+// created in that case.
+//
+// Concurrent writers for the same digest race on the rename; the last
+// rename wins. Both yield identical content, so the LRU bookkeeping is
+// expected to dedupe via singleflight before reaching writeAtomic.
+func writeAtomic(dir string, dgst digest.Digest, maxSize int64, r io.Reader) (path string, n int64, err error) {
```
fabianburth: I don't think the named return parameters are needed here

fabianburth (changes requested):
Generally solid implementation! (Did not have a closer look at the reference cache yet)

bindings/go/oci/cache/reference_cache.go
fabianburth: likely obsolete

bindings/go/oci/cache/blob_cache.go
fabianburth: we are using verify readers upstream

bindings/go/oci/cache/blob_cache.go
```diff
@@ -205,23 +218,20 @@ func (c *BlobCache) Fetch(ctx context.Context, upstream content.ReadOnlyStorage,
...
+	// every waiter is served an independent reader over that copy.
 	v, err, _ := c.sf.Do(target.Digest.String(), func() (any, error) {
-		upstreamRC, err := upstream.Fetch(ctx, target)
-		if err != nil {
-			return nil, err
-		}
-		defer func() { _ = upstreamRC.Close() }()
-
-		data, err := io.ReadAll(upstreamRC)
+		data, err := content.FetchAll(ctx, upstream, target)
```
fabianburth: didn't we want to stream here?

bindings/go/oci/cache/blob_cache.go
```diff
@@ -179,19 +183,28 @@ func (c *BlobCache) Populate(ctx context.Context, dgst digest.Digest, size int64
...
 // descriptor fails either filter (or c is nil), the call falls
-// straight through to upstream.Fetch and nothing is buffered.
+// straight through to upstream.Fetch and nothing is buffered. In the
+// oversize case a debug log is emitted so operators can see why the
+// cache did not engage.
 //
 // Concurrent Fetches for the same digest are collapsed via
 // singleflight into a single upstream.Fetch; every caller receives an
 // independent reader over identical bytes.
 func (c *BlobCache) Fetch(ctx context.Context, upstream content.ReadOnlyStorage, target ociImageSpecV1.Descriptor) (io.ReadCloser, error) {
```
fabianburth: if we hit the cache, we verify, otherwise we don't. seems unreliable somehow

bindings/go/oci/cache/repository.go
```diff
@@ -0,0 +1,117 @@
...
+// Fetch consults [Repository.BlobCache] before delegating to the
+// embedded [*remote.Repository]. See [BlobCache.Fetch] for the exact
+// semantics: cache hit returns the on-disk file directly; miss
+// performs the upstream fetch and tees into the cache. Layer blobs
+// and other non-manifest media types pass through transparently.
+//
+// When BlobCache is nil, this is a direct passthrough.
+func (r *Repository) Fetch(ctx context.Context, target ociImageSpecV1.Descriptor) (io.ReadCloser, error) {
+	return r.BlobCache.Fetch(ctx, r.Repository, target)
+}
```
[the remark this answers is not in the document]
fabianburth: Check the lines in blob cache. 
```
func (c *BlobCache) Fetch(ctx context.Context, upstream content.ReadOnlyStorage, target ociImageSpecV1.Descriptor) (io.ReadCloser, error) {
	if c == nil || !c.Accept(target) {
		return upstream.Fetch(ctx, target)
	}
```

fabianburth (dismissed):

bindings/go/oci/repository/provider/provider.go
```diff
@@ -59,6 +66,45 @@ type CachingComponentVersionRepositoryProvider struct {
...
+	// access-controlled; once you hold a digest you are authorised.
+	//
+	// Note: while an attacker who *guesses* a digest could observe its
+	// existence in the cache, they must first obtain the digest via a
+	// [cache.ReferenceCache.Resolve] under credentials that entitle them
+	// to see the tag — which IS credential-scoped. When we grow the
+	// blob cache to cover resource layers, callers with weaker
+	// credentials must not be able to learn digests from a stronger
+	// scope, so revisit this decision if the cache starts holding
+	// layer blobs (see open-component-model#2833 discussion).
```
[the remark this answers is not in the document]
fabianburth: changed the mechanisms here to an kubernetes imagePullPolicy like pattern for both blob cache and reference cache, please check again!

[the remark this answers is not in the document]
fabianburth: changed the default to always

fabianburth (dismissed):

matthiasbruns (commented):
looking good - can we resolve or tackle fabian's questions?

bindings/go/oci/repository/provider/provider.go
jakobmoellerdev: you mean oci resources?

matthiasbruns (dismissed):

matthiasbruns (approved):
Image

PR #2840, opened 2026-06-12
chore(docs): Expand HTTP client configuration documentation to include retry policies and TLS verification options

### LINKED ISSUE #755: add global option to disable TLS validation
**What would you like to be added**

Add a (global) flag to disable TLS validation for OCM-CLI's commands. Inspired by `curl`, the flag might be named `--insecure`, but any name will do.

**Why is this needed**

For development purposes, there may be cases where no valid certificate is available in a testing environment (e.g. if using a self-signed certificate). Having the option to disable TLS validation will be handy in such cases.

One might also consider productive scenarios, where, through a misconfiguration, TLS validation fails, and OCM-CLI is needed to perform urgent tasks that would otherwise be blocked by TLS validation issues.

Admittedly, those are exceptional and corner-cases. However, most other tooling supports explicit disabling of TLS validation, including e.g. package-managers (apt, apk, pacman), HTTP-APIs for all programming languages, HTTP-tools, such as `curl` or `wget`, ... - even security-aware tools such as `ssh` offer disabling of checks / unsafe mode of operation.

### LINKED ISSUE #976: TLS — Custom CA Certificates
**Description**
Allow users to provide additional trusted CA certificates (PEM files) so OCM can connect to registries using internal/corporate CAs.

**Done Criteria**
- [ ] - Users can list CA cert paths under `tls.rootCAs`
- [ ] Certificates are appended to system cert pool (not replacing it)
- [ ] Clear error message if a cert file is missing or malformed

jakobmoellerdev:

Includes Retry and TLS disablement into the Website docs

fix https://github.com/open-component-model/ocm/issues/755
fix https://github.com/open-component-model/ocm-project/issues/976

##### How to test the changes

build CLI and then run through the tutorial

##### Verification


jakobmoellerdev:
@coderabbitai review

website/content/docs/how-to/configure-http-client-behaviour.md
```diff
@@ -74,28 +78,161 @@ short so a hung server is detected quickly.
...
+# Multiple bundles — point at a directory of *.pem / *.crt files.
+export SSL_CERT_DIR=/etc/ocm/ca.d
+ocm get cv registry.corp/my-org//ocm.software/demos/podinfo:6.8.0
+```
+
+Both variables are additive: certificates loaded this way are merged with the
+system trust store rather than replacing it, so public registries keep
+working in the same invocation. Set the variables only for the OCM process
+(via `env -S`, a systemd unit, or a wrapper script) when you don't want them
+to leak into other tooling.
```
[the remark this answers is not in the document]
jakobmoellerdev: addressed

website/content/docs/how-to/configure-http-client-behaviour.md
```diff
@@ -74,28 +80,198 @@ short so a hung server is detected quickly.
...
+      maxWait: 10s       # Upper bound on backoff between attempts
+```
+
+`maxRetries` counts attempts **after** the initial request, so `maxRetries: 3`
+makes up to four total attempts. Each retry waits for a duration drawn from
+`[minWait, maxWait]` with exponential growth and jitter — set both bounds to
+the same value for a fixed delay.
+
+{{< callout context="caution" >}}
+`timeout` is the **end-to-end** budget for an OCM call, including all retry
```
matthiasbruns: What is "budget for an OCM call" - is it a http call within many during the command execution? I think this is an HTTP call, right?

What I want to say is - I don't understand this :D

website/content/docs/how-to/configure-http-client-behaviour.md
```diff
@@ -74,28 +80,198 @@ short so a hung server is detected quickly.
...
+attempts and their backoff waits — it is not multiplicative with
+`maxRetries`. If `timeout` is reached mid-retry the call fails immediately
+with a context deadline error. Pick `timeout` large enough that the
+**slowest** path you expect (one full attempt + worst-case retries) fits
+inside it; raising `maxRetries` without raising `timeout` simply means later
+retries never run.
+{{< /callout >}}
+
+Disable retries entirely with `maxRetries: -1` when running against a
```
matthiasbruns: Reading this, I think we should have set this differently during the coding phase
`maxRetries=0` sounds more like no retry, not its too late - still feels off

website/content/docs/how-to/configure-http-client-behaviour.md
jakobmoellerdev: Good question — "timeout budget" is unclear. timeout maps to Go's http.Client.Timeout: the maximum time for a single HTTP request end-to-end (TCP connect + TLS handshake + sending the request + reading the full response body). So yes, it's per HTTP call, not per OCM command. I'll rephrase

website/content/docs/how-to/configure-http-client-behaviour.md
jakobmoellerdev: I think 0 means infinite retries. we could have turned it around but meh

matthiasbruns: Yeah as I said, it is how it is now :)

website/content/docs/how-to/configure-http-client-behaviour.md
matthiasbruns: Rest is great - rephrase and I will take another quick look

jakobmoellerdev: double check please

piotrjanik:
*Resolver Configuration* is located under *Reference*, should not HTTP config land in the same place?

matthiasbruns (dismissed):

website/content/docs/how-to/configure-http/_index.md
```diff
@@ -0,0 +1,148 @@
...
+
+### `x509: certificate signed by unknown authority`
+
+**Fix:** See [Configure TLS and Custom CA]({{< relref "tls.md" >}}).
+
+### `SSL_CERT_FILE` set but registry still fails verification
+
+**Fix:** See [Configure TLS and Custom CA — Troubleshooting]({{< relref "tls.md#troubleshooting" >}}).
+
+### `proxyconnect tcp: … connection refused`
```
matthiasbruns: formatting is off here - this should be a header but is not rendered like it
```suggestion
### proxyconnect tcp: … connection refused
```

website/content/docs/how-to/configure-http/tls.md
```diff
@@ -0,0 +1,180 @@
...
+ocm get cv registry.corp/my-org//ocm.software/demos/podinfo:6.8.0
+```
+
+```bash
+# Multiple bundles — point at a directory of *.pem / *.crt files.
+export SSL_CERT_DIR=/etc/ocm/ca.d
+ocm get cv registry.corp/my-org//ocm.software/demos/podinfo:6.8.0
+```
+
+{{< callout context="warning" >}}
```
matthiasbruns: Image

matthiasbruns (commented):
nits and red ci

matthiasbruns (approved):
Love the section intro page very much -
Reference is fire
the rest is good as well - from my side you can merge it like this

morri-son (approved):

PR #2867, opened 2026-06-15
feat(cli): Wire transfer config into the cli

frewilhelm:

Wires the new transfer-config (#2587) into the cli and adds respective tests.


frewilhelm:
@coderabbitai review

cli/cmd/transfer/component-version/cmd.go
```diff
@@ -109,13 +127,18 @@ transfer component-version --transfer-spec spec.yaml
 	cmd.Flags().Bool(FlagDryRun, false, "build and validate the graph but do not execute")
 	cmd.Flags().BoolP(FlagRecursive, "r", false, "recursively discover and transfer component versions")
```
Skarlso, line 131: Shouldn't we align this with the int config? Don't want to start that debate, but maybe worth doing now while we are at it.

frewilhelm: I thought about this as well but decided against it because we (afaik) do not have any other option than no-recursive or full-recursive. I believe it would make sense to adjust this, when we offer more possibilities to actually fine-tune the recursiveness.

Skarlso (dismissed):

jakobmoellerdev (commented):
can you please also adjust the examples module with the new transfer config?

frewilhelm:
> can you please also adjust the examples module with the new transfer config?

oh, good idea! Will take a look

frewilhelm:
@coderabbitai review

frewilhelm:
@coderabbitai review

jakobmoellerdev (dismissed):

frewilhelm:
@coderabbitai review

jakobmoellerdev (approved):
awesome!

cli/integration/transfer_with_transfer_config_test.go
```diff
@@ -0,0 +1,327 @@
...
+// Test_Integration_TransferWithTransferConfig_FlagOverridesFileRecursion
+// proves the override branch in buildGraphDefinitionFromArgs: a recursion
+// setting in the central OCM config can be overridden by an explicit
+// --recursive flag, and that override wins.
+//
+// The signal: a parent component references a child component, both in one
+// source CTF. The config says `recursive: 0` (no recursion) and the CLI passes
+// `--recursive`. With the override honored, the child component lands in the
+// target.
+func Test_Integration_TransferWithTransferConfig_FlagOverridesFileRecursion(t *testing.T) {
```
piotrjanik, line 183: nit: missing test for other flags

matthiasbruns (commented):
So - that I understand this correctly

you dropped the manual transfer config loading and use `LookupConfig` now, since it's already available and partially override the global config with per-transfer settings?

piotrjanik (approved):
LGTM, one nit

cli/integration/transfer_with_transfer_config_test.go
frewilhelm: The tests only test if the transfer configuration is passed through, the cli flags take precedence, and invalid values are rejected. I believe the actual behaviour change of set values from a transfer configuration are tested in `bindings/go/transfer/internal/graph_test.go`

frewilhelm:
> you dropped the manual transfer config loading and use LookupConfig now, since it's already available and partially override the global config with per-transfer settings?

@matthiasbruns I don't think I am overwriting the global config. Can you point me to the location please

PR #2870, opened 2026-06-15
chore!: Isolate unit tests from host config files

### LINKED ISSUE #802: CLI flag to ignore ocmconfig
**Description**
The goal of this issue is to introduce a CLI flag to ignore the `ocmconfig`.

In the next-gen ocm cli tests, we frequently ran into "works on my machine" issues. This was due to the cli loading the `ocmconfig` from a default location - even within the tests. 

**Done Criteria**
- [ ] Ignore Flag added to cli unit tests to prevent accidental loading of local ocm configs

jneisener:
Test introduced in #2752 fails on some dev machines because the test is affected by the `.ocmconfig` under `$HOME`.

Isolates unit test runs from all relevant config files that might get auto-discovered

##### Verification


cli/cmd/internal/test/test.go
```diff
@@ -68,6 +68,11 @@ func WithLogFormat(format string) Option {
 func OCM(tb testing.TB, opts ...Option) (*cobra.Command, error) {
 	tb.Helper()
 
+	// Isolate from host config files
+	tb.Setenv("HOME", tb.TempDir())
+	tb.Setenv("XDG_CONFIG_HOME", "")
+	tb.Setenv("OCM_CONFIG", "")
```
frewilhelm: I am a bit unsure about this.
1. what if we want to test something with those environment variables?
2. According to the docs we cannot parallelize  the tests anymore. This might be fine now, but I am unsure if we want to reconsider this, when more and more tests are added.
```
// Because Setenv affects the whole process, it cannot be used
// in parallel tests or tests with parallel ancestors.
```

Would it be feasible to add an empty config for the failing test? I believe we even have an issue for that somewhere (I will try to find it)

frewilhelm: https://github.com/open-component-model/ocm-project/issues/802

jneisener: >     1. what if we want to test something with those environment variables?

I can make make it configurable via a test option. Though, I don't really see the usecase of a test that depends on these variables honestly.

>     2. According to the docs we cannot parallelize  the tests anymore. This might be fine now, but I am unsure if we want to reconsider this, when more and more tests are added.

Yes, if we try to run tests in parallel this will cause a panic. But we apparently don't do that anywhere and if we ever do, the cause of the error should be clear.

> Would it be feasible to add an empty config for the failing test? I believe we even have an issue for that somewhere (I will try to find it)

Sure its feasible, we could even remove the test. My point is rather that all tests should be isolated from the host environment, I don't think this particular test is the issue. In principle any test could rely on the config somehow and be affected by this unexpected behavior. I'd like to isolate all tests against this case, not just the ones we think of when writing them. Does that make sense?

jneisener: Alternatively, we could replace `Getenv` with a function that we can stub during test execution. That would work for all tests, work in parallel, and still enable testing of a particular env var (by stubbing a custom function).

Like here: https://github.com/open-component-model/open-component-model/pull/2793/changes#diff-8fa454e86c80eb0f74295021b966141ddd746ed2850fff67899f72a7bb4409c4R24-R26

jneisener: Actually, this wouldn't work in parallel either as its a global variable. But the previous solution would https://github.com/open-component-model/open-component-model/pull/2793/changes/06e6b43b79c48c816cea6f35478d35b2a68876bd

frewilhelm: > In principle any test could rely on the config somehow and be affected by this unexpected behavior. I'd like to isolate all tests against this case, not just the ones we think of when writing them. Does that make sense?

absolutely. I don't know yet what the best solution would be :/

jneisener: I'm not sure what I'd like to have is completely realizable. But what do you think about https://github.com/open-component-model/open-component-model/pull/2870/commits/849e9025027e0d78fa5acfadeff66f0e44de30d7?

This way all calls through `test.OCM` get an empty environment, its parallelizable, and specific values can still be tested.
There is some redundancy to avoid cyclic import though and calling a command directly is still possible in other tests, which would then get the os environment as before

cli/cmd/configuration/ocm_config.go
```diff
@@ -22,8 +23,37 @@ const (
...
+// Environment abstracts OS functions the module relies on, dependency injected during testing.
+type Environment struct {
+	Stat        func(string) (os.FileInfo, error)
+	Getenv      func(string) string
+	UserHomeDir func() (string, error)
+	Getwd       func() (string, error)
+	Executable  func() (string, error)
+}
+
+func DefaultEnvironment() *Environment {
```
matthiasbruns: I get the pattern, but did you only introduce this for testing? Since `DefaultEnvironment` basically is the old implementation, whats the benefit here?

cli/cmd/configuration/ocm_config.go
```diff
@@ -73,8 +103,8 @@ func GetOCMConfigForCommand(cmd *cobra.Command) (*genericv1.Config, error) {
 // Returns:
 //   - *v1.Config: The parsed configuration file.
 //   - error: An error if no valid configuration file is found or if decoding fails.
-func GetOCMConfig(additional ...string) (*genericv1.Config, error) {
-	paths, err := GetOCMConfigPaths()
+func GetOCMConfig(ctx context.Context, additional ...string) (*genericv1.Config, error) {
```
matthiasbruns: Since this is a breaking change, we need to rename the or title to 

`chore!: Isolate unit tests from host config files`

cli/cmd/configuration/ocm_config.go
```diff
@@ -22,8 +23,37 @@ const (
 	OCMConfigCommandArgument = "config"
 )
 
-// statFunc is the file stat function used for config discovery. Overridden in tests.
-var statFunc = os.Stat
```
matthiasbruns, line 26: we could have used this pattern to override more things, right?

jneisener: Yes, but then it doesn't work with `t.Parallel` which was one of the initial criticisms, see https://github.com/open-component-model/open-component-model/pull/2870#discussion_r3414648832

cli/cmd/configuration/ocm_config.go
jneisener: Yes, its only for testing. By default the environment behavior is the same as before, but it can be exchanged for a different implementation for testing

cli/cmd/configuration/ocm_config_test.go
```diff
@@ -5,69 +5,13 @@ import (
...
 
-	"github.com/spf13/cobra"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	genericv1 "ocm.software/open-component-model/bindings/go/configuration/generic/v1/spec"
 	"ocm.software/open-component-model/bindings/go/runtime"
 )
 
-func TestGetOCMConfigForCommand(t *testing.T) {
```
jneisener, line 16: Moved out because using `test.OCM` here creates a cyclic dependency

cli/cmd/setup/setup.go
```diff
@@ -146,6 +146,21 @@ func CredentialGraph(cmd *cobra.Command) error {
 	return nil
 }
 
+func Syscalls(cmd *cobra.Command) {
```
matthiasbruns, line 149: the name is kinda weird- why not OsEnvironment or the old one?

jneisener: So it was `Environment` before, but `Getenv` on `Environment` doesn't really make sense. So currently I think `Syscalls` best captures what is actually in there. 
But open to better suggestions

matthiasbruns: RuntimeContext maybe? No idea - I wouldnt block this PR because of the name, the impl looks fine

matthiasbruns (approved):

jakobmoellerdev (approved):

PR #2871, opened 2026-06-15
feat(repository): add OwnershipAwareRepository interface

fabianburth:
## Summary
- Adds `OwnershipAwareRepository` as an optional capability interface in `bindings/go/repository`.
- `AddOwnership` attaches ownership info (component name + version) to a resource so tooling can trace a resource back to its owning component version (ADR 0016).
- Foundation for stacked PRs that wire OCM repositories (OCI first) to record ownership and the constructor to drive it via `ownershipPolicy`.

## Test plan
- [x] `go vet ./bindings/go/repository/...`
- [x] `go test ./bindings/go/repository/...`

PR #2872, opened 2026-06-15
feat(oci)!: support ownership referrers and CTF referrer index

### LINKED ISSUE #1025: Transfer Ownership Referrers

fabianburth:
## Summary
- Adds `AddOwnership` on the OCI `Repository` and `ResourceRepository` to attach an ownership referrer linking a resource back to its owning component version (ADR 0016). Dispatches on access type:
  - **by-value (`LocalBlob`)**: subject is the just-pushed manifest in the component's own store, resolved by digest.
  - **by-reference (`OCIImage`)**: subject is the referenced image in its hosting registry; only the referrer manifest (and the shared empty config/layer blob) is pushed, so OCI-level signatures over the image stay valid.
- The referrer is content-addressed and deterministic — re-running converges on the same digest. No-op when the subject is not an OCI manifest (raw blobs get no referrer).
- CTF gains an index-aware referrer lookup aligned with the oras implementation, plus tag-schema fallback for registries without the referrers API.
- Layout copy migrates to `ExtendedCopyGraph` so referrers travel with their subjects.

## Notes
- Constructor wiring (driving `AddOwnership` via `ownershipPolicy`) is a separate, stacked PR.

## Test plan
- [x] `go build ./bindings/go/oci/...`
- [x] `go test ./bindings/go/oci/...`


bindings/go/oci/ctf/referrers.go
```diff
@@ -0,0 +1,371 @@
...
+	"oras.land/oras-go/v2/content"
+	"oras.land/oras-go/v2/errdef"
+	"oras.land/oras-go/v2/registry"
+
+	v1 "ocm.software/open-component-model/bindings/go/ctf/index/v1"
+	ociblob "ocm.software/open-component-model/bindings/go/oci/blob"
+	"ocm.software/open-component-model/bindings/go/oci/internal/introspection"
+)
+
+// A CTF archive is structurally a pre-v1.1 OCI registry: tags and digests in
```
jakobmoellerdev: this docmment should probably be somewhere in a doc.go because currently its lost

bindings/go/oci/ctf/referrers.go
```diff
@@ -0,0 +1,371 @@
...
+// store's write lock, which fully serializes the read-modify-write of the
+// referrers index. Across multiple [Store] instances for the same path, or
+// across processes, the CTF (like the rest of artifact-index.json handling)
+// is single-writer; see also the store cache in the oci repository provider.
+
+// maxManifestBytes caps how much manifest or referrers-index content is read
+// into memory, both when buffering a pushed manifest to inspect its subject
+// and when loading a referrers index. It matches oras-go's
+// defaultMaxMetadataBytes.
+const maxManifestBytes = 4 * 1024 * 1024 // 4 MiB
```
jakobmoellerdev: can you please link the referred size?

bindings/go/oci/ctf/referrers.go
```diff
@@ -0,0 +1,371 @@
...
+		// the referrer is already indexed and the stored index is clean;
+		// skip the write entirely, making referrer re-pushes idempotent.
+		return nil
+	}
+
+	newIndexDesc, newIndexJSON, err := generateIndex(updated)
+	if err != nil {
+		return err
+	}
+	if err := s.archive.SaveBlob(ctx, ociblob.NewDescriptorBlob(io.NopCloser(bytes.NewReader(newIndexJSON)), newIndexDesc)); err != nil {
```
jakobmoellerdev, line 162: dont you need to drop / delete the old index?

bindings/go/oci/ctf/referrers.go
```diff
@@ -0,0 +1,371 @@
...
+	if err := s.applyTag(ctx, idx, newIndexDesc, referrersTag); err != nil {
+		return fmt.Errorf("unable to retag referrers index for referrers tag %q: %w", referrersTag, err)
+	}
+	return nil
+}
+
+// artifactManifest is the subset of the deprecated OCI artifact manifest
+// (introspection.MediaTypeArtifactManifest) needed for referrer indexing.
+// oras-go/v2 defines the full type in internal/spec (not importable).
+type artifactManifest struct {
```
jakobmoellerdev: why do we need this? I thought we can use a modern manifest?

bindings/go/oci/ctf/referrers.go
```diff
@@ -0,0 +1,371 @@
...
+// was changed.
+//
+// Inspired by oras remote.applyReferrerChanges.
+func addReferrer(referrers []ociImageSpecV1.Descriptor, referrer ociImageSpecV1.Descriptor) ([]ociImageSpecV1.Descriptor, bool) {
+	type key struct {
+		mediaType string
+		digest    digest.Digest
+		size      int64
+	}
+	keyOf := func(d ociImageSpecV1.Descriptor) key {
```
jakobmoellerdev: why is this not just a normal function and type?

bindings/go/oci/ctf/referrers.go
```diff
@@ -0,0 +1,371 @@
...
+	}
+	indexDesc := content.NewDescriptorFromBytes(index.MediaType, indexJSON)
+	return indexDesc, indexJSON, nil
+}
+
+// filterReferrers filters a slice of referrers by artifactType in place.
+// The returned slice contains matching referrers.
+//
+// Copied from oras remote.filterReferrers.
+func filterReferrers(refs []ociImageSpecV1.Descriptor, artifactType string) []ociImageSpecV1.Descriptor {
```
jakobmoellerdev, line 309: looks weird, can you use slices funcs here?

bindings/go/oci/tar/blob_io.go
```diff
@@ -128,133 +124,81 @@ func CopyOCILayoutWithIndex(ctx context.Context, dst content.Storage, src blob.R
...
+	// Subject descriptor is not mutated.
+	// Swap the Subject descriptor with the mutated one.
+	extendedCopyOpts := opts.ExtendedCopyGraphOptions
+	extendedCopyOpts.FindSuccessors = func(ctx context.Context, fetcher content.Fetcher, desc ociImageSpecV1.Descriptor) ([]ociImageSpecV1.Descriptor, error) {
+		var (
+			successors []ociImageSpecV1.Descriptor
+			err        error
+		)
+		if opts.FindSuccessors != nil {
+			successors, err = opts.FindSuccessors(ctx, fetcher, desc)
```
jakobmoellerdev, line 131: to double check if reassignment

bindings/go/oci/repository.go
```diff
@@ -719,6 +714,85 @@ func (repo *Repository) uploadOCIImage(ctx context.Context, newAccess runtime.Ty
...
+
+	store, subject, err := repo.resolveOwnershipSubject(ctx, component, version, resource)
+	if err != nil {
+		return err
+	}
+	return repo.buildAndPushOwnershipReferrer(ctx, store, subject, resource, component, version)
+}
+
+func (repo *Repository) resolveOwnershipSubject(ctx context.Context, component, version string, resource *descriptor.Resource) (spec.Store, ociImageSpecV1.Descriptor, error) {
+	if v2.IsLocalBlob(resource.Access) {
```
jakobmoellerdev: should be a normal scheme.New with type switch instead of if statement

bindings/go/oci/ctf/referrers.go
```diff
@@ -0,0 +1,371 @@
...
+// on manifest push.
+// References:
+//   - https://github.com/opencontainers/distribution-spec/blob/v1.1.1/spec.md#pushing-manifests-with-subject
+//
+// CTF does not implement content.Deleter. Therefore, we hard code the semantics
+// of SkipReferrerGC. Since we cannot delete, we cannot end up with empty
+// referrers lists anyway.
+//
+// The caller MUST hold a write lock and MUST persist the index after the call.
+func (s *repository) updateReferrersIndex(ctx context.Context, idx v1.Index, subject, referrer ociImageSpecV1.Descriptor) error {
```
fabianburth, line 140: this probably creates a number of orphaned manifests now (to be confirmed)

bindings/go/oci/tar/blob_io.go
```diff
@@ -128,133 +124,81 @@ func CopyOCILayoutWithIndex(ctx context.Context, dst content.Storage, src blob.R
...
-	// instead we can use whatever top level descriptor (manifest or index) is located as singleton in the layout index.
+// pickTopLevelDescriptor selects the single top-level manifest from the
+// layout's index.json. With one manifest in the index it returns that
+// manifest; with many it returns the one tagged via
+// `org.opencontainers.image.ref.name`. Returns an error if neither rule
+// uniquely identifies a top-level descriptor.
+//
+// We need this specifically for docker (one manifest), and oras / ocm
+// packaging compat (many manifests, exactly one ref.name).
+func pickTopLevelDescriptor(ociStore *CloseableReadOnlyStore) (ociImageSpecV1.Descriptor, error) {
```
fabianburth, line 158: check whether this still works reliably now that we also have tagged index manifests for the referrers

bindings/go/oci/repository/resource/resource_repository.go
```diff
@@ -146,6 +149,34 @@ func (p *ResourceRepository) DownloadResource(ctx context.Context, resource *des
 	return b, nil
 }
 
+// AddOwnership attaches ownership information (i.e. the
+// component name and version) to a resource. The ownership is attached as a
+// referrer manifest pointing at the resource.
+func (p *ResourceRepository) AddOwnership(ctx context.Context, component, version string, resource *descriptor.Resource, credentials runtime.Typed) error {
```
jakobmoellerdev, line 156: this needs more testing on adding multiple owners etc

bindings/go/oci/repository/resource/resource_repository.go
```diff
@@ -49,6 +49,9 @@ type ResourceRepository struct {
 // make sure that ResourceRepository implements the oci ResourceRepository interface
 var _ repository.ResourceRepository = (*ResourceRepository)(nil)
 
+// TODO: re-enable once repository.OwnershipAwareRepository lands in main.
```
jakobmoellerdev: todo is lost, needs an issue and user as well as be put into a method. In general please dont open docs that are dangling outside of either method, function or package scope

bindings/go/oci/integration/asset_to_owner_integration_test.go
jakobmoellerdev, line 1: why was the asset to owner integration test removed? Im missing integration tests now

bindings/go/oci/internal/introspection/manifest.go
```diff
@@ -12,6 +12,11 @@ const (
 	MediaTypeDockerManifestList = "application/vnd.docker.distribution.manifest.list.v2+json"
 )
 
+// MediaTypeArtifactManifest is the deprecated OCI artifact manifest media type
+// (image-spec v1.1.0-rc series). oras-go/v2 defines it in internal/spec
+// (not importable).
+const MediaTypeArtifactManifest = "application/vnd.oci.artifact.manifest.v1+json"
```
jakobmoellerdev: we agreed to drop this AFAIK

bindings/go/oci/tar/top_level_artifacts.go
```diff
@@ -9,52 +9,47 @@ import (
...
+	excluded := make(map[digest.Digest]struct{}, len(candidates))
 
 	var wg sync.WaitGroup
 	wg.Add(len(candidates))
-	// For each artifact in the index, find all the artifacts it references
 	for i := range candidates {
 		go func() {
 			defer wg.Done()
-			resolveReferences(candidates[i])
+			subject, successors, err := classify(ctx, fetcher, candidates[i])
```
jakobmoellerdev: what does classify mean? this looks very strange as a logic

bindings/go/oci/tar/top_level_artifacts.go
```diff
@@ -9,52 +9,47 @@ import (
...
+// TopLevelArtifacts returns the main top-level artifacts from a list of
+// candidates. A candidate is excluded when it is a referrer (declares a
+// subject) or when another candidate contains it as a successor. The remaining
+// candidates are returned in input order.
+//
+// Each candidate is fetched and decoded once, in parallel, to determine both
+// its subject and its containment successors. A fetch or decode error for a
+// candidate is treated as "not a referrer, contributes no edges" so a transient
+// failure cannot silently drop a real top-level artifact.
 func TopLevelArtifacts(ctx context.Context, fetcher content.Fetcher, candidates []ociImageSpecV1.Descriptor) []ociImageSpecV1.Descriptor {
```
jakobmoellerdev, line 16: the second paragraph is ai slop imho

bindings/go/oci/repository.go
```diff
@@ -47,8 +49,10 @@ import (
 )
 
 var (
-	_            ComponentVersionRepository = (*Repository)(nil)
-	versionRegex                            = regexp.MustCompile(compref.VersionRegex)
+	_ ComponentVersionRepository = (*Repository)(nil)
+	// TODO: re-enable once repository.OwnershipAwareRepository lands in main.
```
jakobmoellerdev: same as with the other todo

bindings/go/oci/integration/asset_to_owner_integration_test.go
piotrjanik: It could have been a rebase issue. In this PR https://github.com/open-component-model/open-component-model/pull/2676/changes I renamed `bindings/go/oci/integration/asset_to_owner_integration_test.go` to `cli/integration/ownership_integration_test.go`.

bindings/go/oci/ctf/referrers.go
fabianburth: I removed it. I'd assume that the content in a ctf is typically rather trusted, so I think keeping this simple and lenient is fine.

bindings/go/oci/tar/blob_io.go
```diff
@@ -128,205 +104,131 @@ func CopyOCILayoutWithIndex(ctx context.Context, dst content.Storage, src blob.R
...
+	mediaTypeDockerManifest     = "application/vnd.docker.distribution.manifest.v2+json"
+	mediaTypeDockerManifestList = "application/vnd.docker.distribution.manifest.list.v2+json"
+)
+
+// classify decodes desc once and returns its subject (nil if desc is not a
+// referrer) and its containment successors (config+layers, child manifests, or
+// blobs depending on media type). Docker manifest types have no subject and
+// are forwarded to [content.Successors]. Any other media type returns
+// (nil, nil, nil) — it is not fetched and contributes no edges.
+func classify(ctx context.Context, fetcher content.Fetcher, desc ociImageSpecV1.Descriptor) (*ociImageSpecV1.Descriptor, []ociImageSpecV1.Descriptor, error) {
```
piotrjanik: Do we still need that as we switch to `ExtendedCopyGraph`? The `successorsWithoutSubject` was used to ignore manifests with subjects when they were added as successors of a root descriptor. 

This logic is now removed:
```
rootChildren := append(append([]ociImageSpecV1.Descriptor{}, index.Manifests...), referrerDescriptors...)` )
```

`classify` looks almost identical to `successorsWithoutSubject` - is it needed for another reason?

bindings/go/oci/tar/oci_layout_tar_reader_test.go
```diff
@@ -174,3 +178,110 @@ func TestReadOCILayout_Close(t *testing.T) {
 	require.NoError(t, err)
 	assert.NoError(t, store.Close())
 }
+
+// ownershipArtifactType is the ADR-0016 artifact type carried by an ownership
+// referrer. Referrer detection keys off the subject, not the artifact type, so
+// the test pairs it with a referrer carrying a different type.
+const ownershipArtifactType = "application/vnd.ocm.software.ownership.v1+json"
```
piotrjanik: nit: It's already defined in https://github.com/fabianburth/open-component-model/blob/2e55fb243d4ecd06b0984e4497c62ffebbd6708e/bindings/go/oci/spec/annotations/ownership.go#L27

bindings/go/oci/ctf/referrers.go
fabianburth: I'm now dropping the old index. I didn't want to extend the ctf index interface again, so I'm using the `RemoveTag` BEFORE I create the actual new index.

bindings/go/oci/ctf/referrers.go
fabianburth: was related to the rfc media type, as we dropped support for it, we were also able to remove this struct

bindings/go/oci/ctf/referrers.go
fabianburth: probably, but figured I just leave it like that as it is currently a verbatim copy of the oras implementation

bindings/go/oci/tar/blob_io.go
fabianburth: Function is only used in OCI Layout. OCI Layout does not use referrer tag schema. So, this should still be fine!

bindings/go/oci/tar/top_level_artifacts.go
fabianburth: We want to retrieve the top level artifact(s) that are not referrers. This essentially just fetches each descriptor and extracts its subject and successors. Renamed it to `extractSubjectAndSuccessors`, is this sufficient?

bindings/go/oci/integration/asset_to_owner_integration_test.go
fabianburth: Yeah, I think the issue is that that's not really just a rename right? 
This means the ownership scenario can no longer be tested standalone for the module.

piotrjanik: I moved it there because I wanted to test scenarios where `OwnershipAwareRepository` isn’t implemented, just to see how things would work in that case.

bindings/go/oci/ctf/doc.go
```diff
@@ -0,0 +1,31 @@
+// Package ctf implements a structurally pre-v1.1 OCI registry: tags and digests
```
jakobmoellerdev: that doesnt make sense. CTF doesnt implement an OCI registry

bindings/go/oci/ctf/doc.go
```diff
@@ -0,0 +1,31 @@
...
+// Reference: https://github.com/opencontainers/distribution-spec/blob/v1.1.1/spec.md#unavailable-referrers-api
+//
+// Concurrency: the spec notes that maintaining the referrers tag is the
+// client's responsibility and that concurrent updates can lose data. Within
+// a process, every repository derived from the same [Store] shares the
+// store's write lock, which fully serializes the read-modify-write of the
+// referrers index. Across multiple [Store] instances for the same path, or
+// across processes, the CTF (like the rest of artifact-index.json handling)
+// is single-writer; see also the store cache in the oci repository provider.
+package ctf
```
jakobmoellerdev, line 7: there is way too much focus on referrers here because the package does way more than this.

jakobmoellerdev (dismissed):

Skarlso:
>The repository.OwnershipAwareRepository type assertions are commented out for now — they depend on https://github.com/open-component-model/open-component-model/pull/2871. Will be re-enabled in a follow-up once that lands.

This is no longer true, right? Or am I missing something. You have type assertion on the interface. 🤔

bindings/go/oci/repository.go
```diff
@@ -598,19 +588,19 @@ func (repo *Repository) getLocalBlobFromIndexOrManifest(
...
-		return tar.CopyToOCILayoutInMemory(ctx, store, artifact, tar.CopyToOCILayoutOptions{
-			CopyGraphOptions: repo.resourceCopyOptions.CopyGraphOptions,
-			Tags:             []string{version},
-			TempDir:          repo.tempDir,
+		graph, ok := store.(content.ReadOnlyGraphStorage)
+		if !ok {
+			return nil, fmt.Errorf("store %T does not support predecessor walks", store)
+		}
+		return tar.CopyToOCILayoutInMemory(ctx, graph, artifact, tar.CopyToOCILayoutOptions{
+			ExtendedCopyGraphOptions: oras.ExtendedCopyGraphOptions{
```
Skarlso, line 596: `ExtendedCopyGraphOptions` is now everywhere regardless if referrels is enabled. This adds an API call to the root node I think once if I get this right. But there is no opt-out. If that's okay, then it's okay. Just wanted to flag this.

bindings/go/oci/spec/store.go
```diff
@@ -6,7 +6,7 @@ import (
 
 // Store defines the interface for interacting with an OCI store.
 type Store interface {
-	content.ReadOnlyStorage
+	content.ReadOnlyGraphStorage
```
Skarlso, line 9: This is a breaking change, right? Should we mark it as `!`?

bindings/go/oci/repository.go
```diff
@@ -598,19 +588,19 @@ func (repo *Repository) getLocalBlobFromIndexOrManifest(
...
-		// into an in-memory OCI layout. This is used when the descriptor refers
-		// to another OCI-compliant manifest instead of a single layer.
-		return tar.CopyToOCILayoutInMemory(ctx, store, artifact, tar.CopyToOCILayoutOptions{
-			CopyGraphOptions: repo.resourceCopyOptions.CopyGraphOptions,
-			Tags:             []string{version},
-			TempDir:          repo.tempDir,
+		graph, ok := store.(content.ReadOnlyGraphStorage)
+		if !ok {
+			return nil, fmt.Errorf("store %T does not support predecessor walks", store)
+		}
```
Skarlso, line 594: I believe since you changed the store.Spec, the code will just not compile if you try to use something that isn't a `content.ReadOnlyGraphStorage`, right? So you can pass in `store` as is.

Skarlso (commented):
Just a couple of small remarks. :) 🙇

bindings/go/oci/spec/store.go
fabianburth: Although, I think actually all our usages actually are `ReadOnlyGraphStorage`.

jakobmoellerdev (approved):

PR #2877, opened 2026-06-15
chore(release): remove path filter for direct pushes to main to keep the floating tag updated

Skarlso:

The path filter on main pushes prevented our floating `main` tag from being properly updated on main updates.

This will increase the runtime of ci builds on main ( because the controller is also building ) but that's the price of the combined monorepo build.

##### Verification


matthiasbruns (commented):
That makes sense, but wont we stack up the build-queue if cancel-in-progress is false? Or will gh be smart enough to skip all builds except the latest, if we merge too fast and the main build queue will fill up?

Skarlso:
@matthiasbruns cancel in progress is only there for pull requests. So all main builds should go through.

matthiasbruns:
> @matthiasbruns cancel in progress is only there for pull requests. So all main builds should go through.

https://docs.github.com/en/enterprise-cloud@latest/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency
> To also cancel any currently running job or workflow in the same concurrency group, specify cancel-in-progress: true. To conditionally cancel currently running jobs or workflows in the same concurrency group, you can specify cancel-in-progress as an expression with any of the allowed expression contexts.
> 
> To allow more than one pending job or workflow run to wait in the same concurrency group, use the optional queue property. The queue property accepts the following values:
> 
> single (default): At most one job or workflow run can be pending in the concurrency group. When a new job or workflow run is queued, any existing pending job or workflow run in the same group is canceled and replaced.
> max: Up to 100 jobs or workflow runs can be pending in the concurrency group. When the queue is full, any additional jobs or workflow runs are canceled.
> The combination of queue: max and cancel-in-progress: true is not allowed and will result in a workflow validation error.

I interpret this as:
- not only for PRs
- will only have the latest in queue

matthiasbruns (approved):

Skarlso:
```yaml
concurrency:
  group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
  cancel-in-progress: ${{ github.event_name == 'pull_request' }}

```

We set it to true only on even name pull_request. :)

frewilhelm (approved):

matthiasbruns:
> ```yaml
> concurrency:
>   group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
>   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
> ```
> 
> We set it to true only on even name pull_request. :)

We might have meant different things :D I meant:

in general they also apply for non-pr workflows, but yeah in this case it does not matter

PR #2881, opened 2026-06-16
feat: Make resolver cache TTL configurable

### LINKED ISSUE #868: expose TTL in controller args for cache
**Description**
Currently, the ttl of cached components is hardcoded.

`kubernetes/controller/cmd/main.go`
```
	const unlimited = 0
	ttl := time.Minute * 30
	resolverCache := expirable.NewLRU[string, *workerpool.Result](unlimited, nil, ttl)
```

To mitigate potential cache invalidation issues, the ttl should be configurable.

This configuration should also surface in our helm chart values.yaml file.

**Done Criteria**

jneisener:
Implements https://github.com/open-component-model/ocm-project/issues/868

Makes resolver cache ttl configurable in controller 

Fixes https://github.com/open-component-model/ocm-project/issues/868

##### How to test the changes
Indirectly tested via e2e tests, similar to the already existing flags

##### Verification


fabianburth (approved):

fabianburth (changes requested):
Actually, I just thought about our recent performance issues and discussions due to users setting the interval on the controllers to very low values.

Now that it is configurable, we should probably add a lower bound and at least throw a warning or something if you are below that lower bound. WDYT?

jneisener:
> Actually, I just thought about our recent performance issues and discussions due to users setting the interval on the controllers to very low values.
> 
> Now that it is configurable, we should probably add a lower bound and at least throw a warning or something if you are below that lower bound. WDYT?

Probably makes sense. Current lower bound is 1 minute, what lower bound did you have in mind?

Also, should it be a warning at runtime or should the docstring advice caution (I think the docstring is more likely to actually be read)

morri-son:
do we even want to set it lower than one minute? It may make sense to even set it higher and then introduce the warning?

fabianburth:
I agree, the docstring is probably more useful. We might just want to say that setting it lower than the 30 minutes is discouraged for production and may cause performance problems (or something like that).

fabianburth (approved):

matthiasbruns (approved):

PR #2886, opened 2026-06-16
chore: Make `sed` invocation BSD/GNU agnostic

jneisener:

> ## How GNU sed Handles In-Place Editing
> 
> GNU sed accepts -i without an argument to edit files in place without creating a backup. GNU sed also accepts -i.bak to create a backup with the .bak extension. The -i flag and the backup extension are a single argument.
> ## How BSD sed Handles In-Place Editing
>
>BSD sed requires a separate argument after -i— even if the argument is an empty string. sed -i '' 's/old/new/' file edits in place without a backup. sed -i .bak 's/old/new/' file creates a backup. Omitting the argument after -i causes BSD sed to interpret the next argument as the backup extension.

https://console9.com/wiki/articles/sed/explanations/gnu-vs-bsd/

Which is why running `task helm/validate` I get an error due to an unexpected new file `README.md-e`.

Using an explicit temp file and avoiding the `-i` parameter should make it work on both.

Incompatibility of `task helm/validate` with MacOS `sed`

#### Testing
-

##### Verification

### DIFF
--- a/kubernetes/controller/Taskfile.yml
+++ b/kubernetes/controller/Taskfile.yml
@@ -256,7 +256,7 @@ tasks:
       - '"{{.HELM_DOCS}}"'
       # Strip SHA digests from README.md for cleaner documentation
       # The digest is kept in values.yaml for security, but removed from docs for readability
-      - sed 's/@sha256:[a-f0-9]\{64\}//g' README.md > README.md.tmp && mv README.md.tmp README.md
+      - tmp=$(mktemp) && sed 's/@sha256:[a-f0-9]\{64\}//g' README.md > "$tmp" && mv "$tmp" README.md
     sources:
       - chart/values.yaml
       - chart/README.md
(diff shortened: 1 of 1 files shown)


kubernetes/controller/Taskfile.yml
```diff
@@ -256,7 +256,7 @@ tasks:
       - '"{{.HELM_DOCS}}"'
       # Strip SHA digests from README.md for cleaner documentation
       # The digest is kept in values.yaml for security, but removed from docs for readability
-      - sed -i'' -e 's/@sha256:[a-f0-9]\{64\}//g' README.md
+      - sed 's/@sha256:[a-f0-9]\{64\}//g' README.md > README.md.tmp && mv README.md.tmp README.md
```
jakobmoellerdev: instead of readme.md.tmp you could use mktemp no?

morri-son (approved):

jakobmoellerdev (approved):

PR #2894, opened 2026-06-18
fix(release): fix website update script

frewilhelm:
This PR tackles two bugs

- Parts of the module part were written in the `version` field that should only hold [version queries](https://gohugo.io/configuration/module/#version)
- The script failed when modules defined in [`CLI_DERIVED_MODULES`](https://github.com/open-component-model/open-component-model/blob/d2b3aa3a403141f9f1db977d0ee0ed8d60f0046b/website/scripts/register-docs-version.js#L196) were not part of the `cli-go.mod` used to get the modules from th current release.

I tested this locally with `0.9.0` and also with a mock `0.10.0` version which worked fine. Additionally, I checked out the website locally showing the different how-tos/concepts/tutorials between `0.9.0` and `main`.


frewilhelm:
@coderabbitai review

matthiasbruns (dismissed):
We really need to refactor this- but it looks good to me right now

morri-son (dismissed):

Skarlso:
I'm looking at this too. :)

.github/workflows/website-verify-scripts.yml
```diff
@@ -16,9 +17,9 @@ permissions:
   contents: read
 
 jobs:
-  lint:
+  verify:
```
Skarlso, line 20: I think this breaks the branch protection setup. 🤔 We can't just rename jobs as far as I know.

website/scripts/register-docs-version.js
```diff
@@ -186,7 +198,11 @@ function resolveGoModVersions(goModPath, modulePaths) {
 
     const missing = modulePaths.filter(p => !result[p]);
     if (missing.length) {
-        fail(`Could not resolve versions for: ${missing.join(', ')} in ${goModPath}`);
```
Skarlso, line 189: This was intentional. We really want a failure here otherwise we might skip problems that are real. I get that it's annoying, but we should prevent accidental misspelling of stuff. This would now only warn instead of fail loudly. 🤔

.github/workflows/website-verify-scripts.yml
```diff
@@ -8,6 +8,7 @@ on:
     paths:
       - '.github/workflows/website-verify-scripts.yml'
       - 'website/assets/js/**'
+      - 'website/scripts/**'
```
Skarlso, line 11: Its' nice that you added this, but you also need to update the lint/script JS target in website package js, because that only checks assets/js/**.

```json
"lint:scripts": "eslint --cache assets/js",
```

If you didn't want that, that's fine then. :)

Skarlso (commented):
Just some questions and concerns about the job rename.

website/scripts/register-docs-version.js
frewilhelm: This is actually the reason why its currently [failing](https://github.com/open-component-model/open-component-model/actions/runs/26824468590/job/81703261602). `CLI_DERIVED_MODULES` includes `ocm.software/open-component-model/bindings/go/http` but `ocm.software/open-component-model/bindings/go/http` is not part of the `cli-go.mod` of release `0.9.0`:

```
hugo.yaml: defaultContentVersion changed from 'main' to '0.9'.
hugo.yaml: added version 0.9 (weights reassigned).
Error:  Could not resolve versions for: ocm.software/open-component-model/bindings/go/http in /home/runner/work/_temp/cli-go.mod
Error:  Could not resolve versions for: ocm.software/open-component-model/bindings/go/http in /home/runner/work/_temp/cli-go.mod
Error: Process completed with exit code 1.
```

Skarlso: Can we put it in there...?

frewilhelm: What do you mean? Put the `http` module into the cli `go.mod` of the release `0.9.0`?

.github/workflows/website-verify-scripts.yml
frewilhelm: uh, nice catch

website/scripts/register-docs-version.js
Skarlso: Ah wait, my memory was that we list these things somewhere, I might be wrong.

Skarlso: Well then that's it. :D

.github/workflows/website-verify-scripts.yml
frewilhelm: > We can't just rename jobs as far as I know.

Are you sure? this would be new for me

website/scripts/register-docs-version.js
frewilhelm: But this would mean we need to backport modules to older releases. Should we then create another release too?

IMO the warning should be fine. Additionally, only the PR is created - nothing is merged automatically. Hence, the release-coordinator has to verify the doc-bump too. If something looks suspicious, then he/she would take a look at the logs and would see the warning

Skarlso: Yeah, it's probably okay. Don't worry about it.

.github/workflows/website-verify-scripts.yml
Skarlso: Yep. It doesn't update it automatically. It needs to be done manually.

frewilhelm: You mean "Require status checks to pass" right?

frewilhelm: I think that workflow is not required (yet?)
image

website/scripts/register-docs-version.test.js
```diff
@@ -88,20 +103,25 @@ test('hasAllImportsForVersion: returns false for wrong version', () => {
...
     const truncated = imports.map(i =>
         i.path.endsWith('/bindings/go/sigstore')
             ? { ...i, mounts: [i.mounts[0]] }
             : i
     );
-    assert.equal(hasAllImportsForVersion({ imports: truncated }, '0.3'), false);
+    assert.equal(hasAllImportsForVersion({ imports: truncated }, '0.3', ALL_DEPS), false);
+});
+
+test('hasAllImportsForVersion: returns true when deps-threaded full import set is present', () => {
```
jneisener: What does `deps-threaded` mean?

website/scripts/register-docs-version.js
```diff
@@ -378,30 +406,28 @@ function updateImportTags(parsed, version, fullVersion, deps) {
...
+        if (imp.path.endsWith('/website') ||
+            imp.path.endsWith('/cli') ||
+            imp.path.endsWith('/kubernetes/controller')) {
+            newTag = `v${fullVersion}`;
         } else if (deps && imp.path.endsWith('/bindings/go/constructor')) {
-            newTag = `bindings/go/constructor/${deps['ocm.software/open-component-model/bindings/go/constructor']}`;
+            newTag = deps['ocm.software/open-component-model/bindings/go/constructor'];
         } else if (deps && imp.path.endsWith('/bindings/go/descriptor/v2')) {
-            newTag = `bindings/go/descriptor/v2/${deps['ocm.software/open-component-model/bindings/go/descriptor/v2']}`;
+            newTag = deps['ocm.software/open-component-model/bindings/go/descriptor/v2'];
```
jneisener: I see its the pre-existing pattern, but is there any reason why in the detection `endsWith` is used instead of the fully qualified name?

Also, I think replacing the magic strings with constants would improve the readability. These constants could then be used in both the condition and the assignment.

.github/workflows/website-verify-scripts.yml
Skarlso: ah then it's perfect 👍

Skarlso (dismissed):

jneisener (approved):

matthiasbruns (approved):

Skarlso (approved):

PR #2904, opened 2026-06-19
feat(controller): add ArgoCD as alternative GitOps deployment option in e2e examples

sk31337:

The existing examples only showed FluxCD as the GitOps delivery mechanism. This PR adds ArgoCD as a parallel path across all six helm-based kro examples and proves it works end-to-end in CI. 

Relates ocm-project#969
Relates ocm-project#1074
##### How to test the changes

```bash
task test/e2e/setup/teardown && task test/e2e/setup/local && task test/e2e
```
##### Verification

- [x] I have tested the changes locally by running `task test/e2e/teardown && task test/e2e/setup/local && task test/e2e`


.gitignore
frewilhelm, line 1: I believe this file does not need a change

kubernetes/controller/Taskfile.yml
```diff
@@ -83,12 +83,52 @@ tasks:
...
+          # it through verbatim (e.g. from cmd/shard CI output). Otherwise
+          # anchor it so "helm-nested" matches exactly that scenario
+          # and not "helm-nested-signed".
+          if [[ "$cli" == ^* ]]; then
+            focus_args=(-ginkgo.focus="$cli")
+          else
+            focus_args=(-ginkgo.focus="^.*${cli}$")
+          fi
+        fi
+        PROJECT_DIR='{{.TASKFILE_DIR}}' go test '{{.TASKFILE_DIR}}/test/e2e/' -v -timeout 30m -ginkgo.v "${focus_args[@]}"
```
frewilhelm: can you please take another look at this. I don't know why this should be needed

kubernetes/controller/Taskfile.yml
```diff
@@ -83,12 +83,52 @@ tasks:
       - 'KIND_NODE_IMAGE_VERSION={{.KIND_NODE_IMAGE_VERSION}} bash test/e2e/hacks/setup.sh'
       - 'kind load docker-image {{.CONTROLLER_IMG}}'
 
+  test/e2e/teardown:
+    desc: "Delete the local Kind cluster and registry used by E2E tests"
+    cmds:
+      - 'kind delete cluster 2>/dev/null || true'
+      - 'docker rm -f image-registry 2>/dev/null || true'
```
frewilhelm: This is a dangerous task. Users might not expect that their cluster gets deleted when executing it. I would recommend to drop this task

kubernetes/controller/test/e2e/e2e_examples_test.go
```diff
@@ -116,15 +138,22 @@ var _ = Describe("controller", func() {
...
+					)).To(Succeed())
 					By("validating the localization")
 					Expect(utils.CompareResourceField(ctx,
 						"pod -l app.kubernetes.io/name="+example.Name()+"-podinfo",
 						"'{.items[0].spec.containers[0].image}'",
 						strings.TrimLeft(imageRegistry, "http://")+"/stefanprodan/podinfo:6.9.1",
 					)).To(Succeed())
-					By("validating the configuration")
+
+					By("validating the ArgoCD configuration (ui.message)")
```
frewilhelm, line 164: Do we need to check for the localisation of argo cd too?

kubernetes/controller/test/e2e/e2e_examples_test.go
```diff
@@ -102,6 +103,27 @@ var _ = Describe("controller", func() {
...
+
+					By("validating the ArgoCD-managed deployment in default-argocd")
+
+					Expect(utils.WaitForResource(ctx, "create", timeout, name, "-n", "default-argocd")).To(Succeed())
+					Expect(utils.WaitForResource(ctx, "condition=Available", timeout, name, "-n", "default-argocd")).To(Succeed())
+					Expect(utils.WaitForResource(
+						ctx, "condition=Ready=true",
+						timeout,
+						"pod", "-l", "app.kubernetes.io/name="+example.Name()+"-argocd-podinfo", "-n", "default-argocd",
+					)).To(Succeed())
```
frewilhelm, line 127: Why is this required in this `if` clause. Can't we validate it after `By("validating the example")` like we do it for the flux deployments?

kubernetes/controller/test/e2e/e2e_examples_test.go
frewilhelm, line 132: We might add a suffix to the fluxcd release name too to clearly distinguish between both deployments

kubernetes/controller/test/e2e/e2e_examples_test.go
sk31337: Yes, if fluxed localisation is checked, so must argocd as well.

kubernetes/controller/Taskfile.yml
sk31337: I needed this task a lot of time. Executing a command without reading the description is not recommended at all. Therefore, it should stay.

kubernetes/controller/Taskfile.yml
sk31337: Without this functionality, I would have wasted soo much more time re-running single failed tests by having to run all tests again. Therefore this is a much welcome functionality for everyone putting their time into OCM k8s controller tests.

kubernetes/controller/test/e2e/e2e_examples_test.go
sk31337: namespace makes the separation

kubernetes/controller/test/e2e/e2e_examples_test.go
sk31337: Sry, i don't see any `if` clause from line 107 to 125.

morri-son: it's in line 99 :)

kubernetes/controller/Taskfile.yml
morri-son: I second the statement about descriptions, but what about calling the task "teardown-cluster"?

sk31337: @frewilhelm okay for you as well?

kubernetes/controller/test/e2e/e2e_examples_test.go
sk31337: @frewilhelm there are two examples without `kro.run` instance and `flux`:
  - applyset-pruning/
  - k8s-manifest-simple/

therefore ArgoCD needs to stay within the `if` statement (to my humble understanding) :)

matthiasbruns: maybe we should invert this and skip the example instead of wrapping it - my bad

sk31337: Nobodies bad! 

I didn't want to update existing working implementation to keep review-ability as good as possible.

kubernetes/controller/Taskfile.yml
frewilhelm: > Executing a command without reading the description is not recommended at all.

The user who has read all the documentation for ocm v1 commands and subcommands before running it shall raise their hands :D I understand what you are saying. In my experience, the reality is very different. Users try things out and users derive the intent from the naming.

The task `test/e2e/teardown` sounds like the complement to `test/e2e`. Hence, I would expect it to uninstall the ocm controller from the cluster. Not delete kinds default cluster.
The task `test/e2e/setup/local` is setting up the whole environment - the cluster, the image registry, ...

I think I could be fine with a task like `test/e2e/setup/teardown` or the like. BUT if we do this, we absolutely need to change the defaults. We should use a named cluster for the e2e test instead of the default one and probably adjust the name of the image registry too. We do not want to delete someones local setup that is used for other things. That said, I don't know if this should be in scope of this PR.

BTW the need to delete the whole cluster and start it again to run the e2e tests sounds a bit like an overkill. I believe there is an easier way depending on the use-case.

kubernetes/controller/test/e2e/e2e_examples_test.go
frewilhelm: and where are we doing this? :D

kubernetes/controller/test/e2e/e2e_examples_test.go
frewilhelm: nice idea! But then we do not need the suffix for the argocd release, right?

kubernetes/controller/test/e2e/e2e_examples_test.go
frewilhelm: I just had a look at the code and realized that a "generic" validation is currently not possible. With the current setup, we need an extra validation for argoCD because its in another namespace that cannot be derived from the example name or the like.

IMHO - lets live with until we overhaul the e2e test suite.

kubernetes/controller/Taskfile.yml
frewilhelm: Okay, I see and I think thats a valid point. However, why do we need to do all the args-parsing? It makes the code harder to read.

I would propose to only append `{{.CLI_ARGS}}` to the original command.

```
PROJECT_DIR='{{.TASKFILE_DIR}}' go test '{{.TASKFILE_DIR}}/test/e2e/' -v -timeout 30m -ginkgo.v {{.CLI_ARGS}}
```

Then, you can just pass any ginkgo flag you like, e.g. `task test/e2e -- -ginkgo.focus=helm-simple` and we let ginkgo do the parsing for us.

wdyt?

kubernetes/controller/Taskfile.yml
morri-son: `test/e2e/setup/teardown` sounds reasonable and better as it is the counterpart to `test/e2e/setup/local`. Using a named cluster in any case makes sense. Since kind is based on Docker, killing and respawning a cluster is much faster than deleting resources for a proper cleanup (for sure depending on how many objects you need to re-deploy), so it can be a valid way and not an overkill 😊

kubernetes/controller/test/e2e/e2e_examples_test.go
sk31337: We do not need, but ArgoCD tests told me, they want to feel special.... I try to change it without causing a (personal) melt down ;)

kubernetes/controller/Taskfile.yml
sk31337: I try to change it without causing a (personal) melt down ;)

kubernetes/controller/Taskfile.yml
sk31337: I try to change it without causing a (personal) melt down ;)

kubernetes/controller/test/e2e/e2e_examples_test.go
sk31337: Line 155-159 was transformed from FluxCD to ArgoCD and fluxed newly created at line 141-146.

kubernetes/controller/test/e2e/e2e_examples_test.go
sk31337: Changed with commit 044ae436

kubernetes/controller/Taskfile.yml
sk31337: Solved in commit 3afcee82.

kubernetes/controller/Taskfile.yml
sk31337: Solve with commit f64bf9bb10515366a86483bf40e2544d13bcbca5

kubernetes/controller/test/e2e/hacks/setup.sh
```diff
@@ -83,27 +83,37 @@ add_hosts_toml() {
 EOF
 }
 
-for node in $(kind get nodes); do
+for node in $(kind get nodes --name ocm-e2e); do
   add_hosts_toml "${node}" "${CONTAINERD_CONFIG_PATH}/${reg_name}:${reg_port}" "http://${reg_name}:${reg_port}"
+  # Also register the "image-registry" DNS alias so containerd skips TLS for that hostname.
+  # bootstrap.yaml and OCM resource status both surface the registry as image-registry:5000;
+  # without this entry containerd falls back to HTTPS and fails the image pull.
+  add_hosts_toml "${node}" "${CONTAINERD_CONFIG_PATH}/image-registry:${reg_port}" "http://image-registry:${reg_port}"
```
frewilhelm: I believe you should adjust the bootstrap.yaml then instead of having two registry names

kubernetes/controller/test/e2e/e2e_examples_test.go
frewilhelm: Still it just tests the configuration but not the localization. Take a look at the validation of the FluxCD example on L147-L152 (in your PR)

```
					By("validating the localization")
					Expect(utils.CompareResourceField(ctx,
						"pod -l app.kubernetes.io/name="+example.Name()+"-podinfo",
						"'{.items[0].spec.containers[0].image}'",
						strings.TrimLeft(imageRegistry, "http://")+"/stefanprodan/podinfo:6.9.1",
					)).To(Succeed())
```

sk31337: Added at commit [a3073dc](https://github.com/open-component-model/open-component-model/pull/2904/commits/a3073dc5697a7dd2b2c67ec371fe3f7111eadcc9)

kubernetes/controller/test/e2e/hacks/setup.sh
sk31337: Solved with commit [c5b7142](https://github.com/open-component-model/open-component-model/pull/2904/commits/c5b71426a7e8c9c2cc83506a0a3ec0d9cd40c6c4)

kubernetes/controller/test/e2e/hacks/setup.sh
```diff
@@ -83,29 +83,42 @@ add_hosts_toml() {
 EOF
 }
 
-for node in $(kind get nodes); do
+for node in $(kind get nodes --name ocm-e2e); do
   add_hosts_toml "${node}" "${CONTAINERD_CONFIG_PATH}/${reg_name}:${reg_port}" "http://${reg_name}:${reg_port}"
+  # Also register the "image-registry" DNS alias so containerd skips TLS for that hostname.
+  # bootstrap.yaml and OCM resource status both surface the registry as ocm-e2e-image-registry:5000;
+  # without this entry containerd falls back to HTTPS and fails the image pull.
+  #add_hosts_toml "${node}" "${CONTAINERD_CONFIG_PATH}/image-registry:${reg_port}" "http://image-registry:${reg_port}"
```
frewilhelm: This comment is not needed anymore

kubernetes/controller/test/e2e/hacks/setup.sh
```diff
@@ -83,29 +83,42 @@ add_hosts_toml() {
...
   add_hosts_toml "${node}" "${CONTAINERD_CONFIG_PATH}/localhost:31002" "registry-internal.default.svc.cluster.local:5002"
   add_hosts_toml "${node}" "${CONTAINERD_CONFIG_PATH}/localhost:31003" "registry-internal.default.svc.cluster.local:5003"
 done
 
 # Connect the registry to the cluster network if not already connected.
 ## This allows kind to bootstrap the network but ensures they're on the same network.
+## The --alias keeps "image-registry" as the in-cluster DNS name so bootstrap.yaml
+## and other manifests that reference ocm-e2e-image-registry:5000 continue to work without changes.
```
frewilhelm: This is also not needed

kubernetes/controller/test/e2e/hacks/setup.sh
sk31337: Removed with commit [8149bb6](https://github.com/open-component-model/open-component-model/pull/2904/commits/8149bb6bdcced27d3a6e9641a46c51eb45021e19)

kubernetes/controller/test/e2e/hacks/setup.sh
sk31337: Removed with commit [8149bb6](https://github.com/open-component-model/open-component-model/pull/2904/commits/8149bb6bdcced27d3a6e9641a46c51eb45021e19)

morri-son (approved):

Skarlso:
task: Task "test/e2e/teardown" does not exist for a start... :D 

I'm running this locally. :)

Skarlso:
Ran 13 of 13 Specs in 441.032 seconds
SUCCESS! -- 13 Passed | 0 Failed | 0 Pending | 0 Skipped
--- PASS: TestE2E (441.03s)

🎉 🎉 🎉

Skarlso (approved):

PR #2917, opened 2026-06-23
chore(GHA): Move label config to omit checkout action

frewilhelm:

Move the label config to its default location, so we do not need the checkout action that needs adjusted permissions due to the `pull_request_target` trigger (that is needed for the label action because of GitHubs `GITHUB_TOKEN` permission settings, see [doc](https://github.com/actions/labeler#recommended-permissions)).

Additionally, this PR removes a lot of the label config that is meant for ocm v1, e.g. `flake` ... (I believe this file was just copy-pasted without being adjusted)


morri-son (dismissed):

frewilhelm:
@coderabbitai review

matthiasbruns (dismissed):

frewilhelm:
Actually, we do not need to move it. We can just point to it and the action will get the config from the GitHub API. I will adjust this

EDIT:

> If the file doesn't exist at the specified path on the runner, action will read from the source repository via the Github API.

(see [inputs](https://github.com/actions/labeler#inputs))

Hence, we can safely remove the checkout action.

Skarlso (approved):

matthiasbruns (approved):

PR #2919, opened 2026-06-23
chore(deps): override undici transitive package version

frewilhelm:

override undici transitive package version because of a reported security vulnerability.


morri-son (dismissed):

.github/scripts/package.json
```diff
@@ -7,5 +7,13 @@
   "devDependencies": {
     "@actions/github-script": "github:actions/github-script",
     "@octokit/graphql": "9.0.3"
+  },
+  "overrides": {
+    "@actions/github": {
+      "undici": "^6.27.0"
+    },
```
jneisener: This override should be unnecessary
```
├─┬ @actions/github-script@9.0.0 (git+ssh://git@github.com/actions/github-script.git#3a2844b7e9c422d3c10d287c895573f7108da1b3)
│ ├─┬ @actions/core@1.11.1
│ │ └─┬ @actions/http-client@2.2.3
│ │   └── undici@5.29.0
│ └─┬ @actions/github@9.1.1
│   ├─┬ @actions/http-client@3.0.2
│   │ └── undici@6.25.0 deduped
│   └── undici@6.25.0
```

i.e. the `@actions/github` dependency path already pulls `6.25.0` and the dependency range (`"undici": "^6.23.0"`) includes `6.27.0`. So only `@actions/http-client@2.2.3` needs to have `undici` overwritten

frewilhelm: you are right! Thank you

jneisener (approved):

matthiasbruns (approved):

PR #2924, opened 2026-06-24
fix(deps): migrate js-yaml usage to v5 API

jneisener:
## Summary

- Replaces `import yaml from 'js-yaml'` (default import) with named imports (`load`, `dump`, `loadAll`) — js-yaml v5 no longer provides a default ESM export
- Removes the `@types/js-yaml` dev dependency since v5 ships its own TypeScript declarations

Fixes the Netlify deploy failure on #2923.

## Test plan

- [x] `.github/scripts` tests pass (5/5)
- [x] `website/scripts/register-docs-version.test.js` passes (53/53)
- [x] ESLint passes on changed files
- [x] Hugo's ESBuild bundling will resolve the named import correctly

### DIFF
--- a/.github/scripts/prepare-registry-constructor.js
+++ b/.github/scripts/prepare-registry-constructor.js
@@ -1,6 +1,6 @@
 // @ts-check
 import fs from 'fs';
-import yaml from 'js-yaml';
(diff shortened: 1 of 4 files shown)


jneisener:
#2925

PR #2926, opened 2026-06-24
chore(website): fix location of crds

frewilhelm:
@jneisener reported that the CRD schema are not displayed on the website for `0.9`. Instead an error is shown:

```
Failed to load schema: HTTP 404
```

The reason for this is, that the plain CRD schemas are not stored in the repository. There are only the helm-template CRDs in `kubernetes/controller/chart/templates/crd` that are not useable for the yaml-loader (as they contain helm-templates which is no valid yaml).

For the `main` version the plain CRD schemas are created on the fly in [`website-publish-site.yaml` ](https://github.com/open-component-model/open-component-model/blob/2a4f044f984af5a507b7dc3243996e2eb2e4aab5/.github/workflows/website-publish-site.yaml#L78). However, generating the CRDs for each (future) version might be possible but would add a lot of complexity.

This is why we decided to (re-) add the plain CRDs to the repository in `kubernetes/controller/config/crd/bases`. The location was chosen because its the default for kubebuilder projects. Although we do not plan to add further "configurations" to that directory, it fits the context.

Changes:
- Adjust the `Taskfile` to generate plain CRDs into `kubernetes/controller/config/crd/bases` as well as follow-up changes for functions and docs
- While testing I noticed that the `release.yml` workflow will fail on a second run for `0.9.1` which is why I added some more debugging possibilities
  - Do not fail when detecting an incomplete block (see `website/scripts/register-docs-version.js`). This might happen for releases that do not include all modules defined in `CLI_DERIVED_MODULES`.
  - Put the output of the script in the PR description, so the reviewer can make sure everything is fine
- Remove the plain CRD creation "on-the-fly" in `.github/workflows/website-publish-site.yaml`

To make the schema-display work on the website for release `0.9` we will need to backport this PR. This will also include the replication controller CRD. However, this will only be a file in a directory and is not used anywhere, so this should be fine.


frewilhelm:
@coderabbitai review

.github/workflows/release.yml
```diff
@@ -683,10 +683,25 @@ jobs:
...
+            --cli-gomod "$CLI_GOMOD" \
+            2>&1 | tee "$LOG_FILE"
+          # Write the logs to the output so they can be included in the PR body. There might be warnings about missing
+          # modules (diff between CLI_DERIVED_MODULES and CLI_GOMOD) that are not fatal but need to be validated
+          # manually.
+          {
+            echo 'log<<__REGISTER_EOF__'
+            cat "$LOG_FILE"
+            echo '__REGISTER_EOF__'
+          } >> "$GITHUB_OUTPUT"
```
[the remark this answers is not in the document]
frewilhelm: I understand the approach but actually, we do not want to fail here. The following step is creating a pull request with the logs in the body, so the reviewer can check for any warnings and failures and validate them

kubernetes/controller/config/crd/bases/delivery.ocm.software_repositories.yaml
```diff
@@ -0,0 +1,245 @@
...
+  - additionalPrinterColumns:
+    - description: Indicates if the Resource is Ready
+      jsonPath: .status.conditions[?(@.type=="Ready")].message
+      name: Ready
+      priority: 1
+      type: string
+    - description: Displays the Age of the Resource
+      jsonPath: .metadata.creationTimestamp
+      name: Age
+      type: date
```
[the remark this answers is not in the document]
frewilhelm: This change is not valid for this PR

frewilhelm:
@coderabbitai review

frewilhelm:
@coderabbitai review

jneisener (dismissed):

morri-son (dismissed):

morri-son:
@frewilhelm conflicts...

jneisener (dismissed):

website/netlify.toml
```diff
@@ -11,11 +11,11 @@
   ignore = "git diff --quiet origin/main HEAD -- . ../cli/docs/reference ../bindings/go/constructor/spec/v1/resources ../bindings/go/descriptor/v2/resources ../kubernetes/controller/api/v1alpha1"
 
 [context.deploy-preview]
-  command = "(cd ../kubernetes/controller && mkdir -p bin/gen/crd && go run sigs.k8s.io/controller-tools/cmd/controller-gen@$CONTROLLER_TOOLS_VERSION crd paths=./... output:crd:artifacts:config=bin/gen/crd) && npm ci && npm run build -- --baseURL=$DEPLOY_PRIME_URL --buildDrafts --buildFuture"
```
Skarlso, line 14: `CONTROLLER_TOOLS_VERSION` is no longer used.

website/netlify.toml
Skarlso: This is now no longer relevant, right?

website/netlify.toml
frewilhelm: ah i missed that. Thanks!

website/netlify.toml
frewilhelm: So, the comment is outdated but the reasoning is still valid I believe. Even if the CRDs are now in another directory, the base for the CRD is `api/v1alpha1` so this should stay. WDYT?

frewilhelm: I adjusted the comment :)

Skarlso (approved):

jneisener (approved):

PR #2930, opened 2026-06-24
chore(docs): go work and binding release adr

### LINKED ISSUE #1087: Create ADR to decide what direction we want to go with our module bindings
**Description**

Our binding release and "independence" setup starts to heavily slow us down and increased the pr-review overhead by a lot. We have discussed in a warroom, if and how we can improve the developer process for the better, or if no better alternative can be found, how to improve our team processes.

Create an ADR based on the options and questions below and weigh, which direction we should pursue.

- Enable go.work by default - also in the pull-requests. This would allow us to push changes into the code-base with way less PRs and give the reviewers a more concise block of changes. Currently, PRs are lacking context of the bigger picture of such change.
- Rethink the release process. How to we want to release bindings? When do we want to pin versions? How should the controller & cli release consume the binding changes? A possible inspiration could be the kubernetes release process, where on release, each binding gets pinned to a version and the dependencies are being replaces inflight during the release process
- Ditch the go modules completely and use the bindings as a part of the code. So instead of having n bindings, we would have a shared core/lib code, that can be consumed by cli and controller. We would lose the binding boundaries completely though.

### Timebox: 5 day(s) + 5 additional days (because we are convinced the time-invest is worth it)

**Done Criteria**
- [ ] PoC your approach

matthiasbruns:
This ADR compares a few approaches on how we can move away from the one-binding-pr restriction we currently have.
They all focus on the use of go work within our workflows.
The ADR also covers how we can release & tag bindings in our release flow.

Check https://github.com/matthiasbruns/open-component-model/pull/115

Fixes: https://github.com/open-component-model/ocm-project/issues/1087

##### How to test the changes

Verified and tested in the linked PoC branch: Check https://github.com/matthiasbruns/open-component-model/pull/115
Workflows for the binding-release: https://github.com/matthiasbruns/open-component-model/actions/workflows/release-bindings.yaml

- release workflow without the needs for tags: https://github.com/matthiasbruns/open-component-model/actions/runs/28328113275/job/83921826060
- release workflow with the need of a binding release: https://github.com/matthiasbruns/open-component-model/actions/runs/28328574202/job/83924638329


matthiasbruns:
@coderabbitai review and also check the linked issue as well as the poc branch

docs/adr/0025_bindings_ci_and_release_strategy.md
```diff
@@ -0,0 +1,369 @@
...
+
+* **Manual per-module release** via `release-go-submodule.yaml`.
+* **Automated phased bulk release** via `release-bindings.yaml` (plan, test, gate, release).
+
+---
+
+## Decision Outcome
+
+Keep Go modules with `go.work`, always test all bindings in CI, and use the automated phased bulk release as the
+canonical path. The manual per-module release is retained for isolated hotfixes. The sparse-checkout strategy is an
```
Skarlso: How will hot fixes influence the overall version that is produced if ONE of the modules starts to tag-drift? 🤔

docs/adr/0025_bindings_ci_and_release_strategy.md
```diff
@@ -0,0 +1,369 @@
...
+open decision (see *Pros and Cons* below).
+
+**Why keep Go modules over a shared library:** Ditching Go modules would eliminate independent versioning, making it
+impossible for external consumers to take only the bindings they need at a specific version. The binding boundary model
+has value; the friction was in the tooling around it, not in the model itself. Adding `go.work` (gitignored, generated
+in CI from the checked-out tree) recovers the developer-experience benefit without sacrificing the boundary model.
```
Skarlso: We will have to generate go.work locally as well, right? In case there is a new module or whatever, I will need to change the go.work file to include that.

docs/adr/0025_bindings_ci_and_release_strategy.md
```diff
@@ -0,0 +1,369 @@
...
+
+**Why always test all bindings over change-based filtering:** Bindings are interdependent via `go.work`. Testing only
+changed modules misses the case where an API change in module A silently breaks module B (not touched in the PR). The
+filtering logic to handle this correctly is complex and has multiple reachable gaps; every correctness gap discovered
+requires another exception rule. All binding source is already checked out in full (`bindings/`) during the unit test
+job, so testing all bindings costs only runner time, not checkout overhead.
+
+**Why the phased bulk release over manual per-module releases:** A phased bulk release computes next tags in dependency
+order, runs tests, requires human review of the plan before any tags are pushed, and pins consumer `go.mod` files
+(`cli`, `kubernetes/controller`) atomically in the same release commit. Manual per-module releases cannot guarantee
```
Skarlso: Same questions. This means that an individual module update will have to update this as well, right?

docs/adr/0025_bindings_ci_and_release_strategy.md
```diff
@@ -0,0 +1,369 @@
...
+  signals, and special-cased expansion rules. Complexity grows with every new correctness gap discovered.
+
+### Sparse-checkout strategy
+
+**Keep sparse-checkout + gitignore `go.work` (selected)**
+
+* **Pros:** Each job downloads only what it needs. Clearly worthwhile for `golangci_lint` (parallel jobs, each scoped to
+  one module + `bindings/`) and for controller/e2e/conformance jobs (avoid pulling all binding source for builds that
+  don't need it). `go.work` is not committed — it is generated in CI via `task init/go.work` after checkout, producing a
+  workspace scoped to the checked-out tree automatically.
```
Skarlso: Hm, this means that we never really test if go.mod file is correct since everything is overshadowed by the go.work file.

Skarlso: Probably would have to add an extra job with `GOWORK=off` just to verify if all the pins are correct.

Skarlso (commented):
Couple of questions first. :) 🙇 Well done, otherwise. :)

docs/adr/0025_bindings_ci_and_release_strategy.md
```diff
@@ -0,0 +1,369 @@
...
+changed modules misses the case where an API change in module A silently breaks module B (not touched in the PR). The
+filtering logic to handle this correctly is complex and has multiple reachable gaps; every correctness gap discovered
+requires another exception rule. All binding source is already checked out in full (`bindings/`) during the unit test
+job, so testing all bindings costs only runner time, not checkout overhead.
+
+**Why the phased bulk release over manual per-module releases:** A phased bulk release computes next tags in dependency
+order, runs tests, requires human review of the plan before any tags are pushed, and pins consumer `go.mod` files
+(`cli`, `kubernetes/controller`) atomically in the same release commit. Manual per-module releases cannot guarantee
+dependency ordering and create a consistency window (see *Release Strategy* below). The manual workflow is kept as an
+escape hatch for genuine single-module hotfixes with no consumers to update.
```
jneisener: While not necessary, it should also be possible to implement change-based filtering. Since we need to construct the dependency graph already (for the release), we might as well use it to decide what needs to be build / what change can impact which modules. 
Any gaps in this logic could also affect the releases - since those *need* to be correct, I don't see clear reasoning against change-based filtering. Depending on build-time it might be a feasible optimization in the future

docs/adr/0025_bindings_ci_and_release_strategy.md
```diff
@@ -0,0 +1,369 @@
...
+  workspace scoped to the checked-out tree automatically.
+* **Cons:** Every CI job must run `task init/go.work` before Go tooling; requires `arduino/setup-task` (or equivalent)
+  before `setup-go`. The authoritative Go version is stored in `.go-version` rather than `go.work`.
+
+**Full checkout everywhere (not selected)**
+
+* **Pros:** Simpler; fewer moving parts; easier to reason about.
+* **Cons:** Every job downloads the full repository. For jobs that only need `bindings/`, this pulls in `cli/`,
+  `kubernetes/controller/`, `website/`, `conformance/` source unnecessarily. For the `golangci_lint` matrix (35 parallel
+  jobs), this overhead multiplies.
```
jneisener: I don't follow the logic, seems like the tail wagging the dog to me.

The 35 parallel jobs for linting are not a given, rather I suspect that they are a result of the absence of a `go.work` file. Its not just the full repository checkout that happens 35 times, `go` and `golangci-lint` setup happens each time as well - in addition to a sparse checkout.

Why not do **one** full checkout and then lint everything in one pass?

`go work edit -json | jq -r '.Use[].DiskPath'  | xargs -I{} golangci-lint run {}/...` takes about a minute on my machine, so I'm not sure what exactly is gained by running 35 jobs in parallel. 

Since this appears to be the main - perhaps only - argument in favor of the sparse checkout it brings the conclusion into question for me

matthiasbruns: Yeah all of this is assuming we keep sparse-checkout. The more I think about it, the less it makes sense to keep it. If we just checkout out everything and run the tests, lint etc on HEAD, we should be just fine and would simplify the workflows by a lot.

docs/adr/0025_bindings_ci_and_release_strategy.md
matthiasbruns: not 100% sure. In my mind there is one case that might not be detected, but maybe I am wrong here.
Imagine we change something in one binding that breaks another. Dev is lazy and does not test the thing locally, pushed the changes. Now the change-based-filtering would only detect the one binding change, but might not detect the breakage that will happen in the other binding... maybe.

docs/adr/0025_bindings_ci_and_release_strategy.md
matthiasbruns: thats one of the open questions in the ADR.
we will have to follow the same logic as in our release. Fix in main, cherry pick into the release branch. But I don't know about the versioning yet.

If we only release bindings in the release flow, they are tied to a cli/controller release, what we might not want. But then, if a binding is broken, how good can the released cli/controller be? Maybe we do not need the manual workflow at all - only to trigger a 0.0.1 release of a non-released binding.. hmm

docs/adr/0025_bindings_ci_and_release_strategy.md
matthiasbruns: yes, this is what we would have to do locally. In the CI this should happen automatically based on the checked out bindings and modules we have. When we run task init/go.work, is should auto-detect all go.mod files and build the go.work accordingly

docs/adr/0025_bindings_ci_and_release_strategy.md
```diff
@@ -0,0 +1,369 @@
...
+resolves bindings from the local tree, not from the module proxy. The `go.mod` pins written by `pinDeps` are never
+exercised during the release build itself.
+
+The pins serve a different audience: **external consumers** who `go get` `cli` or `controller` outside the monorepo.
+They have no `go.work`, so Go falls through to `go.mod`, which must reference the correct binding versions for the
+module proxy to assemble the right dependency graph.
+
+The build and the pins are consistent because they both point to the same commit: `release-bindings.yaml` pushes the
+pin commit, `tag_rc` creates the RC tag at that commit, and the release build checks out at that tag. `go.work` and
+`go.mod` resolve to identical code; they just serve different consumers.
```
fabianburth: > `go.work` and `go.mod` resolve to identical code; they just serve different consumers

I believe that is not entirely correct.
If we use `go.work` (or equivalent replace statements), the build of the individual module (with go.mod) might be different from the build in our CI (with go.work). 

This comes down to go's [MVS](https://go.dev/ref/mod#minimal-version-selection).
If my understanding is correct, if we have `module A` requiring `module C v0.2.0` and `module B` requiring `module C v0.1.0`, if we build them standalone, `module B` will build with `module C v0.1.0`, if you build them with go.work, `module B` will build with `module C v0.2.0`.

Besides, I believe it changes the graph pruning behaviour. But I'm not too familiar with the details here. I think it comes down to a rule like - every tests dependency that is not part of the `main` module or a module required by the main module is pruned. But `go.work` promotes the other modules to main modules, so you the set of test dependencies considered change and might again influence the MVS.

fabianburth: Please take this w/ a grain of salt and verify. I haven't played around with this in quite some time!

matthiasbruns: I will verify and check everything that is written here - the more input the better :)

docs/adr/0025_bindings_ci_and_release_strategy.md
matthiasbruns: I removed the sparse-checkout in the lint case and will investigate further, if the other cases still make sense

docs/adr/0025_bindings_ci_and_release_strategy.md
matthiasbruns: and we only release bindings that actually have a diff to the last tag - maybe I need to make this clearer in the ADR in the release-flow section

docs/adr/0025_bindings_ci_and_release_strategy.md
matthiasbruns: the cli & controller release do not use go-work. So we would have that, but the question is, if this is too late

docs/adr/0025_bindings_ci_and_release_strategy.md
```diff
@@ -0,0 +1,369 @@
...
+## Release Strategy
+
+### Consistency window in manual per-module releases
+
+If `bindings/go/helm` depends on `bindings/go/credentials` and a developer manually releases
+`bindings/go/credentials@v1.2.0`, the
+`go.mod` of `bindings/go/helm` still pins the old version until a follow-up PR updates it. During that window,
+`bindings/go/helm` is internally consistent via `go.work` in local development, but its published `go.mod` references
+the old API. Any consumer resolving via `go get` (not `go.work`) gets a mixed build. The
+`concurrency: group: binding-release` guard prevents concurrent releases from racing but does not close this window.
```
jneisener: How is that different from the current state? 

As I understand it the exact same problem can happen right now - no `go.work` required. Currently its not 'internally consistent' but rather internally pinned. But a customer updating via `go get` could still get an mixed build just like sketched here. Or am I misunderstanding something?

matthiasbruns: This is actually what the manual binding release workflow reflects. 
1. I don't know if we really need the workflow at all anymore once we migrated
2. I don't know how to circumvent this problem

docs/adr/0025_bindings_ci_and_release_strategy.md
Skarlso: Good question, hmm hmm.

docs/adr/0025_bindings_ci_and_release_strategy.md
Skarlso: 👍

docs/adr/0025_bindings_ci_and_release_strategy.md
Skarlso: This is in partial why I'm saying that we need tests that verify pins and versions in the go mod file. Because go.work will somewhat mess with those.

matthiasbruns: It's in as well - the cli & controller will run without go work enabled.

docs/adr/0025_bindings_ci_and_release_strategy.md
jakobmoellerdev: I dont like assuming things without trying. I think you guys should try running the linter across a go work enabled repo. The reason there are multiple different runs right now is that there were in the past some linters on go.mod files that could only run per-module and broke with workspaces.

jakobmoellerdev: Another thing, just because a 14 core Dev Mac can run a golangci lint run of all modules in a minute (with your local dev cache), does not mean this can be translated to CI. The goal of the CI thus far has always been extremely fast response time. If we can keep job times low while merging them, lets do that, but we should definetly see how fast a lint run really is when merging it. It might be that we need some caching here to help us.

docs/adr/0025_bindings_ci_and_release_strategy.md
```diff
@@ -0,0 +1,401 @@
...
+
+* **PR overhead**: a single logical change (e.g., adding a new credential type that spans multiple bindings) required
+  a strict sequential chain: merge a PR for the base binding, release it, then open the next PR for the dependent
+  binding combining the `go.mod` pin update with the functional changes, merge it, release it, and repeat for every
+  binding in the dependency chain. Reviewers saw each PR in isolation without the context of the broader change, making
+  review harder and slower.
+* **CI complexity**: the initial CI design used change-based filtering to avoid running all binding tests on every PR.
+  This required `dorny/paths-filter`, separate signals for `.env` and CI workflow changes, and special-cased expansion
+  rules to catch cross-module regressions. Each correctness gap discovered required another exception. Adding a new
+  binding also required a CI config change.
```
jakobmoellerdev: `Adding a new binding also required a CI config change.`

this is not correct. the only special case that was introduced was for non binding modules such as the kubernetes controller as they were migrated from outside the monorepo.

docs/adr/0025_bindings_ci_and_release_strategy.md
```diff
@@ -0,0 +1,401 @@
...
+* **Release friction**: releasing a set of related bindings required manually triggering one workflow per binding in the
```
jakobmoellerdev: This is the same point as the first one (PR overhead)

docs/adr/0025_bindings_ci_and_release_strategy.md
```diff
@@ -0,0 +1,401 @@
...
+  releases or against a manual release conflicting with an ongoing bulk release.
+
+The friction was in the tooling around the module boundary model, not in the model itself. This ADR documents the three
+candidate directions evaluated and explains how the chosen approach resolves each pain point.
+
+### Context and Problem Statement
+
+Four concrete problems had to be solved:
+
+1. **Scalable CI enrollment**: the binding set grows over time; CI must cover all bindings without requiring a config
```
jakobmoellerdev: which config change are you referring to here?

docs/adr/0025_bindings_ci_and_release_strategy.md
```diff
@@ -0,0 +1,401 @@
...
+   change each time one is added.
+2. **Cross-module regressions**: a change to `bindings/go/credentials` can silently break `bindings/go/helm` or `cli` if
+   only the changed module is tested.
+3. **Workspace management**: the repo-wide `go.work` at the root is always present; it must reflect the checked-out
```
jakobmoellerdev: this is not the case, its not always present and there are tasks to create it on demand.

docs/adr/0025_bindings_ci_and_release_strategy.md
```diff
@@ -0,0 +1,401 @@
...
+   tree so that workspace-aware commands resolve all dependencies locally during development and CI.
+4. **Coordinated releases**: bindings must be released in dependency order; a manual per-module release can leave
```
jakobmoellerdev: this is again the same point as above, feels repeated

docs/adr/0025_bindings_ci_and_release_strategy.md
```diff
@@ -0,0 +1,401 @@
...
+
+* **Change-based filtering**: discover all modules dynamically; on PR, test only the modules that changed.
+* **Always test all bindings**: discover all modules dynamically; test every binding on every PR.
+
+### Sparse-checkout strategy
+
+* **Full checkout for lint/unit, sparse for integration**: lint and unit tests check out the full repository so
+  `go.work` covers all modules. Integration tests use sparse-checkout per module since they are I/O-bound and
+  parallelism helps.
+* **Per-module matrix with sparse-checkout everywhere**: each CI job checks out only its module.
```
jakobmoellerdev: on sparse checkout: the reason I added sparse checkout is to prevent inexperienced engineers in the codebase to reference files from other modules. for example, when writing the helm bindings, I dont want anyone to reference test files from the runtime. Why? because that would break modularization and starts to introduce coupling over time which is undesired.

Thus, sparse checkout is not a technical requirement but only for formalization of the boundary.

docs/adr/0025_bindings_ci_and_release_strategy.md
```diff
@@ -0,0 +1,401 @@
...
+### Release strategy
+
+* **Manual per-module release** via `release-go-submodule.yaml`.
+* **Automated phased bulk release** via `release-bindings.yaml` (plan, test, gate, release).
+
+---
+
+## Decision Outcome
+
+Keep Go modules with a committed `go.work`, always test all bindings in CI, and use the automated phased bulk release
```
jakobmoellerdev: How does a module behave that is referenced in go.work but then fetched externally with `go get ocm.software/...@main`. Will this cause different dependencies? If yes, IMHO we have a problem here and checking in go.work is a problem for us.

docs/adr/0025_bindings_ci_and_release_strategy.md
```diff
@@ -0,0 +1,401 @@
...
+as the canonical path. The manual per-module release is retained for isolated fixes and for bootstrapping new
+bindings. Lint and unit tests use a full checkout; integration tests use sparse-checkout in a parallel matrix.
+
+**Why keep Go modules over a shared library:** Ditching Go modules would eliminate independent versioning, making it
+impossible for external consumers to take only the bindings they need at a specific version. The binding boundary model
+has value; the friction was in the tooling around it, not in the model itself. Adding `go.work` recovers the
+developer-experience benefit without sacrificing the boundary model.
```
jakobmoellerdev: I disagree. If we use go.work all engineers must be aware of what version a model is in without developing on it. This will move the problem from "what do i need to release first" to a "when do I need to release in order to not break anyone that does not rely on go work"

docs/adr/0025_bindings_ci_and_release_strategy.md
```diff
@@ -0,0 +1,401 @@
...
+
+**Why always test all bindings over change-based filtering:** Unit tests for all bindings run as a single
```
jakobmoellerdev: This does not work with or without go work:

```
go test ./bindings/go/...
# ./bindings/go/...
pattern ./bindings/go/...: directory prefix bindings/go does not contain main module or its selected dependencies
FAIL    ./bindings/go/... [setup failed]
```

Im not sure how we came to this without testing. Do you mind sharing?

docs/adr/0025_bindings_ci_and_release_strategy.md
```diff
@@ -0,0 +1,401 @@
...
+    participant Build as cli.yml / controller.yml
+    participant Proxy as module proxy
+    participant Ext as external consumer
+
+    RB->>RB: pinDeps — go mod edit in bindings (topo order) + cli + controller
+    RB->>RB: git commit -s (pin commit) + push binding semver tags
+    RC->>RC: checkout branch HEAD (= pin commit)
+    RC->>RC: create RC tag at pin commit
+    Build->>Build: checkout cli/ + bindings/ at RC tag
+    Build->>Build: compile (GOWORK=off → resolves go.mod pins directly)
```
jakobmoellerdev: The one issue I have with this is that the default in the monorepo is to not use the modules as they are in the go mod. Also I dont see how we solve with this the problem that engineers wanna test modules at differing versions. Maybe im missing something?

docs/adr/0025_bindings_ci_and_release_strategy.md
```diff
@@ -0,0 +1,401 @@
...
+* **PR overhead**: `go.work` (git-ignored, generated in CI and locally via `task init/go.work`) allows developers
+  to make cross-binding changes in a single PR. The workspace ensures all interdependent modules resolve against
+  the current tree, not published versions, so a change spanning `bindings/go/credentials` and `bindings/go/helm`
+  can be reviewed and merged together.
+* **CI complexity**: module discovery is two steps with no filtering logic. New bindings enroll automatically. All
+  binding tests always run, eliminating the correctness gaps that came with change-based filtering.
+* **Release friction**: the phased bulk release computes dependency order, runs tests, gates on human approval, and
+  pins all consumers atomically. Developers no longer need to know or manually enforce the release sequence.
+
+The approach trades a small amount of CI compute (running all binding tests on every PR) for correctness and
```
jakobmoellerdev: this is not a "small amount of CI compute". you are effectively suggesting to test all modules always, including integration tests. This will elevate a PR running 6-10 jobs to a structure where every commit on every PR will have 50-60 checks even if we optimize (current main has 80 checks).

docs/adr/0025_bindings_ci_and_release_strategy.md
```diff
@@ -0,0 +1,401 @@
...
+
+## Conclusion
+
+The chosen approach directly addresses each identified pain point:
+
+* **PR overhead**: `go.work` (git-ignored, generated in CI and locally via `task init/go.work`) allows developers
+  to make cross-binding changes in a single PR. The workspace ensures all interdependent modules resolve against
+  the current tree, not published versions, so a change spanning `bindings/go/credentials` and `bindings/go/helm`
+  can be reviewed and merged together.
+* **CI complexity**: module discovery is two steps with no filtering logic. New bindings enroll automatically. All
```
jakobmoellerdev: This is already happening today, only requirement is the taskfile setup

docs/adr/0025_bindings_ci_and_release_strategy.md
```diff
@@ -0,0 +1,400 @@
...
+// For each binding module:
+internalDeps = go mod edit -json | Require[]
+    .filter(path => path.startsWith("ocm.software/open-component-model/bindings/go/"))
+
+// Topological sort (Kahn's algorithm) — deps precede their dependents
+ordered = topoSort(modules, depsMap)
+```
+
+Using `go mod edit -json` (toolchain-authoritative) rather than regex parsing ensures all `go.mod` syntax
+(replace directives, indirect markers, multi-line blocks) is handled correctly.
```
Skarlso: Why is any of this needed? 🤔

docs/adr/0025_bindings_ci_and_release_strategy.md
```diff
@@ -0,0 +1,400 @@
...
+```
+
+- **Released this run**: `go mod edit -require` to the new tag computed by `planRelease`.
+- **Skipped/unchanged with an existing tag**: `go mod edit -require` to `latestTag()`, ensuring consumers stay
+  current with upstream releases that happened outside the current run.
+- **Never tagged**: skipped; see *New binding lifecycle* below.
+
+`pinDeps` applies the result with `go mod edit -require` — a pure file edit, no network access. `go.sum` is not
+updated during the release; it is computed on the first standalone (non-workspace) build after the proxy has indexed
+the new tags.
```
Skarlso: I don't get his. 🤔 This means that none of our actual tags and dependencies are verified doesn't it? Also means that is someone would check out the tags and would build using `-mod=readonly` it could literally fail with missing go.sum entry error. 🤔  That would be super bad.

docs/adr/0025_bindings_ci_and_release_strategy.md
matthiasbruns: Going to verify this today in the PoC branch

docs/adr/0025_bindings_ci_and_release_strategy.md
Skarlso: Also, there is a problem here if we have module dependency cycles. go.work does not detect those. So if  A->B->A. Package level cyclic import is disallowed by go but module boundaries are fudged with go work and Khan's sorting requires a DAG.

docs/adr/0025_bindings_ci_and_release_strategy.md
matthiasbruns: Yeah I will drop this - I just wanted to say that we should not have to touch anything CI related, if we added a binding, but this was the state already as you have mentioned

docs/adr/0025_bindings_ci_and_release_strategy.md
```diff
@@ -0,0 +1,401 @@
...
+A binding that has never been tagged is skipped by `planRelease` — it appears in the gate summary as
+`(no prior tag; skipped)` and receives no semver tag in the bulk release. During this pre-release period the
+binding is still usable: dependents can reference it via a Go pseudo-version
+(`v0.0.0-<timestamp>-<commit>`) which the Go toolchain derives from the commit SHA. In workspace mode
+(`go.work`) this is transparent to developers — the local tree is used directly regardless of what version
+is declared in `go.mod`.
+
+When the binding is ready for stable versioning:
+
+1. Merge the binding's implementation to `main`.
```
Skarlso: So, we push to main and create a tag...? Where does the mod change land? In the tag or main or both?

docs/adr/0025_bindings_ci_and_release_strategy.md
matthiasbruns: I thought I updated all "keep commited" yesterday, but saw I haven't - we wont commit .work at all - also because of go.work.sum conflicts in the PRs

docs/adr/0025_bindings_ci_and_release_strategy.md
matthiasbruns: We have two binding tag releases

1. the manual one that we already have (explained here) - if we want to life a binding from a 0.0.0-pseudo-version to a 0.0.1, we need to use the manual workflow to do a first release of a binding
2. the release-flow, will only lift already released bindings. If we did not release a binding before, the release flow should behave as we currently work - just use the pseudo versions

we can also change that and always pin bindings, even if they were never released before, but I did not want to change they way we release bindings

docs/adr/0025_bindings_ci_and_release_strategy.md
matthiasbruns: yeah I will consolidate the ADR sections

docs/adr/0025_bindings_ci_and_release_strategy.md
matthiasbruns: Do they? Isn't that the goal of the release-flow changes, that we do not have to think about this in the first place?

docs/adr/0025_bindings_ci_and_release_strategy.md
matthiasbruns: Cleaned the left-over from the "keep go.work" comited approach

docs/adr/0025_bindings_ci_and_release_strategy.md
matthiasbruns: Its actually `task bindings/test` as in https://github.com/matthiasbruns/open-component-model/actions/runs/28328574202/job/83924328432
will update

docs/adr/0025_bindings_ci_and_release_strategy.md
matthiasbruns: Yeah maybe not "small" - but what's the alternative without going bonkers to test dependency chains if one binding is modified?

`running all binding tests on every PR` I have no better solution.

Imagine we have a change in `credentials` that would break `oci` (but we do not update the code there) in a PR but only test change-based tests, could the issue slip through if we do not test `oci`?

docs/adr/0025_bindings_ci_and_release_strategy.md
jneisener: > I dont like assuming things without trying. I think you guys should try running the linter across a go work enabled repo. The reason there are multiple different runs right now is that there were in the past some linters on go.mod files that could only run per-module and broke with workspaces.

Yes, I think thats the intention behind https://github.com/matthiasbruns/open-component-model/pull/115.
Also the local runs were done with that branch (at a point where it had a top-level `go.work` file).


> Another thing, just because a 14 core Dev Mac can run a golangci lint run of all modules in a minute (with your local dev cache), does not mean this can be translated to CI. 

We discussed this in more depth during the daily. Run time did variate. I did clean the cache though. Neither `golangci-lint cache clean`  nor `GOLANGCI_LINT_CACHE=$(mktemp);` had a significant impact (plus/minus 10-20 seconds).
Regarding the 14 cores, I did not observe CPU usage spiking and running the command with `GOMAXPROCS=1` increased the time to 1min 30 seconds.



> The goal of the CI thus far has always been extremely fast response time. 

We discussed this too. My argument would be that for 'extremely' fast response time linting can be executed locally. Anything that runs in the pipeline will be slowed down by all the setup around it. What use is super fast linting if setting up the runner takes longer than the linting?

I think this boils down to: What is the expected turnaround time of a PR? If it is <3 minutes then we probably need to lint in parallel.

But maybe the entire discussion is overblown. Locally the difference between sparse checkout and full-checkout (fetch-depth: 1) is 1.5 seconds vs 4.5 seconds.

docs/adr/0025_bindings_ci_and_release_strategy.md
matthiasbruns: It's more about the filtering setup the CI does, with the path-filter chain. I will rephrase it, but this depends on if we are going for the "test-everything" direction

docs/adr/0025_bindings_ci_and_release_strategy.md
matthiasbruns: technically, we do not need this I think - let me try that out.
We should be able, since we already know which versions are being pinned, to just override them instead of doing this in a fixed order - but I am not too sure about go.sum and tidy - let me fiddle around a bit

matthiasbruns: but on the circular dependencies - is this even allowed in go? feels really wrong to me

docs/adr/0025_bindings_ci_and_release_strategy.md
matthiasbruns: Valid - we are basically only testing the go.work setup, but not the go.mod one, except for the final release of cli and controller.
What we could do is add another parallel job that checks if the monorepo setup without go.work is still valid - this could double or compute usage, depending on how deep we want to go. 
We need to decide what jobs should test what

e.g. conformance tests - go.work or no go.work - or both in parallel?

matthiasbruns:
I will move this back to draft, since there are too many open questions that I did not think about

docs/adr/0025_bindings_ci_and_release_strategy.md
matthiasbruns: Yeah didn't think of somebody actually checking out the release branch locally - in this case, this would break.

So we only have one option
- ditch the single commit pinning of bindings
- bring back topo-sort and release in order, meaning a pinning commit per binding + tag
- calculate go.sum in each step when we change the go.mod

docs/adr/0025_bindings_ci_and_release_strategy.md
Skarlso: It's allowed between _modules_ since they don't know about each other basically... There is no check for modules. There is check for packages in the same module.

docs/adr/0025_bindings_ci_and_release_strategy.md
jakobmoellerdev: IMHO this is something that we accept and that's also why modules lower in the hierarchy (having many dependents) need to be tested more vigorously. Upstream projects face this problem consistently and the correct answer from my view has always been to generalize a test case and move it into the module

matthiasbruns: After talking to the team in the WR, we decided to try to minimize the compute use while not overcomplicating everything - meaning, we will try to only test modules and the dependent (also transitive) ones

docs/adr/0025_bindings_ci_and_release_strategy.md
```diff
@@ -0,0 +1,409 @@
...
+1. **Cross-module pull-requests**: changes relating to multiple bindings can now be provided in one PR instead of having
+   a PR per binding change with binding releases in-between.
+2. **Cross-module regressions**: a change to `bindings/go/credentials` can silently break `bindings/go/helm` or `cli` if
+   only the changed module is tested.
+3. **Workspace management**: the repo-wide `go.work` at the root is always present during development or is generated
+   within the CI workflows.
+   It must reflect the checked-out tree so that workspace-aware commands resolve all dependencies locally during
+   development and CI.
+4. **Coordinated releases**: bindings must be released in dependency order. A manual per-module release can leave
+   dependent modules in an inconsistent state.
```
fabianburth: If we'd go for a lockstep release, does the order really matter?

docs/adr/0025_bindings_ci_and_release_strategy.md
fabianburth: Assuming we update `bindings/runtime` with a breaking change and adjust `bindings/credentials` in the same PR. If I now go get ocm.software/...@main, either way, `go.work` committed or not, this won't work, will it?

matthiasbruns:
Added all discussed topics to the new ADR update

jneisener:
We've had trouble visualizing the monorepo dependency tree we are talking about, so I took a stab at it:
image

Take with a grain of salt as this was generated. But superficially it looks correct

matthiasbruns:
> We've had trouble visualizing the monorepo dependency tree we are talking about, so I took a stab at it: 
> ...
> Take with a grain of salt as this was generated. But superficially it looks correct

This matches this output pretty well: https://github.com/matthiasbruns/open-component-model/actions/runs/28385613778/job/84099456472#step:9:15

matthiasbruns:
closing due to take over from johannes and prossible pivot/delay/drop of the idea

PR #2939, opened 2026-06-25
chore(GHA): remove blackduck scan workflow

frewilhelm:
The workflow is deleted in the source repository and disabled in this repository since some time. Hence, deleting it is just a clean up.


jneisener (approved):

morri-son (approved):
wech mit 😊

PR #3015, opened 2026-07-01
fix(oci): include SubPath in OCI credential consumer identity

### LINKED ISSUE #2990: Configuring Credentials for multiple registries fails, if they live on the same host
**Context**
We're using STACKIT's Container Registry, which splits registries by using paths, rather than subdomains. Credentials are scoped to the hostname, with no way to split them by path. Due to this, a transfer from Registry A to Registry B using `ocm transfer` isn't possible, as the CLI can't use different credentials for it. It can be worked around by first transfering from Registry A to a local CTF file with one `.ocmconfig`, then transfering that local CTF file to Registry B with another `.ocmconfig`.

**Version**
`{"major":"0","minor":"9","patch":"1","gitVersion":"0.9.1","goVersion":"go1.26.3","compiler":"gc","platform":"darwin/arm64"}`

Also occurs on `linux-amd64`.

**To Reproduce**
1. Setup two separate registries, which are served from the same FQDN, but on different paths
2. Follow the official documentation on how to [Configure Credentials for Multiple Registries](https://ocm.software/docs/how-to/configure-credentials-for-multiple-registries/) with different paths

**Actual behavior**
- Credentials are used based on hostname only; path overrides are ignored

**Expected behavior**
- Credentials are used based on hostname and path

**Additional Comments**
Tests seem to confirm this behavior: https://github.com/open-component-model/open-component-model/blob/e2671641205bbad5ea80c849e56cc6ca73a846ea/bindings/go/examples/06_oci_test.go#L280-L288

As reference: Using this .ocmconfig resulted in all requests using `REDACTED3:REDACTED3`:

```yaml
configurations:
  - type: credentials.config.ocm.software
    consumers:
      - identities:
          - type: OCIRegistry
            hostname: "registry.onstackit.cloud"
            path: k7e/*
        credentials:
          - type: Credentials/v1
            properties:
              username: "REDACTED1"
              password: "REDACTED1"
      - identities:
          - type: OCIRegistry
            hostname: "registry.onstackit.cloud"
            path: k7e-components-pre/*
        credentials:
          - type: Credentials/v1
            properties:
              username: "REDACTED2"
              password: "REDACTED2"
      - identities:
          - type: OCIRegistry
            hostname: "registry.onstackit.cloud"
        credentials:
          - type: Credentials/v1
            properties:
              username: "REDACTED3"
              password: "REDACTED3"
```

Somewhat related, logs also contain this output with no clear way to resolve them; the warning appears to be scoped at developers, rather than users of the OCM CLI:
```
time=2026-06-25T11:49:16.037Z level=WARN msg="resolved credentials for identity using direct credential resolution, consider configuring a CredentialTypeSchemeProvider" identity="hostname=registry.onstackit.cloud,type=OCIRegistry"
time=2026-06-25T11:49:16.037Z level=DEBUG msg="failed to parse identity type, defaulting to versioned type" error="missing identity attribute \"type\""
```

jakobmoellerdev:

`IdentityFromOCIRepository` and `OCIRegistryIdentityFromOCIRepository` built the credential lookup identity solely from `repository.BaseUrl`, ignoring the `SubPath` field entirely. This meant path-scoped consumer entries in `.ocmconfig` were unreachable — the lookup identity never carried a `path` attribute, so the credential graph always matched the first unscoped (hostname-only) entry regardless of the configured path patterns.

The fix applies the same normalization already used by \`buildResolver\` for the actual OCI client:

- **Explicit `SubPath`**: takes precedence over any path component already extracted from `BaseUrl`
- **Empty `SubPath`**: `ParseURLToIdentity` already auto-extracts the path from `BaseUrl`, so no additional code is needed for this case

This allows users to configure separate credentials per registry path on a shared hostname, e.g.:

```yaml
type: credentials.config.ocm.software
consumers:
  - identities:
      - type: OCIRegistry
        hostname: registry.onstackit.cloud
        path: k7e-components-pre/*
    credentials:
      - type: Credentials/v1
        properties:
          username: user-a
          password: pass-a
  - identities:
      - type: OCIRegistry
        hostname: registry.onstackit.cloud
        path: k7e/*
    credentials:
      - type: Credentials/v1
        properties:
          username: user-b
          password: pass-b
```

Fixes #2990

#### ⚠️ Breaking Change / Migration Notes

Before this fix, `SubPath` was never included in the credential lookup identity. As a result, credential entries with a `path` attribute in `.ocmconfig` were silently ignored — the lookup identity never carried a path, so the path-pattern match in `IdentityMatchesPath` was skipped and the entry matched all requests for that hostname regardless of path.

**After this fix**, path-scoped credential entries behave as intended: a configured entry with `path: some/pattern` only matches requests whose path matches that glob pattern.

**Who is affected**: users who have a credential config with a `path`-scoped identity entry and relied on it matching all requests to that hostname (because the path was never checked). Example of a config that previously matched all paths on the host but now only matches the stated pattern:

```yaml
- identities:
    - type: OCIRegistry
      hostname: registry.onstackit.cloud
      path: k7e-components-pre/*   # was previously ignored; now enforced
  credentials:
    - type: Credentials/v1
      properties:
        username: user-a
        password: pass-a
```

**Migration**: to apply credentials to all paths on a hostname, remove the `path` attribute from the identity entry. A hostname-only identity (no `path`) still matches all requests for that host. To scope credentials per path, add separate entries per pattern as shown in the multi-entry example above.

##### How to test the changes

Configure two credential entries for the same hostname with different `path` patterns as shown above. Run `ocm transfer` from `registry.onstackit.cloud/k7e-components-pre/...` to `registry.onstackit.cloud/k7e/...` — each path should resolve its own credentials instead of falling through to a hostname-only fallback.

##### Verification

### DIFF
--- a/bindings/go/oci/spec/identity/v1/type.go
+++ b/bindings/go/oci/spec/identity/v1/type.go
@@ -13,6 +13,12 @@ func IdentityFromOCIRepository(repository *oci.Repository) (runtime.Identity, er
 	if err != nil {
 		return nil, fmt.Errorf("could not parse OCI repository URL: %w", err)
 	}
+	// When SubPath is explicitly set, use it directly (BaseUrl is host-only).
+	// When SubPath is empty, ParseURLToIdentity already extracted any path
+	// component from BaseUrl into identity[IdentityAttributePath].
+	if repository.SubPath != "" {
+		identity[runtime.IdentityAttributePath] = repository.SubPath
+	}
 	identity.SetType(Type)
 	return identity, nil
 }
@@ -21,6 +27,9 @@ func OCIRegistryIdentityFromOCIRepository(repository *oci.Repository) (*OCIRegis
 	identity, err := runtime.ParseURLToIdentity(repository.BaseUrl)
 	if err != nil {
 		return nil, fmt.Errorf("could not parse OCI repository URL: %w", err)
+	}
+	if repository.SubPath != "" {
+		identity[runtime.IdentityAttributePath] = repository.SubPath
 	}
 	return FromIdentity(identity), nil
 }
--- a/bindings/go/oci/spec/identity/v1/type.go
+++ b/bindings/go/oci/spec/identity/v1/type.go
@@ -2,23 +2,32 @@ package v1
 
 import (
 	"fmt"
+	"strings"
 
 	"ocm.software/open-component-model/bindings/go/oci/spec/repository/v1/ctf"
 	"ocm.software/open-component-model/bindings/go/oci/spec/repository/v1/oci"
 	"ocm.software/open-component-model/bindings/go/runtime"
 )
 
+// applySubPath sets the path identity attribute from subPath when non-empty,
+// overriding any path already extracted from BaseUrl. The leading slash is
+// stripped to match the normalization applied by ParseURLToIdentity.
+func applySubPath(identity runtime.Identity, subPath string) {
+	if subPath != "" {
+		identity[runtime.IdentityAttributePath] = strings.TrimPrefix(subPath, "/")
+	}
+}
+
 func IdentityFromOCIRepository(repository *oci.Repository) (runtime.Identity, error) {
 	identity, err := runtime.ParseURLToIdentity(repository.BaseUrl)
 	if err != nil {
 		return nil, fmt.Errorf("could not parse OCI repository URL: %w", err)
 	}
-	// When SubPath is explicitly set, use it directly (BaseUrl is host-only).
+	// When SubPath is explicitly set, it takes precedence over any path
+	// component already extracted from BaseUrl.
 	// When SubPath is empty, ParseURLToIdentity already extracted any path
 	// component from BaseUrl into identity[IdentityAttributePath].
-	if repository.SubPath != "" {
-		identity[runtime.IdentityAttributePath] = repository.SubPath
-	}
+	applySubPath(identity, repository.SubPath)
 	identity.SetType(Type)
 	return identity, nil
 }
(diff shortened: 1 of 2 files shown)


bindings/go/oci/spec/identity/v1/type.go
```diff
@@ -13,6 +13,12 @@ func IdentityFromOCIRepository(repository *oci.Repository) (runtime.Identity, er
 	if err != nil {
 		return nil, fmt.Errorf("could not parse OCI repository URL: %w", err)
 	}
+	// When SubPath is explicitly set, use it directly (BaseUrl is host-only).
+	// When SubPath is empty, ParseURLToIdentity already extracted any path
+	// component from BaseUrl into identity[IdentityAttributePath].
+	if repository.SubPath != "" {
+		identity[runtime.IdentityAttributePath] = repository.SubPath
```
matthiasbruns: to understand the impact

if we only have one identity in our config, where `path` is set, this identity will not work for all paths in the domain, but only for the `path`?

Could this be a problem for users with existing configs?

jakobmoellerdev: why would that be the case? my assumption is that the identity should match with subset matching. is that not the case?

matthiasbruns: thats my question :) 
I will check tomorrow

matthiasbruns (dismissed):

bindings/go/oci/spec/identity/v1/type.go
matthiasbruns: Okay so...

The only thing that could happen is the following:

Since we never checked `path`, a config like this that worked before

```yml
  - identities:
      - type: OCIRegistry
        hostname: registry.onstackit.cloud
        path: k7e-components-pre/*
    credentials:
      - type: Credentials/v1
        properties:
          username: user-a
          password: pass-a
```

for every route with the domain `registry.onstackit.cloud`, wouldn't work anymore.
I think this is not a big deal, but we should write this into the release/pr notes somewhere, that the fix might lead to breaking previously wrongly working configs.

frewilhelm (approved):

Skarlso (approved):

PR #3028, opened 2026-07-03
chore(deps): configure renovate to update`bindings/go/examples` too

frewilhelm:

Currently, renovate ignores `bindings/go/examples` because [`config:recommended`](https://docs.renovatebot.com/presets-config/#configrecommended) introduces [`ignoreModulesAndTests`](https://docs.renovatebot.com/presets-default/#ignoremodulesandtests) which sets `ignorePaths` that includes `**/examples/**`.

The proposed change disables the `ignorePaths` setting for renovates gomod package manager completely.

I ran renovate on my [local fork](https://github.com/frewilhelm/open-component-model/pulls?q=is%3Apr+is%3Aopen) and it [updated](https://github.com/frewilhelm/open-component-model/pulls?q=is%3Apr+is%3Aopen+examples) the `bindings/go/examples` module too, see for example [here](https://github.com/frewilhelm/open-component-model/pull/582/changes).


.github/renovate.json5
```diff
@@ -14,6 +14,15 @@
...
     enabled: true,
   },
+  // config:recommended transitively pulls in :ignoreModulesAndTests, which
+  // sets ignorePaths to include **/examples/**. That causes Renovate to skip
+  // bindings/go/examples/go.mod entirely, so its dependencies drift out of
+  // sync with the rest of the Go modules. ignorePaths is not mergeable, so
+  // we scope the override to the gomod manager and empty the list: every
+  // go.mod in this repo is a real, maintained module and should be scanned.
+  gomod: {
+    ignorePaths: [],
```
matthiasbruns: will this have more side-effect? cant we just whitelist the examples binding instead of purging the blacklist?

frewilhelm: I did not find an option to whitelist it. So, we could just reset the default list and only remove the `examples` part.

These are the paths that are ignored by default

```
  "ignorePaths": [
    "**/node_modules/**",
    "**/bower_components/**",
    "**/vendor/**",
    "**/examples/**",
    "**/__tests__/**",
    "**/test/**",
    "**/tests/**",
    "**/__fixtures__/**"
```

I don't think we need any of these paths for the `gomod` match manager.

matthiasbruns: https://docs.renovatebot.com/configuration-options/#includepaths
also path is a glob, maybe in `ignorePaths` you can use `!`

frewilhelm: > Include package files only within these defined paths.

I read this like, if you define `includePackage` than only that one package is checked for updates. If this is true, then we would need to specify every path we want updates for which seems more complex than overwriting paths that are not necessarily go-related and not relevant for our project. Nevertheless, I will play around with this a bit

frewilhelm: If I get this right, `ignorePaths` is evaluated after `includePaths` (see [code](https://github.com/renovatebot/renovate/blob/cb9a5ae71bb645eefcc3eb812080a030eb34d0a3/lib/workers/repository/extract/file-match.ts#L47)). Hence, if we use `includePaths`, we would need to specify all paths we want to update and use `ignorePaths`, too. So, `includePaths` does not seem like an option for me. Regarding `ignorePaths`, we could re-set all the paths I unset in this PR but I don't see a reason why we should do this as these paths are currently not relevant in our monorepository.

frewilhelm: I ran it again with

```
   gomod: {
     includePaths: ["bindings/go/examples", "**"],
   },
```

and it did [not update](https://github.com/frewilhelm/open-component-model/pull/547/changes) the examples package.

frewilhelm: We discussed this and decided to be less invasive

frewilhelm: Seems to work :) https://github.com/frewilhelm/open-component-model/pull/582/changes

jakobmoellerdev (approved):

jneisener (approved):

PR #3029, opened 2026-07-03
fix: Make version specific references dynamic

jneisener:
Recent version change broke these hardcoded links in the code examples, see https://github.com/open-component-model/ocm-project/issues/1160 for more context.

Also, the install instructions would always install the latest version and thus not match the currently selected version.

https://ocm.software/latest/schemas/bindings/go/constructor/schema-2020-12.json -> `404`

https://ocm.software/main/schemas/bindings/go/constructor/schema-2020-12.json -> `200`

Decision was to leave the current links on 0.9 and 0.10 broken (i.e. they still point to latest). But from 0.11 forward (or if we would choose to backport this) the changes of this PR will also affect the versioned website and not just main.

#### Testing
Tested locally with a fictitious 0.11 version:
image
image
image

Install script tested locally:
```
$ OCM_VERSION=0.10 bash static/install-cli.sh 
[INFO]  Downloading metadata https://api.github.com/repos/open-component-model/open-component-model/releases?per_page=100
[INFO]  Using 0.10.0 as release
[INFO]  Downloading binary https://github.com/open-component-model/open-component-model/releases/download/v0.10.0/ocm-darwin-arm64
```

##### Verification


website/layouts/_markup/render-codeblock-shell.html
```diff
@@ -0,0 +1,23 @@
+{{- $inner := .Inner -}}
+{{- with .Page.Site.Version -}}
+  {{/* "latest" appears in pinned Go module content (v0.9.1, v0.10.0) that cannot be updated */}}
```
jneisener: Well technically we can update it, but we would have to backport and release that change, which might not be worth it

frewilhelm (commented):
I hoped there would be some kind of variable containing the current version provided by Hugo. Do you know if there is anything like that?

jakobmoellerdev:
yes there should be, please lets make this content dynamic!

jneisener:
> I hoped there would be some kind of variable containing the current version provided by Hugo. Do you know if there is anything like that?

Yeah, the `.Page.Site.Version` that I'm using atm - or what do you mean? 
~I can't use a shortcode directly in the yaml block because they don't expand inside of fences~ Actually, it appears this was incorrect, they do get rendered inside the codeblock, so the simple approach would work, but only if we backport the `.Site.Version.Name`, because otherwise I don't know how I would replace the `latest` here: https://github.com/open-component-model/open-component-model/blob/website/v0.10.0/website/content/docs/getting-started/create-component-version.md#define-the-component

jneisener:
> yes there should be, please lets make this content dynamic!

Could you clarify what you mean by dynamic? Because by my definition it would be dynamic right now.
Do you mean rendered and resolved in the browser?

jakobmoellerdev (approved):

website/static/install-cli.sh
```diff
@@ -147,9 +148,32 @@ extract_stable_version() {
...
 }
 
+# Extract the latest stable version matching a MAJOR.MINOR prefix from a releases JSON file.
+extract_stable_version_for_minor() {
+    local minor_prefix="$1"
+    grep '"tag_name":' "$2" \
+        | grep -E "\"v${minor_prefix}\.[0-9]+\"" \
+        | sed -E "s|.*\"v([^\"]+)\".*|\1|" \
+        | sort -V \
+        | tail -1 \
```
frewilhelm, line 158: what is this for? Is it a safeguard in case the results from github are not sorted?

frewilhelm (commented):
only a question. overall lgtm

website/static/install-cli.sh
jneisener: Its a reuse of the pattern from `extract_stable_version`

https://github.com/open-component-model/open-component-model/blob/0df7ec288b2e65096be9c5f847624877c8038e7c/website/static/install-cli.sh#L134-L148

frewilhelm (approved):
do we want to backport this to 0.11 so we can see it working next week?

jneisener:
> do we want to backport this to 0.11 so we can see it working next week?

If its not too much work I think that would be cool :)

PR #3030, opened 2026-07-03
chore(website): Add a section about how we handle spillover

frewilhelm:
This is a proposal on how to handle sprint-item spillovers in the future. Usually, we just put the spillover-item into the next sprint. However, this skews our actual work amount for the last and the current sprint because the work done in the last sprint is not reflected in the last sprint and the work for the next sprint should not be the same amount/complexity as some work was already done.

### DIFF
--- a/website/content/community/how-we-work.md
+++ b/website/content/community/how-we-work.md
@@ -38,7 +38,7 @@ All work is tracked on the
 | [Roadmap](https://github.com/orgs/open-component-model/projects/10/views/5) | Timeline view of planned work |
 | [Epics](https://github.com/orgs/open-component-model/projects/10/views/16) | High-level feature tracks |
 
-## How Issues Enter the Sprint
+### How Issues Enter the Sprint
 
 Anyone can propose work for an upcoming sprint. The process is:
 
@@ -51,6 +51,23 @@ Anyone can propose work for an upcoming sprint. The process is:
 4. **Refinement discussion** - during the weekly refinement meeting the team discusses the issue. If everyone
    understands it, it is story-pointed and its priority is evaluated.
 5. **Ready for sprint** - once refined, the issue moves to the "Next-Up" column and is available for sprint planning.
+
+### Handling Sprint Spillover
+
+Items that don't finish in their sprint are not carried over. We close the original
+issue and, if the remaining work is still worth doing, track it as a new sub-issue
+that goes through refinement.
+
+**At sprint end**, for any issue that is not "Done":
+
+1. Add a closing comment: what shipped, what didn't, what was learned.
+2. Close the issue.
+3. If the remaining work is still worth doing, open a
+   [sub-issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/adding-sub-issues)
+   under the closed one, scoped to what is left.
+4. The sub-issue enters `Needs Refinement` and gets its own story-point vote based on
+   the remaining scope, not the original estimate.
+5. It is prioritized at Planning against every other candidate.
 
 ## Meetings
 


website/content/community/how-we-work.md
```diff
@@ -52,6 +52,23 @@ Anyone can propose work for an upcoming sprint. The process is:
...
+
+Items that don't finish in their sprint are not carried over. We close the original
+issue and, if the remaining work is still worth doing, track it as a new sub-issue
+that goes through refinement.
+
+**At sprint end**, for any issue that is not "Done":
+
+1. Add a closing comment: what shipped, what didn't, what was learned.
+2. Close the issue.
+3. If the remaining work is still worth doing, open a
```
matthiasbruns, line 65: >  under the closed one

why not simply duplicating the issue that would be carried over and update the AC und description with a link to the old issue

website/content/community/how-we-work.md
```diff
@@ -52,6 +52,23 @@ Anyone can propose work for an upcoming sprint. The process is:
    understands it, it is story-pointed and its priority is evaluated.
 5. **Ready for sprint** - once refined, the issue moves to the "Next-Up" column and is available for sprint planning.
 
+### Handling Sprint Spillover
+
+Items that don't finish in their sprint are not carried over. We close the original
+issue and, if the remaining work is still worth doing, track it as a new sub-issue
+that goes through refinement.
```
matthiasbruns, line 59: but during planning right? all of this happens in the planning, doesn't it?

frewilhelm: I don't think we have enough time in the planning to do prioritisation, sprint goals, and refining spill-overs. Usually, we already need longer. I think it would make sense to do this async before the planning. WDYT?

website/content/community/how-we-work.md
frewilhelm: We can do a duplicate as well. We just need to make sure that the connection between both issues is noticeable in the list of sub-issue in the epic. Any idea how we could do this? A prefix on the issue title maybe?

website/content/community/how-we-work.md
matthiasbruns: I mean, if its a spillover, cant this be something like
"what do you thing - what is remaining" - and the dev who does it says just the points?
prio wont change or might be increased - only in certain cases we would drop it or remove it from the next sprint, this should be 15-30 second max per ticket in common cases

website/content/community/how-we-work.md
matthiasbruns: why not just linking them? or mention it in the issue description.

frewilhelm: I was thinking about having a structured overview on the epic that lists those issues as sub-tasks. Without visible indicator, you need to open each issue and check the description for each. Hence, it would just be a bit more cumbersome. But I won't die on that hill :D. 

But I believe we should have an idea on how to title those spill-over tasks. Should we just use the same name?

matthiasbruns: maybe as geralds prs? [172] and then [2/2] as a suffix?

website/content/community/how-we-work.md
frewilhelm: We kind of did this in the last planning with https://github.com/open-component-model/ocm-project/issues/1087 and https://github.com/open-component-model/ocm-project/issues/1162. I did not have the impression it only took 15-30 seconds.

I am really concerned that we will start discussions about spill-overs and this will exceed the time we have in the planning.

What is the reason you would rather do it sync than async?

matthiasbruns: My bad, didn't read "before the planning" - I would ditch the voting though - every dev should be able to set the remaining story points by himself before the planning, if not sure, they can still vote - wdyt?

website/content/community/how-we-work.md
```diff
@@ -52,6 +52,23 @@ Anyone can propose work for an upcoming sprint. The process is:
    understands it, it is story-pointed and its priority is evaluated.
 5. **Ready for sprint** - once refined, the issue moves to the "Next-Up" column and is available for sprint planning.
 
+### Handling Sprint Spillover
+
+Items that don't finish in their sprint are not carried over. We close the original
```
jakobmoellerdev, line 57: This is not true and I dont know how we got to this :D

matthiasbruns: I think that is the desired goal :)

jakobmoellerdev: I understand thats the proposal, but tbh I want to push back on this. Ive been in setups where this has been enforced and it often leads to issue churn that is very hard to track, so I usually favor accepting the trade-off that an issue gets carried over. That said, happy to hear other thoughts and get overruled

frewilhelm: I believe we wanted to add this because the amount of "work" (although story-points do not reflect the work necessarily) was not tracked between sprints. Take the bindings ADR as an example:
- Matthias and Johannes did put a lot of work into it in Sprint 65
- If we just move their [issue](https://github.com/open-component-model/ocm-project/issues/1087) into the next sprint, those 5 story-points are not reflected in the "statistics" for the Sprint 65 but would potentially overflow the amount of story-points for Sprint 66
- Additionally, if we would decide to not put 5 story-points on the remaining work of the spill-over, because most of the work is already done, the 5 story-points would skew the amount of story-points for the next sprint.

Does this make sense?

jakobmoellerdev: I understand and accept these tradeoffs in favor of not having to create even more issues than we have already. its a matter of trading correctness on tracking in one sprint vs duplicating issues that carry over. if we wouldnt do it that often I wouldnt mind but I feel like we have a lot of carry over so the effect is quite large.

matthiasbruns: Fair - maybe, we can keep this in draft and first figure out a process we can actually live with instead of doing this beforehand?

frewilhelm: Yeah, thats why I opened this as a proposal to discuss. I will create a short meeting where we can discuss this

frewilhelm:
Closed after a discussion with the result that we don't want to pursue this proposal. We will observe how we handle spill-overs in the future and might adjust the process.

PR #3031, opened 2026-07-03
feat(release): add publish OCM component-versions for CLI, controller and product to release workflow

morri-son:
Publishes three OCM component-versions to `ghcr.io` on every release (RC and final):

- `ocm.software/cli` — 6 binaries (local blobs) + multi-arch CLI image by digest
- `ocm.software/controller` — controller image + Helm chart, both by digest
- `ocm.software/ocm` — product wrapper referencing the two via `componentReferences`

All three carry the same semver as the GitHub release. Both RC and final use `--component-version-conflict-policy replace` for idempotent reruns.

New reusable workflow `.github/workflows/publish-components.yml` called twice from `release.yml` (RC + final). All three constructors generated inline via shell heredoc. `promote_and_release_final` gets a new output `final_chart_digest` as the chart is repackaged at promotion, so the digest differs from RC.

Updated `RELEASE_PROCESS.md` with a new "OCM components produced" section. ADR 0010 was updated, too.

I validated this end-to-end on my `morri-son/open-component-model` fork (I needed to tweak the workflow a bit and replace OCMBot and GPG signing requirements to be able to go without the secrets we use on main)

### DIFF
--- a/.github/workflows/publish-components.yml
+++ b/.github/workflows/publish-components.yml
@@ -1,0 +1,274 @@
+name: Publish OCM components
+
+# Publishes ocm.software/{cli,controller,ocm} component-versions to GHCR.
+# The CLI binary used to invoke `ocm add component-version` is the one this
+# release just built — downloaded from `cli_artifact_name`. Image and chart
+# artefacts referenced by the constructors are pinned by digest, not fetched.
+
+on:
+  workflow_call:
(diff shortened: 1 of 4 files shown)


jakobmoellerdev:
closing in favor of https://github.com/open-component-model/open-component-model/pull/3071

PR #3050, opened 2026-07-06
chore: add ci script testing pipeline

matthiasbruns:
This PR adds a github CI script verification workflow. We have a lot of unit tests in the scripts but are never testing them.
During a renovate pr https://github.com/open-component-model/open-component-model/pull/3048 I flagged this.

https://github.com/open-component-model/open-component-model/pull/3048

- missing unit tests of scripts
- missing linting of scripts


jakobmoellerdev (dismissed):

.github/scripts/.gitignore
```diff
@@ -0,0 +1,2 @@
+node_modules/
```
frewilhelm: We already have `.github/scripts/node_modules/` in `/.gitignore`. Do we want to create another `.gitignore` or use the root-level file?

matthiasbruns: Yeah I am actually a fan of local/scopes .gitignores but having the node one in the root makes this obsolete

frewilhelm (approved):

jakobmoellerdev (approved):

PR #3058, opened 2026-07-07
chore(releases-011): Cherry-pick #3015 and #3029

frewilhelm:
#3015 (fix in OCI module) includes a fix for #2990 that get introduced to the cli through #3033 (monorepo bump PR). Instead of cherry-picking the whole PR, I only checked out the relevant files for `cli` (using the whole PR introduces tons of merge conflicts as there are a lot of commits missing in-between for various modules. Additionally, I had to bump the go version of the module `generator`).
We want to have a RC that the user can test with. 

#3029 contains a change to have dynamic version strings on the website depending on the version that is chosen for the website. We want to include this to test the new feature with the promotion of `0.11.0` next week.


frewilhelm:
As the CI is failing (because of go 1.26.3 != 1.26.4) I am thinking about only bumping the OCI module instead of using the whole file in `cli/go.mod`

jneisener (approved):

jakobmoellerdev (approved):

PR #3059, opened 2026-07-07
chore(docs): ADR: Bindings CI and Release Strategy

### LINKED ISSUE #1087: Create ADR to decide what direction we want to go with our module bindings

jneisener:

Introduces and ADR to summarize our considered approaches and possible solutions for improving the developer experience for the bindings. 

Fixes: https://github.com/open-component-model/ocm-project/issues/1087

##### Verification


docs/adr/0025_bindings_ci_and_release_strategy.md
```diff
@@ -0,0 +1,314 @@
...
+Most tests would run on every commit, a shared downside with the `go.work` approach, though harder to optimize since there's no module-level dependency graph to derive an affected set from.
+
+Consumers could still choose to consume only individual modules of the bundled library, and in principle Go would optimize the unused parts away. 
+
+#### Security scans and reflection
+
+This approach could be feasible, except that it breaks the consumer experience in two key ways:
+
+* Due to our internally used reflection, Go cannot optimize away the unused libraries. Consumers would receive all our (transitive) dependencies into their BOM.
+* Security scans will trigger on anything inside our entire dependency tree. This was and is a big pain-point in OCM v1. With this approach a consumer that e.g. only needs `descriptor/v2` would now also receive the entire helm SDK and thus be impacted by any security vulnerabilities discovered within.
```
jneisener: Verify by generating the actual dependency tree, this finding was true in v1 but might not be true for v2

jneisener: So in v1 it might not be about the reflection alone, it seems like a primary cause is the `init.go` in `compdesc` (which you probably can't do much without): https://github.com/open-component-model/ocm/blob/main/api/ocm/compdesc/init.go

So it does appear as if v2 would not have that problem: https://github.com/jneisener/open-component-model/blob/a58aa4816fc7b42d56a82009f72d6070c99869ae/docs/adr/0025_evidence_monolithic/results.md

PoC: https://github.com/jneisener/open-component-model/commit/a58aa4816fc7b42d56a82009f72d6070c99869ae

frewilhelm: Just as a test: What happens when you remove the imports in `init.go`? Does this change something for ocm v1 or are the imports still there due to

```
compdesc -> credentials -> config -> datacontext -> ...
          -> signing/handlers -> sigstore (+ cosign + rekor + fulcio)
          -> credentials/extensions/repositories -> vault + dockerconfig + gardener
```

jneisener: > Just as a test: What happens when you remove the imports in `init.go`? Does this change something for ocm v1 or are the imports still there due to

Good point. Removing the `init.go` only drops the `sigstore` dependency. We still import credentials which has its own [`init.go`](https://github.com/open-component-model/ocm/blob/db3775efda0d935880e2bfd964c6de62ae0447d3/api/credentials/init.go#L4) and pulls in vault, dockerconfig, gardener, and npm.

frewilhelm: Do you know what the reason for the `init.go` was? EDIT: Rather the imports in that file

jneisener: We discussed this in the warroom and @jakobmoellerdev shared some of the history. My one-line (hopefully fair) summary: This was a pattern for type registering adopted from the kubernetes controller world.

jakobmoellerdev (approved):
possibly one of, if not the best simplifications the v2 codebase has seen this year. thanks to everyone contributing here.

frewilhelm (approved):
<3

docs/adr/0025_bindings_ci_and_release_strategy.md
```diff
@@ -0,0 +1,460 @@
...
+* Increases CI complexity; what the CI tests is not what consumers receive
+* Would require a large time investment
+
+### Option 2: Monolithic binding library(ies)
+
+Pros:
+
+* Solves inter-module development friction & inter-module regressions
+* Simple to implement, simple release process
+* Simple migration - no consumer import path changes required
```
jneisener, line 332: Expected change in `go.mod` files, before:
```go
  require (
      ocm.software/open-component-model/bindings/go/descriptor/v2 v2.0.3-alpha3
      ocm.software/open-component-model/bindings/go/runtime v0.0.8
  )
```

after:
```go
 require ocm.software/open-component-model/bindings/go v0.1.0
```

PR #3062, opened 2026-07-07
feat: add input to wget bindings

### LINKED ISSUE #1104: wget input method for component constructor
**Description**
Implement a `wget` input method in `bindings/go/wget/input/` that allows users to reference HTTP/S URLs as resource inputs in `component-constructor.yaml`.

During component construction the input method fetches the remote resource via HTTP/S and either:
- stores it as a local blob (for air-gapped / self-contained CTF), or
- sets a `wget` access spec pointing at the original URL

The implementation must implement the `ResourceInputMethod` interface from `bindings/go/constructor/interface.go` and be registered in the constructor input method registry.

Example usage in `component-constructor.yaml`:
```yaml
resources:
  - name: my-binary
    type: blob
    input:
      type: wget
      url: https://example.com/releases/v1.0.0/binary.tar.gz
      mediaType: application/gzip
```

Reuses the HTTP client from `bindings/go/http` (`http.config.ocm.software` config type). Credential resolution follows the same consumer identity pattern as the access type (see `bindings/go/wget/`).

Part of EPIC #958. Depends on #1103 (wget access type merged).

**Done Criteria**
- [ ] `wget` input method implemented in `bindings/go/wget/input/`
- [ ] Implements `ResourceInputMethod` interface from `bindings/go/constructor/interface.go`
- [ ] Registered in constructor input method registry
- [ ] CLI test cases added
- [ ] Supports both local-blob and access-spec output modes
- [ ] Uses `bindings/go/http` client with config resolution

matthiasbruns:

Input method addition to the wget binging.
Full implementation with CLI tests can be found here https://github.com/matthiasbruns/open-component-model/pull/116

Contributes: https://github.com/open-component-model/ocm-project/issues/1104

##### How to test the changes

See linked [PR](https://github.com/matthiasbruns/open-component-model/pull/116) in my fork

##### Verification


bindings/go/wget/internal/download/download.go
```diff
@@ -0,0 +1,243 @@
...
+	for _, opt := range opts {
+		opt(o)
+	}
+
+	if req.URL == "" {
+		return nil, fmt.Errorf("url is required")
+	}
+	client := o.Client
+	if client == nil {
+		client = http.DefaultClient
```
piotrjanik, line 56: We recently developed the HTTP client `bindings/go/http`. Perhaps it's worth considering using it to standardise HTTP clients across OCM.

matthiasbruns: That's why we allow injecting it in the ResourceRepo

https://github.com/matthiasbruns/open-component-model/blob/6dda47da19a4789f24993e0da5b8066c57ac9f5c/bindings/go/wget/repository/resource_repository.go#L24

and in input
https://github.com/open-component-model/open-component-model/pull/3062/changes#diff-951b1cc99ed2bad2211e8658293474da44140f02e408fef5a897ed07c08bcd3aR80

download.go doesn't care about that and takes what's being passed to it through the options 
https://github.com/open-component-model/open-component-model/pull/3062/changes#diff-efdf2bab3263d371db0d4adf0d4b640bf956a53110bb86f09d668987546a1a85R32

bindings/go/wget/access/access.go
jakobmoellerdev, line 1: can we move this entire function as its on top level and super confusing for me?

bindings/go/wget/input/method.go
```diff
@@ -0,0 +1,125 @@
...
+
+// InputMethod implements the [constructor.ResourceInputMethod] interface for wget-based inputs.
+// It downloads a resource from an HTTP/S URL declared in the component constructor
+// and returns it as a local blob to be stored in the component version.
+type InputMethod struct {
+	// HTTPConfig configures the HTTP client (timeouts, retries, TLS, routing) used for
+	// downloads. When nil, a default client is used.
+	HTTPConfig *httpv1alpha1.Config
+	// MaxDownloadSize limits the number of bytes read from a response body. When zero,
+	// the download package default is used. A negative value disables the limit.
```
jakobmoellerdev: what is that default?

bindings/go/wget/input/method.go
```diff
@@ -0,0 +1,125 @@
...
+
+var _ constructor.ResourceInputMethod = (*InputMethod)(nil)
+
+// InputMethod implements the [constructor.ResourceInputMethod] interface for wget-based inputs.
+// It downloads a resource from an HTTP/S URL declared in the component constructor
+// and returns it as a local blob to be stored in the component version.
+type InputMethod struct {
+	// HTTPConfig configures the HTTP client (timeouts, retries, TLS, routing) used for
+	// downloads. When nil, a default client is used.
+	HTTPConfig *httpv1alpha1.Config
```
jakobmoellerdev, line 27: I think before we used fields by value and did not inser the HTTP Config here, do you mind checking this again? In general the behavior is okay for me just want to make sure the contracts fit

bindings/go/wget/input/method.go
```diff
@@ -0,0 +1,125 @@
...
+// In the default local-blob mode it downloads the resource and returns it as local blob data.
+// When the input sets Reference, it instead returns a processed resource
+// carrying a wget access specification pointing at the URL, without downloading anything.
+func (i *InputMethod) ProcessResource(ctx context.Context, resource *constructorruntime.Resource, credentials runtime.Typed) (*constructor.ResourceInputMethodResult, error) {
+	wget := v1.Wget{}
+	if err := i.GetInputMethodScheme().Convert(resource.Input, &wget); err != nil {
+		return nil, fmt.Errorf("error converting resource input spec: %w", err)
+	}
+
+	if wget.URL == "" {
```
jakobmoellerdev, line 76: shouldnt this call validate here?

bindings/go/wget/input/method.go
```diff
@@ -0,0 +1,125 @@
...
+		return nil, fmt.Errorf("url is required in wget input spec")
+	}
+
+	// Access-spec mode: do not download. Store the resource with a wget access specification
+	// pointing at the URL so the content is resolved lazily when the resource is accessed.
+	if wget.Reference {
```
jakobmoellerdev: this is not a valid case because then you can just add the wget access directly instead of having to use the input method right?

bindings/go/wget/internal/download/download.go
```diff
@@ -0,0 +1,247 @@
...
+//     certificate, applied to the transport independently of the header auth below
+//   - identityToken: Bearer token in the Authorization header
+//   - username + password: HTTP Basic Authentication
+//
+// The mTLS client certificate composes with the header-based auth. Bearer and
+// Basic both use the Authorization header and are mutually exclusive; the bearer
+// token takes precedence when both are set.
+//
+// Both WgetCredentials/v1 and legacy DirectCredentials/v1 are accepted.
+func applyCredentials(ctx context.Context, req *http.Request, client **http.Client, credentials runtime.Typed) error {
```
jakobmoellerdev, line 168: this looks like duplicated logic from the HTTP package. do you mind checking and unifying?

bindings/go/wget/doc.go
```diff
@@ -31,6 +32,24 @@
...
+// [ocm.software/open-component-model/bindings/go/wget/input.InputMethod], which
+// references an HTTP/S URL declared on a resource in a component-constructor.yaml
+// through a
+// [ocm.software/open-component-model/bindings/go/wget/spec/input/v1.Wget] input
+// spec. The input spec carries the same request details as the access spec and
+// supports two output modes:
+//
+//   - local blob (default): the content is downloaded during construction and
+//     stored as a local blob, making the component version self-contained.
+//   - access spec (the input sets Reference): the content is not downloaded; the
```
jakobmoellerdev: this second case is not needed imho, see above.

bindings/go/wget/input/method.go
matthiasbruns: Maybe I misunderstood the issue - I thought the behaviour should have been configurable in the input itself - let me change this

bindings/go/wget/internal/download/download.go
matthiasbruns: This was basically moved out of repository https://github.com/open-component-model/open-component-model/blob/main/bindings/go/wget/repository/resource_repository.go#L193, but sure I can check

bindings/go/wget/doc.go
matthiasbruns: see above, will reiterate

bindings/go/wget/input/method.go
matthiasbruns: Sure I can validate there as well, it would fail down the line latest in the `downloader.go`

bindings/go/wget/access/access.go
matthiasbruns: ehm.. what 😅? 

this was merged like that into main - not released yet, so we can still move things around
https://github.com/open-component-model/open-component-model/blob/main/bindings/go/wget/access/access.go

what can I change or improve in your view?

bindings/go/wget/input/method.go
jakobmoellerdev: Ah i guess maybe you missed that in a constructor you can directly add an access, so thats why im saying its not really required to add an input method for this case. You can simply encode the access directly

bindings/go/wget/access/access.go
jakobmoellerdev: I think it was bad in the first iteration already tbh, just didnt occur to me. I think it should move to a similar structure as helm, or is it the same there?

matthiasbruns: In Helm we added the methods to the spec struct instead - which makes more sense. I actually bamboozled myself with `oci<->helm `back and forth comparison and what the status quo should be.

`bindings/go/helm/spec/access/v1/helm.go`
```go
type Helm struct {
	// +ocm:jsonschema-gen:enum=Helm/v1,helm/v1
	// +ocm:jsonschema-gen:enum:deprecated=Helm,helm

	// HelmRepository is the URL of the helm repository to load the chart from.
	HelmRepository string `json:"helmRepository"`

	// HelmChart is the name of the helm chart and its version separated by a colon.
	HelmChart string `json:"helmChart"`

	// Version can either be specified as part of the chart name or separately.
	Version string `json:"version,omitempty"`
}

func (h *Helm) ChartReference() (string, error) {
	repo := h.HelmRepository
	chartName := h.GetChartName()
	version := h.GetVersion()

	if chartName == "" {
		return "", fmt.Errorf("chart name is required to construct chart reference")
	}

	u, err := url.Parse(repo)
	if err != nil {
		return "", fmt.Errorf("error parsing helm repository URL: %w", err)
	}
	u.Path, err = url.JoinPath(u.Path, chartName)
	if err != nil {
		return "", fmt.Errorf("error joining chart name to helm repository URL: %w", err)
	}
	ref := u.String()
	if version != "" {
		ref += ":" + version
	}
	return ref, nil
}
```

jakobmoellerdev: also favor that tbh

matthiasbruns: inlined it - didn't make sense to reuse the identity handling anywhere

bindings/go/wget/input/method.go
matthiasbruns: docs updated

bindings/go/wget/input/method.go
matthiasbruns: Just checked `helm`, where we do the same thing for the `input` path

https://github.com/open-component-model/open-component-model/blob/main/bindings/go/helm/input/method.go#L39

bindings/go/wget/input/method.go
matthiasbruns: done

bindings/go/wget/internal/download/download.go
matthiasbruns: Do you mean the TLS things? They never landed in http afaik - I cannot find what you meant

bindings/go/wget/repository/resource_repository.go
```diff
@@ -1,28 +1,27 @@
...
+	"ocm.software/open-component-model/bindings/go/wget/internal/download"
 	accessspec "ocm.software/open-component-model/bindings/go/wget/spec/access"
 	"ocm.software/open-component-model/bindings/go/wget/spec/access/v1"
-	credv1 "ocm.software/open-component-model/bindings/go/wget/spec/credentials/v1"
 )
 
-var wgetAccess = access.NewWgetAccess()
+const (
+	// hashAlgorithmSHA256 is the hash algorithm used for wget resource digests.
+	hashAlgorithmSHA256 = "SHA-256"
```
frewilhelm, line 21: why do we hardcode on `SHA-256`? what about other hash algorithms?

bindings/go/wget/input/method.go
```diff
@@ -0,0 +1,106 @@
...
+	if err := i.GetInputMethodScheme().Convert(resource.Input, &wget); err != nil {
+		return nil, fmt.Errorf("error converting resource input spec: %w", err)
+	}
+
+	if wget.URL == "" {
+		return nil, fmt.Errorf("url is required")
+	}
+
+	// validate URL
+	if _, err := looseref.ParseReference(wget.URL); err != nil {
```
frewilhelm: ParseReference was made for OCI references. Is this safe to use on URLs that potentially contain queries or the like?

bindings/go/wget/doc.go
frewilhelm: I think this path does not exist anymore

frewilhelm (commented):
overall lgtm. only some questions

bindings/go/wget/input/method.go
matthiasbruns: every time I want to use urls, I am being sent to looseref  - I would actually use `url` here too

jakobmoellerdev: looseref is only for OCI / CTF links, not for generic urls!

bindings/go/wget/repository/resource_repository.go
matthiasbruns: we are doing the same in helm and something similar in oci - I think we should improve the overall pattern, but for now I would keep them aligned

bindings/go/wget/input/method.go
matthiasbruns: migrated to `url`

bindings/go/wget/doc.go
matthiasbruns: fixed

bindings/go/wget/internal/download/download.go
jakobmoellerdev: im saying that TLS was standardized: https://ocm.software/docs/how-to/configure-http/tls-and-custom-ca/
youre right its not directly in the module but should still work like that right?

matthiasbruns: Ok, but this is mTLS here - as your previous PR introduced it - I am going to check again, but custom ca shouldn't be of concern here

frewilhelm (commented):
The CI skips the integration tests. Do we need an extra `Taskfile.yaml` in wget/integration as well as an adjustment in the roots `Taskfil.yaml`?

bindings/go/wget/internal/download/download.go
```diff
@@ -0,0 +1,247 @@
...
+		}
+
+		// Clone the client and install an mTLS transport, preserving the
+		// original transport's settings (proxy, timeouts, connection pooling).
+		existing := *client
+		var baseTransport *http.Transport
+		if t, ok := existing.Transport.(*http.Transport); ok && t != nil {
+			baseTransport = t.Clone()
+		} else {
+			baseTransport = &http.Transport{}
```
piotrjanik: -> `http.DefaultTransport.(*http.Transport).Clone()`
`&http.Transport{}` would be bare transport without default Proxy logic

piotrjanik:
LGTM, one comment left

bindings/go/wget/internal/download/download.go
matthiasbruns: I have now changed it to

```go
var baseTransport *http.Transport
if t, ok := existing.Transport.(*http.Transport); ok && t != nil {
	baseTransport = t.Clone()
} else {
	if t, ok = http.DefaultTransport.(*http.Transport); ok && t != nil {
		baseTransport = t.Clone()
	} else {
		baseTransport = &http.Transport{}
	}
}
```

matthiasbruns:
> Member

Fixed: https://github.com/open-component-model/open-component-model/actions/runs/29082823398/job/86329506061?pr=3062

frewilhelm (approved):

piotrjanik (approved):
LGTM

bindings/go/wget/internal/download/download.go
matthiasbruns: https://github.com/matthiasbruns/open-component-model/pull/116/changes#diff-e39ab83ba5715cc3949ac282daa26244ea43defdeef7876dfc9540a94a7923a8
works fine - but there seems to be another problem with https://ocm.software/docs/how-to/configure-http/tls-and-custom-ca/

// Note darwin (and windows, ios) are not in that list.                                                                                                                                    
`/src/crypto/x509/root_unix.go`
```go
//go:build aix || dragonfly || freebsd || (js && wasm) || linux || netbsd || openbsd || solaris || wasip1                                                                               
...                                                                                                                                                                                     
certFileEnv = "SSL_CERT_FILE"   // read at line 36: os.Getenv(certFileEnv)                                                                                                              
certDirEnv  = "SSL_CERT_DIR"    // read at line 53                                                                                                                                      
```

`/src/crypto/x509/root_darwin.go`
In that file, there is no `SSL_CERT_FILE` or `SSL_CERT_DIR`     

So our custom ca support only works for linux and the others from the list above.
This came up during the integration tests on my machine.

We said, we wont invest more into this unless its needed. I will add this comment to https://github.com/open-component-model/ocm-project/issues/976 to not forget this.

PR #3068, opened 2026-07-08
docs(ADR): ocm-k8s-toolkit discovery CR

### LINKED ISSUE #1221: K8S Discovery CRD and Controller
**Description**

Design and prototype a new **discovery controller** (CRD + reconciler) for the OCM Kubernetes toolkit that resolves an OCM resource *across multiple component versions*, driven by the OpenControlPlane (OCP) service-provider use case.

Today the existing `Component` and `Resource` controllers bind to exactly one component version and one resource. OCP service providers (e.g. flux at `v2.2.0` and `v2.3.0`) instead need the access information (helm chart / image reference + pull secrets) for a *range* of versions from a single umbrella "releasechannel" component, so that a `ProviderConfig` can reference a discovery object rather than hardcoding URLs per version.

References:
- OCP ADR "discoverable-components": https://github.com/openmcp-project/docs/blob/main/adrs/2026-06-08-discoverable-components.md
- Discussion / agreed direction: https://github.com/open-component-model/ocm-project/issues/1154#issuecomment-4917294703
- Existing controllers: `kubernetes/controller` (`Component`, `Resource`, `Repository`)

Rough shape agreed in the discussion (subject to API design):
```yaml
spec:
  umbrellaComponent: releasechannel
  component: github.tools.sap/openmcp/releasechannel/flux
  componentSemver: "v2.2.0<v2.3.0"
  filter:
    byName: [chart, image]
status:
  versions:
    - version: v2.2.0
      resources:
        - name: chart
          imageRef: chart-url:2.2.0
          pullSecret: <ref/value>
```

Scope of this issue:
1. Design the CRD in collaboration with the OCP team (n:m component↔discovery relation, filter/selector semantics, per-version resource list, pull-secret propagation, optional `additionalFields` per resource).
2. Implement a controller prototype that traverses the umbrella component and its references and populates the status.
3. Verify the prototype by running the actual OCP flux service-provider scenario end-to-end (`ProviderConfig` selecting the discovery result).

**Potential problems**
- **Performance**: discovery must traverse the whole component reference tree across a semver range. Descriptors can be large (gardenlinux example: ~37k YAML lines / ~884KB compact JSON). Must avoid re-traversing on every reconcile — needs caching / conditional resolution keyed on resolved versions + interval, similar to how `Component` gates on `ObservedGeneration`.
- **Behaviour duplication**: resolution logic overlaps (partially) with the existing `Component` and `Resource` controllers (repository resolution, OCM config/credential handling, verification, status propagation). Decide whether to reuse the existing `internal/ocm` / `internal/resolution` machinery or fork it, to avoid divergence.

**Open questions**
- Kubernetes layer vs. direct OCM consumption — the ADR leans toward direct OCM consumption (Approach A); is a controller/CRD the right long-term surface, or only a fallback? (relevant also for ODG's synchronous-resolution UX concern).
- Output format: expose the full component descriptor, or a curated per-version resource list in status? (status size limits vs. usability).
- Pull secrets: mirror through as-is, or add a flag to convert `ocm.config` credentials into referenced `dockerconfigjson` secrets (raised by OCP)?
- Behaviour when a requested `filter.byName` / `additionalField` is absent in some versions/resources — fail, skip, or partial status?
- Do we materialize child `Resource` objects from the discovery result, or keep everything in one status? (affects credential/registry-per-resource handling).
- Dependency on upstream OCM features (umbrella-component discovery, local-YAML fallback) — availability and timeline.

**Done Criteria**
- [ ] CRD/API designed and agreed with the OCP team
- [ ] Controller prototype traverses the umbrella component tree and populates per-version status
- [ ] Prototype verified against the real OCP flux service-provider scenario

frewilhelm:

OCP and ODG voiced interest in having a discovery CR publishing informational content about OCM related objects. This ADR lists the requirements, available options, and chosen result.

Fixes https://github.com/open-component-model/ocm-project/issues/1221


docs/adr/0026_ocm_k8s_toolkit_discover.md
```diff
@@ -0,0 +1,340 @@
...
+
+- `referenceSelector` filters **references** in the graph. A reference is a `component.references[]` entry: it carries
+  its own identity (local `name`, target `componentName`, `version`, extra identity) and its own labels.
+- `componentSelector` filters **components**: the descriptors themselves, by their own identity and labels.
+- `resourceSelector` filters **resources within a surviving component**. Applied in place; empty selector keeps all.
+
+Traversal semantics under filters: Filters are applied as post-order predicates over the full reachable graph,
+not as per-edge pruning during traversal. Per-edge pruning would drop deep matches whose ancestor edges happen not
+to match. This trades a broader graph walk for correctness under nested reference structures. The graph walk is
+concurrent and unbounded today; a concurrency limit might be a follow-up.
```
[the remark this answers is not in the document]
frewilhelm: I understand the concern but this adr is not about improving the graph traversal. there are already some todo's and follow-ups in place

docs/adr/0026_ocm_k8s_toolkit_discover.md
```diff
@@ -0,0 +1,340 @@
...
+  [`discover_options.go`](../../bindings/go/dag/sync/discover_options.go) referencing
+  [ocm-project#705](https://github.com/open-component-model/ocm-project/issues/705)), then threaded through
+  Discovery's caller.
+- `cel.CostLimit` on `Selector.Expression` and `Extract.Expression`: Discovery has no `Interval` and no default
+  per-reconcile deadline. A pathological CEL program (e.g. a regex-`matches` on user-controlled input, or a nested
+  `.filter()` over a large graph) currently has no static bound on evaluation cost. `cel-go` exposes
+  `cel.CostLimit(n)`, the same mechanism `x-kubernetes-validations` uses, which counts AST-cost units at
+  evaluation time and aborts on overrun. Adding it makes worst-case eval time bounded and surfaces the error as
+  `Ready=False` with reason `ExtractFailed` (and, when non-retriable, also `Stalled=True`) instead of runaway CPU.
+  Needs benchmark input to pick sensible defaults per binding scope.
```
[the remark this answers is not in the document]
frewilhelm: This is already a follow up if i am not mistaken

docs/adr/0026_ocm_k8s_toolkit_discover.md
```diff
@@ -0,0 +1,340 @@
...
+  # ... shared spec fields (ocmConfig, suspend) omitted; see other CRDs.
+
+  # Required. Same-namespace reference to a Component CR.
+  componentRef:
+    name: releasechannel
+
+  # Optional. Filters references. A component is kept if it is the target of
+  # at least one matching reference somewhere below the root. The root itself
+  # is not a reference target, so it is dropped when this selector is set.
+  referenceSelector:
```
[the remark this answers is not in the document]
frewilhelm: I adjusted the wording

docs/adr/0025_ocm_k8s_toolkit_discover.md
```diff
@@ -0,0 +1,343 @@
+# Component Graph Discovery in the OCM Kubernetes Controller Toolkit
+
+* **Status**: proposed
+* **Deciders**: @frewilhelm @fabianburth
```
fabianburth:
```suggestion
* **Deciders**: OCM Maintainer Team
```

docs/adr/0026_ocm_k8s_toolkit_discover.md
```diff
@@ -0,0 +1,343 @@
...
+* **Deciders**: @frewilhelm @fabianburth
+* **Date**: 2026-07-28
+
+Technical Story:
+
+- EPIC: [ocm-project#1153](https://github.com/open-component-model/ocm-project/issues/1153)
+- Spike: [ocm-project#1154](https://github.com/open-component-model/ocm-project/issues/1154)
+- POC branch: [`feat-discovery`](https://github.com/frewilhelm/open-component-model/tree/feat-discovery)
+- Related upstream work: [PR #2833 (`feat(oci): blob and resolution cache`)](https://github.com/open-component-model/open-component-model/pull/2833),
+  [ocm-project#296 (`ResourceID.BySelector`)](https://github.com/open-component-model/ocm-project/issues/296)
```
fabianburth: Please replace with a 

> brief description of the technical context and need for the decision

You may still share the links in the context and problem statement below

docs/adr/0025_ocm_k8s_toolkit_discover.md
```diff
@@ -0,0 +1,343 @@
...
+
+- EPIC: [ocm-project#1153](https://github.com/open-component-model/ocm-project/issues/1153)
+- Spike: [ocm-project#1154](https://github.com/open-component-model/ocm-project/issues/1154)
+- POC branch: [`feat-discovery`](https://github.com/frewilhelm/open-component-model/tree/feat-discovery)
+- Related upstream work: [PR #2833 (`feat(oci): blob and resolution cache`)](https://github.com/open-component-model/open-component-model/pull/2833),
+  [ocm-project#296 (`ResourceID.BySelector`)](https://github.com/open-component-model/ocm-project/issues/296)
+
+## Context and Problem Statement
+
+The OCM Kubernetes Controller Toolkit currently exposes an address-based consumption model: a user creates a
```
fabianburth:
```suggestion
The OCM Kubernetes Controller Toolkit currently exposes an identity-based consumption model: a user creates a
```

docs/adr/0025_ocm_k8s_toolkit_discover.md
```diff
@@ -0,0 +1,343 @@
...
+`Repository`, then a `Component` pinned to a single semver constraint, then one `Resource` per artifact they want to
+consume. Every level of that chain requires the caller to know the exact identity of what they are pulling:
+repository spec, component name, resource identity, and (for nested references) the full `referencePath`.
+
+Two external consumer projects asked for a query-based mode on top of that address-based baseline:
```
fabianburth:
```suggestion
Two external consumer projects asked for a query-based mode on top of that identity-based baseline:
```

docs/adr/0025_ocm_k8s_toolkit_discover.md
```diff
@@ -0,0 +1,343 @@
...
+
+## Decision Drivers
+
+- Familiar shapes where possible: `matchLabels`-shaped label filtering and CEL for structured predicates:
+  the same shapes users see in other CRDs. OCM-specific identity attributes get their own field (`matchIdentity`).
+- Kubernetes etcd payload cap: Objects have a soft limit of [~1.5 MiB](https://etcd.io/docs/v3.6/dev-guide/limit/)
+  per resource; component descriptors can approach or exceed it on their own (e.g.
+  `europe-docker.pkg.dev/gardener-project/releases//github.com/gardenlinux/gardenlinux:2150.6.0` is ~884 KB compact
+  JSON).
+- Additive to the existing address-based CRDs: `Repository`, `Component`, and `Resource` are already consumed
```
fabianburth:
```suggestion
- Additive to the existing identity-based CRDs: `Repository`, `Component`, and `Resource` are already consumed
```

docs/adr/0026_ocm_k8s_toolkit_discover.md
```diff
@@ -0,0 +1,343 @@
...
+
+Dedicated `Discovery` CRD (chosen): See [Decision Outcome](#decision-outcome). Earlier CRD sketches from the
+spike (`ResourceRange`, `ResourceSet`, `ComponentDescriptor`) are on
+[ocm-project#1153](https://github.com/open-component-model/ocm-project/issues/1153); each covers one query shape
+and needed siblings to cover the others. The chosen CRD subsumes all three by covering all three filtering
+steps in a single kind.
+
+## Decision Outcome
+
+Chosen option: **a dedicated `Discovery` CRD** with three selectors and a CEL projection. Targets
```
fabianburth: > CEL projection

what do you mean by projection?

docs/adr/0026_ocm_k8s_toolkit_discover.md
```diff
@@ -0,0 +1,343 @@
...
+
+  # Optional. Filters references. A component is kept if it is the target of
+  # at least one matching reference somewhere below the root. The root has no
+  # incoming reference and is therefore dropped when this selector carries at
+  # least one predicate; an empty `{}` selector is treated the same as an
+  # omitted field and preserves the full graph including the root.
+  referenceSelector:
+    matchIdentity:
+      componentName: ghcr.io/example/releasechannel/flux
+    expression: semverCheck(identity.version, ">=2.7.0, <2.10.0")
```
fabianburth, line 97: Is this a `cel` expression? Do we already expose `semverCheck` as a `cel` function? What's the input to the `cel` environment here?

docs/adr/0026_ocm_k8s_toolkit_discover.md
```diff
@@ -0,0 +1,343 @@
...
+
+  # Required. Same-namespace reference to a Component CR.
+  componentRef:
+    name: releasechannel
+
+  # Optional. Filters references. A component is kept if it is the target of
+  # at least one matching reference somewhere below the root. The root has no
+  # incoming reference and is therefore dropped when this selector carries at
+  # least one predicate; an empty `{}` selector is treated the same as an
+  # omitted field and preserves the full graph including the root.
```
fabianburth, line 93: Do we stop the traversal on a non-matching reference (so will all components below a non-matching reference be ignored?)

fabianburth: answered below, we do not

docs/adr/0026_ocm_k8s_toolkit_discover.md
```diff
@@ -0,0 +1,343 @@
...
+Traversal semantics under filters: Filters are applied as post-order predicates over the full reachable graph,
+not as per-edge pruning during traversal. Per-edge pruning would drop deep matches whose ancestor edges happen not
+to match. This trades a broader graph walk for correctness under nested reference structures. The graph walk is
+concurrent and unbounded today; a concurrency limit might be a follow-up.
+
+Short-circuit: When a selector fixes a component identity, `name` and `version` set as concrete equality, no
+other constraints active, the controller extracts that target and terminates DAG traversal at the vertex's subtree
+boundary. Component identity `{name, version}` is [globally unique per the OCM
+spec](https://github.com/open-component-model/ocm-spec/blob/main/doc/01-model/02-elements-toplevel.md#component-identity),
+so nothing below the target needs to be walked. Label and expression constraints disable the short-circuit; the
```
fabianburth: > Label and expression constraints disable the short-circuit

But the `name` and `version` are still unique? So why would additional constraints disable the short circuit?

docs/adr/0026_ocm_k8s_toolkit_discover.md
```diff
@@ -0,0 +1,343 @@
...
+  its own identity (local `name`, target `componentName`, `version`, extra identity) and its own labels.
+- `componentSelector` filters **components**: the descriptors themselves, by their own identity and labels.
+- `resourceSelector` filters **resources within a surviving component**. Applied in place; empty selector keeps all.
+
+Traversal semantics under filters: Filters are applied as post-order predicates over the full reachable graph,
+not as per-edge pruning during traversal. Per-edge pruning would drop deep matches whose ancestor edges happen not
+to match. This trades a broader graph walk for correctness under nested reference structures. The graph walk is
+concurrent and unbounded today; a concurrency limit might be a follow-up.
+
+Short-circuit: When a selector fixes a component identity, `name` and `version` set as concrete equality, no
```
fabianburth: Do you have concrete plans on how to implement the short circuit? If we would be willing to drop "reachability guarantee" (so, that you can reach that component from the root), we could just fetch it. But that seems off.

docs/adr/0026_ocm_k8s_toolkit_discover.md
```diff
@@ -0,0 +1,343 @@
...
+- `matchIdentity`: equality on identity attributes (`name`, `version`, and per-element extras like `componentName` on
+  references). First-class map on the CR surface so the controller can read it statically for the short-circuit
+  optimisation.
+- `matchLabels`: string-equality on labels whose value is a string. OCM labels may hold arbitrary JSON; non-string
+  values are not comparable this way and are silently non-matching.
+- **`expression`**: a CEL boolean predicate. A CEL binding is a named value the expression can reference; for
+  selectors the controller binds `identity` to the element's identity map and `labels` to its label map, so users
+  write `identity.version` or `labels["tier"]`. `expression` covers structured labels, cross-field predicates, and
+  semver ranges (`semverCheck(v, c)`). Missing attribute or label references evaluate to `false` rather than raising,
+  so users don't have to guard every access with `has()`.
```
fabianburth: > Missing attribute or label references evaluate to `false` rather than raising,
  so users don't have to guard every access with `has()`.

How are you planning to implement that?

docs/adr/0026_ocm_k8s_toolkit_discover.md
```diff
@@ -0,0 +1,343 @@
...
+
+Adding `matchExpressions []metav1.LabelSelectorRequirement`, the `In` / `NotIn` / `Exists` / `DoesNotExist`
+operators from `metav1.LabelSelector`, was considered because it's a familiar k8s shape and would let admission
+tooling that walks `LabelSelector`-shaped fields work on the string subset without changes. Not adopted, for three
+reasons:
+
+- Two ways to write the same query: `env in [prod, staging]` works in both `matchExpressions` and CEL.
+- `LabelSelector` values are string-only: OCM labels are arbitrary JSON. `matchExpressions` covers only the
+  string subset; structured labels still need `expression`. Adding it doesn't remove any surface, only adds.
+- The one range operator users actually need, semver, is already in CEL as `semverCheck`.
```
fabianburth: How is it already in `cel`?

docs/adr/0026_ocm_k8s_toolkit_discover.md
```diff
@@ -0,0 +1,343 @@
...
+In map modes (`byResources` / `byComponents`), a field expression that hits a missing attribute, field, or map
+key is treated as null and the field is dropped from that entry. This lets a single map span heterogeneous access
+types (e.g. some resources carry `access.imageReference`, others don't) without CEL `?` on every path, matching the
+tolerance selectors have for missing attributes (see [Selector shape](#selector-shape) above). To emit a
+placeholder instead of dropping, use `has(...)` with a conditional, e.g.
+`has(resource.access.imageReference) ? resource.access.imageReference : "n/a"`.
+
+`expression` mode has no per-field container to drop into: the whole payload is a single CEL expression, so an
+unguarded miss surfaces as `ExtractFailed`. Guard with `has(...)` or CEL optional access
+(`foo.?bar.orValue(...)`) if the expression may traverse fields that are not always present.
```
fabianburth: Didn't you just specify above that you just drop misses?

docs/adr/0026_ocm_k8s_toolkit_discover.md
```diff
@@ -0,0 +1,343 @@
...
+
+`Extract` is optional. When absent, the raw v2 descriptor list is emitted, subject to the etcd soft limit.
```
fabianburth, line 251: Even if we filtered resources?

docs/adr/0026_ocm_k8s_toolkit_discover.md
```diff
@@ -0,0 +1,343 @@
...
+
+Status payloads are subject to the etcd soft limit (~1.5 MiB). Discovery does not store descriptors outside etcd.
+Selectors and `Extract` let users shape the payload down. When the result still exceeds the limit, the API server
+rejects the status write; the controller clears `status.discovery`, repatches a `Ready=False` /
+`PayloadTooLarge` condition carrying the API server's rejection message, and returns without requeueing.
+
+The large-descriptor navigation case is descoped. Out-of-status storage (encoded payload,
+sibling ConfigMap, artifact CR) is an additive path if a concrete need appears.
+
+#### Deterministic ordering: `(name, version)` lexicographic
```
fabianburth, line 298: would semver ordering be much more difficult? (it's a nit, but would be more intuitive to users I believe)

docs/adr/0026_ocm_k8s_toolkit_discover.md
```diff
@@ -0,0 +1,343 @@
...
+  descriptor set to empty. In that state `status.discovery` is set to an explicit empty JSON array (`[]`) so
+  consumers can distinguish "query ran and matched nothing" from "payload never computed". Distinct reasons per
+  stage so consumers can distinguish "no reference matched" from "reference matched but the component didn't".
+
+Consumers of `status.discovery` follow `status.effectiveOCMConfig` when they need credentials (pull-secrets, etc.)
+for the discovered resources.
+
+## Caching Strategy
+
+Discovery has no cross-reconcile descriptor cache today; every reconcile refetches every reachable descriptor from
```
fabianburth, line 320: it's a blob cache, right? let's just try to get it in soon!

docs/adr/0026_ocm_k8s_toolkit_discover.md
```diff
@@ -0,0 +1,343 @@
...
+  per-reconcile deadline. A pathological CEL program (e.g. a regex-`matches` on user-controlled input, or a nested
+  `.filter()` over a large graph) currently has no static bound on evaluation cost. `cel-go` exposes
+  `cel.CostLimit(n)`, the same mechanism `x-kubernetes-validations` uses, which counts AST-cost units at
+  evaluation time and aborts on overrun. Adding it makes worst-case eval time bounded and surfaces the error as
+  `Ready=False` with reason `ExtractFailed` (and, when non-retriable, also `Stalled=True`) instead of runaway CPU.
+  Needs benchmark input to pick sensible defaults per binding scope.
+- Unify `Selector` with `ResourceID.BySelector`: Once
+  [ocm-project#296](https://github.com/open-component-model/ocm-project/issues/296) lands, verify the same
+  `Selector` shape serves both consumers without further extension. Any divergence should be resolved on the
+  shared type.
```
fabianburth, line 364: Im wondering whether that is even still a valid feature after implementing this API. Generally, thinking ahead, would it make sense to unify those apis?

docs/adr/0026_ocm_k8s_toolkit_discover.md
frewilhelm: Not yet. I added one in my PoC, see [here](https://github.com/frewilhelm/open-component-model/blob/81854c3e627bbcc4fef045fb974e9ad833bbb41a/kubernetes/controller/internal/cel/functions/semver.go).

docs/adr/0026_ocm_k8s_toolkit_discover.md
frewilhelm: In a first iteration I added a k8s-like `matchExpressions` field with an additional `semverCheck` operation.

This section is about adding such a `matchExpressions` field to the spec. However, I rejected the idea because CEL is more flexibel and I can add a `semverCheck` CEL-function. Thus, I would have two fields `matchExpressions` and the CEL `expressions` which would be able to do the same thing.

What I wanted to say with that - I don't need that field `matchExpressions` because I can do everything else and more in CEL. But I will take another look on how to simplify this!

docs/adr/0026_ocm_k8s_toolkit_discover.md
```diff
@@ -0,0 +1,343 @@
...
+#### `Extract` and its three modes
+
+`Extract` is the projection stage: it runs *after* filtering, on the already-narrowed descriptor list, and produces
+the final payload landing in `status.discovery`. Three mutually exclusive modes cover the three join shapes a graph
+query produces, enforced by CRD-level validation:
+
+- `byResources: {field: expr}`: flat map, evaluated once per surviving resource of each surviving component.
+  Bindings: `resource`, `component`. Emits `[]object`, one entry per `(component, resource)` pair.
+- `byComponents: {field: expr}`: same shape, evaluated once per surviving component. Binding: `component`.
+  Emits `[]object`.
```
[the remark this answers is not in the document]
frewilhelm: I believe this is part of a follow up

docs/adr/0026_ocm_k8s_toolkit_discover.md
frewilhelm: I changed the wording and clarified it

docs/adr/0026_ocm_k8s_toolkit_discover.md
frewilhelm: I believe I had such a scenario in mind:
Suppose we look for a component version using the reference selector with `componentName=C4`, `version=1.0.0`, and `label=key=value`.

`C4`, however, is referenced by `C1 -> C2 -> C4` and `C1 -> C3 -> C4`. But only `C3` has the `label=key=value`. Now, the graph traversal begins and we first reach `C4` through `C2` and we would match and short-circuit because `name=C4` and `version=1.0.0` matches. Then, we would not have `C3` in our collection for our post-filtering and we would neglect `C4` because `C2` does not contain the `label=key=value`.

Does this make sense? :D

Nevertheless, I need to clarify this in the ADR a bit.

docs/adr/0026_ocm_k8s_toolkit_discover.md
frewilhelm: I thought about the graph `Discovery` like

```
func (r *resolverAndDiscoverer) Discover(ctx context.Context, parent *descruntime.Descriptor) ([]string, error) {
	logger := log.FromContext(ctx)

	if r.shortCircuitKey != "" {
		parentKey := parent.Component.ToIdentity().String()
		if parentKey == r.shortCircuitKey {
			logger.V(1).Info("short-circuit: target found, stopping traversal", "target", parentKey)
			return nil, errShortCircuit
		}
	}
...
```

So, currently it would just check if the identity matches the single identity (name + version) from the `shortCircuitKey`.

docs/adr/0026_ocm_k8s_toolkit_discover.md
frewilhelm: A "bit" "hacky": 

```go
// IsMissingAttributeErr recognises CEL runtime errors caused by referencing a
// map key / attribute / field that the activation doesn't carry. cel-go
// surfaces these as unwrapped *types.Err values distinguishable only by their
// message. Callers that want "missing evaluates to no-value rather than
// raising" semantics use this to convert the error into a null / non-match.
func IsMissingAttributeErr(err error) bool {
	if err == nil {
		return false
	}
	msg := err.Error()
	return strings.Contains(msg, "no such key") ||
		strings.Contains(msg, "no such attribute") ||
		strings.Contains(msg, "no such field")
}
```

WDYT?

docs/adr/0026_ocm_k8s_toolkit_discover.md
frewilhelm: I don't want to opinionate on that but rather give the user the possibility to specify `Extract` and chose what they want. 

What are you thinking about? Something like, if we filter for resources, we should just return the resource-spec for these filtered resources?

I think I would argue against that because there might be a use-case where users want to have component-descriptors that contain certain resources.

docs/adr/0026_ocm_k8s_toolkit_discover.md
frewilhelm: what do you mean by "semver ordering"?

docs/adr/0026_ocm_k8s_toolkit_discover.md
frewilhelm: yes

docs/adr/0026_ocm_k8s_toolkit_discover.md
frewilhelm: I don't think it makes sense to unify the `Resource` and `Discovery` API because of the separation of concerns. `Discovery` is for discovering stuff through component descriptors. `Resource` is for getting metadata for one specific resource. Additionally, for `Resource` we verify (if possible) the integrity of the component version + component reference as well as the integrity of the resource itself through its digest processor. Having those integrity-checks in the `Discovery` would impact performance significantly, e.g. running the digest processor for every (filtered) resource.

docs/adr/0026_ocm_k8s_toolkit_discover.md
```diff
@@ -0,0 +1,343 @@
...
+for the discovered resources.
+
+## Caching Strategy
+
+Discovery has no cross-reconcile descriptor cache today; every reconcile refetches every reachable descriptor from
+the upstream repository. The plan is to rely on the component descriptor cache landing upstream in
+[PR #2833](https://github.com/open-component-model/open-component-model/pull/2833), which caches descriptor bytes
+on disk across all consumers of the OCI provider. The existing resolution service is not wired in, its removal is
+contingent on the upstream cache landing; if the upstream cache is delayed, the resolution service remains available
+as a fallback.
```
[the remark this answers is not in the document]
frewilhelm: Caching will be a follow up. I don't think we need to detail out how it will work in the ADR as this ADR is about the API contract first

docs/adr/0026_ocm_k8s_toolkit_discover.md
frewilhelm: The section above is about the filtering with `Selector.expression`. There it makes sense to me to not require the user to specify `has()` for every resource.spec field (for example) as the fields differ between resource types.

For the `Extract.expression`, however, I was thinking about informing the user if an attribute is missing, so results do not land in the void without the users noticing it. But I am unsure if thats the right decision.

docs/adr/0026_ocm_k8s_toolkit_discover.md
```diff
@@ -0,0 +1,371 @@
...
+
+Adding `matchExpressions []metav1.LabelSelectorRequirement`, the `In` / `NotIn` / `Exists` / `DoesNotExist`
+operators from `metav1.LabelSelector`, was considered because it's a familiar k8s shape and would let admission
+tooling that walks `LabelSelector`-shaped fields work on the string subset without changes. Not adopted, for three
+reasons:
+
+- Two ways to write the same query: `env in [prod, staging]` works in both `matchExpressions` and CEL.
+- `LabelSelector` values are string-only: OCM labels are arbitrary JSON. `matchExpressions` covers only the
+  string subset; structured labels still need `expression`. Adding it doesn't remove any surface, only adds.
+- The one range operator users actually need, semver, will already be available as CEL function `semverCheck`.
```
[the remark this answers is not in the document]
frewilhelm: The semverCheck implementation does not need its own section in this adr.

docs/adr/0026_ocm_k8s_toolkit_discover.md
```diff
@@ -0,0 +1,371 @@
...
+  anyway.
+
+- **`referenceSelector`**: `matchLabels` and `expression` must be empty. Reference labels live on edges, and the
+  same target can be reached via multiple parents. The early-exit signal propagates through the discoverer's
+  `errgroup` and cancels sibling goroutines, some of which may still be resolving alternative parents of the
+  target. Those parents' descriptors then never enter `filterByReferenceSelector`, which iterates each surviving
+  descriptor's `Component.References` to decide which targets match. Cancelling a parent hides its reference
+  edge to the target from the post-filter, so a `matchLabels`/`expression` clause that would have matched on
+  the cancelled edge produces a false negative. Requiring empty label/expression clauses keeps short-circuit to
+  cases where the post-filter is a no-op (any surviving edge to the target is a match).
```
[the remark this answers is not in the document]
frewilhelm: I believe this was tackled right? https://github.com/open-component-model/open-component-model/pull/3068/changes/7bd47dd7c2fd335b977b9c5efcafcee9514f8ef0

docs/adr/0026_ocm_k8s_toolkit_discover.md
```diff
@@ -0,0 +1,378 @@
...
+- Open Delivery Gear (ODG): The [Open Delivery Gear](https://github.com/open-component-model/odg-core) project, needs to
+  resolve component descriptors on demand, enumerate the versions available for a component, filter by publication date,
+  and download resources for external scanners (SBOM, vulnerability tooling). ODG's pipelines are synchronous and the
+  CR-reconcile round-trip is a blocker. That is why ODGs requirements are out of scope for this ADR.
+  Details in [ocm-project#1154](https://github.com/open-component-model/ocm-project/issues/1154).
+
+Neither workflow is expressible today. A caller who does not know the exact
+`(component, version, resource, referencePath)` tuple up front has no way to ask the controller
+"give me all versions of resource R across umbrella C's references" or
+"publish this component's descriptor so I can navigate it".
```
Skarlso: So I would reframe this a tiny bit from the identity based thing. Because the deployment will always be identity based. The end result will always need to be specific. But how you GET to that can be a dynamic way of discovery. 🤔 The first segment talks about this like the deployment will no longer require a specific identity, which is not quite true, right?

n3rdc4ptn (approved):
lgtu from OpenControlPlane

docs/adr/0026_ocm_k8s_toolkit_discover.md
```diff
@@ -0,0 +1,378 @@
...
+    matchIdentity:
+      componentName: ghcr.io/example/releasechannel/flux
+    expression: semverCheck(identity.version, ">=2.7.0, <2.10.0")
+
+  # Optional. Filters components by their own identity and labels. Applied to
+  # whatever components survived referenceSelector; when referenceSelector is
+  # unset (or empty), that's the full graph including the root.
+  componentSelector:
+    matchLabels:
+      tier: platform
```
Skarlso, line 104: I can't do OR or AND here, right? What decides how these are combined?

docs/adr/0026_ocm_k8s_toolkit_discover.md
```diff
@@ -0,0 +1,378 @@
...
+
+  # Optional. Projects the discovered content. Exactly one of byResources,
+  # byComponents, or expression may be set (enforced by CRD validation).
+  extract:
+    byResources:
+      imageRef:        resource.access.imageReference
+      resourceName:    resource.name
+      resourceVersion: resource.version
+      name:            component.name
+      version:         component.version
```
Skarlso, line 119: this looks like it should be able to select multiple but supports only a single object? 

Also name is `component.name` i assume you means `resource.name`?

docs/adr/0026_ocm_k8s_toolkit_discover.md
```diff
@@ -0,0 +1,378 @@
...
+
+- `referenceSelector` filters **references** in the graph. A reference is a `component.references[]` entry: it carries
+  its own identity (local `name`, target `componentName`, `version`, extra identity) and its own labels.
+- `componentSelector` filters **components**: the descriptors themselves, by their own identity and labels.
+- `resourceSelector` filters **resources within a surviving component**. Applied in place; empty selector keeps all.
+
+Traversal semantics under filters: Filters are applied as post-order predicates over the full reachable graph,
+not as per-edge pruning during traversal. Per-edge pruning would drop deep matches whose ancestor edges happen not
+to match. This trades a broader graph walk for correctness under nested reference structures. The graph walk is
+concurrent and unbounded today; a concurrency limit might be a follow-up.
```
Skarlso, line 158: In what way is this information relevant? 🤔 I mean, of course you don't want to delete items before you check them. I feel like this is an implementation details that is irrelevant from the ADR's perspective. WDYT?

docs/adr/0026_ocm_k8s_toolkit_discover.md
```diff
@@ -0,0 +1,378 @@
...
+
+Short-circuit: When a selector fixes a component identity `name` and `version` set as concrete equality on
+`matchIdentity` the controller marks that identity as a short-circuit target. Once the DAG traversal resolves it,
+the target's `Discover` call returns an early-exit signal that propagates through the discoverer's `errgroup` and
+stops the walk. Component identity `{name, version}` is [globally unique per the OCM
+spec](https://github.com/open-component-model/ocm-spec/blob/main/doc/01-model/02-elements-toplevel.md#component-identity),
+so at most one vertex can match and nothing else needs to be walked.
```
Skarlso, line 165: I do not understand any of this. :D Also, again, how or why is this relevant information? :D

Why is it `fix`-ing anything? How is that a fix? :D

docs/adr/0026_ocm_k8s_toolkit_discover.md
```diff
@@ -0,0 +1,378 @@
...
+  exactly `componentName` and `version` (no other keys). Reference labels and per-edge identity attributes
+  (local `name`, extra identity) live on edges, and the same target can be reached via multiple parents. The
+  early-exit signal propagates through the discoverer's `errgroup` and cancels sibling goroutines, some of
+  which may still be resolving alternative parents of the target. Those parents' descriptors then never enter
+  `filterByReferenceSelector`, which iterates each surviving descriptor's `Component.References` to decide
+  which targets match. Cancelling a parent hides its reference edge to the target from the post-filter, so a
+  `matchLabels` / `expression` clause, or a `matchIdentity` clause on a per-edge attribute, that would have
+  matched on the cancelled edge produces a false negative. Restricting the short-circuit trigger to a
+  `matchIdentity` of exactly `{componentName, version}` with no label/expression clauses keeps it to cases
+  where the post-filter is a no-op (any surviving edge to the target is a match).
```
Skarlso, line 186: Jesus :D

docs/adr/0026_ocm_k8s_toolkit_discover.md
```diff
@@ -0,0 +1,378 @@
...
+#### Selector shape
+
+Selectors have three fields, all ANDed. An empty selector matches everything.
+
+- `matchIdentity`: equality on identity attributes (`name`, `version`, and per-element extras like `componentName` on
+  references). First-class map on the CR surface so the controller can read it statically for the short-circuit
+  optimization.
+- `matchLabels`: string-equality on labels whose value is a string. OCM labels may hold arbitrary JSON; non-string
+  values are not comparable this way and are silently non-matching.
+- **`expression`**: a CEL boolean predicate. A CEL binding is a named value the expression can reference; for
```
Skarlso, line 201: Why is this bolded and the others aren't?

docs/adr/0026_ocm_k8s_toolkit_discover.md
```diff
@@ -0,0 +1,378 @@
...
+  selectors the controller binds `identity` to the element's identity map and `labels` to its label map, so users
+  write `identity.version` or `labels["tier"]`. `expression` covers structured labels, cross-field predicates, and
```
Skarlso, line 203: What's a cross-field predicate? 🤔

docs/adr/0026_ocm_k8s_toolkit_discover.md
```diff
@@ -0,0 +1,378 @@
...
+  semver ranges (`semverCheck(v, c)`). Missing attribute or label references evaluate to `false` rather than raising,
+  so users don't have to guard every access with `has()`. Without this behaviour, mismatches would return an error
```
Skarlso, line 205: This is actually dangerous because you think that you correctly defined something and then the selection is just not coming up with something as opposed to failing with missing field. And you would spend some minutes trying to figure out why it doesn't show up in the search.

Or worse, you think it shouldn't so up, but it didn't show up for the wrong reason and you silently accept this as a success.

Skarlso (commented):
I'm not completely done yet. Will continue later, but have to jump now. :D Didn't want to leave you off without anything. :)

docs/adr/0026_ocm_k8s_toolkit_discover.md
frewilhelm: I think, I understand where you are coming from. Addressed in 32a7ba10406a962a8a2d906d237c6c40030f26e0

docs/adr/0026_ocm_k8s_toolkit_discover.md
frewilhelm: `byResources` will iterate over every (filtered) component descriptor and its resources and output a list with an entry for every resource. You can then specify which fields you want to see in the output. `component.name` could be one of them, so you can still connect the resources in the output to the cv they are coming from.

docs/adr/0026_ocm_k8s_toolkit_discover.md
frewilhelm: So, when filtering for component versions, we potentially need to check the whole graph for matching descriptors. There are components with nested and nested and ... references. So, in case the filter gives us a unique target (component name + version), we can stop the graph traversal because, we know we found our one and only component version.

docs/adr/0027_ocm_k8s_toolkit_discover.md
```diff
@@ -0,0 +1,378 @@
...
+    matchIdentity:
+      componentName: ghcr.io/example/releasechannel/flux
+    expression: semverCheck(identity.version, ">=2.7.0, <2.10.0")
+
+  # Optional. Filters components by their own identity and labels. Applied to
+  # whatever components survived referenceSelector; when referenceSelector is
+  # unset (or empty), that's the full graph including the root.
+  componentSelector:
+    matchLabels:
+      tier: platform
```
frewilhelm: All the labels would be ANDs. However, you can specify ORs and a lot of more stuff using the `expression`

docs/adr/0027_ocm_k8s_toolkit_discover.md
```diff
@@ -0,0 +1,378 @@
...
+
+- `referenceSelector` filters **references** in the graph. A reference is a `component.references[]` entry: it carries
+  its own identity (local `name`, target `componentName`, `version`, extra identity) and its own labels.
+- `componentSelector` filters **components**: the descriptors themselves, by their own identity and labels.
+- `resourceSelector` filters **resources within a surviving component**. Applied in place; empty selector keeps all.
+
+Traversal semantics under filters: Filters are applied as post-order predicates over the full reachable graph,
+not as per-edge pruning during traversal. Per-edge pruning would drop deep matches whose ancestor edges happen not
+to match. This trades a broader graph walk for correctness under nested reference structures. The graph walk is
+concurrent and unbounded today; a concurrency limit might be a follow-up.
```
frewilhelm: I believe the discovery has implications on the performance especially regarding x-nested components. However, you are right in that sense that I wanted to focus on the API contract and not internal implementations. But IMO it does not hurt and those questions were asked before, so I would let it in.

docs/adr/0026_ocm_k8s_toolkit_discover.md
frewilhelm: Sometimes I am at war with my AI

docs/adr/0026_ocm_k8s_toolkit_discover.md
frewilhelm: a predicate from different fields. I clarified it in 028fa1d25e78f43193daf63b6c9b2c192dcf9004

docs/adr/0026_ocm_k8s_toolkit_discover.md
frewilhelm: I believe I understand where you are coming from. But this would mean that users need to guard every expression (that might not be present) with `has()`. `has()` would then result in the same behaviour. As we have different field names depending on types, I think users would use `has()` pretty often and its not the best UX.

But honestly, I can understand why we maybe want users to explicitly use `has()` because doing it automatically, kind of feels like a bit of magic. 

I don't know yet...

Skarlso: I'm okay with this right now, maybe this is even an implementation nit. 🤔 Maybe we could output some kind of warning and call it a day. I'm not going to block the ADR on this.

Skarlso (dismissed):
LGTM

jakobmoellerdev:
We will have a check on requirements on Maestro here as well since we need it for discovery purposes too

docs/adr/0026_ocm_k8s_toolkit_discover.md
```diff
@@ -0,0 +1,379 @@
...
+
+- No selectors, no extract: JSON *array* of surviving descriptors (always an array, even when there is exactly
+  one, consumers can iterate uniformly).
+- `referenceSelector`: array of target descriptors (root excluded; it has no incoming reference).
+- `componentSelector`: array of matched descriptors; runs after `referenceSelector` (if set), so it operates on
+  whatever survived that stage.
+- `resourceSelector`: array of descriptors with `.component.resources` filtered in place.
+- `extract.byResources`: flat array of per-resource records.
+- `extract.byComponents`: flat array of per-component records.
+- `extract.expression`: verbatim return value of the CEL expression (any JSON type).
```
ntnn: I don't think the status being schemaless is a good idea long-term. 

If size scaling is the concern I'd rather follow the way kube scales potentially large objects - e.g. Service->EndpointSlice. Endpoint was hitting the etcd cap so EndpointSlice was added as a scalable typed object.
`Discovery` -> `DiscoverySlice` or `DiscoveredResourceSlice`

For the extract shape I'd rather split the output into different CRs. When a consumer defines extract mode X they can also expect the CR for mode X - `ExtractedByResources`, `ExtractedByComponents` (naming is hard).

docs/adr/0027_ocm_k8s_toolkit_discover.md
```diff
@@ -0,0 +1,343 @@
...
+
+Status payloads are subject to the etcd soft limit (~1.5 MiB). Discovery does not store descriptors outside etcd.
+Selectors and `Extract` let users shape the payload down. When the result still exceeds the limit, the API server
+rejects the status write; the controller clears `status.discovery`, repatches a `Ready=False` /
+`PayloadTooLarge` condition carrying the API server's rejection message, and returns without requeueing.
+
+The large-descriptor navigation case is descoped. Out-of-status storage (encoded payload,
+sibling ConfigMap, artifact CR) is an additive path if a concrete need appears.
+
+#### Deterministic ordering: `(name, version)` lexicographic
```
Skarlso: Semver odering would sort 1.10.0 and 1.9.0 differently. Lexicographic ordering orders 1.10.0 before 1.9.0. Which would be problematic.

docs/adr/0026_ocm_k8s_toolkit_discover.md
Skarlso: Hm, removing that would mean that expression needs to be dropped. 🤔 I guess the question is, is it worth it? Is the size concern real, or should we measure it and see? I feel like having a lot of CRs might be also making it more difficult to track things. Back in OCM v1 we had over 8 or so... It got really messy. :)

ntnn: I can imagine :D I also don't like adding a million types, but I also don't like schemaless types 🤷 
You can also have the status typed and only fill in what makes sense and see if there are problems with large data sets.

Skarlso: What if we would change this to 
```go
Discovery []map[string]apiextensionsv1.JSON `json:"discovery,omitempty"`
```

We could skip the schemaless and still get the expressions.

ntnn: How about this?

```go
   Components []v2.ComponentDescriptor           `json:"components,omitempty"`
   Extracted  []map[string]apiextensionsv1.JSON  `json:"extracted,omitempty"`
```

Then discovery with no cel extraction stil gets a fully typed status and with cel extraction a list of results per discovered component

Skarlso: Hmmm, I feel like, two entries would make this ambiguous, no? Like, this means that you'll have to check both of them all the time, not?

ntnn: If you request discovery with extraction I'd expect you check `.extracted`, if you didn't you check `.components` for the full typed spec.

Skarlso: Oh, I see. Actually, I like it. In case there are no extracts the raw list is emitted.

ntnn: Yeah exactly; I see the value of extraction but in most cases I'd want to use Discovery just need a filter and then I'd want to use the typed values directly instead of having to convert to the descriptor.
And when I do extraction I'd make a type to convert to to access the extracted values.

Skarlso: Yep, I'm writing things up right now. :)

Skarlso: Actually, this does not work. :) And now I understand why this was a JSON thing in here, but it wasn't written down properly.

First, the v2.Descriptor is not an object you can JSON marshal by Kubernetes easily, because it contains a runtime.Raw object that has a Type that is like this:

```go
type Type struct {
	Version string
	Name    string
}
```

Our custom marshaller makes this work, but Kubernetes immediately fails with 
```go
bindings/go/runtime/type.go:24:2: encountered struct field "Version" without JSON tag in type "Type"
bindings/go/runtime/type.go:25:2: encountered struct field "Name" without JSON tag in type "Type"
```

One alternative is to create our own type that is just ComponentInfo but then that's not a real descriptor and cannot be properly used by anyone without using a conversion of some kind.

What does work nicely is []JSON and it fits the current need too.

I would say that's a scarifies we should take as is.

So we can have:

```go
// +optional
Components []apiextensionsv1.JSON `json:"components,omitempty"`
// +optional
Extracted []map[string]apiextensionsv1.JSON `json:"extracted,omitempty"`
```

And that's it. :)

Skarlso: Okay, updated this with the latest changes. Should now use these both fields. :)

ntnn: Ahhh okay, that's sad but then thats the best option we have yes. Thanks for going through the paces!

ntnn (approved):

docs/adr/0026_ocm_k8s_toolkit_discover.md
Skarlso: With pleasure. Thank you for the ideas! :) 🙇‍♀️

fabianburth (approved):

docs/adr/0027_ocm_k8s_toolkit_discover.md
```diff
@@ -0,0 +1,476 @@
...
+  # components that did resolve and records the rest in `status.unresolved`.
+  onResolutionError: Fail
+
+status:
+  # ... shared status fields (observedGeneration, effectiveOCMConfig,
+  # standard Ready/Reconciling/Stalled conditions) omitted; see other CRDs.
+
+  # Written only when `spec.extract` is set. Free-form records.
+  extracted:
+    - imageRef: ghcr.io/fluxcd/flux:2.8.1
```
piotrjanik, line 132: Could that be in the same shape as the access type? For example:

```yaml
- name: "ghcr.io/example/releasechannel/flux"
   version: "2.8.1"
   type: "ociArtifact"
   relation: "external"
   access:
     imageReference: "ghcr.io/fluxcd/flux:2.8.1",
   digest: ...
```

Maybe then we would not have to define the optional `extract` field as above.

Skarlso: This isn't an access spec though. 🤔 It's just plain information about resource and component.

piotrjanik: I understand it is not the same thing, but it could. Notice what we have above:

```yaml
extract:
byResources:
imageRef:        resource.access.imageReference
```

We could simplify that by just returning `resource.access`. User would not be allowed to project discovered content. That would simplify implementation.

Skarlso: This doesn't work for several reasons. For one, access is runtime.Raw which has the same problem I outlined above. We cannot marshal it. If we would only give the access, the resource entry has no parent and the look up functions in the controller wouldn't work, like `byComponents`. And lastly, it's huge. :) It could contain multiple fields that the ADR specifically removes because of the 1.5Mib limit. :)

Skarlso (approved):

PR #3070, opened 2026-07-08
fix: config discovery within `.config` folders

### LINKED ISSUE #3061: `ocm` CLI does not read config from `$HOME/.config/ocm/config`
**Context**

The `ocm` CLI does not seem to pick up my config from `$HOME/.config/ocm/config` which is listed as a well known location in `ocm --help`.

**Version**

```
{"major":"0","minor":"10","patch":"0","gitVersion":"0.10.0","goVersion":"go1.26.3","compiler":"gc","platform":"darwin/arm64"}
```

**To Reproduce**

- Place the config in `$HOME/.config/ocm/config`
- Confirm using `ls $HOME/.config/ocm/config`
- Run `ocm transfer --loglevel debug component-version a b`

**Actual behavior**

The CLI cannot find the config file.

```
time=2026-07-07T16:29:53.181+02:00 level=DEBUG msg="could not get configuration" error="ocm config not found in any known locations, see --help for details on how to supply configuration files"
```

Copying it to another well known location works:

```
cp $HOME/.config/ocm/config $HOME/.config/.ocmconfig
ocm transfer --loglevel debug component-version a b
```

```
time=2026-07-07T16:31:38.664+02:00 level=DEBUG msg="ocm config was loaded successfully" path=/Users/my-user/.config/.ocmconfig
```

**Expected behavior**

The CLI should find the config in all well known locations.

jneisener:
Config directories (`$XDG_CONFIG_HOME`, `$HOME/.config`) now look for `ocm/config` (no dot) - following XDG convention where you don't prefix with a dot inside a directory that's already dedicated to config.

Non-config directories (`$HOME`, `$PWD`, `$EXE_DIR`) still look for `.ocm/config` (with the dot) - because hidden directories are the convention there.

Both still also check for `.ocmconfig` as a flat-file alternative everywhere.
Fixes https://github.com/open-component-model/open-component-model/issues/3061

##### Verification


cli/cmd/configuration/ocm_config.go
```diff
@@ -18,6 +18,7 @@ import (
 const (
 	OCMConfigDirectoryName   = ".ocm"
 	OCMConfigFileName        = OCMConfigDirectoryName + "/config"
+	XDGOCMConfigFileName     = "ocm/config"
```
jakobmoellerdev, line 21: nit: should this be joined by path separator because on win it would be `\` right?

jneisener: Should be fine. They all go through `checkConfigPaths` which has:
https://github.com/open-component-model/open-component-model/blob/a4566d589af38f619942f6727446a5ebafa3c833/cli/cmd/configuration/ocm_config.go#L256

and in `filepath.Clean`:

> any occurrences of slash are replaced by Separator. 

https://pkg.go.dev/path/filepath#Clean

jakobmoellerdev: I do think you omitted the finally, so would "//" be replaced?

jneisener: > I do think you omitted the finally, so would "//" be replaced?

I'm not sure I get the question. Yes, I omitted the finally from the doc quote. Is there some problem in the ordering that I'm not seeing where 'finally' becomes important?

Where would a `//` come from though?

Either way
> 1. Replace multiple [Separator](https://pkg.go.dev/path/filepath#Separator) elements with a single one.

should take care of it, double slashes are simply skipped: https://github.com/golang/go/blob/go1.23.0/src/internal/filepathlite/path.go#L93-L95

jakobmoellerdev: well if the separator replacement comes first, then if we are on windows separator would be `\`, so `//` would not be recognized as needing to be deduped right?

jneisener: `IsPathSeparator` on Windows treats `\` and `/` as equivalent: https://github.com/golang/go/blob/master/src/internal/filepathlite/path_windows.go#L19-L21 

So I don't think the logic changes either way

matthiasbruns: We had this windows path discussion already in another PR - I would want you to test this in a windows vm and optionally also with windows targeted tests. Would that be okay?

jneisener: > We had this windows path discussion already in another PR - I would want you to test this in a windows vm and optionally also with windows targeted tests. Would that be okay?

Sure, I can do that. But wouldn't we just be re-testing the correctness of the go stdlib? 
What would you like to cover that's not already covered by e.g. https://github.com/golang/go/blob/go1.23.0/src/path/filepath/path_windows_test.go#L370-L399 or https://github.com/golang/go/blob/go1.23.0/src/path/filepath/path_windows_test.go#L333-L347?

frewilhelm (approved):

jakobmoellerdev (approved):

PR #3071, opened 2026-07-08
feat(release): add publish OCM component-versions for CLI, controller and product to release workflow

### LINKED ISSUE #1157: Create OCM components during release process for CLI and Controllers and the OCM product wrapper
**Description**
Create OCM component versions during the release process. Whenever a new GH release of CLI and controller is created, an OCM component version for each, CLI and Controller version should be created and pushed to ghcr.io. In addition to that a wrapping OCM product component that references these two components should be created.

All components should be created on every new release, for both minor and patch releases.

**Done Criteria**
- [x] CLI component gets created during release
- [x] Controller component gets created during release
- [x] OCM product component gets created during release
- [x] Code has been reviewed by team

jakobmoellerdev:
Publishes three OCM component-versions to `ghcr.io` on every release (RC and final):

- `ocm.software/cli` — 6 binaries (local blobs) + multi-arch CLI image by digest
- `ocm.software/controller` — controller image + Helm chart, both by digest
- `ocm.software/ocm` — product wrapper referencing the two via `componentReferences`

All three carry the same semver as the GitHub release. Both RC and final use `--component-version-conflict-policy replace` for idempotent reruns.

New reusable workflow `.github/workflows/publish-components.yml` called twice from `release.yml` (RC + final). `promote_and_release_final` gets a new output `final_chart_digest` as the chart is repackaged at promotion, so the digest differs from RC.

Updated `RELEASE_PROCESS.md` with a new "OCM components produced" section. ADR 0010 was updated, too.

Validated end-to-end on`jakobmoellerdev/open-component-model` fork — all three component descriptors landed in GHCR \`component-descriptors/ocm.software/ocm`, `ocm.software/ocm/cli`, `ocm.software/ocm/controller`).

---

**Difference from #3031 (@morri-son):** instead of generating three constructors inline via shell heredoc and calling `ocm add cv` three times, this PR uses a single checked-in `.github/components/ocm-release/component-constructor.yaml` with `${VAR}` placeholders and one `add cv` call. The OCM CLI expands env vars natively (`os.Expand`) and resolves component dependency order automatically within a single constructor. All other content from #3031 is preserved unchanged.


.github/components/ocm-release/component-constructor.yaml
```diff
@@ -0,0 +1,123 @@
+components:
+  - name: ocm.software/cli
```
jakobmoellerdev, line 5: unsure about the names, lets discuss those

docs/adr/0010_release_strategy.md
```diff
@@ -179,6 +178,27 @@ git tag -v <tag>
...
+
+**Conflict policy:** Both RC and final publications use `--component-version-conflict-policy replace`. This keeps reruns after a transient failure idempotent (mid-run flakes leave part of the three component-versions written; the rerun completes the rest without operator intervention). The release workflow's `concurrency` group and `create-tag.js`'s tag-existence checks prevent concurrent or diverging runs for the same version, so `replace` cannot silently overwrite a different commit's artifact.
+
+**Resource form:**
+
+* OCI artefacts (CLI image, controller image, Helm chart) are referenced **by digest** via `access.ociArtifact`. The image and chart are pushed earlier in the same release run; the component-version pins their digests, not their tags. Storage is not duplicated.
+* CLI binaries are embedded **by value** via `input.type: file`. There is no per-platform OCI store for the six binaries, so digest-pinning is not available. This duplicates ~180 MB per release between the GitHub release assets and the component-version local blobs — an acceptable cost for offline transportability of the component-version.
```
jakobmoellerdev: maybe we want to use the wget access here, that we recently added. that way we could do reference by url

.github/components/ocm-release/component-constructor.yaml
matthiasbruns: It is `ocm.software/ocmcli` in v1 right? now cli?
We cannot have the same naming, cause the versions might conflict? `cli` feels a bit too generic for me

.github/components/ocm-release/component-constructor.yaml
```diff
@@ -0,0 +1,123 @@
...
+          path: "${CLI_ASSETS}/bin/ocm-windows-arm64"
+      - name: image
+        type: ociImage
+        version: "${VERSION}"
+        relation: local
+        access:
+          type: ociArtifact
+          imageReference: "${CLI_IMAGE_REF}"
+
+  - name: ocm.software/controller
```
frewilhelm: If we want to use the same package name, I would propose to use `ocm.software/kubernetes/controller`

.github/components/ocm-release/component-constructor.yaml
frewilhelm: IMO `cli` is fine as it matches the pattern of the cli-image `ghcr.io/open-component-model/cli`

.github/workflows/publish-components.yml
```diff
@@ -0,0 +1,123 @@
...
+          EOF
+
+      - name: Publish component-versions
+        env:
+          VERSION: ${{ inputs.version }}
+          CONFLICT_POLICY: ${{ inputs.conflict_policy }}
+          CONSTRUCTOR: ${{ github.workspace }}/.github/components/ocm-release/component-constructor.yaml
+        run: |
+          set -euo pipefail
+          "$CLI_BIN" add component-version \
```
frewilhelm: I am wondering if we could use the cli image instead of deriving the name of the `CLI_BIN` executable from the runner-arch. I am concerned that this cli-name-construction is not stable (e.g. when `dpkg` or the runners arch changes)

.github/workflows/release.yml
```diff
@@ -315,6 +315,25 @@ jobs:
               .write();
   # === cli-release.yml::release_rc + controller-release.yml::release_rc ===
 
+  # Publishes ocm.software/{cli,controller,ocm} for the RC, referencing the
+  # just-pushed CLI image, controller image and chart by digest.
+  publish_components_rc:
```
frewilhelm: I would put this before the `Summarize RC release` step and put the cv name and reference into the summary. Then, as a release manager, I can verify if everything is in place

.github/workflows/release.yml
```diff
@@ -567,7 +591,9 @@ jobs:
           VERSION: ${{ needs.prepare.outputs.base_version }}
         run: |
           set -euo pipefail
-          echo "digest=$(oras resolve $CONTROLLER_CHART:${VERSION})" >> "$GITHUB_OUTPUT"
+          digest=$(oras resolve "$CONTROLLER_CHART:${VERSION}")
+          [[ "$digest" == sha256:* ]] || { echo "::error::oras resolve returned unexpected digest: $digest"; exit 1; }
```
frewilhelm, line 598:
```suggestion
          [[ "$digest" == sha256:* ]] || { echo "::error::oras resolve returned unexpected digest algorithm: $digest"; exit 1; }
```

Is there any concern that this could not be `sha256`? Can we pin that?

frewilhelm (commented):
Do we want to add a section into our [release notes](https://github.com/open-component-model/open-component-model/blob/main/cliff.toml) about the cvs?

.github/workflows/release.yml
jakobmoellerdev: since github only publishes sha256, I dont see a reason for that codepath right now

.github/workflows/publish-components.yml
jakobmoellerdev: agreed on this.

.github/workflows/release.yml
jakobmoellerdev: I will think about this

.github/components/ocm-release/component-constructor.yaml
morri-son: Ok, I sticked to ghcr.io/open-component-model/cli

docs/adr/0010_release_strategy.md
morri-son: changed to Wget/v1 access type, but this changes the download/upload of the release artifacts. We first need a URL to use in the constructor and then use it instead of the local binaries in the former version.

morri-son:
@frewilhelm 
>Do we want to add a section into our [release notes](https://github.com/open-component-model/open-component-model/blob/main/cliff.toml) about the cvs?

added a section that lists the components and describes an `ocm get cv`and then an `ocm download resource` to get the ocm binary for the end user architecture.

.github/components/ocm-release/component-constructor.yaml
morri-son: oki, used ocm.software/kubernetes/controller

.github/workflows/publish-components.yml
morri-son: changed to doing all ocm CLI related actions with the CLI image

.github/workflows/release.yml
morri-son: I kept the order, but added the (calculated) CV into the RC summary.

morri-son:
@coderabbitai review

.github/workflows/release.yml
```diff
@@ -297,6 +297,14 @@ jobs:
             });
             const assetsCount = release.data.assets?.length || 0;
 
+            // OCM component-versions the publish_components_rc job pushes for this RC,
+            // so the release manager can verify the expected coordinates are in place.
+            // The // separates the OCM repository from the component name (the CLI
+            // maps it to the component-descriptors path internally).
```
frewilhelm: I understand that comment but I do not understand why its there

frewilhelm: I think you mean the trailing `/` in the `repo` variable. Nevertheless, I find it a bit confusing here because there is no instance of `//`. I would remove the comment

.github/workflows/release.yml
frewilhelm: Why? Does this mean the summary is written before the components are pushed? What happens when `publish-components` fails? The summary would be written before right?

docs/adr/0010_release_strategy.md
```diff
@@ -179,6 +178,24 @@ git tag -v <tag>
 * If a final release has a defect, ship a corrective patch (vX.Y.Z+1) and mark the previous final as superseded in the notes.
 * RC defects found after promotion follow the same corrective-patch route.
 
+### OCM components produced
+
+Every release publishes three OCM component-versions to `ghcr.io/<owner>`, implementing the originally out-of-scope "Root ocm component" item from this ADR.
```
frewilhelm: I think the owner is `open-component-model`

frewilhelm: > implementing the originally out-of-scope "Root ocm component" item from this ADR.

This feels off especially regarding the fact that we remove that line about a `Root ocm component` above

.github/components/ocm-release/component-constructor.yaml
```diff
@@ -0,0 +1,107 @@
...
+# access type. This requires the GitHub release carrying ${RELEASE_TAG}
+# to already exist when this constructor is published.
+_cli_executable_defaults: &cli_executable_defaults
+  name: cli
+  type: executable
+  version: "${VERSION}"
+  labels:
+    - name: downloadName
+      value: ocm
+  relation: external
```
frewilhelm: what is this? This feels like template magic from a long time ago o.O

docs/adr/0010_release_strategy.md
```diff
@@ -179,6 +178,24 @@ git tag -v <tag>
...
+* **`ocm.software/kubernetes/controller`** — controller image and Helm chart, both by digest.
+* **`ocm.software/ocm`** — product wrapper; pure `componentReferences` to the two above.
+
+All three carry the same bare semver as the GitHub release, and are published on both RC and final phases so consumers can validate the shape against an RC first.
+
+**Conflict policy:** both phases use `--component-version-conflict-policy replace`, so reruns after a transient failure are idempotent. The `concurrency` group and `create-tag.js` tag-existence checks prevent diverging runs for the same version, so `replace` cannot overwrite a different commit's artifact.
+
+**Resource form:** OCI artefacts (images, chart) are pinned by digest via `access.ociArtifact`. CLI binaries use `wget` access (`relation: external`) pointing at the release download URLs — no bytes are duplicated, but publishing depends on the GitHub release existing first (see the RC/final job wiring in `release.yml`).
+
+**Constructor evolution:** adding resources within a MAJOR is additive; removing/renaming a resource or changing `extraIdentity` is breaking and needs a MAJOR bump.
```
frewilhelm: I don't understand this. Do we plan to release majors?

cliff.toml
frewilhelm, line 1: While you are at that file: Can you please adjust the examples above (installing the cli, running the cli image, and installing the controller) with the respective version for that release.

RELEASE_PROCESS.md
```diff
@@ -108,6 +111,29 @@ The website integrates into the same workflow run, after `promote_and_release_fi
 
 There is no `website/v0.X.Y-rc.N` tag. The website has no RC artifacts to validate (it's a docs site, not a binary or image), so the RC variant would have no consumer.
 
+## OCM components produced
+
+Every release publishes three OCM component-versions to `ghcr.io/<owner>`, all carrying the same version as the GitHub release. Both RC and final publications use the `replace` conflict policy so that reruns after a transient failure are idempotent. The release workflow's `concurrency` group prevents parallel runs for the same version.
```
frewilhelm: `owner` should be known

RELEASE_PROCESS.md
```diff
@@ -108,6 +111,29 @@ The website integrates into the same workflow run, after `promote_and_release_fi
...
+```bash
+# The whole product (CLI + controller) for a given release
+ocm get component-version ghcr.io/open-component-model//ocm.software/ocm:0.7.0
+
+# Pull a CLI binary for your platform
+ocm download resource ghcr.io/open-component-model//ocm.software/cli:0.7.0 \
+  --identity name=cli,os=linux,architecture=amd64
+```
+
+The `ocm.software/ocm` product version is the recommended entry point — it pins the exact CLI and controller versions that were released together.
```
frewilhelm: I don't think this should be part of the release process.

jakobmoellerdev:
@morri-son can you take care of this on my fork pls?

.github/components/ocm-release/component-constructor.yaml
morri-son: not ancient, anchors are part of the spec and [used across DevOps teams](https://support.atlassian.com/bitbucket-cloud/docs/yaml-anchors/). But maybe a bit overambitious 😂 I  wanted to save 30 lines of duplicate attributes in resources where only the os/arch is changing. But you're right, this is anyway a template, so why trying to optimize it? I de-deduplicated the constructor template.

docs/adr/0010_release_strategy.md
morri-son: exchanged owner by open-component-model and removed the second part of the sentence with the root comp.

morri-son: changed to ghcr.io/open-component-model

.github/workflows/release.yml
morri-son: comment removed

RELEASE_PROCESS.md
morri-son: replaced with open-component-model

RELEASE_PROCESS.md
morri-son: removed the "consume" part.

cliff.toml
morri-son: adjusted the parts to use the release's version.

docs/adr/0010_release_strategy.md
morri-son: removed **Constructor evolution:**

.github/workflows/release.yml
morri-son: didn't think this through and wanted to avoid creating a new job. But now I did this and create a new job for the RC summary, so now the order should be correct.

.github/workflows/release-test.yml
```diff
@@ -0,0 +1,788 @@
...
+          path: ${{ runner.temp }}/cli-assets
+
+      - name: Download controller chart artifact
+        uses: actions/download-artifact@3e5f45b2cfb9172054b4087a40e8e0b5a5461e7c # v8
+        with:
+          name: ${{ needs.build_and_test.outputs.controller_chart_artifact_name }}
+          path: ${{ runner.temp }}/chart-assets
+
+      - name: Create RC pre-release
+        uses: softprops/action-gh-release@718ea10b132b3b2eba29c1007bb80653f286566b # v3
```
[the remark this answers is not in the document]
morri-son: Removed test workflow

.github/workflows/release-test.yml
```diff
@@ -0,0 +1,788 @@
...
+      - name: Checkout
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6
+        with:
+          fetch-depth: 0
+          ref: ${{ github.ref_name }}
+          token: ${{ steps.get_token.outputs.token }}
+          fetch-tags: true
+          # persist-credentials kept intentionally — create-tag.js does
+          # `git push origin refs/tags/...` which needs the app token persisted
+          # in .git/config to authenticate. Zizmor flags this; accepted.
```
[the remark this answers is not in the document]
morri-son: Removed test workflow

.github/components/ocm-release/component-constructor.yaml
```diff
@@ -0,0 +1,132 @@
...
+        version: "${VERSION}"
+        labels:
+          - name: downloadName
+            value: ocm
+        relation: external
+        extraIdentity:
+          os: linux
+          architecture: amd64
+        access:
+          type: wget/v1
```
matthiasbruns: nit: UpperCamelCase for `type` :>

.github/workflows/release.yml
```diff
@@ -309,18 +334,19 @@ jobs:
                 ['Controller Image Digest', (process.env.IMAGE_DIGEST || 'N/A').substring(0, 19) + '...'],
                 ['Chart Digest', (process.env.CHART_DIGEST || 'N/A').substring(0, 19) + '...'],
                 ['Uploaded Assets', String(assetsCount)],
+                ['OCM Product Component', `${repo}/ocm.software/ocm:${ver}`],
+                ['OCM CLI Component', `${repo}/ocm.software/cli:${ver}`],
+                ['OCM Controller Component', `${repo}/ocm.software/kubernetes/controller:${ver}`],
```
Skarlso: Errr... shouldn't these be with `//ocm.software`?

RELEASE_PROCESS.md
```diff
@@ -77,18 +77,21 @@ flowchart TD
     end
 
     Pipeline --> ReleaseRC[release_rc<br/>GitHub pre-release<br/>binaries + OCI tarballs + chart + changelog]
+    Pipeline --> PubCompRC[publish_components_rc<br/>ocm.software/cli, /controller, /ocm<br/>conflict: replace]
```
Skarlso: Fairly sure according to the comment in release this should be `ReleaseRC --> PubCompRC` and not `Pipeline --> PubCompRC`.

.github/workflows/publish-components.yml
```diff
@@ -0,0 +1,131 @@
...
+          CONFLICT_POLICY: ${{ inputs.conflict_policy }}
+          WORKSPACE: ${{ github.workspace }}
+          CONSTRUCTOR: ${{ github.workspace }}/.github/components/ocm-release/component-constructor.yaml
+        run: |
+          set -euo pipefail
+          docker run --rm \
+            -v "${HOME}/.docker/config.json:/.docker/config.json:ro" \
+            -v "${HOME}/.ocmconfig:/.ocmconfig:ro" \
+            -v /etc/ssl/certs/:/etc/ssl/certs/:ro \
+            -v "${WORKSPACE}:${WORKSPACE}" \
```
Skarlso: nit: I believe this mount can also be `:ro` right?

.github/workflows/release.yml
morri-son: yes, just corrected :)

RELEASE_PROCESS.md
morri-son: corrected

matthiasbruns (approved):

Skarlso (approved):

PR #3085, opened 2026-07-10
feat(github): implement download for the GitHub access method

### LINKED ISSUE #1069: Implement github download functionality for the GitHub access method
**Description**

Implement the get transformer that downloads a GitHub commit tarball and buffers it to the filesystem. Depends on #1068 (spec).

Use `github.com/google/go-github/v66/github` for the GitHub REST API client. Follow the transformer pattern in [`bindings/go/oci/transformer/`](../bindings/go/oci/transformer/).

Credential consumer identity must resolve a GitHub token from the OCM credential store — follow the credential resolution pattern in existing access methods.

**Done Criteria**
- [ ] Credential consumer identity resolution for GitHub tokens
- [ ] Get transformer downloads commit tarball from GitHub API, buffers to filesystem
- [ ] GitHub Enterprise supported via `apiHostname` field
- [ ] Unit tests with `go-github` mock server
- [ ] Integration test exercises download against a real (or stubbed) GitHub repository

piotrjanik:

This PR introduces the GitHub resource repository with download support.

- A download fetches the pinned commit's archive via the GitHub REST API and returns it as an application/x-tgz blob.
- Credentials resolve through the GitHubRepository consumer identity into the new typed `GitHubCredentials/v1` credential, which carries a single token applied as bearer auth; GitHub Enterprise is reached via apiHostname.
- The digest processor pins an unpinned ref to the commit it currently resolves to and computes a genericBlobDigest/v1 SHA-256 over the archive. A resource may be declared with only a ref: the digest processor pins it to a commit, and a download without a pinned commit resolves the ref on the fly.
- Sources may carry the same access spec, but only as provenance metadata: sources have no digest in the component descriptor, so their content is never fetched or validated during construction. The constructor now warns about this and advises declaring content as a resource when it must be verifiable. A dedicated source repository is therefore not part of this module.

Fixes: https://github.com/open-component-model/ocm-project/issues/1069


bindings/go/github/internal/archive/archive.go
```diff
@@ -0,0 +1,156 @@
+// Package archive fetches the source archive of a GitHub repository at a
```
matthiasbruns: in other bindings we called this simply "download" :D

bindings/go/github/internal/tempblob/tempblob.go
```diff
@@ -0,0 +1,115 @@
...
+// api/utils/accessobj.CachedBlobAccessForWriter), whose stated purpose is to
+// store large remote blobs "in the filesystem, instead of the memory, to avoid
+// blowing up the memory consumption".
+//
+// Unlike that cache, cleanup here does not rely on the consumer: no caller of
+// repository.ResourceRepository.DownloadResource closes the blob it receives,
+// so a blob that is never closed removes its file once it becomes unreachable.
+// Close remains available for callers that know when they are done and want
+// the file reclaimed immediately.
+package tempblob
```
matthiasbruns: did you check what we already have in our filesystem bindings?

bindings/go/github/internal/access.go
```diff
@@ -0,0 +1,34 @@
...
+
+// AccessFromSource converts the source's access into a typed *v1.GitHub.
+func AccessFromSource(source *descriptor.Source) (*v1.GitHub, error) {
+	if source == nil || source.Access == nil {
+		return nil, fmt.Errorf("source access is required")
+	}
+	return convertAccess(source.Access)
+}
+
+func convertAccess(access runtime.Typed) (*v1.GitHub, error) {
```
matthiasbruns: this could all be one method instead of three - just pass in the Access instead of source or resource

bindings/go/github/internal/credentials.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	typed, err := credentialScheme.NewObject(credentials.GetType())
+	if err != nil {
+		return "", fmt.Errorf("error converting credential type: %w", err)
+	}
+	if err := credentialScheme.Convert(credentials, typed); err != nil {
+		return "", fmt.Errorf("error converting credential type: %w", err)
+	}
+	direct, ok := typed.(*credv1.DirectCredentials)
```
matthiasbruns: why direct credentials?

bindings/go/github/internal/archive/archive.go
```diff
@@ -0,0 +1,156 @@
...
+}
+
+// clientFor parses repoURL into owner and repository and builds a GitHub REST
+// client for it, authenticated with token when non-empty.
+func clientFor(repoURL, apiHostname, token string) (gh *github.Client, owner, repo string, err error) {
+	owner, repo, err = parseOwnerRepo(repoURL)
+	if err != nil {
+		return nil, "", "", err
+	}
+	gh, err = newGitHubClient(repoURL, apiHostname, token, nil)
```
matthiasbruns: http client is never used besides tests - dont we want the http binding in here somehow? or the http config?

bindings/go/github/repository/resource/resource_repository.go
```diff
@@ -0,0 +1,118 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error resolving GitHub credentials: %w", err)
+	}
+
+	switch {
+	case gitHub.Commit == "":
+		if err := gitHub.Validate(); err != nil {
+			return nil, fmt.Errorf("invalid GitHub access: %w", err)
+		}
+		resolved, err := archive.ResolveCommit(ctx, gitHub.RepoURL, gitHub.APIHostname, gitHub.Ref, token)
```
matthiasbruns: you should use the github credentials typed cred (one is needed still) internally, instead of just passing the token

matthiasbruns: (still not resolved, just moved)

bindings/go/github/internal/credentials.go
piotrjanik: introduced `GitHubCredentials` type

bindings/go/github/spec/credentials/v1/convert.go
```diff
@@ -0,0 +1,66 @@
...
+	credentialKeyAccessToken = "accessToken"
+)
+
+var convertScheme = runtime.NewScheme()
+
+func init() {
+	// Register through the same helper as the public Scheme, so the two cannot
+	// disagree about which spellings resolve.
+	MustRegisterCredentialType(convertScheme)
+	directcredsv1.MustRegister(convertScheme)
```
matthiasbruns, line 23: what is this for?

piotrjanik: @matthiasbruns  That was some "code simplification" :) At least that was my intention. I changed it to match the same style as in the helm binding.

bindings/go/github/internal/credentials.go
```diff
@@ -0,0 +1,49 @@
...
+func CredentialsFrom(credentials runtime.Typed) (*credsv1.GitHubCredentials, error) {
+	gitHubCredentials, err := credsv1.ConvertToGitHubCredentials(credentials)
+	if err != nil {
+		return nil, err
+	}
+	if gitHubCredentials == nil {
+		return nil, nil
+	}
+	if gitHubCredentials.Token == "" {
+		return nil, fmt.Errorf("credentials were provided but contain no github token; refusing to fall back to anonymous access")
```
matthiasbruns: why? what if this is a public repo or am I missing something?

bindings/go/github/repository/source/doc.go
```diff
@@ -0,0 +1,5 @@
+// Package source implements the OCM source repository for the GitHub access
+// type: it downloads a GitHub repository's source archive at a pinned commit
+// via the GitHub REST API and serves it as a gzipped tar blob
+// (application/x-tgz).
+package source
```
matthiasbruns: we usually have a bit of example code in package docs

bindings/go/github/internal/download/client.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+	return parts[0], strings.TrimSuffix(parts[1], ".git"), nil
+}
+
+// clientFor splits the access's repository URL into owner and repository and
+// builds a GitHub REST client for it; nil credentials leave it anonymous. For
+// github.com the public API is used; for any other host, or when the access sets
+// APIHostname, the client targets that GitHub Enterprise API host. httpClient,
+// when nil, falls back to defaultHTTPClient.
+func clientFor(gitHub *v1.GitHub, credentials *credsv1.GitHubCredentials, httpClient *http.Client) (gh *github.Client, owner, repo string, err error) {
```
matthiasbruns: if we need `owner, repo`, cant we just get them in the caller side and pass them as params into this function instead of this weird return for `clientFor`
I would expect `clientFor` only returning the client, not owner nor repo.

so my suggestion - move 

```go
	u, err := parseRepoURL(gitHub.RepoURL)
	if err != nil {
		return nil, "", "", err
	}
	if owner, repo, err = ownerRepo(u, gitHub.RepoURL); err != nil {
		return nil, "", "", err
	}
````
one layer up and expect url, owner and rep as params, or just url and do a second `ownerRepo` call in here

**Review by matthiasbruns — COMMENTED** [review]
conceptionally I think this is fine, just code-nits

File: bindings/go/github/internal/credentials.go (hunk 8 above)
Comment (piotrjanik) [reply bindings/go/github/internal/credentials.go]: what is the point of gh credentials without a token? would not that be a malformed credential config?
Do you think we should just pass token as-is? I'm fine with that

Comment (matthiasbruns) [reply bindings/go/github/internal/credentials.go]: ok, so if a credential is resolved, but invalid, we warn / error out - thats fair. But  don't we have a Validate() func for that?

Comment (piotrjanik) [reply bindings/go/github/internal/credentials.go]: I checked how GitHub api works and in case of incorrect token it returns "401- Bad credentials". So, I think droping this validation is ok.

File: bindings/go/github/repository/source/doc.go (hunk 9 above)
Comment (piotrjanik) [reply bindings/go/github/repository/source/doc.go]: done

File: bindings/go/github/internal/download/client.go (hunk 10 above)
Comment (piotrjanik) [reply bindings/go/github/internal/download/client.go]: done

File: bindings/go/github/internal/download/client.go [hunk 11]
```diff
@@ -0,0 +1,168 @@
...
+	if gitHub.APIHostname == "" && u.Hostname() == "github.com" {
+		return gh, nil
+	}
+
+	enterpriseHost := gitHub.APIHostname
+	if enterpriseHost == "" {
+		enterpriseHost = u.Host
+	}
+	base := (&url.URL{Scheme: u.Scheme, Host: enterpriseHost}).String()
+	gh, err := gh.WithEnterpriseURLs(base, base)
```
matthiasbruns: whats the difference to https://github.com/google/go-github/blob/v89.0.0/github/github.go#L452 ?

matthiasbruns:
Works after wiring it into the cli

```bash
#!/bin/zsh
set -euo pipefail

# Test: Download a resource using the GitHub access type.
# This creates a CTF with a GitHub/v1 access spec pinned to a public commit,
# then downloads and verifies the archive content.
#
# Note: the download hits the public GitHub REST API unauthenticated (rate
# limited to 60 requests/hour). Export GITHUB_TOKEN and configure a credential
# for the GitHubRepository identity to raise the limit if needed.

alias OCM='go run ../../main.go'

SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
WORK_DIR="$SCRIPT_DIR/test-github-resource-output"
rm -rf "$WORK_DIR"
mkdir -p "$WORK_DIR"
CTF_DIR="$WORK_DIR/ctf"
DOWNLOAD_DIR="$WORK_DIR/downloaded"
CONSTRUCTOR_PATH="$WORK_DIR/constructor.yaml"

echo "=== Test: Download GitHub Resource ==="
echo "Working directory: $WORK_DIR"

# 1. Create constructor with a GitHub access pinned to a known public commit
cat <<EOF > "$CONSTRUCTOR_PATH"
components:
- name: ocm.software/github-test
  version: 1.0.0
  provider:
    name: ocm.software
  resources:
  - name: hello-world-source
    version: 1.0.0
    type: blob
    access:
      type: GitHub
      repoUrl: https://github.com/octocat/Hello-World
      commit: 7fd1a60b01f91b314f59955a4e4d4e80d8edf11d
EOF

# 2. Add component version to CTF
echo "--- Adding component version to CTF ---"
OCM add cv --repository "ctf::$CTF_DIR" --constructor "$CONSTRUCTOR_PATH" --skip-reference-digest-processing

# 3. Download the resource via the GitHub access type
echo "--- Downloading GitHub resource ---"
OCM download resource "ctf::$CTF_DIR//ocm.software/github-test:1.0.0" \
  --identity name=hello-world-source,version=1.0.0 \
  --output "$DOWNLOAD_DIR"

# 4. Verify the download produced a non-empty file
if [ ! -e "$DOWNLOAD_DIR" ]; then
  echo "FAIL: Downloaded file does not exist at $DOWNLOAD_DIR"
  exit 1
fi

FILE_SIZE=$(stat -f%z "$DOWNLOAD_DIR" 2>/dev/null || stat -c%s "$DOWNLOAD_DIR" 2>/dev/null)
if [ "$FILE_SIZE" -eq 0 ]; then
  echo "FAIL: Downloaded file is empty"
  exit 1
fi

echo "--- Downloaded file size: $FILE_SIZE bytes ---"

# 5. Verify it is a valid gzip archive (GitHub returns the commit archive as x-tgz)
if ! file "$DOWNLOAD_DIR" | grep -q "gzip"; then
  echo "FAIL: Downloaded file is not gzip compressed"
  exit 1
fi

# 6. Unpack the archive and verify it contains files.
# GitHub tarballs nest everything under a single "<owner>-<repo>-<shortsha>/" directory.
EXTRACT_DIR="$WORK_DIR/extracted"
mkdir -p "$EXTRACT_DIR"
echo "--- Unpacking archive into $EXTRACT_DIR ---"
tar xzf "$DOWNLOAD_DIR" -C "$EXTRACT_DIR"

EXTRACTED_FILES=$(find "$EXTRACT_DIR" -type f | wc -l | tr -d ' ')
if [ "$EXTRACTED_FILES" -eq 0 ]; then
  echo "FAIL: Archive unpacked to no files"
  exit 1
fi

echo "--- Extracted $EXTRACTED_FILES file(s): ---"
find "$EXTRACT_DIR" -type f -print

echo ""
echo "=== PASS: GitHub resource download and unpack succeeded ==="

```

matthiasbruns (dismissed):

bindings/go/github/internal/download/download.go
```diff
@@ -0,0 +1,79 @@
...
+	if err != nil {
+		return nil, err
+	}
+	defer func() {
+		if err := stream.Close(); err != nil {
+			slog.WarnContext(ctx, "error closing GitHub archive stream", "error", err)
+		}
+	}()
+
+	tmpFile, err := os.CreateTemp("", "github-archive-*.tgz")
```
Skarlso: Needs to use the filesystem config temp folder. :/

Skarlso:
Nothing else found other than using the right TEMP folder.

bindings/go/github/internal/download/download.go
piotrjanik: Please check :)

bindings/go/github/go.mod
```diff
@@ -3,16 +3,37 @@ module ocm.software/open-component-model/bindings/go/github
 go 1.26.4
 
 require (
+	github.com/google/go-github/v66 v66.0.0
```
frewilhelm: Is there a specific reason why you use `v66` if `v89` is available?

bindings/go/github/repository/resource/resource_repository.go
```diff
@@ -0,0 +1,148 @@
...
+	}
+
+	gitHubCredentials, err := credsv1.ConvertToGitHubCredentials(credentials)
+	if err != nil {
+		return nil, fmt.Errorf("error resolving GitHub credentials: %w", err)
+	}
+
+	switch {
+	case gitHub.Commit == "":
+		if err := gitHub.Validate(); err != nil {
```
frewilhelm: why don't we call validate outside of the switch-case? is a valid access spec not relevant when `github.Ref != ""`?

bindings/go/github/spec/credentials/v1/convert.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"fmt"
+
+	directcredsv1 "ocm.software/open-component-model/bindings/go/credentials/spec/config/v1"
+	"ocm.software/open-component-model/bindings/go/runtime"
+)
+
+const (
+	credentialKeyToken       = "token"
+	credentialKeyAccessToken = "accessToken"
```
frewilhelm: I was just wondering what the difference between a `token` and an `accessToken` is. Is this some kind of compatibility-issue? We might want to add a comment to clarify it

bindings/go/github/repository/source/source_repository.go
```diff
@@ -0,0 +1,95 @@
...
+	"ocm.software/open-component-model/bindings/go/github/internal/download"
+	ocmhttp "ocm.software/open-component-model/bindings/go/http"
+	httpv1alpha1 "ocm.software/open-component-model/bindings/go/http/spec/config/v1alpha1"
+	"ocm.software/open-component-model/bindings/go/repository"
+	"ocm.software/open-component-model/bindings/go/runtime"
+)
+
+// SourceRepository implements a source repository for GitHub repositories. It
+// serves the same commit archive as the resource repository, but anonymously:
+// [repository.SourceRepository] carries no credentials.
```
frewilhelm: Can't the `httpConfig` or `httpClient` carry credentials?

bindings/go/github/internal/download/client.go
```diff
@@ -0,0 +1,168 @@
...
+// archiveMaxRedirects bounds redirect following when resolving the archive
+// link. GitHub answers with a 302 that GetArchiveLink returns as a Location
+// without following; this only guards against a 301 chain.
+const archiveMaxRedirects = 1
+
+// ownerRepo extracts the owner and repository from a parsed GitHub repository
+// URL, whose path must have the form <owner>/<repo> (with or without a .git
+// suffix). repoURL is the caller's original, un-normalized URL, used only so a
+// rejection quotes the string the caller actually supplied.
+func ownerRepo(u *url.URL, repoURL string) (owner, repo string, err error) {
```
frewilhelm, line 36: Shouldn't `github.OwnerInfo{}` or `github.Repository` contain these information?

frewilhelm (commented):
I am missing an entry in `bindings/go/README.md` and a toplevel `doc.go` in `bindings/go/github`.

Additionally, I have the impression that a lot of the test mock-ups are redundant. However, I am unsure if there is any other way to do that.

bindings/go/github/digest/digest.go
```diff
@@ -0,0 +1,145 @@
...
+	}()
+
+	resolvedDigest, err := godigest.FromReader(reader)
+	if err != nil {
+		return nil, fmt.Errorf("error digesting downloaded github archive: %w", err)
+	}
+
+	if res.Digest == nil {
+		res.Digest = &descriptor.Digest{
+			HashAlgorithm:          hashAlgorithmSHA256,
```
frewilhelm, line 129: We do trust `godigest.FromReader` to return a SHA256 hash right? We do not set it for `godigest` or validate if it is a SHA256. I am just wondering what will happen if `godigest` will change the default to SHA512. I don't think we will notice. Is there any possibility to set the digest algorithm to SHA256?

bindings/go/github/digest/digest.go
```diff
@@ -0,0 +1,145 @@
...
+		}
+		return res, nil
+	}
+
+	if res.Digest.HashAlgorithm != hashAlgorithmSHA256 {
+		return nil, fmt.Errorf("hash algorithm mismatch: expected %s, got %s", hashAlgorithmSHA256, res.Digest.HashAlgorithm)
+	}
+	if res.Digest.NormalisationAlgorithm != normalisationGenericBlobDigestV1 {
+		return nil, fmt.Errorf("normalisation algorithm mismatch: expected %s, got %s", normalisationGenericBlobDigestV1, res.Digest.NormalisationAlgorithm)
+	}
```
frewilhelm: Is this check necessary? We do not do any form of normalisation do we? So, in theory, a digest for a GitHub repository without `res.Digest.NormalisationAlgorithm` should be valid too. WDYT?

frewilhelm (commented):
Sorry, for the re-review. I accidentally submitted the previous one

bindings/go/github/digest/digest.go
```diff
@@ -0,0 +1,145 @@
...
+	// just resolved the ref, or the caller pinned it — so carrying the ref into
+	// the download would only make DownloadResource re-resolve it to check for
+	// drift, a redundant API call on the digest path. The returned resource
+	// keeps the ref informationally; the download copy drops it.
+	dlAccess := gitHub.DeepCopy()
+	dlAccess.Ref = ""
+	dlResource := res.DeepCopy()
+	dlResource.Access = dlAccess
+
+	downloaded, err := p.resourceRepository.DownloadResource(ctx, dlResource, credentials)
```
frewilhelm, line 113: Do we want to remove the file after processing the digest or is this the responsibility of the user?

frewilhelm (commented):
this one is the last - i promise :D

bindings/go/github/spec/credentials/v1/convert.go
Skarlso: These are different fields for short lived access tokens vs tokens that are dev tokens if I recall correctly. One is used to generate short lived access the other is for proper access with a dev token. 🤔

bindings/go/github/repository/source/source_repository.go
piotrjanik: @frewilhelm This relates to:
- https://github.com/open-component-model/ocm-project/issues/857 which was replaced by:
- https://github.com/open-component-model/ocm-project/issues/774

I would keep it like that as a tech-debt and handle it in dedicated task

bindings/go/github/digest/digest.go
piotrjanik: @frewilhelm i think it safe to remove it when digest processor finished its job. Wdyt?

bindings/go/github/digest/digest.go
piotrjanik: It should be pinned to sha256 now

bindings/go/github/go.mod
piotrjanik: No :) bumped

bindings/go/github/repository/source/source_repository.go
frewilhelm: I am not sure I understand. What will happen and how can we make sure this tech-debt in this code-path will be addressed

bindings/go/github/digest/digest.go
frewilhelm: I am unsure :D Can you please raise this in the warroom

bindings/go/github/internal/download/client.go
piotrjanik: @frewilhelm  `ownerRepo(u, gitHub.RepoURL)` gets `gitHub.RepoURL` where `gitHub` is access type, so I dont know how `github` could help here

bindings/go/github/digest/digest.go
piotrjanik: @frewilhelm 
We have it in wget, so I just followed.. 
https://github.com/open-component-model/open-component-model/blob/d818a3dd7cde8c67c09f7e63f32c4de79f18c871/bindings/go/wget/repository/resource_repository.go#L161

`genericBlobDigest/v1` is - correct me if I am wrong - a no-op normalisation.

bindings/go/github/spec/credentials/v1/convert.go
piotrjanik: Helm converts `accessToken` from `DirectCredentials`, so I assumed this is a convention.
https://github.com/open-component-model/open-component-model/blob/e4d3ffbaf1f3aec637b2f965fe77af52979a1f9a/bindings/go/helm/spec/credentials/v1/convert.go#L46-L53

Shall I remove it? I think it is safe to remove. Old ocm requires `token`

bindings/go/github/repository/source/source_repository.go
piotrjanik: Because we have a task for it, otherwise this PR would get even bigger. I would say we can add such task to GitHub epic. wdyt?

bindings/go/github/repository/resource/resource_repository.go
piotrjanik: done

bindings/go/github/spec/credentials/v1/convert.go
matthiasbruns: what has helm to do with that 🤔 :) 
if you dont need it, drop it - your credentials should serve your impl

bindings/go/github/digest/digest.go
matthiasbruns: digest processor is download-and-forget

jakobmoellerdev: To me it is questionalble if an automatic download for digest calculation is really the best behavior. I think we should at the very least print a fat warning here as discussed. If you dont want to go that route, the alternative is not calculating a digest at all (no digest processor capability)

bindings/go/github/internal/download/client.go
frewilhelm: I believe I mismatched `github` and `gitHub`

bindings/go/github/digest/digest.go
frewilhelm: AFAIK the digest process is usually called when a component constructor is used to create a component descriptor.
Imagine a user has typed down a resource.Digest in his component constructor. The user would not necessarily know what normalisation algorithm to put in because the blob is not getting normalized. Accordingly, an unset field `resource.Digest.NormalisationAlgorithm` should be fine I think. Additionally, on re-calculating the digest, we could check if `resource.Digest.HashAlgorithm` is set and use that info to calculate the digest.

bindings/go/github/digest/digest.go
piotrjanik: I have implemented similar approach as in helm

bindings/go/github/spec/credentials/v1/convert.go
piotrjanik: I have removed accessToken

piotrjanik:
> I am missing an entry in `bindings/go/README.md` and a toplevel `doc.go` in `bindings/go/github`.


@frewilhelm both added: the github row is in bindings/go/README.md and bindings/go/github/doc.go now has the package docs.

bindings/go/github/repository/resource/doc.go
```diff
@@ -0,0 +1,29 @@
...
+// The [ResourceRepository] downloads a GitHub repository's source archive at a
+// pinned commit via the GitHub REST API and returns it as a gzipped tar blob
+// (application/x-tgz). A resource may be pinned to a commit or carry only a ref,
+// which [ResourceRepository.ResolveCommit] resolves to the commit it currently
+// points at; when both are set the commit wins.
+//
+// # Usage
+//
+//	// A nil filesystem config buffers archives in the OS temp directory.
+//	repo := resource.NewResourceRepository(nil)
```
Skarlso: This is no longer true I believe since the NewResourceRepository needs a filesystem config.

bindings/go/constructor/construct.go
```diff
@@ -314,6 +314,14 @@ func (c *DefaultConstructor) processDescriptor(
 		})
 	}
 
+	if len(component.Sources) > 0 {
+		// Sources carry no digest in the component descriptor, so unlike
+		// resources their content is never validated during construction and
+		// cannot be verified afterwards. The limitation is the same for every
+		// source, so one warning per component version is enough.
+		logger.Warn("source content is recorded without a digest and is not verifiable; declare the artifact as a resource if content validation is needed")
```
Skarlso, line 322: This will warn EVERY component version and type. Are we okay with that? Just making sure. :)

bindings/go/github/internal/download/client.go
```diff
@@ -0,0 +1,172 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error creating github archive download request for %s/%s@%s: %w", owner, repo, commit, err)
+	}
+	resp, err := httpClient.Do(req)
+	if err != nil {
+		// For a private repository GitHub signs the link with a short-lived token
+		// in its query, and the *url.Error net/http returns renders the full
+		// request URL. Report the cause against the commit coordinates instead,
+		// so the token cannot reach a log.
+		return nil, fmt.Errorf("error downloading github archive %s/%s@%s: %w", owner, repo, commit, errors.Unwrap(err))
```
Skarlso: Why Unwrap? 🤔  I don't see anyone trying to assert this error in any way? ( Although effectively it will populate to DownloadResource, do you expect anyone to read this in some meaningful way? )

bindings/go/github/digest/digest.go
```diff
@@ -0,0 +1,163 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error reading downloaded github archive: %w", err)
+	}
+	defer func() {
+		if err := reader.Close(); err != nil {
+			slog.WarnContext(ctx, "error closing downloaded github archive", "error", err)
+		}
+	}()
+
+	resolvedDigest, err := godigest.SHA256.FromReader(reader)
```
Skarlso: Why are we doing this again if download.Download already passed in a digester and has the digest. Can't you use downloaded.Digest() here? 🤔

Skarlso (commented):
Just three tiny questions. Otherwise this appears to be working as advertised. :)

bindings/go/github/repository/resource/doc.go
piotrjanik: It does not need it anymore. I changed my approach. I am not buffering it to a file; I am just streaming. Initially, I wanted to download and process, but then when should the file be removed? It was solvable with some extra code. For calculating the digest, I think it is unnecessary.

bindings/go/constructor/construct.go
piotrjanik: The idea was to indicate that if sources>0, no digest calculation is performed; thus, resources might be a better option.

bindings/go/github/internal/download/client.go
piotrjanik: I changed that a bit. GH provides s3-like authenticated url, so I added redact func + some tests.
My thinking was download must be performed "in a secure way" :)

bindings/go/github/repository/resource/doc.go
Skarlso: So you need to rewrite this?

bindings/go/github/digest/digest.go
piotrjanik: Fixed that by checking whether the blob supports the DigestAware interface and then calling `Digest()`

bindings/go/github/repository/resource/doc.go
piotrjanik: yes, I just did that

Skarlso (approved):

matthiasbruns (approved):
🌻

PR #3116, opened 2026-07-14
feat: s3 access, credentials & identity (download only) (#1178)

### LINKED ISSUE #1178: S3 resource repository with authentication
**Description**

Define the S3Bucket access type struct in `bindings/go/s3/spec/access/v1/` following the OCI/Helm pattern.

- Define `S3Bucket` struct with fields backwards-compatible with OCM v1 (there exists a v1 and v2 of the S3 access that will have to be evaluated and aligned with the team how to deal with it)
- Add `Validate()` method verifying required fields
- Register via `MustAddToScheme()` with versioned (`S3Bucket/v1`) and unversioned (`S3Bucket`) aliases
- Apply code generation annotations: `+ocm:typegen=true`, `+ocm:jsonschema-gen=true`, `+k8s:deepcopy-gen=true`

Reference: [`bindings/go/helm/spec/access/`](https://github.com/open-component-model/open-component-model/tree/main/bindings/go/helm/spec/access)

---

Implement `repository.ResourceRepository` in `bindings/go/s3/repository/resource/resource_repository.go` to handle download, ~upload~, and authenticated access for S3 artifacts.

- `DownloadResource()` — fetch artifact bytes from a S3 bucket
- ~`UploadResource()` — push artifact bytes to S3 bucket, return updated resource descriptor~ (wait for uploader concept)
- `GetResourceCredentialConsumerIdentity()` — return `S3Bucket` identity so OCM credential providers can supply accessKey / secret auth
- Credentials /Token passed through to all AWS SDK requests

Reference:
- [`bindings/go/helm/repository/resource/`](https://github.com/open-component-model/open-component-model/tree/main/bindings/go/helm/repository/resource)
- [`bindings/go/repository/interface.go`](https://github.com/open-component-model/open-component-model/blob/main/bindings/go/repository/interface.go)
- [`bindings/go/credentials/`](https://github.com/open-component-model/open-component-model/tree/main/bindings/go/credentials)
- [`website/content/docs/reference/credential-consumer-identities.md`](https://github.com/open-component-model/open-component-model/blob/main/website/content/docs/reference/credential-consumer-identities.md)

**Done Criteria**
- [x] `DownloadResource()` fetches artifact from S3 bucket
- [ ] ~`UploadResource()` pushes artifacts and returns updated resource descriptor~
- [x] `S3Bucket` credential consumer identity constant defined
- [x] AWS Credentials/Token wired into SDK requests

matthiasbruns:
Creates a new binging for s3 access.
This PR focusses on the downloading part. Upload will be tackled later.

One thing to note - the aws sdk handles retries on a higher level. I will still use http config retries for lower level calls - so retries might multiply.

Contributes: https://github.com/open-component-model/ocm-project/issues/1178

#### Testing

Fully implemented and tested in https://github.com/matthiasbruns/open-component-model/pull/118

##### Verification


matthiasbruns:
@coderabbitai review

bindings/go/s3/spec/credentials/v1/convert.go
```diff
@@ -0,0 +1,77 @@
...
+		AccessKeyID:     accessKeyID,
+		SecretAccessKey: secretAccessKey,
+		SessionToken:    sessionToken,
+	}
+}
+
+// ConvertToS3Credentials converts runtime.Typed into S3Credentials.
+// Direct conversion as well as converting from DirectCredentials/v1 is supported.
+func ConvertToS3Credentials(creds runtime.Typed) (*S3Credentials, error) {
+	typed, err := convertScheme.NewObject(creds.GetType())
```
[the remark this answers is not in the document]
matthiasbruns: should be fixed

bindings/go/s3/internal/download/download.go
```diff
@@ -0,0 +1,290 @@
...
+// InsecureSkipTLSVerify from the access spec is folded into the TLS config so the
+// shared client applies it through its own transport, which warns once per host
+// that verification is off instead of disabling it silently.
+func httpConfig(cfg *httpv1alpha1.Config, insecureSkipTLSVerify bool) *httpv1alpha1.Config {
+	out := cfg.DeepCopy()
+	if out == nil {
+		out = &httpv1alpha1.Config{}
+	}
+
+	out.Retry = noRetry()
```
matthiasbruns: TODO @matthiasbruns  instead of resetting it and set it to -1, we should take the default settings from aws sdk's client

matthiasbruns:
Tested with

**.ocmconfig**
```yaml
configurations:
  - type: credentials.config.ocm.software
    consumers:
      - identity:
          type: S3Bucket
          hostname: 127.0.0.1
          port: "$MINIO_PORT"
        credentials:
          - type: S3Credentials/v1
            accessKeyId: $MINIO_USER
            secretAccessKey: $MINIO_PASSWORD
```

**constructor.yaml**
```yaml
components:
- name: ocm.software/s3-test
  version: 1.0.0
  provider:
    name: ocm.software
  resources:
  - name: payload
    version: 1.0.0
    type: blob
    access:
      type: S3Bucket/v1
      endpoint: http://127.0.0.1:$MINIO_PORT
      bucketName: $BUCKET
      objectKey: $OBJECT_KEY
      usePathStyle: true
      mediaType: text/plain
```

against my local minio docker container. Ran `add cv` and ` download resource`

bindings/go/s3/internal/download/download.go
```diff
@@ -0,0 +1,224 @@
...
+// AWS requires a region even when a custom endpoint is targeted; S3-compatible
+// stores usually ignore it.
+const defaultRegion = "us-east-1"
+
+const tempFilePattern = "ocm-s3-download-*"
+
+// UnversionedVersionID is the versionId S3 reports for objects that carry no real
+// version. It is a valid input to GetObject but pins nothing — overwriting such an
+// object leaves the versionId unchanged.
+const UnversionedVersionID = "null"
```
Skarlso: Use

```go
const ()
```

instead.

bindings/go/s3/internal/download/download.go
```diff
@@ -0,0 +1,224 @@
...
+	Blob blob.ReadOnlyBlob
+	// VersionID is the object version that was read, as reported by S3. It is
+	// [UnversionedVersionID] for an unversioned object and empty for stores that do
+	// not report a version, so it identifies immutable content only when it is neither.
+	VersionID string
+}
+
+// Request describes a single S3 object download.
+type Request struct {
+	// Region is the bucket region. Optional for custom endpoints.
```
Skarlso: This does not coincide with `// AWS requires a region even when a custom endpoint is targeted;`.

Mention that it's defaulted in case of custom endpoint.

bindings/go/s3/internal/download/download.go
```diff
@@ -0,0 +1,224 @@
...
+	if getter == nil {
+		client, err := newClient(ctx, req, o)
+		if err != nil {
+			return nil, err
+		}
+		getter = client
+	}
+
+	in := &s3.GetObjectInput{
+		Bucket: aws.String(req.BucketName),
```
Skarlso: Instead of all the `aws.String` things you can use `new(req.BucketName)` now.

bindings/go/s3/internal/download/download.go
```diff
@@ -0,0 +1,224 @@
...
+func httpConfig(cfg *httpv1alpha1.Config, insecureSkipTLSVerify bool) *httpv1alpha1.Config {
+	out := cfg.DeepCopy()
+	if out == nil {
+		out = &httpv1alpha1.Config{}
+	}
+
+	if insecureSkipTLSVerify {
+		skip := true
+		out.InsecureSkipVerify = &skip
+	}
```
Skarlso:
```suggestion
	if insecureSkipTLSVerify {
		out.InsecureSkipVerify = new(insecureSkipTLSVerify)
	}
```

bindings/go/s3/internal/download/download.go
```diff
@@ -0,0 +1,224 @@
...
+
+	// S3 reports the size up front, so an oversized object is rejected before any
+	// of it is transferred.
+	if maxDownloadSize > 0 && out.ContentLength != nil && *out.ContentLength > maxDownloadSize {
+		return nil, fmt.Errorf("s3 object %s/%s exceeds maximum allowed size of %d bytes", req.BucketName, req.ObjectKey, maxDownloadSize)
+	}
+
+	body := io.Reader(out.Body)
+	if maxDownloadSize > 0 {
+		body = io.LimitReader(out.Body, maxDownloadSize+1)
```
Skarlso: I noticed this in wget too but I don't understand what's the purpose of the `+1`. Especially because if MaxInt is given and you +1 that it will result in an overflow and nothing is guarding against that.

bindings/go/s3/internal/download/download.go
```diff
@@ -0,0 +1,224 @@
...
+	if closeErr := file.Close(); err == nil {
+		err = closeErr
+	}
+	if err != nil {
+		return nil, fmt.Errorf("error writing s3 object %s/%s to %s: %w", req.BucketName, req.ObjectKey, path, err)
+	}
+
+	if maxDownloadSize > 0 && written > maxDownloadSize {
+		return nil, fmt.Errorf("s3 object %s/%s exceeds maximum allowed size of %d bytes", req.BucketName, req.ObjectKey, maxDownloadSize)
+	}
```
Skarlso, line 177: This... is too late, right? 🤔 We already wrote it. The damage has been done. 🤔

bindings/go/s3/internal/download/download.go
```diff
@@ -0,0 +1,224 @@
...
+		region = defaultRegion
+	}
+
+	httpClient := o.HTTPClient
+	if httpClient == nil {
+		httpClient = ocmhttp.New(ocmhttp.WithConfig(httpConfig(o.HTTPConfig, req.InsecureSkipTLSVerify)))
+	}
+
+	// Deliberately no retry option: the SDK resolves its own attempt count from
+	// AWS_MAX_ATTEMPTS and AWS_RETRY_MODE.
```
Skarlso: I'm not a 100% sure here. Like AWS has it's own retry-er even and `RetryMaxAttempts` and `RetryMode`. Why can't we configured that from the configured http options?

bindings/go/s3/internal/download/download.go
```diff
@@ -0,0 +1,224 @@
...
+	if o.Credentials != nil {
+		s3creds, err := credv1.ConvertToS3Credentials(o.Credentials)
+		if err != nil {
+			return nil, fmt.Errorf("error converting s3 credentials: %w", err)
+		}
+		if s3creds.AccessKeyID != "" {
+			loadOpts = append(loadOpts, config.WithCredentialsProvider(
+				credentials.NewStaticCredentialsProvider(s3creds.AccessKeyID, s3creds.SecretAccessKey, s3creds.SessionToken),
+			))
+		}
```
Skarlso: What happens if I ONLY supply the access key/session token? That should be a valid credential configuration I believe?

bindings/go/s3/repository/resource_repository.go
```diff
@@ -0,0 +1,255 @@
...
+	"ocm.software/open-component-model/bindings/go/runtime"
+	"ocm.software/open-component-model/bindings/go/s3/internal/download"
+	accessspec "ocm.software/open-component-model/bindings/go/s3/spec/access"
+	v1 "ocm.software/open-component-model/bindings/go/s3/spec/access/v1"
+	identityv1 "ocm.software/open-component-model/bindings/go/s3/spec/identity/v1"
+)
+
+const (
+	hashAlgorithmSHA256 = "SHA-256"
+	genericBlobDigestV1 = "genericBlobDigest/v1"
```
Skarlso, line 27: Shouldn't this be uppercase? 🤔

bindings/go/s3/repository/resource_repository.go
```diff
@@ -0,0 +1,255 @@
...
+
+// pinAccess points the resource's access at the object version the digest was computed
+// over, satisfying the [repository.ResourceDigestProcessor] requirement that a processed
+// access "MUST always reference the content described by the digest and cannot be mutated".
+// An access that already names a version was pinned by whoever wrote it and is left alone.
+//
+// Unversioned buckets report no version, or the placeholder
+// [download.UnversionedVersionID], neither of which survives an overwrite. Those are
+// logged rather than pinned or rejected, because erroring would make digests unusable
+// for every unversioned bucket.
```
Skarlso: I think that would be the point, no? 🤔 I 100% think it should error and and not do anything other than just Warn that hey, your shit might be overwritten but go ahead. 🤔

bindings/go/s3/spec/identity/v1/type.go
```diff
@@ -0,0 +1,119 @@
...
+// its bucket, key and optional endpoint. Without an endpoint the identity carries only
+// the object path, because AWS is reached at a well-known host; with one, the endpoint's
+// URL attributes are added so consumer entries can be scoped to that store. The
+// unversioned [Type] is used, consistent with the other consumer identities.
+func IdentityFromObject(bucketName, objectKey, endpoint string) (runtime.Identity, error) {
+	if bucketName == "" {
+		return nil, fmt.Errorf("bucketName is required")
+	}
+
+	loc := path.Join(bucketName, objectKey)
```
Skarlso: path.Join runs path.Clean messing up valid s3 key syntax. Add tests for this:

```go
	s1 := "asdf//asdf"
	s2 := "asdf/./asdf"
	s3 := "asdf/../asdf"
	fmt.Println(path.Join("bucket", s1))
	fmt.Println(path.Join("bucket", s2))
	fmt.Println(path.Join("bucket", s3))
```

becomes:
```go
bucket/asdf/asdf
bucket/asdf/asdf
bucket/asdf
```

bindings/go/s3/repository/resource_repository.go
```diff
@@ -0,0 +1,255 @@
...
+
+	if versionID == "" || versionID == download.UnversionedVersionID {
+		slog.WarnContext(ctx, "s3 object carries no version, so its access cannot be pinned to the digested content and may later resolve to a different object",
+			slog.String("bucket", spec.BucketName),
+			slog.String("objectKey", spec.ObjectKey))
+		return
+	}
+
+	spec.Version = versionID
+	resource.Access = spec
```
Skarlso: Hm. This updates the access to a specifically typed access. Meaning if it was previously Raw this now becomes S3Bucket. Are you SURE that's correct?

bindings/go/s3/repository/resource_repository.go
```diff
@@ -0,0 +1,255 @@
...
+// matching rules.
+func (r *ResourceRepository) GetResourceCredentialConsumerIdentity(ctx context.Context, resource *descriptor.Resource) (runtime.Identity, error) {
+	if resource == nil {
+		return nil, fmt.Errorf("resource is required")
+	}
+	if resource.Access == nil {
+		return nil, fmt.Errorf("resource access is required")
+	}
+
+	spec := v1.S3Bucket{}
```
Skarlso: Aren't your already doing this in converAccess? 🤔

Skarlso (commented):
Okay, I think this is enough for now. :DDD

I'll do another pass a bit later. I need to clear my head. :D

bindings/go/s3/repository/resource_repository.go
matthiasbruns: hmmmmmmmmmmmm

github and wget both do lowercase 🥲

matthiasbruns: and our json as well
https://github.com/open-component-model/open-component-model/blob/a7fdc8c5bb860375a0b5d945a4131031caf6fd28/bindings/go/descriptor/normalisation/json/v4alpha1/normalisation.go#L28

bindings/go/s3/internal/download/download.go
matthiasbruns: I will never get used to that fancy modern go with it's generics and functions :D

bindings/go/s3/internal/download/download.go
matthiasbruns: `EOF` magic - we need to figure out if it's a real EOF or just one by `io.LimitReader`
We read one more byte from the stream - if there is none, the size wont be `+1`, if there is more, we find it like this.
Overflow capped now

bindings/go/s3/spec/identity/v1/type.go
matthiasbruns: fixed, back to manual concat :D

bindings/go/s3/repository/resource_repository.go
matthiasbruns: that's okay, I would want a real typed instead of Raw anyway - and it's being done basically in all bindings afaik

bindings/go/s3/internal/download/download.go
matthiasbruns: updated docs

bindings/go/s3/internal/download/download.go
matthiasbruns: hope I got all of them

bindings/go/s3/repository/resource_repository.go
matthiasbruns: checking - not 100% sure yet. If we can use ETAG or something when it's unversioned, I will have to update the Digest pinning

bindings/go/s3/repository/resource_repository.go
Skarlso: Ah okay. :)

bindings/go/s3/repository/resource_repository.go
matthiasbruns: so the worst thing that could happen for unversioned data is that we wont accept new versions of the file thus existing cv's would become useless

Skarlso: Perfect 👍

bindings/go/s3/internal/download/download.go
matthiasbruns: we can, but http is per request and the sdk is per operation. I was also unsure if we cant just set this in the sdk from the http config

bindings/go/s3/internal/download/download.go
matthiasbruns: I am pretty sure I checked this - let me get the docs again

matthiasbruns: But actually, we should maybe leave the validation to the sdk and do not check it at all

bindings/go/s3/internal/download/download.go
matthiasbruns: we do a Content-Length check before downloading, and then have a fallback case here which will stop writing after we hit a limit

bindings/go/s3/internal/download/download.go
matthiasbruns: ditched the validation

bindings/go/s3/internal/download/download.go
```diff
@@ -0,0 +1,292 @@
...
+	o := &option{}
+	for _, opt := range opts {
+		opt(o)
+	}
+
+	if req.BucketName == "" {
+		return nil, fmt.Errorf("bucketName is required")
+	}
+	if req.ObjectKey == "" {
+		return nil, fmt.Errorf("objectKey is required")
```
Skarlso: Replace with errors.New

bindings/go/s3/repository/resource_repository.go
```diff
@@ -0,0 +1,272 @@
...
+	}
+	// The blob never leaves this function, so its file is released right away instead
+	// of waiting for the caller or the cleanup to reclaim it.
+	defer func() {
+		if closeErr := result.Blob.Close(); closeErr != nil {
+			slog.WarnContext(ctx, "failed to remove temporary file after digest processing", "err", closeErr)
+		}
+	}()
+
+	raw, _ := result.Blob.Digest()
```
Skarlso: Oh is this okay to ignore here? 🤔

bindings/go/s3/internal/download/download.go
Skarlso: Actually. This should call Validate on S3Bucket. Better yet, put Validate into the convertAccess method. Otherwise it's never really called anywhere unless I'm missing some interface call in convert. 🤔

bindings/go/s3/spec/access/v1/s3_bucket.go
```diff
@@ -0,0 +1,74 @@
...
+	// Endpoint is the base endpoint of an S3-compatible store (e.g. MinIO, Ceph,
+	// R2). When empty, AWS S3 is targeted.
+	Endpoint string `json:"endpoint,omitempty"`
+
+	// UsePathStyle enables path-style addressing (bucket in the path instead of the
+	// host). Required by most self-hosted S3-compatible stores.
+	UsePathStyle bool `json:"usePathStyle,omitempty"`
+
+	// InsecureSkipTLSVerify disables TLS certificate verification for the endpoint.
+	InsecureSkipTLSVerify bool `json:"insecureSkipTLSVerify,omitempty"`
```
Skarlso: I'm pretty sure we shouldn't be exposing this. This means that a component author would be able to disable tls verification, right? That's definitely not something that we would like to allow people to do?

Skarlso (commented):
Some further comments, nothing much now! Well done. :)

matthiasbruns:
> Some further comments, nothing much now! Well done. :)

I will be on my MacBook this evening (80% chance) and update this or one more time. image

Skarlso:
@matthiasbruns I'm taking this over and fix things up with regards to my comments. :)

Skarlso:
>I will be on my MacBook this evening (80% chance) and update this or one more time.

Looks like that didn't happen. :D It's fine, you're on vacation. :) Leave this with me. :)

bindings/go/s3/internal/download/download.go
```diff
@@ -0,0 +1,284 @@
...
+
+	// Retrying happens in the SDK alone; see [httpConfig]. The ocm retry configuration
+	// drives it, so one setting means one thing, and an unset one leaves the SDK on its
+	// own default rather than silently adopting the transport's.
+	loadOpts := []func(*config.LoadOptions) error{
+		config.WithRegion(region),
+		config.WithHTTPClient(httpClient),
+	}
+	if attempts := sdkRetryAttempts(o.HTTPConfig); attempts > 0 {
+		loadOpts = append(loadOpts, config.WithRetryMaxAttempts(attempts))
```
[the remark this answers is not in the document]
Skarlso: The behavior here is inconsistent with wget/oci. Our own integration test would break too.

matthiasbruns:
@Skarlso 

> > I will be on my MacBook this evening (80% chance) and update this or one more time.
> 
> Looks like that didn't happen. :D It's fine, you're on vacation. :) Leave this with me. :)

My connection was basically non-existent.. Germany has some huge flaws in areas we travelled (Starlink ordered) - thanks for taking care, I will check the remaining comments :)

bindings/go/s3/internal/download/download.go
matthiasbruns: [bindings/go/s3/internal/download/download.go](https://github.com/open-component-model/open-component-model/pull/3116/files/b4a9b9318d94e7244f0db1a0318592e718cb81a7#diff-6df00dc441e15ace22b732be4150ba0040e5802b3cb2a22b4293d8146e944e55) is another layer, but I will call `Validate` before that

matthiasbruns: `Validate` is called here -> https://github.com/matthiasbruns/open-component-model/blob/fa7445648b1b395a5a46172db7e79a6bc9b0081f/bindings/go/s3/repository/resource_repository.go#L101

Skarlso (dismissed):

bindings/go/s3/spec/access/scheme.go
```diff
@@ -0,0 +1,32 @@
...
+var Scheme = runtime.NewScheme()
+
+func init() {
+	MustAddToScheme(Scheme)
+}
+
+func MustAddToScheme(scheme *runtime.Scheme) {
+	spec := &v1.S3Bucket{}
+
+	lowerCaseAccessType := strings.ToLower(v1.Type)
```
fabianburth: nit: I'd rather have a constant for this

bindings/go/s3/spec/access/scheme.go
```diff
@@ -0,0 +1,32 @@
...
+import (
+	"strings"
+
+	"ocm.software/open-component-model/bindings/go/runtime"
+	v1 "ocm.software/open-component-model/bindings/go/s3/spec/access/v1"
+)
+
+const (
+	S3BucketConsumerType = "S3Bucket"
+)
```
fabianburth: nit: don't we have a dedicated identity package for this?

bindings/go/s3/internal/download/download.go
```diff
@@ -0,0 +1,325 @@
...
+// are streamed rather than buffered, so memory use stays flat regardless of object
+// size; the file is created under the directory given by [WithTempDir] and outlives
+// this call.
+//
+// The returned [Blob] owns that file: callers should [Blob.Close] it once they are
+// done, and an unclosed blob has its file removed when it becomes unreachable.
+//
+// The S3 client, credentials and maximum size are supplied via options; see
+// [WithClient], [WithCredentials] and [WithMaxDownloadSize].
+func Download(ctx context.Context, req Request, opts ...Option) (_ *Result, err error) {
```
fabianburth: nit: why does this need a named error return?

bindings/go/s3/internal/download/options.go
```diff
@@ -0,0 +1,72 @@
...
+	Client          ObjectGetter
+	HTTPClient      *http.Client
+	Credentials     runtime.Typed
+	MaxDownloadSize *int64
+	TempDir         string
+	HTTPConfig      *httpv1alpha1.Config
+}
+
+// Option configures a download.
+type Option func(*option)
```
fabianburth, line 24: nit: do we really need a fully fledged option API for unexported options?

bindings/go/s3/repository/resource_repository.go
```diff
@@ -0,0 +1,278 @@
...
+	maxDownloadSize  *int64
+	httpConfig       *httpv1alpha1.Config
+	httpClient       *http.Client
+	filesystemConfig *filesystemv1alpha1.Config
+}
+
+// NewResourceRepository creates a new S3 resource repository. If filesystemConfig
+// is non-nil, its TempFolder is used for the files downloaded objects are streamed
+// into; otherwise os.CreateTemp's default directory is used.
+func NewResourceRepository(filesystemConfig *filesystemv1alpha1.Config, opts ...Option) *ResourceRepository {
```
fabianburth, line 44: is there a particular reason the filesystem config is required API - even though we even document a defaulting - but the rest are options?

bindings/go/s3/repository/options.go
```diff
@@ -0,0 +1,63 @@
...
+	"ocm.software/open-component-model/bindings/go/s3/internal/download"
+)
+
+// Options holds configuration for the S3 resource repository.
+type Options struct {
+	// client injects a pre-built S3 client in tests. It is deliberately not settable
+	// from outside this package: an S3 client bakes in one endpoint, region and
+	// identity, so injecting one would route every resource to the same place
+	// regardless of its access spec and credentials.
+	client download.ObjectGetter
```
fabianburth: isn't your resource repository anyway just an adapter for your internal download without a lot of real logic? do we really need the client injection here?

bindings/go/s3/internal/download/blob.go
```diff
@@ -0,0 +1,102 @@
...
+// Blob is the file-backed blob returned by [Download]. The file behind it is a
+// temporary download artifact owned by the blob: [Blob.Close] removes it, and a
+// cleanup attached to the blob removes it once the blob becomes unreachable.
+// A caller that drops the blob without closing it therefore does not keep the
+// file around for the lifetime of the process, which matches the reclamation an
+// in-memory blob got for free.
+type Blob struct {
+	*filesystem.Blob
+	path string
+}
```
fabianburth, line 24: why do we need this wrapper around filesystem blob?

fabianburth (changes requested):

bindings/go/s3/repository/resource_repository.go
matthiasbruns: Kept the original pattern from other repos, but I think we can do options here

bindings/go/s3/repository/options.go
matthiasbruns: Testing - and the downloader will be reused for input as well (see helm) - I can remove the injection though

bindings/go/s3/internal/download/blob.go
matthiasbruns: to auto-clean the tmp files - we did not do this in helm e.g. and are usually leaking the blobs all the time - I can either drop this completely or move it into our filesystem binding and update the others as well later

bindings/go/s3/spec/access/scheme.go
matthiasbruns: we started adding this in other bindings - but sure I can roll this back to a const

bindings/go/s3/spec/access/scheme.go
matthiasbruns: left over <3

bindings/go/s3/internal/download/options.go
matthiasbruns: yep - I would keep this to align with the existing patterns in helm - there are a lot of optional bits in the download package that can we either set or not

matthiasbruns:
@coderabbitai cleaned up a bit - based on the open comments and the issue, can you do another round of reviews?

Taskfile.yml
```diff
@@ -49,6 +49,14 @@ includes:
     optional: true
     taskfile: ./bindings/go/wget/integration/Taskfile.yml
     dir: ./bindings/go/wget/integration
+  bindings/go/s3:
+    optional: true
+    taskfile: ./bindings/go/s3/Taskfile.yml
+    dir: ./bindings/go/s3
```
frewilhelm: Is there a Taskfile?

bindings/go/s3/internal/download/blob.go
matthiasbruns: removed

fabianburth (dismissed):

fabianburth (dismissed):

bindings/go/s3/doc.go
```diff
@@ -0,0 +1,170 @@
...
+// attempt counts, so transport retry is switched off for the client handed to the SDK,
+// globally and per host.
+//
+// The ocm retry configuration drives that one layer, so a single setting means a single
+// thing. retry.maxRetries counts the attempts after the first, the SDK counts the first
+// as well:
+//
+//	maxRetries: 3    // 4 attempts
+//	maxRetries: -1   // 1 attempt, no retrying
+//	maxRetries: 0    // infinite, bounded only by ctx and the SDK's retry quota
```
frewilhelm: This feels kind of unintuitive to me. `-1` feels like it should be infinite, while `0` means no-retries == 1 attempt. Is this the usual behaviour?

bindings/go/s3/spec/access/v1/s3_bucket.go
```diff
@@ -0,0 +1,70 @@
...
+	// version is read.
+	Version string `json:"version,omitempty"`
+
+	// Endpoint is the base endpoint of an S3Bucket-compatible store (e.g. MinIO, Ceph,
+	// R2). When empty, AWS S3Bucket is targeted.
+	Endpoint string `json:"endpoint,omitempty"`
+
+	// UsePathStyle enables path-style addressing (bucket in the path instead of the
+	// host). Required by most self-hosted S3Bucket-compatible stores.
+	UsePathStyle bool `json:"usePathStyle,omitempty"`
```
frewilhelm, line 50: Do we plan to add those fields to the spec?

bindings/go/s3/spec/access/v1/s3_bucket.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+// Validate verifies that the required fields of the S3Bucket access are set.
+func (t *S3Bucket) Validate() error {
+	if t.BucketName == "" {
+		return errors.New("bucketName is required")
+	}
+	if t.ObjectKey == "" {
+		return errors.New("objectKey is required")
+	}
```
frewilhelm, line 60: Following the spec the `mediaType` should also be required

bindings/go/s3/integration/integration_test.go
```diff
@@ -0,0 +1,172 @@
...
+
+	filesystemv1alpha1 "ocm.software/open-component-model/bindings/go/configuration/filesystem/v1alpha1/spec"
+	descriptor "ocm.software/open-component-model/bindings/go/descriptor/runtime"
+	"ocm.software/open-component-model/bindings/go/s3/repository"
+	accessspec "ocm.software/open-component-model/bindings/go/s3/spec/access"
+	accessv1 "ocm.software/open-component-model/bindings/go/s3/spec/access/v1"
+	credv1 "ocm.software/open-component-model/bindings/go/s3/spec/credentials/v1"
+)
+
+const minioImage = "minio/minio:RELEASE.2024-01-16T16-07-38Z"
```
frewilhelm: Why is this release pinned?

matthiasbruns: unpinned

bindings/go/s3/spec/access/v1/s3_bucket.go
matthiasbruns: I don't see this to be required at all - even ocmv1 did automatic lookup and felt back when none was set

bindings/go/s3/spec/access/v1/s3_bucket.go
matthiasbruns: v2 is not even in the spec yet afaik - I will create a ticket

matthiasbruns: https://github.com/orgs/open-component-model/projects/10/views/16?pane=issue&itemId=5203086544&issue=open-component-model%7Cocm-project%7C1256

bindings/go/s3/spec/access/v1/s3_bucket.go
matthiasbruns: https://github.com/orgs/open-component-model/projects/10/views/16?pane=issue&itemId=5203086544&issue=open-component-model%7Cocm-project%7C1256

bindings/go/s3/internal/download/download.go
```diff
@@ -0,0 +1,324 @@
...
+	}
+
+	if req.BucketName == "" {
+		return nil, errors.New("bucketName is required")
+	}
+	if req.ObjectKey == "" {
+		return nil, errors.New("objectKey is required")
+	}
+
+	getter, err := newClient(ctx, req, o)
```
jakobmoellerdev, line 91: why do we produce a new client per request download?

bindings/go/s3/internal/download/download.go
```diff
@@ -0,0 +1,324 @@
...
+func httpConfig(cfg *httpv1alpha1.Config) *httpv1alpha1.Config {
+	out := cfg.DeepCopy()
+	if out == nil {
+		out = &httpv1alpha1.Config{}
+	}
+
+	// Retrying is left to the SDK: it retries the whole operation, re-signs every attempt
+	// and classifies S3's error codes, and retrying here too would multiply both counts.
+	// Per-host entries override the global policy, so they are switched off as well.
+	out.Retry = &httpv1alpha1.RetryConfig{MaxRetries: new(disableRetry)}
```
jakobmoellerdev, line 197: we probably want to log this overwrite if it happens!

bindings/go/s3/internal/download/download.go
```diff
@@ -0,0 +1,324 @@
...
+	}
+
+	return out
+}
+
+// hostKeys returns the per-host configuration keys to try for endpoint, most specific
+// first. It mirrors the host matching of the ocm HTTP client, where an entry keyed by
+// host:port wins over one keyed by the bare hostname.
+//
+// TODO: drop this once http config exposes per-host resolution. Until then, this duplicates
```
jakobmoellerdev: http config should already today allow per host resolution tbh, im a bit confused here

bindings/go/s3/internal/download/download.go
```diff
@@ -0,0 +1,324 @@
...
+		s3creds, err := credv1.ConvertToS3Credentials(o.Credentials)
+		if err != nil {
+			return nil, fmt.Errorf("error converting s3 credentials: %w", err)
+		}
+		if provider := staticCredentials(s3creds); provider != nil {
+			loadOpts = append(loadOpts, config.WithCredentialsProvider(*provider))
+		}
+	}
+
+	awsCfg, err := config.LoadDefaultConfig(ctx, loadOpts...)
```
jakobmoellerdev, line 288: i was expecting the config parameters to be routed through somehow, is this not done on purpose?

fabianburth (dismissed):

bindings/go/s3/repository/resource_repository.go
```diff
@@ -0,0 +1,268 @@
...
+	}
+
+	return resource, nil
+}
+
+// pinningVersion returns versionID when it identifies immutable content, and the empty
+// string when it does not. An unversioned bucket reports either no version or the
+// placeholder [download.UnversionedVersionID], neither of which survives an overwrite,
+// whether the store reported it or an author wrote it into the spec.
+func pinningVersion(versionID string) string {
```
jakobmoellerdev, line 256: is a function here really necessary with 4 lines of comment? :D

bindings/go/s3/spec/identity/v1/type.go
```diff
@@ -0,0 +1,131 @@
...
+	identity := runtime.Identity{}
+	if endpoint != "" {
+		id, err := runtime.ParseURLToIdentity(endpoint)
+		if err != nil {
+			return nil, fmt.Errorf("error parsing s3 endpoint to identity: %w", err)
+		}
+		identity = id
+		// Only the endpoint is URL-parsed; the location is appended to its base path.
+		if prefix := identity[runtime.IdentityAttributePath]; prefix != "" {
+			loc = strings.TrimSuffix(prefix, "/") + "/" + loc
```
jakobmoellerdev: this looks very strange, are you sure this is correct behavior?

bindings/go/s3/doc.go
```diff
@@ -0,0 +1,170 @@
+// Package s3 provides access to OCM resources stored as objects in an S3 or
```
jakobmoellerdev, line 1: nit: I think this doc needs to be shortened a bit, too agentic

bindings/go/s3/spec/identity/v1/type.go
matthiasbruns: That's the reason why 
https://github.com/open-component-model/open-component-model/pull/3116#discussion_r3672455207

but let me double check

bindings/go/s3/internal/download/download.go
matthiasbruns: Because e.g. credentials can change per request and target - I did not want to reuse the same client and do some locking just to keep the created object count down

bindings/go/s3/internal/download/download.go
matthiasbruns: Do we? In the end, we set the retry from the http config in the sdk and not on http layer - I don't see why we need to tell the user that - the result stays somewhat the same

bindings/go/s3/internal/download/download.go
matthiasbruns:
```go
func effectiveRetry(cfg *httpv1alpha1.Config, endpoint string) *httpv1alpha1.RetryConfig {
	if cfg == nil {
		return nil
	}

	for _, key := range hostKeys(endpoint) {
		if hostCfg := cfg.Hosts[key]; hostCfg != nil {
			return httpv1alpha1.MergeRetryConfig(cfg.Retry, hostCfg.Retry)
		}
	}

	return cfg.Retry
}
```

This is the lookup I use - but since we can now modify all bindings at once, I will move the lookup into a method on the config itself

bindings/go/s3/internal/download/download.go
matthiasbruns: All configs are mapped into `loadOpts` a few lines above - or am I getting your question wrong?

bindings/go/s3/internal/download/download.go
matthiasbruns: moved it into the http binding for reuse

bindings/go/s3/internal/download/download.go
jakobmoellerdev: because the user configuration in ocm config wouldnt work anymore even if it is custom configured right?

bindings/go/s3/internal/download/download.go
jakobmoellerdev: this should be decoupled by the client no? would you really need locking?

matthiasbruns: `region` is set in the client via `loaderOpts`
credentials could be dynamically retrieved - but honestly, I don't see this.


So some _could_ by retrieved dynamically, but not all - I would rather keep it that way than hack around it

https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/config#LoadOptions

bindings/go/s3/internal/download/download.go
matthiasbruns: So what happens:

if the user configures httpconfig, we take maxretries from it, pass it into the aws sdk but remove it from the underlying http client transport to prevent retry * retry retries happening.

https://github.com/open-component-model/open-component-model/blob/26ede43f421968d4047c71418713a95e3c6b879e/bindings/go/s3/internal/download/download.go#L223

We are simply moving the http retry from the http layer into the sdk layer.

bindings/go/http/spec/config/v1alpha1/config_test.go
```diff
@@ -873,3 +873,95 @@ func TestMerge_TLS(t *testing.T) {
...
 }
+
+func TestHostKeys(t *testing.T) {
+	tests := []struct {
+		name   string
+		host   string
+		expect []string
+	}{
+		{name: "empty host matches nothing", host: "", expect: nil},
+		{name: "bare hostname", host: "example.com", expect: []string{"example.com"}},
```
piotrjanik: Im a bit confused. It allows providing host:port where entry with a port gets higher priority but then port is removed as host is the only part that matters.

bindings/go/http/spec/config/v1alpha1/config_test.go
```diff
@@ -873,3 +873,95 @@ func TestMerge_TLS(t *testing.T) {
...
+			"nil-entry.com": nil,
+		},
+	}
+
+	t.Run("nil config yields zero value", func(t *testing.T) {
+		var nilCfg *httpspec.Config
+		assert.Equal(t, httpspec.HostConfig{}, nilCfg.ResolveHost("example.com"))
+	})
+
+	t.Run("unknown host yields global values", func(t *testing.T) {
```
piotrjanik: Does it mean that in case dns resolution failure we would apply defaults? Shouldn't we just return an error?

bindings/go/http/spec/config/v1alpha1/config_test.go
```diff
@@ -873,3 +873,95 @@ func TestMerge_TLS(t *testing.T) {
...
+
+	t.Run("host entry merges onto global values", func(t *testing.T) {
+		resolved := cfg.ResolveHost("example.com")
+		require.NotNil(t, resolved.Retry)
+		assert.Equal(t, 2, *resolved.Retry.MaxRetries, "host overrides global")
+		assert.Equal(t, httpspec.Timeout(200*time.Millisecond), *resolved.Retry.MinWait, "global preserved where host is nil")
+		assert.Equal(t, httpspec.Timeout(30*time.Second), *resolved.Timeout, "global preserved where host is nil")
+	})
+
+	t.Run("host:port entry wins over bare hostname", func(t *testing.T) {
```
piotrjanik: I think it's been already tested.

bindings/go/http/spec/config/v1alpha1/config_test.go
matthiasbruns: clarified

bindings/go/http/spec/config/v1alpha1/config_test.go
matthiasbruns: clarified

bindings/go/http/spec/config/v1alpha1/config_test.go
matthiasbruns: clarified

bindings/go/s3/spec/identity/v1/type.go
matthiasbruns: fixed + tests + cli tests (not commited)

fabianburth (approved):

jakobmoellerdev (approved):

PR #3118, opened 2026-07-14
fix: Adjust `FlatMap` merging logic

jneisener:
Part of addressing https://github.com/open-component-model/ocm-project/issues/1169

`FlatMap` would return a reverse list, which was opposite to the expected contract of 'last-wins'

Fixing this in the central interface and documenting the intention is step one for addressing https://github.com/open-component-model/ocm-project/issues/1169
Next step would be adding merge tests that actually test the flow through `FlatMap` in all consumers and finally a e2e test in the CLI (and potentially the controller).

##### Verification


bindings/go/configuration/generic/v1/spec/flat_map.go
```diff
@@ -1,14 +1,29 @@
...
+//	configurations:
+//	  - type: generic
+//	    configurations:
+//	      - type: transfer
+//	        copyMode: localBlob
+//	  - type: transfer
+//	    copyMode: allResources
+//
+// FlatMap produces: [transfer(allResources), transfer(localBlob)].
+//
```
jneisener, line 26: While the reversal didn't make sense, the rest of the function does - in a way - because it mirrors the v1 behavior.


In v1, config application is somewhat obtuse. It appears to work like this:

1. The generic config type's [`ApplyTo`](https://github.com/open-component-model/ocm/blob/465dcba9/api/config/extensions/config/type.go#L57-L69) iterates `c.Configurations` in order, calling [`cctx.ApplyConfig`](https://github.com/open-component-model/ocm/blob/465dcba9/api/config/internal/context.go#L237-L266) for each entry.
2. [`ApplyConfig`](https://github.com/open-component-model/ocm/blob/465dcba9/api/config/internal/context.go#L237-L266) first calls [`Evaluate`](https://github.com/open-component-model/ocm/blob/465dcba9/api/config/internal/store.go#L71-L80) on the entry (decoding `GenericConfig` → typed config), then [stores it](https://github.com/open-component-model/ocm/blob/465dcba9/api/config/internal/store.go#L112-L120) in the config store, then enters an [update loop](https://github.com/open-component-model/ocm/blob/465dcba9/api/config/internal/context.go#L254-L263) that calls [`updater.Update()`](https://github.com/open-component-model/ocm/blob/465dcba9/api/config/internal/updater.go#L79-L99).
3. [`updater.Update()`](https://github.com/open-component-model/ocm/blob/465dcba9/api/config/internal/updater.go#L79-L99) has an [`inupdate` re-entrancy guard](https://github.com/open-component-model/ocm/blob/465dcba9/api/config/internal/updater.go#L81-L84). When the outer generic's children call `ApplyConfig`, they're within the `Update` triggered by their parent - so the inner update loop [breaks immediately](https://github.com/open-component-model/ocm/blob/465dcba9/api/config/internal/context.go#L256) (because `in=true`). All direct children are stored without triggering expansion.
4. After the outer generic's `ApplyTo` finishes iterating all its children and the first `Update()` returns, the [loop re-checks](https://github.com/open-component-model/ocm/blob/465dcba9/api/config/internal/context.go#L256): `updater.lastGeneration < configs.Generation()` - there are new entries to process. A **second** `Update()` fires, which replays the newly stored entries. When it hits the nested generic, its [`ApplyTo`](https://github.com/open-component-model/ocm/blob/465dcba9/api/config/extensions/config/type.go#L57-L69) expands its children via [`ApplyConfig`](https://github.com/open-component-model/ocm/blob/465dcba9/api/config/internal/context.go#L237-L266), storing them **after** all direct siblings.

So the result for 
```yaml
configurations:
  - type: Dummy
    alice: first
  - type: Dummy
    alice: second
  - type: generic.config.ocm.software/v1
    configurations:
      - type: Dummy
        alice: from-nested
  - type: Dummy
    alice: fourth
```
is `["first", "second", "fourth", "from-nested"]` - not declaration order. Nested generic children are always evaluated last, regardless of where the nested generic appears.

`FlatMap` reproduces this by collecting direct entries first, then appending recursively flattened nested entries.

jakobmoellerdev (approved):

matthiasbruns (approved):
as discussed in the wr thats fine for a fix

PR #3128, opened 2026-07-15
chore(GHA): remove timestamp from pr name (ocm cli v1 docs creation)

frewilhelm:
The drama:
- We have a branch protection rule requiring an up-to-date branch for merging a PR
- We create a PR for updating the ocm cli v1 docs via workflow
- The workflow adds a timestamp to the branch-name that is used for the PR. Hence, every branch-name to update the ocm cli v1 docs is different
- When another PR gets merged after the ocm cli v1 docs PR gets created, we cannot merge that PR because of the branch protection rule. Currently, I can just retrigger the workflow (or do it manually) and hope that no other PR gets merged after. This is inefficient.

The solution:
- Do not append the timestamp to the branch-name. Hence, we can rerun the workflow to "create" ocm cli v1 docs which will now just force-push on the current branch with the same name.


.github/workflows/website-update-cli-docs.yaml
```diff
@@ -70,8 +70,7 @@ jobs:
             website/content_versioned/version-legacy/
             website/hack/generate-cli-docs/go.mod
             website/hack/generate-cli-docs/go.sum
-          branch: update-cli-docs/auto
-          branch-suffix: timestamp
```
matthiasbruns, line 74: I think this is to prevent branch collisions right? can we have something else like an uuid or so? or dont we need this at all?

frewilhelm: I think a collision is super unlikely considering the circumstance that this PR is created every two weeks. But an uuid is safer

morri-son (approved):

matthiasbruns (approved):

PR #3142, opened 2026-07-17
chore(deps): group digest bump prs

frewilhelm:
Groups digest-only re-pins (same tag, new SHA) for GitHub Actions and Docker base images into one weekly `Digest pins` PR instead of one each - they were ~18% of Renovate PRs. Pinning is unchanged; only the batching is.

`gomod` is excluded on purpose: sweeping it in would span multiple `go.mod` files (cross-module PR) and break the `go.sum` post-upgrade step. Go digests keep their own PRs.

See https://github.com/frewilhelm/open-component-model/pull/808 as an example (contains a lot of bumps because I outdated my `main` branch on purpose).

### DIFF
--- a/.github/renovate.json5
+++ b/.github/renovate.json5
@@ -59,6 +59,24 @@
       automerge: true,
     },
     {
+      // Collapse digest-only re-pins (same tag, new SHA) for Actions and
+      // Docker base images into one weekly PR. gomod is excluded so Go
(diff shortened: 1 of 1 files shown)


morri-son (approved):

frewilhelm:
@coderabbitai review

matthiasbruns (approved):

PR #3144, opened 2026-07-17
feat(release): sign published OCM component-versions with keyless Sigstore

### LINKED ISSUE #1236: Sign published OCM component versions with keyless Sigstore during release
## Description

Sign the three OCM component versions (CLI, controller, product wrapper) published during the release process using keyless Sigstore signing.

## What needs to be done

- Add `id-token: write` permission to the release jobs to mint a GitHub Actions OIDC token
- Install cosign on the runner (version pinned via `.env`, tracked by Renovate)
- After publishing, sign each component version using `ocm sign component-version --signer-spec` inside the CLI container

## Done Criteria

- [x] All three component versions are signed during every release (minor and patch)
- [ ] Signing identity is the GitHub Actions OIDC token (no stored credentials)
- [ ] Cosign version is sourced from the existing `.env` file (single source of truth)
- [ ] Signatures verified locally with `ocm verify component-version`
- [ ] Code reviewed by team

## References

- Implements: open-component-model/open-component-model#3144
- Follow-up to #1157 (component creation during release)

morri-son:
Follow-up to #3071, which added the `publish-components.yml` workflow. This PR adds keyless Sigstore signing of the three published component-versions.

## What changes

**`publish-components.yml`**
- Add `id-token: write` permission to the job (required to mint the GHA OIDC token)
- Install `cosign` on the runner via `sigstore/cosign-installer`
- After publishing, sign each component-version with `ocm sign component-version --signer-spec`.

**`release.yml`**
- Add `id-token: write` to both `publish_components_rc` and `publish_components_final` call sites (callers must forward this permission explicitly)

## Verification

E2E tested on the PR fork against existing `v0.7.0-rc.1` artifacts. All three component-versions signed and verified locally with `ocm verify component-version --verifier-spec`:


.github/workflows/publish-components.yml
```diff
@@ -53,11 +46,14 @@ jobs:
     permissions:
       contents: read
       packages: write
+      id-token: write # mint the GitHub Actions OIDC token used as the Sigstore signing identity
```
jakobmoellerdev: minting is ai slop

.github/workflows/publish-components.yml
```diff
@@ -53,11 +46,14 @@ jobs:
...
     steps:
       - name: Checkout component constructor
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
         with:
-          sparse-checkout: .github/components
+          sparse-checkout: |
+            .github/components
+            bindings/go/sigstore/signing/handler/internal/.env
```
jakobmoellerdev, line 57: why do we need the internal .env? EDIT: just saw its used for the sigstore version, good practice actually but probably deserves a comment

.github/workflows/publish-components.yml
```diff
@@ -129,3 +125,52 @@ jobs:
...
+              -v "${HOME}/.docker/config.json:/.docker/config.json:ro" \
+              -v "${HOME}/.ocmconfig:/.ocmconfig:ro" \
+              -v /etc/ssl/certs/:/etc/ssl/certs/:ro \
+              -v "${WORKSPACE}:${WORKSPACE}:ro" \
+              -w "${WORKSPACE}" \
+              "$CLI_IMAGE_REF" \
+                sign component-version \
+                --config "/.ocmconfig" \
+                --signer-spec "${WORKSPACE}/sigstore-sign.yaml" \
+                --force \
```
jakobmoellerdev: why do we need force? this should be the first time we sign right?

jakobmoellerdev:
I dont think we ever published them before right? Wanted to check this out somewhere

morri-son:
@jakobmoellerdev 
> I dont think we ever published them before right? Wanted to check this out somewhere

Correct, I tested this only on my fork using a copy of the workflow which does not contain our OCMBot/GPG sign secrets: https://github.com/morri-son/open-component-model/actions/runs/29498491185. Components produced:

ghcr.io/morri-son//ocm.software/ocm:0.7.0-rc.1
ghcr.io/morri-son//ocm.software/ocmcli:0.7.0-rc.1
ghcr.io/morri-son//ocm.software/kubernetes/controller:0.7.0-rc.1

.github/workflows/publish-components.yml
morri-son: changed

.github/workflows/publish-components.yml
morri-son: idea is to use the same version across bindings and workflow. The .env file is watched by renovate already. added a comment

.github/workflows/publish-components.yml
morri-son: correct, I thought in case we created a version and then need to restart workflow for the same release. But I removed "force" now.

jakobmoellerdev (approved):

Skarlso (approved):

PR #3185, opened 2026-07-24
feat(cli): auto-configure ocm from docker config on first startup

jakobmoellerdev:
## What

Adds a best-effort, first-startup auto configuration to the OCM CLI. When no OCM
configuration exists yet, the CLI autodetects a Docker config file and, if found,
writes an initial OCM config that wires it in as a credential repository so Docker
credentials are reused transparently.

## Why

New users typically already have working Docker credentials (`docker login`). Without
any OCM config they hit auth errors against OCI registries and must hand-write an
`.ocm/config`. This removes that first-run friction.

## How

`setup.AutoConfigure` runs in the persistent pre-run hook (after syscalls, before
config load) and:

- skips if `--config` was passed, if `OCM_DISABLE_AUTO_CONFIG` is truthy, or if any
  OCM config already exists in a known location (so it only runs on genuine first startup);
- detects a Docker config via `$DOCKER_CONFIG/config.json`, falling back to
  `$HOME/.docker/config.json`;
- writes a `generic.config.ocm.software/v1` document embedding a
  `credentials.config.ocm.software/v1` config with a `DockerConfig/v1` repository entry
  pointing at the detected file;
- writes to `$XDG_CONFIG_HOME/.ocm/config` or `$HOME/.ocm/config` (a location the normal
  lookup discovers on subsequent runs).

Failures are non-fatal — they log a warning and never block the command.

Generated config:

```yaml
configurations:
- repositories:
  - repository:
      dockerConfigFile: /home/user/.docker/config.json
      type: DockerConfig/v1
  type: credentials.config.ocm.software/v1
```

## Opt-out

Set `OCM_DISABLE_AUTO_CONFIG=true` or pass `--config`.

## Tests

`cli/cmd/setup/setup_autoconfig_test.go` covers: writing from `~/.docker`,
`XDG_CONFIG_HOME` target, `DOCKER_CONFIG` env detection, no-docker no-op, skip when a
config exists, disabled via env, and skip when `--config` is set.


cli/cmd/setup/setup_autoconfig.go
```diff
@@ -0,0 +1,195 @@
...
+		slog.String("config", target),
+		slog.String("dockerConfig", dockerConfigPath),
+	)
+
+	return nil
+}
+
+// detectDockerConfig locates a docker config.json using the standard docker resolution order:
+// $DOCKER_CONFIG/config.json first, then $HOME/.docker/config.json.
+func detectDockerConfig(s *ocmctx.Syscalls) (string, bool) {
```
matthiasbruns, line 103: We already have docker lookup in the code, don't we? At least we utilize `oras` `NewStoreFromDocker` for our oci docker lookup - can't we use this as well here?

matthiasbruns (commented):
I like the idea - just one question

cli/cmd/setup/setup_autoconfig.go
jakobmoellerdev: The way that the oras lookup works is implicit while the auto discover and setup would be explicit. I dont actually think docker config is pulled in by default but happy to stand corrected!

jakobmoellerdev: I think it came from a discussion during our recent hackathon so I assumed our default behavior ignored docker config

matthiasbruns: Our implementation forces the fields to be set - true

```go
func getStore(ctx context.Context, dockerConfig v1.DockerConfig) (remotecredentials.Store, error) {
	// Determine which store creation strategy to use based on the provided configuration
	switch {
	case dockerConfig.DockerConfigFile == "" && dockerConfig.DockerConfig == "":
		return createDefaultStore(ctx)
	case dockerConfig.DockerConfig != "":
		return createInlineConfigStore(ctx, dockerConfig.DockerConfig)
	case dockerConfig.DockerConfigFile != "":
		return createFileBasedStore(ctx, dockerConfig.DockerConfigFile)
	default:
		return nil, fmt.Errorf("invalid docker config: neither default, inline config, nor config file specified")
	}
}
```

but oras  `oras.land/oras-go/v2@v2.6.2/registry/remote/credentials/store.go:123`

```go
// NewStoreFromDocker returns a Store based on the default docker config file.
//   - If the $DOCKER_CONFIG environment variable is set,
//     $DOCKER_CONFIG/config.json will be used.
//   - Otherwise, the default location $HOME/.docker/config.json will be used.
//
// NewStoreFromDocker internally calls [NewStore].
//
// References:
//   - https://docs.docker.com/engine/reference/commandline/cli/#configuration-files
//   - https://docs.docker.com/engine/reference/commandline/cli/#change-the-docker-directory
func NewStoreFromDocker(opt StoreOptions) (*DynamicStore, error) {
	configPath, err := getDockerConfigPath()
	if err != nil {
		return nil, err
	}
	return NewStore(configPath, opt)
}

```

actually supports the same logic as you have added to the cli.

jakobmoellerdev: I think we should decide wether our default plugin behavior should just load from docker config. i believe in the past we didnt want it due to the side effects in testing.

cli/cmd/setup/setup_autoconfig.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+	if info, err := s.Stat(candidate); err != nil || info.IsDir() {
+		return "", false
+	}
+	return candidate, true
+}
+
+// defaultConfigWritePath returns the location the generated config is written to. It mirrors the
+// discovery logic so the file is found on subsequent runs: $XDG_CONFIG_HOME/.ocm/config if
+// XDG_CONFIG_HOME is set, otherwise $HOME/.ocm/config.
```
jneisener: Strictly speaking it does not mirror the discovery logic. It doesn't use `OCM_CONFIG` and it won't write to `$PWD` or `$EXE_DIR`. 

`$OCM_CONFIG` being set to a non-existing file on the first run is probably not a case that needs to be covered and no `$HOME` directory existing is also an edge case. Its slightly strange that we honor `$XDG_CONFIG_HOME` but not `$OCM_CONFIG`, though it makes sense in the first-run context.

But I think the `defaultConfigWritePath` should not be here, but rather in `configuration` as a sibling of `GetOCMConfigPaths`, because it partially duplicates the the lookup logic. WDYT?

jneisener (dismissed):
I have one question / opinion, but its pretty nitpicky.

Current state achieves the goal of improving the first time setup and I'm not sure the edgecases really matter, as the generated config will be merged into any other discovered config anyway

cli/cmd/setup/setup_autoconfig.go
```diff
@@ -0,0 +1,195 @@
...
+	credv1 "ocm.software/open-component-model/bindings/go/credentials/spec/config/v1"
+	ocicredsv1 "ocm.software/open-component-model/bindings/go/oci/spec/credentials/v1"
+	"ocm.software/open-component-model/bindings/go/runtime"
+	"ocm.software/open-component-model/cli/cmd/configuration"
+	ocmctx "ocm.software/open-component-model/cli/internal/context"
+)
+
+const (
+	// AutoConfigEnvVar disables the first-startup auto configuration when set to a truthy value.
+	AutoConfigEnvVar = "OCM_DISABLE_AUTO_CONFIG"
```
frewilhelm, line 24: actually, `autoConfigDisabled` returns even true on non-truthy values like `OCM_DISABLE_AUTO_CONFIG=asd`

```
OCM_DISABLE_AUTO_CONFIG=asd ./cli/tmp/bin/ocm --loglevel debug get cv asdasd
strconv.ParseBool: parsing "asd": invalid syntax
time=2026-08-06T14:04:09.362+02:00 level=DEBUG msg="auto configuration disabled via environment variable"
...
```

cli/cmd/setup/setup_autoconfig.go
```diff
@@ -0,0 +1,195 @@
...
+		return "", false
+	}
+	if info, err := s.Stat(candidate); err != nil || info.IsDir() {
+		return "", false
+	}
+	return candidate, true
+}
+
+// defaultConfigWritePath returns the location the generated config is written to. It mirrors the
+// discovery logic so the file is found on subsequent runs: $XDG_CONFIG_HOME/.ocm/config if
```
frewilhelm: I believe the logic might be wrong here because `XDGOCMConfigFileName` is actually `ocm/config` without the `.`

cli/cmd/setup/setup_autoconfig.go
jneisener: Talked about this in the warroom and decided that while correct, we don't care enough to block this PR while adjusting the configuration module

cli/cmd/setup/setup_autoconfig.go
jneisener: Yes, thats intentional. See https://github.com/open-component-model/open-component-model/pull/3185/changes#diff-b847963cdcbad56d3d1b6229f95bf13635d52293ac660b4cd6b8667ec95ac81bR194-R195

frewilhelm: ah, then i just go confused by that comment about "truthy" values. I expected something like `1, t, T, TRUE, true, True` (see [doc](https://pkg.go.dev/strconv#ParseBool))

jneisener: Fair point, how about
```go
// AutoConfigEnvVar disables the first-startup auto configuration when not empty and not set to a falsy value
```
?

frewilhelm (approved):
LGTM. Might be kind of hard to create an integration test for that functionality because it probably won't work locally as most developers have their configs already on default places. Do you have an idea how we could set something up?

jneisener:
> LGTM. Might be kind of hard to create an integration test for that functionality because it probably won't work locally as most developers have their configs already on default places. Do you have an idea how we could set something up?

Maybe `HOME=$(mktemp) XDG_CONFIG_HOME=$(mktemp) OCM_CONFIG= ocm get config`?

But I'm not sure if this functionality needs to be covered by an integration test

jneisener (approved):

cli/cmd/setup/setup_autoconfig.go
jneisener: @matthiasbruns I don't completely get what the remaining question is here. I hope its alright to resolve it.

If there is anything left to do here, ping me and I'll address it in a follow-up

PR #3201, opened 2026-07-27
feat(constructor-digest): TEST PR for verify user-supplied resource digest at construction

morri-son:
This draft PR contains changes to constructor and cli to be able to test the changes across modules. I'll close it once the single module PRs have been merged.

The PR lets an author pin the expected digest of an access-referenced resource directly in the component constructor. During `ocm add cv` the digest is carried into the descriptor before digest processing runs, so the existing verify path checks the downloaded content against it and fails on mismatch. The JSON schema already accepted `digest` on access-based resources — this wires it through the Go types and conversions.

- New optional `digest` field on constructor resources (spec + runtime + conversions)
- Warns when a resource carries a digest while `--skip-reference-digest-processing` is set (recorded but not verified)
- Round-trip + wget access verification tests

Follow-up: adopt this in the OCM release workflow's component constructor (`.github/components/ocm-release/component-constructor.yaml`), sourcing per-asset SHA-256 from the GitHub attestation subjects. Separate PR, since it needs a released CLI that understands the field.

### DIFF
--- a/bindings/go/constructor/runtime/constructor.go
+++ b/bindings/go/constructor/runtime/constructor.go
@@ -84,6 +84,9 @@ type Resource struct {
 	Relation ResourceRelation `json:"-"`
 	// Options bundles optional, construction-time directives for the resource.
 	Options ResourceOptions `json:"-"`
+	// Digest is the optional digest of the referenced resource.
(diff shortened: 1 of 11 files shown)


morri-son:
leading PR merged, so closing this

PR #3202, opened 2026-07-27
feat(constructor-digest): add verifiable resource digest field

### LINKED ISSUE #1199: Add an optional Digest in the resource spec to enable trust
**Description**
During the discussion here https://github.com/open-component-model/ocm-project/issues/1105#issuecomment-4929110684 we came to the conclusion that it might be worth adding an using digest in the wget resources during input to verify, that the blob we download can be trusted.
The `Digest` if set, must match the calculated digest during construction to enable trust in the wget method.
If if matches, we can continue as usual - the previously downloaded blob was verified and validated by the user and ocm made sure the digest is the same, so nothing changed in between.
If the digest does not match, we must error out and notify the user.

**Technical infos**

- use sha256 as everywhere in ocm
- tell users how to provide it

**Done Criteria**
- [x] Digest added to the spec
- [x] Digest comparison is working
- [x] Document how to calculate the digest so the user can use this hardening feature (done in #3205)
- [x] Enduser Documentation updated (done in #3205)
- [x] Internal technical Documentation created/updated (done in #3205)

morri-son:
## Description

Adds an optional `digest` field to the constructor resource so an author can pin the expected digest of an access-referenced resource. During `ocm add cv` the digest is carried into the descriptor before digest processing runs, so a configured digest processor verifies the referenced content against it and fails on mismatch.

Since the related issue spawns multiple modules, the following land in separate later PRs, not here:
- CLI flag help + skip-verification warning (`cli`), and the wget access verification integration test (`bindings/go/wget/integration`).
- Follow-up: Use new field in OCM release workflow component constructor.

Draft PR #3201, kept open as a test bed.


matthiasbruns (commented):
All good so far, but don't we have to validate the new `Digest` the same way we do for `reference.Digest` somewhere in https://github.com/open-component-model/open-component-model/blob/9fb2a8611bed0640d5d604395d0ec2af33e64b09/bindings/go/constructor/construct.go ?

morri-son:
@matthiasbruns 
> All good so far, but don't we have to validate the new `Digest` the same way we do for `reference.Digest` somewhere in https://github.com/open-component-model/open-component-model/blob/9fb2a8611bed0640d5d604395d0ec2af33e64b09/bindings/go/constructor/construct.go ?

validation happens, but at a different place than for reference.Digest. reference.Digest is checked in construct.go and the constructor calculates the component digest and compares it right there, because only the constructor knows how to digest a component. resource.Digest is checked by the digest-processor for that access type. The constructor calls ProcessResourceDigest. I only added that the new digest in the constructor really reaches that digest-processor. It was silently dropped before.

matthiasbruns:
> @matthiasbruns
> 
> > All good so far, but don't we have to validate the new `Digest` the same way we do for `reference.Digest` somewhere in https://github.com/open-component-model/open-component-model/blob/9fb2a8611bed0640d5d604395d0ec2af33e64b09/bindings/go/constructor/construct.go ?
> 
> validation happens, but at a different place than for reference.Digest. reference.Digest is checked in construct.go and the constructor calculates the component digest and compares it right there, because only the constructor knows how to digest a component. resource.Digest is checked by the digest-processor for that access type. The constructor calls ProcessResourceDigest. I only added that the new digest in the constructor really reaches that digest-processor. It was silently dropped before.

So we will need to update all bindings to use the new digest sooner or later right? 🥲

morri-son:
@matthiasbruns The Digest field in the constructor spec is for construction time only, kind of an input time trust anchor. It ends up in descriptor.Resource.Digest and is passed to the digest-processor ProcessResourceDigest.

Like the one for wget, all other existing ones, github, helm, oci already followed the same pattern: if resource.Digest is nil, compute and set it. if it is already set, verify the computed value against it. Before this PR the constructor digest was silently dropped during conversion, so it never reached the processor. That's the only fix in this PR, the final descriptor digest is unchanged and therefore my guess would be, that we don't require additional changes in the other bindings.

matthiasbruns (approved):

piotrjanik (approved):

Skarlso (approved):

PR #3205, opened 2026-07-28
chore(docs): wget reference docs

### LINKED ISSUE #1107: Website documentation for wget access and input type
**Description**
Write end-user documentation on the OCM website covering the `wget` access type and input type.

Documentation should cover:
- What the `wget` access type is and when to use it
- Full spec reference (fields: `url`, `mediaType`, `header`, `verb`, `body`, `noRedirect`)
- How to reference an HTTP/S resource in a component descriptor
- How to use the `wget` input type in `component-constructor.yaml`
- How to configure HTTP behaviour (`http.config.ocm.software`) — timeouts, retries, TLS
- How credentials work with wget (Basic Auth, Bearer token, mTLS) — cross-reference credential docs
- Migration guide from OCM v1 `wget` access type

Part of EPIC #958. Should be written after #1103 and #1104 are merged.

**Done Criteria**
- [x] wget access type reference page published on website
- [x] wget input type documented with `component-constructor.yaml` example
- [x] HTTP config cross-reference added
- [x] Credential usage documented
- [x] OCM v1 → v2 migration note included

matthiasbruns:
This PR adds `wget` access reference docs to our website.
It also adds identity to wget (currently docs and runtime parsing only as in helm and oci) and fixes the missing filesystem config wiring for `input`

To ship the new `scheme` files, we will need to update the `wget` dependency in `cli` as well or wait for renovate.

The docs also cover the changes in https://github.com/open-component-model/open-component-model/pull/3212

Fixes: https://github.com/open-component-model/ocm-project/issues/1107


matthiasbruns:
@coderabbitai review again

website/content/docs/reference/credential-consumer-identities.md
```diff
@@ -177,6 +178,124 @@ identity is derived from the Helm repository URL using the same URL-based attrib
...
+  credentials:
+    - type: WgetCredentials/v1
+      certificate: |
+        -----BEGIN CERTIFICATE-----
+        MIIDdzCCAl+gAwIBAgIEbGVnYWw...
+        -----END CERTIFICATE-----
+      privateKey: |
+        -----BEGIN PRIVATE KEY-----
+        MIIEvQIBADANBgkqhkiG9w0BAQ...
+        -----END PRIVATE KEY-----
```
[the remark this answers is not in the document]
matthiasbruns: @coderabbitai - what about the existing docs? we have key examples everywhere dont we?

website/content/docs/concepts/resource-repositories.md
```diff
@@ -57,7 +60,8 @@ Each resource repository implements three operations:
 2. **Download**: fetch the resource content as a blob.
 3. **Upload**: push resource content to the backend (not all backends support this).
 
-OCM ships with built-in resource repositories for OCI artifacts and Helm charts. For the full technical details (
+OCM ships with built-in resource repositories for OCI artifacts, Helm charts, and content served over
```
piotrjanik: Are we going to list each repo type here? It will get lengthy at some point. Maybe we should just mention that there are built-in types and inform the user where to find them

website/content/docs/reference/credential-consumer-identities.md
```diff
@@ -177,6 +178,124 @@ identity is derived from the Helm repository URL using the same URL-based attrib
 
 ---
 
+## Wget
```
piotrjanik, line 181: nit: Maybe it's worth considering creating a separate subpage for each type? I mean reference for Wget, OCI, etc.

feel free to resolve this one.

matthiasbruns: not yet :) but worth doing - can you create an issue?

website/content/docs/concepts/resource-repositories.md
matthiasbruns: Yeah will generalize this

website/content/docs/how-to/add-resources-from-http-urls.md
```diff
@@ -0,0 +1,485 @@
...
+username/password pair, both versions behave identically. Split such entries if the server distinguishes the two
+mechanisms.
+
+## Troubleshooting
+
+### Symptom: `401 Unauthorized` from the origin server
+
+**Cause:** No consumer entry matched, so the request went out unauthenticated. The identity type is matched by exact
+string and is unversioned, so neither `Wget/v1` nor the lowercase `wget` used by OCM v1 will match.
+
```
morri-son, line 211:
```suggestion
Another silent cause: an entry that sets only certificateAuthority without certificate. In OCM v2, certificateAuthority is only evaluated alongside a client certificate. It is not applied for server-side certificate verification on its own.

```

website/content/docs/how-to/add-resources-from-http-urls.md
```diff
@@ -0,0 +1,485 @@
...
+### Set the media type explicitly
+
+The media type of the resulting blob is resolved in this order:
+
+1. The `mediaType` field of the specification, if set.
+2. The `Content-Type` header of the HTTP response.
+3. `application/octet-stream`.
+
+{{< callout context="note" >}}
+Unlike OCM v1, the media type is **not** derived from the URL's file extension. Set `mediaType` whenever the server does
```
morri-son: I would remove the "Unlike in OCM v1..." sentence as this is covered in the migration table in line 382. The how-to targets V2 users, so the callout does not need to mention v1.

morri-son (dismissed):

website/content/docs/how-to/add-resources-from-http-urls.md
```diff
@@ -0,0 +1,484 @@
...
+{{< tab "Input type" >}}
+
+```yaml
+resources:
+  - name: release-archive
+    type: blob
+    version: 1.0.0
+    input:
+      type: Wget/v1
+      url: https://downloads.example.com/myapp/1.0.0/myapp-linux-amd64.tar.gz
```
frewilhelm: I believe a real-life example would make sense. For example an URL to our releases `https://github.com/open-component-model/open-component-model/releases/download/v0.12.0/cli.tar`

website/content/docs/how-to/add-resources-from-http-urls.md
```diff
@@ -0,0 +1,484 @@
...
+## Troubleshooting
+
+### Symptom: `401 Unauthorized` from the origin server
+
+**Cause:** No consumer entry matched, so the request went out unauthenticated. The identity type is matched by exact
+string and is unversioned, so neither `Wget/v1` nor the lowercase `wget` used by OCM v1 will match.
+Another silent cause: an entry that sets only certificateAuthority without certificate. In OCM v2, certificateAuthority is only evaluated alongside a client certificate. It is not applied for server-side certificate verification on its own.
+
+**Fix:** Set `type: Wget` in the consumer identity, and drop `pathprefix`. `path` is matched with `path.Match`, whose
+`*` does not cross `/`, so it cannot express a prefix over nested paths. Omitting it matches every path on the host:
```
frewilhelm: I don't understand where `pathprefix` is coming from. If the symptom is `401 Unauthorized` it might be incorrect credentials or no identity was matching. In the latter case, I think I would rather point to the credential-reference section for `wget` and telling the user to make sure it matches instead of listing arbitrary failure-causes (at least it seems arbitrary to me :D)

website/content/docs/how-to/add-resources-from-http-urls.md
frewilhelm, line 1: I believe most of our How-Tos give the user a reproducible task they can follow. This How-To feels more like a general help-page what is possible with the `wget` resource type. Would it make sense to think about a real scenario and lead the user through? WDYT?

website/content/docs/reference/input-and-access-types.md
```diff
@@ -265,6 +326,61 @@ resources:
...
+      type: Wget/v1
+      url: https://downloads.example.com/myapp/1.0.0/myapp-linux-amd64.tar.gz
+      mediaType: application/x-tar+gzip
+```
+
+{{< callout context="note" >}}
+Upload is not supported for this access type, because a plain HTTP endpoint has no standardized write API. Wget
+resources are **always transferred by value**: transferring a component version downloads the content and stores it as a
+[`LocalBlob/v1`]({{< relref "input-and-access-types.md" >}}#localblobv1) in the target repository, so the target no
+longer depends on the origin server.
```
frewilhelm: I don't understand this. If I have a component descriptor with a `wget` access and I transfer it to a registry, I don't see a download and no `LocalBlob`. It just transfers the component descriptor. What do you mean by that?

website/content/docs/how-to/add-resources-from-http-urls.md
matthiasbruns: `pathprefix` is ocmv1 - since we mention the `ocmv1` diff and migration, I think this is worth listing

website/content/docs/how-to/add-resources-from-http-urls.md
matthiasbruns: Maybe I should rewrite this more like "migration from ocmv1 wget"

website/content/docs/reference/input-and-access-types.md
matthiasbruns: https://github.com/open-component-model/open-component-model/blob/a437b47405738f612eb95aba1bc67239f96ad7b2/bindings/go/transfer/internal/wget.go#L18

https://github.com/open-component-model/open-component-model/blob/a437b47405738f612eb95aba1bc67239f96ad7b2/bindings/go/transfer/internal/graph.go#L325

website/content/docs/how-to/add-resources-from-http-urls.md
matthiasbruns: split it into a general 401 and a ocmv1 migration solution

website/content/docs/how-to/add-resources-from-http-urls.md
matthiasbruns: but I do not want to :D

website/content/docs/reference/input-and-access-types.md
frewilhelm: https://github.com/open-component-model/open-component-model/blob/a437b47405738f612eb95aba1bc67239f96ad7b2/bindings/go/transfer/internal/graph.go#L254

morri-son: I guess the confusion simply comes from the fact that **always transferred by value** is used here  The callout in line 54 in the new how-to makes a distinction between transfer and transfer with --copy-resources, but this is missed here. What about 

```
Upload is not supported for this access type: a plain HTTP endpoint has no standardized write API. A `Wget/v1` access therefore has no by-reference form in a target repository. It is copied only when resource copying is requested using `--copy-resources`, and then always by value.  The content is downloaded and stored as a [`LocalBlob/v1`]({{< relref "input-and-access-types.md" >}}#localblobv1).
```

?

website/content/docs/how-to/add-resources-from-http-urls.md
```diff
@@ -0,0 +1,583 @@
...
+`mediaType` is set explicitly because GitHub serves every release asset as `application/octet-stream`. See
+[Setting the media type](#setting-the-media-type) for the full resolution order.
+
+{{< callout context="caution" >}}
+**Never put credentials in `url`, `header`, or `body`.** That includes userinfo (`https://user:token@host/...`) and
+presigned query parameters. Anything that is part of the specification is stored, not just headers.
+
+An **access** specification is stored verbatim in the component descriptor: everything written there is persisted with
+the component version, travels with it through every transfer, is covered by its signature, and is readable by anyone
+who can read the component version. An **input** specification is resolved at construction time and never reaches the
```
morri-son:
```suggestion
who can read the component version.  

An **input** specification is resolved at construction time and never reaches the
```

morri-son: I would add a new line in between input an access type. The callout is anyway quite long so should be fine.

website/content/docs/how-to/add-resources-from-http-urls.md
```diff
@@ -0,0 +1,583 @@
...
+descriptor, but it does live in your `component-constructor.yaml`, which is normally checked into version control.
```
morri-son:
```suggestion
descriptor, but it does live in your `component-constructor.yaml`, which is normally checked into version control.  

```

morri-son: same here. The generic part with the authentication config should be split from the access type section.

website/content/docs/how-to/add-resources-from-http-urls.md
```diff
@@ -0,0 +1,583 @@
...
+
+{{< /tab >}}
+{{< /tabs >}}
+
+`mediaType` is set explicitly because GitHub serves every release asset as `application/octet-stream`. See
+[Setting the media type](#setting-the-media-type) for the full resolution order.
+
+{{< callout context="caution" >}}
+**Never put credentials in `url`, `header`, or `body`.** That includes userinfo (`https://user:token@host/...`) and
+presigned query parameters. Anything that is part of the specification is stored, not just headers.
```
frewilhelm: What is meant by "not just headers"?

website/content/docs/how-to/add-resources-from-http-urls.md
```diff
@@ -0,0 +1,583 @@
...
+
+1. The `mediaType` field of the specification, if set.
+2. The `Content-Type` header of the HTTP response.
+3. `application/octet-stream`.
+
+Set `mediaType` whenever the server does not send a useful `Content-Type`. Otherwise the resource ends up as
+`application/octet-stream`. GitHub release assets are a common case: they are served as `application/octet-stream`
+regardless of what they contain, which is why the walkthrough sets `mediaType: application/x-tar` explicitly.
+
+## Authenticating against a protected server
```
frewilhelm: IMO we should not have a whole section about this because (a) it looks like its own little how-to and (b) it looks like the usual credential handling of ocm. I would just add a call-out and refer the respective `wget` credential reference section.

website/content/docs/how-to/add-resources-from-http-urls.md
```diff
@@ -0,0 +1,583 @@
...
+When the media type is an archive format the CLI can unpack, `--output` is a **directory** holding the extracted
+contents. Other media types are written to a file at that path. `cli.tar` is a `.tar`, so `./ocm-cli` is a directory
+here; its contents happen to be an OCI image layout, because that is what this particular release asset packages.
+{{< /callout >}}
+
+{{< /step >}}
+
+{{< /steps >}}
+
+## Setting the media type
```
frewilhelm: This could also rather be a callout than its own section

website/content/docs/how-to/add-resources-from-http-urls.md
```diff
@@ -0,0 +1,583 @@
...
+HTTP Basic Auth, bearer tokens, and mutual TLS are supported. Basic Auth and a bearer token both set the
+`Authorization` header and are mutually exclusive. When both are configured, the bearer token wins and a warning is
+logged. A client certificate combines with either, but only takes effect during a TLS handshake.
+
+See [Credential Consumer Identities: Wget]({{< relref "docs/reference/credential-consumer-identities.md#wget" >}}) for
+the identity attributes and matching rules, and
+[Credential Types: WgetCredentials/v1]({{< relref "docs/reference/credential-types.md#wgetcredentialsv1" >}}) for the
+full field reference.
+
+## Pinning an expected digest
```
frewilhelm: I believe it would make sense to put this above in the How-To. This is a best-practice that users should be aware of (putting it down here, makes it likely to be skipped)

website/content/docs/how-to/add-resources-from-http-urls.md
frewilhelm: This How-To contains the several sections:
- The actual steps to "Add resources from HTTP URLs"
- Setting the media type
- Authenticating against a protected server
- Pinning an expected digest
- Sending a non-Get request
- Tuning the download
- Migrating from OCM v1

If you have a look at the other How-Tos on our website, then they usually have
- Steps
- Rarely "Tips"
- Rarely another section explaining the concept/background

I think a lot of the stuff you have written down here is important but it is a bit to much for a How-To and it is loosely coupled. For a user who just wants to use that resource and see a small example on how to do so, it feels too much.

Maybe a tutorial would be a better idea, where we can show why it makes sense to pin the expected digest and configure the download etc.pp while following the steps.

morri-son: I second @frewilhelm comment. This is way too much for a simple hopw-to. We should reduce the content to just the current steps plus the media type thingy as this required for the specified resource. I would even cur out the input type and just mention input types and reference the input-and-access-types reference doc. All the other details and the input type usage should go into a tutorial which could look pretty much the same as this current how-to.

Skarlso:
@matthiasbruns Taking this over to push it through the finish-line. :)

matthiasbruns:
> @matthiasbruns Taking this over to push it through the finish-line. :) 

Still cannot access slack but if you have questions ping me on my non-sap email

Skarlso:
@frewilhelm @morri-son I dismissed your comments so you can start with a clean slate. :D

website/content/docs/how-to/add-resources-from-http-urls.md
```diff
@@ -0,0 +1,282 @@
...
+using the `Wget/v1` type.
+
+OCM has two options where the difference is only the end location:
+
+- The [**input type**]({{< relref "docs/reference/input-and-access-types.md#wgetv1-input" >}}) downloads the file while
+  the component version is built and stores it as a localBlob in the component version.
+- The [**access type**]({{< relref "docs/reference/input-and-access-types.md#wgetv1-access" >}}) stores only the
+  **URL** and leaves the file on the remote server.
+
+Both use the same download and credential code, so everything below applies to either one. For the reasoning behind
```
morri-son:
```suggestion
Both use the same download and credential code, so everything below applies to either one. For the full set of options and the reasoning behind them, see the tutorial
```

website/content/docs/how-to/add-resources-from-http-urls.md
```diff
@@ -0,0 +1,282 @@
...
+- That same resource downloaded back out, to confirm the round trip works
+
+**Estimated time:** ~10 minutes
+
+## Prerequisites
+
+- [OCM CLI]({{< relref "docs/getting-started/ocm-cli-installation.md" >}}) installed
+- A working directory for the constructor and the transport archive
+
+This guide fetches a public OCM release asset, so it runs end to end with no registry and no credentials. To point it
```
morri-son:
```suggestion
This guide fetches a public OCM release asset, so it runs end-to-end with no registry access or credentials required. To point it
```

website/content/docs/how-to/add-resources-from-http-urls.md
```diff
@@ -0,0 +1,282 @@
...
+at a server that needs authentication, see [Authenticate against a protected server](#authenticate-against-a-protected-server).
+
+## Steps
+
+{{< steps >}}
+
+{{< step >}}
+
+### Choose the input type or the access type
```
morri-son:
```suggestion
### Choose input or access type
```

website/content/docs/how-to/add-resources-from-http-urls.md
```diff
@@ -0,0 +1,282 @@
...
+
+| Choose the **input type** when                     | Choose the **access type** when                    |
```
morri-son:
```suggestion
| Choose **input type** when                     | Choose **access type** when                    |
```

website/content/docs/how-to/add-resources-from-http-urls.md
```diff
@@ -0,0 +1,282 @@
...
+|----------------------------------------------------|----------------------------------------------------|
+| The file must stay reproducible and work offline   | The URL is the source of truth and should stay so  |
+| The URL might disappear or change what it serves   | The file is large and you don't want to copy it    |
+| You are building an air-gapped repository          | Consumers should fetch directly from the origin    |
+
+Pick either one to follow this guide.
+
+{{< callout context="caution" >}}
+A Wget resource is transferred **by value**. When you move the component version with `ocm transfer cv`, pass
```
morri-son:
```suggestion
A Wget resource is transferred **by value**. When you move the component version with `ocm transfer cv` and pass the 
```

website/content/docs/how-to/add-resources-from-http-urls.md
morri-son: just to not repeat the "choose" from the step title

website/content/docs/how-to/add-resources-from-http-urls.md
```diff
@@ -0,0 +1,282 @@
...
+
+### Choose the input type or the access type
+
+| Choose the **input type** when                     | Choose the **access type** when                    |
+|----------------------------------------------------|----------------------------------------------------|
+| The file must stay reproducible and work offline   | The URL is the source of truth and should stay so  |
+| The URL might disappear or change what it serves   | The file is large and you don't want to copy it    |
+| You are building an air-gapped repository          | Consumers should fetch directly from the origin    |
+
+Pick either one to follow this guide.
```
morri-son:
```suggestion
```

morri-son: move this up ^^ right behind the step title

website/content/docs/how-to/add-resources-from-http-urls.md
morri-son:
```suggestion
### Choose the input type or the access type

Pick either one to follow this guide:
```

website/content/docs/how-to/add-resources-from-http-urls.md
```diff
@@ -0,0 +1,282 @@
...
+[Transfer behavior]({{< relref "docs/tutorials/wget-http-resources.md#how-transfer-works" >}}) for more details.
+{{< /callout >}}
+
+{{< /step >}}
+
+{{< step >}}
+
+### Create the component constructor
+
+Both spellings describe the same resource with the same fields. Write one of them to `component-constructor.yaml`:
```
morri-son:
```suggestion
Both spellings below describe the same resource with the same fields. Write one of them to `component-constructor.yaml`:
```

website/content/docs/how-to/add-resources-from-http-urls.md
```diff
@@ -0,0 +1,282 @@
...
+
+{{< /tab >}}
+{{< /tabs >}}
+
+Set `mediaType` explicitly here because GitHub serves every release asset as `application/octet-stream`. If you leave
+it out, OCM falls back to the server's `Content-Type`, then to `application/octet-stream`. The full order is in
+[Set the media type]({{< relref "docs/tutorials/wget-http-resources.md#set-the-media-type" >}}).
+
+{{< callout context="caution" >}}
+**Never put credentials in `url`, `header`, or `body`.** That includes userinfo (`https://user:token@host/...`) and
```
morri-son:
```suggestion
**Never put credentials in `url`, `header`, or `body`.** That includes user info (`https://user:token@host/...`) and
```

website/content/docs/how-to/add-resources-from-http-urls.md
```diff
@@ -0,0 +1,282 @@
...
+presigned query parameters. An access specification is stored in the component descriptor, signed, and is included in
```
morri-son:
```suggestion
presigned query parameters. An access specification is stored in the component descriptor then signed and finally included in
```

website/content/docs/how-to/add-resources-from-http-urls.md
```diff
@@ -0,0 +1,282 @@
...
+the transfer. An input specification lives in your constructor file, which is usually checked into version control.
+Either way the secret leaks. Use the credential system instead, as shown in
```
morri-son:
```suggestion
Either way the secret leaks. Use the [credential system]({{< relref "docs/concepts/credential-system.md" >}}) instead, as shown in
```

website/content/docs/how-to/add-resources-from-http-urls.md
```diff
@@ -0,0 +1,282 @@
...
+{{< /callout >}}
+
+{{< /step >}}
+
+{{< step >}}
+
+### Build the component version
+
+```bash
+ocm add cv --repository ./transport-archive --constructor component-constructor.yaml
```
morri-son: I guess with our defaults, we can just do an `ocm add cv` and skip the params

website/content/docs/how-to/add-resources-from-http-urls.md
```diff
@@ -0,0 +1,282 @@
...
+
+In both cases, the resource has the same digest, computed over the bytes that were fetched.
+
+{{< /step >}}
+
+{{< step >}}
+
+### Download the resource back
+
+This proves the transport (and, on a protected server, the credentials) work end to end:
```
morri-son:
```suggestion
This proves the transport (and, on a protected server, the credentials) work end-to-end:
```

website/content/docs/how-to/add-resources-from-http-urls.md
```diff
@@ -0,0 +1,282 @@
...
+Other media types are written to a file at that path. `cli.tar` is a `.tar`, so `./ocm-cli` is a directory here.
+{{< /callout >}}
+
+{{< /step >}}
+
+{{< /steps >}}
+
+## Authenticate against a protected server
+
+The release asset above is public. When the URL is behind authentication (an artifact repository, or a private
```
morri-son:
```suggestion
The release asset above is public. When accessing the URL requires authentication (an artifact repository, or a private
```

website/content/docs/how-to/add-resources-from-http-urls.md
```diff
@@ -0,0 +1,282 @@
...
+GitHub asset), configure the credentials instead of putting them in the specification. OCM matches them to the request
```
morri-son:
```suggestion
GitHub asset), [configure credentials]({{< relref "docs/how-to/configure-multiple-credentials.md" >}}) of type `WgetCredentials` instead of putting them in the specification. OCM matches them to the request
```

website/content/docs/how-to/add-resources-from-http-urls.md
```diff
@@ -0,0 +1,282 @@
...
+          hostname: downloads.example.com
+          scheme: https
+        credentials:
+          - type: WgetCredentials/v1
+            username: my-user
+            password: my-password
+EOF
+```
+
+Then pass `--config` when you build (or drop the file in one of the
```
morri-son:
```suggestion
Then pass `--config` when you build the component (or drop the file in one of the
```

website/content/docs/how-to/add-resources-from-http-urls.md
```diff
@@ -0,0 +1,282 @@
...
+[well-known locations]({{< relref "docs/reference/ocm-cli/ocm.md" >}}) the CLI reads automatically):
+
+```bash
+ocm --config .ocmconfig add cv \
+  --repository ./transport-archive \
+  --constructor component-constructor.yaml
+```
+
+HTTP Basic Auth, bearer tokens, and mutual TLS are all supported. For how the identity is matched and how the auth
+methods interact, see [Authentication]({{< relref "docs/tutorials/wget-http-resources.md#authentication" >}}) in the
```
morri-son:
```suggestion
methods interact, see the [Authentication]({{< relref "docs/tutorials/wget-http-resources.md#authentication" >}}) section in the
```

website/content/docs/tutorials/wget-http-resources.md
```diff
@@ -0,0 +1,320 @@
+---
+title: "Work with HTTP Resources"
```
morri-son:
```suggestion
title: "Working with HTTP Resources"
```

morri-son: to be aligned with the other tutorials

website/content/docs/how-to/add-resources-from-http-urls.md
```diff
@@ -0,0 +1,282 @@
...
+weight: 17
+toc: true
+---
+
+## Goal
+
+Add a file served over plain HTTP or HTTPS (a release archive, a checksum, a signed binary) to a component version
+using the `Wget/v1` type.
+
+OCM has two options where the difference is only the end location:
```
morri-son:
```suggestion
OCM provides two types for HTTP resources. They differ in when the file is fetched and where the bytes end up:
```

website/content/docs/how-to/add-resources-from-http-urls.md
```diff
@@ -0,0 +1,282 @@
...
+  **URL** and leaves the file on the remote server.
+
+Both use the same download and credential code, so everything below applies to either one. For the reasoning behind
+each option and the full set of variable, see the tutorial
+[Work with HTTP Resources]({{< relref "docs/tutorials/wget-http-resources.md" >}}).
+
+## You'll end up with
+
+- A component version containing a resource fetched over HTTP or HTTPS, either embedded as a local blob or referenced by URL
+- That same resource downloaded back out, to confirm the round trip works
```
morri-son, line 27:
```suggestion
- The same resource retrieved from the archive, verifying the round trip end to end.
```

website/content/docs/how-to/add-resources-from-http-urls.md
```diff
@@ -0,0 +1,282 @@
...
+- The [**access type**]({{< relref "docs/reference/input-and-access-types.md#wgetv1-access" >}}) stores only the
+  **URL** and leaves the file on the remote server.
+
+Both use the same download and credential code, so everything below applies to either one. For the reasoning behind
+each option and the full set of variable, see the tutorial
+[Work with HTTP Resources]({{< relref "docs/tutorials/wget-http-resources.md" >}}).
+
+## You'll end up with
+
+- A component version containing a resource fetched over HTTP or HTTPS, either embedded as a local blob or referenced by URL
```
morri-son:
```suggestion
- A component version containing a resource fetched over HTTP or HTTPS, either embedded as a local blob or referenced by URL.
```

website/content/docs/tutorials/wget-http-resources.md
```diff
@@ -0,0 +1,320 @@
...
+OCM can add a file served over HTTP or HTTPS to a component version using the `Wget/v1` type. There is one download
+engine underneath, with two front doors:
+
+- The **input type** downloads the file while the component version is built and stores the bytes inside it.
+- The **access type** stores only the URL and leaves the bytes on the remote server.
+
+They share the same download, credential, and configuration code, so the choice between them is about *where the bytes
+live*, not about what you can configure.
+
+If you just want the commands to add a file and download it back, start with the how-to
```
morri-son:
```suggestion
If you just want the commands to add a file and download it back, start with the How-To guide 
```

website/content/docs/tutorials/wget-http-resources.md
```diff
@@ -0,0 +1,320 @@
...
+[Add Resources from HTTP URLs]({{< relref "docs/how-to/add-resources-from-http-urls.md" >}}). This tutorial explains
+the parts that guide leaves out: why you'd pick one type over the other, how credentials are matched, how the media
+type is decided, how digest pinning protects you, and what changed since OCM v1.
+
+## What You'll Learn
+
+- When to embed a file (input type) and when to reference it by URL (access type)
+- What happens to a Wget resource when you transfer the component version
+- How OCM decides a resource's media type
```
morri-son:
```suggestion
- How OCM resolves a resource's media type
```

website/content/docs/tutorials/wget-http-resources.md
```diff
@@ -0,0 +1,320 @@
...
+- How credentials are matched to a request, and how the three authentication methods interact
+- How to pin a digest so a changed file is rejected instead of silently accepted
```
morri-son:
```suggestion
- How to pin a resource by digest so a changed file is rejected instead of silently accepted
```

website/content/docs/tutorials/wget-http-resources.md
```diff
@@ -0,0 +1,320 @@
...
+
+## Prerequisites
+
+- [OCM CLI]({{< relref "docs/getting-started/ocm-cli-installation.md" >}}) installed
+- Comfortable adding a component version with `ocm add cv`; the
+  [how-to]({{< relref "docs/how-to/add-resources-from-http-urls.md" >}}) is a good warm-up
+
+## Input or access: where the bytes live {#choosing-input-or-access}
+
+Both describe the same resource with the same fields. The only difference is *when* the file is fetched and
```
morri-son:
```suggestion
Both types describe the same resource with the same fields. The only difference is *when* the file is fetched and
```

website/content/docs/tutorials/wget-http-resources.md
```diff
@@ -0,0 +1,320 @@
...
+The rule of thumb is: use **input type** when you want a reproducible copy, and the **access type** when the URL
+is the authoritative location and should be used for accessing the resource.
+
+## How transfer works {#how-transfer-works}
+
+A Wget resource is always transferred by value.
+
+When you run `ocm transfer cv`, an access-type resource does not stay a `Wget/v1` access in the target. OCM fetches the
+bytes and writes them into the target as a [`LocalBlob/v1`]({{< relref "docs/reference/input-and-access-types.md#localblobv1" >}}). After a transfer
+both will end up as local blobs.
```
morri-son, line 62: what is meant by "both"?

website/content/docs/tutorials/wget-http-resources.md
```diff
@@ -0,0 +1,320 @@
...
+
+This means:
+
+1. You MUST pass `--copy-resources true` to `ocm transfer cv`. Without it, the resource is skipped, because there is no
```
morri-son:
```suggestion
1. You MUST pass the `--copy-resources` flag to `ocm transfer cv`. Without it, the resource is skipped, because there is no
```

website/content/docs/how-to/add-resources-from-http-urls.md
```diff
@@ -0,0 +1,282 @@
...
+|----------------------------------------------------|----------------------------------------------------|
+| The file must stay reproducible and work offline   | The URL is the source of truth and should stay so  |
+| The URL might disappear or change what it serves   | The file is large and you don't want to copy it    |
+| You are building an air-gapped repository          | Consumers should fetch directly from the origin    |
+
+Pick either one to follow this guide.
+
+{{< callout context="caution" >}}
+A Wget resource is transferred **by value**. When you move the component version with `ocm transfer cv`, pass
+`--copy-resources true`; the `Wget/v1` access is then replaced by a
```
morri-son:
```suggestion
`--copy-resources` flag; the `Wget/v1` access is then replaced by a
```

website/content/docs/tutorials/wget-http-resources.md
```diff
@@ -0,0 +1,320 @@
...
+   way to transfer it without copying the bytes.
+2. The bytes are fetched *again at transfer time* and checked against the resource's digest. If the file behind the URL
+   changed since the component version was built, the transfer fails instead of copying different content. That
+   check is what makes [digest pinning](#pin-a-digest) worth doing on an access-type resource.
+
+"Keep the bytes on the server" property of the access type is only true if the component version stays where it
+was built. During a transfer it will end up as an embedded blob.
```
morri-son:
```suggestion
The file stays on the remote server only if the component version is never transferred. Transfer converts the `Wget/v1` reference into a `LocalBlob/v1`.
```

website/content/docs/how-to/add-resources-from-http-urls.md
```diff
@@ -0,0 +1,282 @@
...
+
+OCM has two options where the difference is only the end location:
+
+- The [**input type**]({{< relref "docs/reference/input-and-access-types.md#wgetv1-input" >}}) downloads the file while
+  the component version is built and stores it as a localBlob in the component version.
+- The [**access type**]({{< relref "docs/reference/input-and-access-types.md#wgetv1-access" >}}) stores only the
+  **URL** and leaves the file on the remote server.
+
+Both use the same download and credential code, so everything below applies to either one. For the reasoning behind
+each option and the full set of variable, see the tutorial
```
morri-son:
```suggestion
Both types share the same download and credential code, so the rest of this guide applies to either one. For the full set of options and the reasoning behind them, see the tutorial 
```

website/content/docs/reference/resource-repositories.md
```diff
@@ -128,6 +128,75 @@ resolves the OCI manifest digest via the registry API.
...
+### Capabilities
+
+| Operation         | Supported |
+|-------------------|-----------|
+| Download          | Yes       |
+| Upload            | No        |
+| Digest Processing | Yes       |
+
+{{< callout context="note" >}}
+Upload is not supported because a plain HTTP endpoint has no standardized write API. Wget resources are therefore always
```
morri-son:
```suggestion
Upload is not supported because a plain HTTP endpoint has no standardized write API. `Wget` resources are therefore always
```

website/content/docs/tutorials/wget-http-resources.md
```diff
@@ -0,0 +1,320 @@
...
+
+## Set the media type {#set-the-media-type}
+
+Every resource has a media type. For a Wget resource, OCM picks it in this order and stops at the first one it finds:
```
morri-son:
```suggestion
Every resource has a media type. For a Wget resource, OCM picks it in the following order and stops at the first one it finds:
```

website/content/docs/tutorials/wget-http-resources.md
```diff
@@ -0,0 +1,320 @@
...
+
+1. The `mediaType` field in your specification, if you set it.
+2. The `Content-Type` header the server returns.
+3. `application/octet-stream`, as a last resort.
+
+Practical advice is: **set `mediaType` yourself whenever the server doesn't return a useful `Content-Type`.**
```
morri-son:
```suggestion
**Practical advice**: set `mediaType` yourself whenever the server doesn't return a useful `Content-Type`.
```

website/content/docs/tutorials/wget-http-resources.md
```diff
@@ -0,0 +1,320 @@
...
+The most common example is GitHub release assets. GitHub serves all of them as `application/octet-stream` no matter what
```
morri-son:
```suggestion
The most common examples are GitHub release assets. GitHub serves all of them as `application/octet-stream`, no matter what
```

website/content/docs/tutorials/wget-http-resources.md
```diff
@@ -0,0 +1,320 @@
...
+application/x-tar` explicitly.
+
+{{< callout context="note" >}}
+OCM v2 does **not** guess the media type from the file extension in the URL. OCM v1 did. See
+[Migrate from OCM v1](#migrating-from-ocm-v1) if you are moving old constructor files across.
+{{< /callout >}}
+
+## Authentication {#authentication}
+
+Credentials are resolved through OCM's normal credential system. Never add authentication to a URL directly. OCM
```
morri-son:
```suggestion
Credentials are resolved through OCM's normal [credential system]({{< relref "docs/concepts/credential-system.md" >}}). Never add authentication to a URL directly. OCM
```

website/content/docs/tutorials/wget-http-resources.md
```diff
@@ -0,0 +1,320 @@
...
+
+`WgetCredentials/v1` supports three methods:
+
+| Method                   | Fields                                                           | Sets                      |
+|--------------------------|------------------------------------------------------------------|---------------------------|
+| HTTP Basic Auth          | `username`, `password`                                           | `Authorization: Basic …`  |
+| Bearer token             | `identityToken`                                                  | `Authorization: Bearer …` |
+| Mutual TLS (client cert) | `certificate`, `privateKey`, and optional `certificateAuthority` | The TLS handshake         |
+
+- **Basic Auth and a bearer token both set the `Authorization` header, so they are mutually exclusive.** If you
```
morri-son:
```suggestion
- Basic Auth and a bearer token both set the `Authorization` header, **so they are mutually exclusive.** If you
```

website/content/docs/tutorials/wget-http-resources.md
```diff
@@ -0,0 +1,320 @@
...
+  configure both, the bearer token wins and OCM logs a warning.
+- **The client certificate is separate.** It is applied during the TLS handshake, not in a header, so it works independent 
+  of the other two.
+
+{{< callout context="caution" >}}
+Put secrets in your credential configuration, never in the specification. Anything you write into `url`, `header`, or
+`body` (including `https://user:token@host/...` and presigned query parameters) is stored with the component version
+(access type) or lives in your constructor file (input type). Both are places a secret should never be at.
```
morri-son:
```suggestion
(access type) or lives in your constructor file (input type). Neither is a safe place for a secret.
```

website/content/docs/tutorials/wget-http-resources.md
```diff
@@ -0,0 +1,320 @@
...
+```
+
+The `digest` block is optional and works on both the input and the access type. If set, these are the exact fields:
+
+- `hashAlgorithm` MUST be `SHA-256`. It is the only algorithm the Wget resource repository computes. `sha256` and
+  `SHA256` do not match.
+- `normalisationAlgorithm` MUST be `genericBlobDigest/v1`, which means "hash the raw bytes, no normalisation".
+- `value` is the bare lowercase hex digest, with **no** `sha256:` prefix.
+
+### Why access-type pinning {#why-access-type-pinning
```
morri-son:
```suggestion
### Why access-type pinning {#why-access-type-pinning}
```

website/content/docs/tutorials/wget-http-resources.md
```diff
@@ -0,0 +1,320 @@
...
+
+For the access type, the digest is checked again every time the component version is transferred, because the bytes are
+re-fetched at transfer time (see [How transfer works](#how-transfer-works)). So a file that changes *after* you publish
```
morri-son:
```suggestion
re-fetched at transfer time (see [How transfer works](#how-transfer-works)). So, a file that changes *after* you publish
```

website/content/docs/tutorials/wget-http-resources.md
```diff
@@ -0,0 +1,320 @@
...
+the component version makes the next transfer fail, rather than silently swapping in different content.
+
+```text
+digest mismatch: expected 6e3205bb…, got 3b1f0c72…                          # access type
+resource blob digest mismatch: resource 6e3205bb… vs blob sha256:3b1f0c72…  # input type
+```
+
+## Send a non-GET request {#non-get-requests}
+
+By default, OCM sends a `GET`. To send something else, set `verb`, and optionally `header` and `body`. `body` is
```
morri-son:
```suggestion
By default, OCM sends a `GET` request. To send something else, set `verb`, and optionally `header` and `body`. `body` is
```

website/content/docs/tutorials/wget-http-resources.md
```diff
@@ -0,0 +1,320 @@
...
+```
+
+## Fine-tuning the download {#fine-tuning-the-download}
+
+**Only successful responses are accepted.** OCM accepts a 2xx status and fails on anything else. Redirects are followed by
+default. If you set `noRedirect: true`, the request fails.
+
+**Downloads are streamed to disk, not held in memory.** The response body is written straight to a temporary file, so
+memory use stays flat no matter how big the file is. There is no size limit by default. That temporary file is created
+under the `tempFolder` of the `filesystem.config.ocm.software/v1alpha1` configuration, falling back to the operating
```
morri-son:
```suggestion
under the `tempFolder` of the `filesystem.config.ocm.software/v1alpha1` attribute in the .ocmconfig configuration file, falling back to the operating
```

website/content/docs/tutorials/wget-http-resources.md
```diff
@@ -0,0 +1,320 @@
...
+        timeout: 5m
+```
+
+See [HTTP Client Configuration]({{< relref "docs/reference/http-client-configuration.md" >}}) for the full schema,
+defaults, and how per-host settings are merged.
+
+## Migrate from OCM v1 {#migrating-from-ocm-v1}
+
+If you are bringing constructor files or credential config over from OCM v1, keep these in mind. The credential changes
+are the most important ones.
```
morri-son:
```suggestion
Three things changed between OCM v1 and v2: credential matching, constructor syntax, and behavior. The credential changes are the most likely to break existing configurations.
```

morri-son: Mentioning the credential changes as the most important ones, you should move them directly below this sentence, as the first.

website/content/docs/tutorials/wget-http-resources.md
```diff
@@ -0,0 +1,320 @@
...
+
+| Area        | OCM v1                                                                             | OCM v2                                                    |
+|-------------|------------------------------------------------------------------------------------|-----------------------------------------------------------|
+| Media type  | `mediaType` → `Content-Type` → **URL file extension** → `application/octet-stream` | `mediaType` → `Content-Type` → `application/octet-stream` |
+| Minimum TLS | TLS 1.3                                                                            | TLS 1.2                                                   |
+
+OCM v2 **dropped** the file-extension guess, so a `.tar.gz` URL that used to resolve to `application/x-gzip` on its own now
+needs an explicit `mediaType`.
+
+### Credential changes {#credential-changes}
```
morri-son, line 183: move this section ^^up as the first of the three points you mention.

website/content/docs/tutorials/wget-http-resources.md
```diff
@@ -0,0 +1,320 @@
...
+If you are bringing constructor files or credential config over from OCM v1, keep these in mind. The credential changes
+are the most important ones.
+
+### Constructor changes {#constructor-changes}
+
+| Field        | OCM v1       | OCM v2                    | What to do                                    |
+|--------------|--------------|---------------------------|-----------------------------------------------|
+| Input `body` | Plain string | Base64-encoded byte slice | Base64-encode the body in the constructor.    |
+
+In OCM v1 the body was an `io.Reader`, which has no YAML form. In OCM v2 it is a byte slice, therefor it needs to be base64.
```
morri-son:
```suggestion
In OCM v1 the body was an `io.Reader`, which has no YAML form. In OCM v2 it is a byte slice, therefore it needs to be base64.
```

website/content/docs/tutorials/wget-http-resources.md
```diff
@@ -0,0 +1,320 @@
...
+
+### Behavior changes{#behavior-changes}
+
+| Area        | OCM v1                                                                             | OCM v2                                                    |
+|-------------|------------------------------------------------------------------------------------|-----------------------------------------------------------|
+| Media type  | `mediaType` → `Content-Type` → **URL file extension** → `application/octet-stream` | `mediaType` → `Content-Type` → `application/octet-stream` |
+| Minimum TLS | TLS 1.3                                                                            | TLS 1.2                                                   |
+
+OCM v2 **dropped** the file-extension guess, so a `.tar.gz` URL that used to resolve to `application/x-gzip` on its own now
```
morri-son:
```suggestion
OCM v2 **dropped** the file-extension guess, so a `.tar.gz` URL that used to resolve to `application/x-gzip` on its own, now
```

website/content/docs/tutorials/wget-http-resources.md
```diff
@@ -0,0 +1,320 @@
...
+| Auth precedence | Basic Auth wins; the bearer token is a fallback | Bearer token wins; Basic Auth is used only when no token is set |
+| Basic Auth      | Needs both `username` and `password`            | Sent as soon as `username` is set                               |
+| Custom CA       | Root CAs from credentials always applied        | `certificateAuthority` applied only alongside `certificate`     |
+
+## Troubleshooting {#troubleshooting}
+
+### `401 Unauthorized` from the server
+
+**Why:** Either the credentials are wrong, or no consumer entry matched the identity OCM built from the URL. No matching
+entry is difficult to debug because there was no entry matched which assumes no credentials were provided.
```
morri-son:
```suggestion
**Why:** Either the credentials are wrong, or no consumer entry matched the identity OCM built from the URL.  The second case is hard to spot: when nothing matches, OCM sends the request without credentials and the server returns the same 401 either way.
```

Skarlso:
Test fix in https://github.com/open-component-model/open-component-model/pull/3280.

website/content/docs/how-to/add-resources-from-http-urls.md
fabianburth: > Both types share the same download and credential code

maybe rather "Both types share the same download and credential functionality"

website/content/docs/how-to/add-resources-from-http-urls.md
```diff
@@ -0,0 +1,282 @@
...
+
+{{< step >}}
+
+### Choose the input type or the access type
+
+| Choose the **input type** when                     | Choose the **access type** when                    |
+|----------------------------------------------------|----------------------------------------------------|
+| The file must stay reproducible and work offline   | The URL is the source of truth and should stay so  |
+| The URL might disappear or change what it serves   | The file is large and you don't want to copy it    |
+| You are building an air-gapped repository          | Consumers should fetch directly from the origin    |
```
fabianburth: I don't think this is the right place for this comparison. I think we should either add something like this to the `input and access types` reference docs or even add a completely separate tutorial (when to use input or acess type?)

website/content/docs/how-to/add-resources-from-http-urls.md
fabianburth: > A Wget resource is transferred **by value**.

Not sure what message you are trying to convey with this "caution"? I mean, it is only transferred by value IF you set --copy-resources.

website/content/docs/how-to/add-resources-from-http-urls.md
fabianburth:
```suggestion
Both variants describe the same resource with the same fields. Write one of them to `component-constructor.yaml`:
```
spellings?

website/content/docs/how-to/add-resources-from-http-urls.md
```diff
@@ -0,0 +1,282 @@
...
+          mediaType: application/x-tar
+EOF
+```
+
+{{< /tab >}}
+{{< /tabs >}}
+
+Set `mediaType` explicitly here because GitHub serves every release asset as `application/octet-stream`. If you leave
+it out, OCM falls back to the server's `Content-Type`, then to `application/octet-stream`. The full order is in
+[Set the media type]({{< relref "docs/tutorials/wget-http-resources.md#set-the-media-type" >}}).
```
fabianburth:
```suggestion
Set `mediaType` explicitly here. Otherwise, OCM will default to the `Content-Type` served by GitHub, which for release assets is always `application/octet-stream`. The full `mediaType` defaulting mechanism of `Wget` is described in
[Set the media type]({{< relref "docs/tutorials/wget-http-resources.md#set-the-media-type" >}}).
```

website/content/docs/how-to/add-resources-from-http-urls.md
Skarlso: >retrieved from the archive

What archive? 🤔

website/content/docs/how-to/add-resources-from-http-urls.md
Skarlso: >registry access or credentials

These are the same thing. 🤔

website/content/docs/how-to/add-resources-from-http-urls.md
```diff
@@ -0,0 +1,282 @@
...
+```bash
+cat > .ocmconfig << 'EOF'
+type: generic.config.ocm.software/v1
+configurations:
+  - type: credentials.config.ocm.software
+    consumers:
+      - identity:
+          type: Wget
+          hostname: downloads.example.com
+          scheme: https
```
fabianburth, line 223: Isn't the consumer identity type versioned? I guess not since we do not have typed consumer identities yet, right?

website/content/docs/how-to/add-resources-from-http-urls.md
Skarlso: True 👍

website/content/docs/how-to/add-resources-from-http-urls.md
```diff
@@ -0,0 +1,282 @@
...
+EOF
+```
+
+Then pass `--config` when you build (or drop the file in one of the
+[well-known locations]({{< relref "docs/reference/ocm-cli/ocm.md" >}}) the CLI reads automatically):
+
+```bash
+ocm --config .ocmconfig add cv \
+  --repository ./transport-archive \
+  --constructor component-constructor.yaml
```
fabianburth: same as mentioned above, `--repository` and `--constructor` could be omitted

website/content/docs/reference/credential-consumer-identities.md
```diff
@@ -177,6 +178,125 @@ identity is derived from the Helm repository URL using the same URL-based attrib
...
+### Credential Properties
+
+| Property               | Description                                                                                                 |
+|------------------------|-------------------------------------------------------------------------------------------------------------|
+| `username`             | Username for HTTP basic authentication                                                                      |
+| `password`             | Password for HTTP basic authentication                                                                      |
+| `identityToken`        | Bearer token sent as `Authorization: Bearer <token>`. Takes precedence over Basic Auth.                     |
+| `certificate`          | PEM-encoded client certificate for mutual TLS                                                               |
+| `privateKey`           | PEM-encoded private key paired with `certificate`                                                           |
+| `certificateAuthority` | PEM-encoded CA certificate used to verify the server certificate. Only applied together with `certificate`. |
```
fabianburth, line 221: why is this only applied together with certificate?

website/content/docs/reference/credential-consumer-identities.md
```diff
@@ -177,6 +178,125 @@ identity is derived from the Helm repository URL using the same URL-based attrib
...
+### Matching Behavior
+
+The same three chained checks as [`OCIRegistry`](#ociregistry) apply: path glob, URL (scheme, hostname, port with
+default-port handling), then exact equality on the remaining attributes.
+
+{{< callout context="caution" >}}
+The identity type is matched by exact string and is **unversioned**, so it must be written as `type: Wget`. Neither
+`Wget/v1` (the name of the
+[access and input type]({{< relref "input-and-access-types.md#wgetv1-access" >}})) nor the lowercase `wget` used by OCM v1
+will match. A non-matching entry fails silently: no credentials are resolved and the request goes out unauthenticated,
```
fabianburth, line 239: > A non-matching entry fails silently

this is such a UX impairment...we'd really need proper typed consumer identity support

website/content/docs/reference/credential-types.md
```diff
@@ -142,6 +143,79 @@ use HTTP/S transport. For OCI-based Helm repositories, use `OCICredentials/v1` i
...
+          -----BEGIN CERTIFICATE-----
+          MIIDQTCCAimgAwIBAgITBmyf...
+          -----END CERTIFICATE-----
+```
+
+{{< callout context="note" >}}
+`username`/`password` and `identityToken` both set the `Authorization` header and are mutually exclusive. When both are
+present, the bearer token is used and a warning is logged. The mutual TLS pair (`certificate` + `privateKey`) is a
+transport-layer credential and can be combined with either. `certificateAuthority` is only evaluated when `certificate`
+is set, and a client certificate has no effect on a plain `http://` URL.
```
fabianburth, line 208: We do have a bit of documentation duplication between consumer identity and credential type documentation. Might be a structural issue of our website?

website/content/docs/reference/resource-repositories.md
```diff
@@ -128,6 +128,75 @@ resolves the OCI manifest digest via the registry API.
...
+transferred by value: the transfer downloads the content and stores it as a
```
fabianburth: I believe "transferred by value" is a misnomer here. If I'd upload it to another http server, it'd still be transfer by value.
Besides, we DO want to support transfer by reference (so not touching the access!)

website/content/docs/how-to/add-resources-from-http-urls.md
Skarlso: I can remove this entire thing then. :)

website/content/docs/how-to/add-resources-from-http-urls.md
Skarlso: Ah, I think I messed this up here. The by value was that it's always downloaded but this is regarding transfer. So yeah.

website/content/docs/reference/credential-consumer-identities.md
Skarlso: Only applied together with `certificate` yes?

website/content/docs/reference/credential-consumer-identities.md
Skarlso: not something in this PR, but yes. :)

website/content/docs/reference/credential-types.md
Skarlso: Maaaybe. Do you want me to do something about that here? :)

website/content/docs/reference/input-and-access-types.md
Skarlso: I like that.

website/content/docs/reference/resource-repositories.md
Skarlso: @fabianburth We do? Because I got the impression that we don't... @morri-son && @frewilhelm ?

website/content/docs/how-to/add-resources-from-http-urls.md
Skarlso: Do you want me to do anything about that here?

website/content/docs/tutorials/wget-http-resources.md
Skarlso: done 👍

website/content/docs/reference/resource-repositories.md
morri-son: I guess the issue with the outdated and also the new content 

```
Upload is not supported because a plain HTTP endpoint has no standardized write API. The transfer downloads the content
and stores it as a [`LocalBlob/v1`]({{< relref "input-and-access-types.md" >}}#localblobv1) in the target repository, regardless of the requested upload type.
```

is that we only state "transfer". But I thought on a transfer without --copy-resources all resources of type access stay untouched and point to the same location as before the transfer. So, I think we should add the --copy-resources here to make it work, or?

website/content/docs/how-to/add-resources-from-http-urls.md
```diff
@@ -0,0 +1,258 @@
...
+{{< callout context="note" >}}
+When the media type is an archive the CLI can unpack, `--output` is a **directory** with the extracted contents.
+Other media types are written to a file at that path. `cli.tar` is a `.tar`, so `./ocm-cli` is a directory here.
+{{< /callout >}}
+
+{{< /step >}}
+
+{{< /steps >}}
+
+## Authenticate against a protected server
```
frewilhelm, line 205: I don't understand why this section is in the How-To. I would've expected it in the tutorial

website/content/docs/tutorials/wget-http-resources.md
```diff
@@ -0,0 +1,317 @@
...
+- How credentials are matched to a request, and how the three authentication methods interact
+- How to pin a resource by digest so a changed file is rejected instead of silently accepted
+- How to send a non-GET request and tune timeouts, retries, and the download directory
+- What changed between OCM v1 and OCM v2, and how to migrate
+
+## Prerequisites
+
+- [OCM CLI]({{< relref "docs/getting-started/ocm-cli-installation.md" >}}) installed
+- Comfortable adding a component version with `ocm add cv`; the
+  [how-to]({{< relref "docs/how-to/add-resources-from-http-urls.md" >}}) is a good warm-up
```
frewilhelm, line 39: This tutorial does not have any steps to follow. Why do I need the prerequisites?

website/content/docs/tutorials/wget-http-resources.md
```diff
@@ -0,0 +1,317 @@
...
+A Wget resource is always transferred by value.
+
+When you run `ocm transfer cv`, an access-type resource does not stay a `Wget/v1` access in the target. OCM fetches the
+bytes and writes them into the target as a [`LocalBlob/v1`]({{< relref "docs/reference/input-and-access-types.md#localblobv1" >}}). After a transfer
+both will end up as local blobs.
+
+This means:
+
+1. You MUST pass the `--copy-resources` flag to `ocm transfer cv`. Without it, the resource is skipped, because there is no
+   way to transfer it without copying the bytes.
```
frewilhelm, line 67: This confuses me :D

Isn't every resource skipped when `--copy-resources` is not present?

website/content/docs/tutorials/wget-http-resources.md
```diff
@@ -0,0 +1,317 @@
...
+2. The bytes are fetched *again at transfer time* and checked against the resource's digest. If the file behind the URL
+   changed since the component version was built, the transfer fails instead of copying different content. That
+   check is what makes [digest pinning](#pin-a-digest) worth doing on an access-type resource.
```
frewilhelm: > The bytes are fetched *again at transfer time* and checked against the resource's digest

The resource's digest from the component descriptor (if any)?

website/content/docs/tutorials/wget-http-resources.md
```diff
@@ -0,0 +1,317 @@
...
+
+The file stays on the remote server only if the component version is never transferred. Transfer converts the `Wget/v1` reference into a `LocalBlob/v1`.
+
+## Set the media type {#set-the-media-type}
+
+Every resource has a media type. For a Wget resource, OCM picks it in the following order and stops at the first one it finds:
```
frewilhelm: > Every resource has a media type. For a Wget resource, ...

What is meant by "every resource"? Every Wget resource? If yes, then this sentence sound weird to me

website/content/docs/tutorials/wget-http-resources.md
```diff
@@ -0,0 +1,317 @@
...
+{{< callout context="caution" >}}
+Put secrets in your credential configuration, never in the specification. Anything you write into `url`, `header`, or
+`body` (including `https://user:token@host/...` and presigned query parameters) is stored with the component version
+(access type) or lives in your constructor file (input type). Neither is a safe place for a secret.
+{{< /callout >}}
+
+For the full field reference, see
+[Credential Types: WgetCredentials/v1]({{< relref "docs/reference/credential-types.md#wgetcredentialsv1" >}}).
+
+## Pin a digest {#pin-a-digest}
```
frewilhelm: I still think this should be part of the How-To to show that this should be done (best practice)

website/content/docs/tutorials/wget-http-resources.md
```diff
@@ -0,0 +1,317 @@
...
+    access:
+      type: Wget/v1
+      url: https://github.com/open-component-model/open-component-model/releases/download/v0.12.0/cli.tar
+      mediaType: application/x-tar
+```
+
+The `digest` block is optional and works on both the input and the access type. If set, these are the exact fields:
+
+- `hashAlgorithm` MUST be `SHA-256`. It is the only algorithm the Wget resource repository computes. `sha256` and
+  `SHA256` do not match.
```
frewilhelm: This sounds like a bug. Is there a reason we cannot use another algo (other than "its not implemented")?

website/content/docs/tutorials/wget-http-resources.md
```diff
@@ -0,0 +1,317 @@
...
+- `normalisationAlgorithm` MUST be `genericBlobDigest/v1`, which means "hash the raw bytes, no normalisation".
```
frewilhelm: This is also weird. Why should the user be required to set this?

(So sorry, you are probably just writing down what is implemented - but this sounds odd to me :D)

website/content/docs/tutorials/wget-http-resources.md
```diff
@@ -0,0 +1,317 @@
...
+|-------------------|------------------------------------|--------------------|--------------------------------------------------------------------------------------------|
+| Consumer identity | `type: wget`                       | `type: Wget`       | Rename the type.                                                                           |
+| Identity path     | `pathprefix`, longest-prefix match | `path`, glob match | Rename the attribute. `*` matches one path segment; omit `path` to match the whole host.   |
+
+Further matching changes:
+
+| Area            | OCM v1                                          | OCM v2                                                          |
+|-----------------|-------------------------------------------------|-----------------------------------------------------------------|
+| Auth precedence | Basic Auth wins; the bearer token is a fallback | Bearer token wins; Basic Auth is used only when no token is set |
+| Basic Auth      | Needs both `username` and `password`            | Sent as soon as `username` is set                               |
```
frewilhelm: What does "Sent as soon as `username` is set" mean?

website/content/docs/tutorials/wget-http-resources.md
```diff
@@ -0,0 +1,317 @@
...
+| Custom CA       | Root CAs from credentials always applied        | `certificateAuthority` applied only alongside `certificate`     |
+
+### Constructor changes {#constructor-changes}
+
+| Field        | OCM v1       | OCM v2                    | What to do                                    |
+|--------------|--------------|---------------------------|-----------------------------------------------|
+| Input `body` | Plain string | Base64-encoded byte slice | Base64-encode the body in the constructor.    |
```
frewilhelm: what is meant with "Input" in "Input `body`"?

frewilhelm (commented):
The How-To is much better!! However, the tutorial still feels like a `wget` misc and help-page. From our websites [contributing](https://github.com/open-component-model/open-component-model/blob/main/website/CONTRIBUTING.md#tutorial-checklist) about tutorials, it does not fit. Its missing a clear scenario where we lead the user through.

However, thats not the first tutorial that is written like that, so I guess this is fine :D

website/content/docs/tutorials/wget-http-resources.md
Skarlso: the body field in the input spec. :)

website/content/docs/tutorials/wget-http-resources.md
Skarlso: Eh this should be username is enough. :D

website/content/docs/tutorials/wget-http-resources.md
Skarlso: No idea, just copied this. I would have to check.

website/content/docs/tutorials/wget-http-resources.md
Skarlso: It's not implemented. :))) that's all.

website/content/docs/tutorials/wget-http-resources.md
Skarlso: 👍 Sure.

website/content/docs/tutorials/wget-http-resources.md
Skarlso: No, I meant here that every resource in your component version has a media type technically. Like a general statement, you know? :D But yeah, it's confusing. I'll remove it. :)

website/content/docs/tutorials/wget-http-resources.md
Skarlso: Sure...? I just copied this over. :D

website/content/docs/tutorials/wget-http-resources.md
Skarlso: Hmmm, yeah, I should add some kind of thing to follow. 🤔

website/content/docs/how-to/add-resources-from-http-urls.md
Skarlso: It's literally a How to authenticate against a protected server... why would this be in the tutorial that expected you to read the "how to do things"?

website/content/docs/tutorials/wget-http-resources.md
Skarlso:
```go
type Digest struct {
	HashAlgorithm          string `json:"hashAlgorithm"`
	NormalisationAlgorithm string `json:"normalisationAlgorithm"`
	Value                  string `json:"value"`
}
```

It's a required field. It must be set. :)

website/content/docs/tutorials/wget-http-resources.md
Skarlso: Decided in the warroom to create a followup to make the tutorial more tutorially later.

website/content/docs/tutorials/wget-http-resources.md
frewilhelm: i saw you added the section back to the how-to. But what I actually meant was setting/pinning the digest in the component-constructor as part of the how-to :D and then maybe a little call out why this is important

Skarlso: Ah then say thaaaat. :DDDD

frewilhelm (approved):
I believe we agreed in a warroom this week to merge this PR like this because the content is alright and we will rework the structure at a later point

website/content/docs/how-to/add-resources-from-http-urls.md
```diff
@@ -0,0 +1,305 @@
...
+        mediaType: application/x-tar
+        type: LocalBlob/v1
+      digest:
+        hashAlgorithm: SHA-256
+        normalisationAlgorithm: genericBlobDigest/v1
+        value: 6e3205bbad194f902ee8bdc5a712c47f7a5443fabfd066ef1e95088c837fe0ae
+      name: ocm-cli
+      relation: local
+      type: blob
+      version: 1.0.0
```
piotrjanik, line 151: 0.12 ?

website/content/docs/how-to/add-resources-from-http-urls.md
```diff
@@ -0,0 +1,305 @@
...
+cat > component-constructor.yaml << 'EOF'
+components:
+  - name: github.com/acme.org/myapp
+    version: 1.0.0
+    provider:
+      name: acme.org
+    resources:
+      - name: ocm-cli
+        type: blob
+        version: 1.0.0
```
piotrjanik, line 83: 0.12, let's make it a bit realistic:)

website/content/docs/reference/resource-repositories.md
Skarlso: Added this sentence btw.

fabianburth (approved):

website/content/docs/how-to/add-resources-from-http-urls.md
Skarlso: Technically, the resource version can be separate from the CLI. :) But yes. I will update these things anyways because I need to refactor the entire tutorial. :)

PR #3208, opened 2026-07-28
fix(ci): extract PR labeler into a tested script

morri-son:
During my work on the constructor I recognized that when using feat(constructor) in the PR title, our "PR based on Conventional Commit" check fails as the conventional commit labeler builds its label list with bracket lookups:

```js
if (type && typeToLabel[type]) { labels.push(typeToLabel[type]); }
if (scope && scopeToLabel[scope]) { labels.push(scopeToLabel[scope]); }
```

A PR whose scope (or type) is a prototype property name, most notably `constructor`, but also `toString`, `hasOwnProperty`, etc.,  resolves to that prototype value instead of `undefined`. For `constructor`, `scopeToLabel["constructor"]` returns `Object.prototype.constructor` which is true and gets pushed as a label:

```
Adding labels: kind/feature,function Object() { [native code] }
```

`addLabels` then rejects it with an HTTP 422 error and the check fails.

This is why PRs scoped `deps`, `docs`, `cli`, or unscoped pass fine, while a PR titled `feat(constructor): …` fails.

## Fix

Use `Object.hasOwn(map, key)` so only own keys match. No behavior change for existing valid scopes/types. The inline script also has been moved into an own script and tests have been added.


frewilhelm (dismissed):

matthiasbruns (changes requested):
Would you be so kind and move this inline script into its own js file and write tests for it? :)

morri-son:
@matthiasbruns 
> Would you be so kind and move this inline script into its own js file and write tests for it? :)

this is not my workflow, dunno who created it. I just stumbled upon it during my last PR. No worries to first merge this and then enhance / change it, but that is worth an own issue (with low prio), isn't it? 😊.

matthiasbruns:
I do not blindly trust untested changes :)

morri-son:
@matthiasbruns 
> I do not blindly trust untested changes :)

I feared that you'll answer like this 😂. But it is a fair point. We did the same for our release flow, so let me work on moving the script out and add tests.

matthiasbruns (approved):
Ran and tested it locally - looks good :) thanks again

frewilhelm (approved):
i trust you both

PR #3212, opened 2026-07-28
feat!: wget download cache

### LINKED ISSUE #1234: Wget - loosen hard download limit and migrate to temp file caching
**Description**
In Wget we currently have a hard file limit which is held in memory.
We need to migrate away from that to allow files being downloaded that are bigger than the limit added to the binding

### Current behaviour

The whole response body is read into a `[]byte` and returned as an in-memory blob, so peak memory is roughly the artifact size per concurrent download:

- Limit: [`DefaultMaxDownloadSize = 100 MiB`](https://github.com/open-component-model/open-component-model/blob/ba2ab4bca27b9715e0f022a1923351bc6f011262/bindings/go/wget/internal/download/options.go#L11)
- Read + hard failure: [`download.go#L125-L132`](https://github.com/open-component-model/open-component-model/blob/ba2ab4bca27b9715e0f022a1923351bc6f011262/bindings/go/wget/internal/download/download.go#L125-L132) (`io.LimitReader`, then `exceeds maximum allowed size`)
- Buffered in RAM: [`download.go#L153`](https://github.com/open-component-model/open-component-model/blob/ba2ab4bca27b9715e0f022a1923351bc6f011262/bindings/go/wget/internal/download/download.go#L153) (`inmemory.New`)

The limit is **not reachable from the CLI**: [`register.go`](https://github.com/open-component-model/open-component-model/blob/ba2ab4bca27b9715e0f022a1923351bc6f011262/cli/internal/plugin/builtin/wget/register.go#L24-L39) wires neither `InputMethod.MaxDownloadSize` nor `repository.WithMaxDownloadSize`, so both fall through to the 100 MiB default. Only Go callers can override it (0 or negative disables the limit entirely).

This affects all three wget paths, since digest and transfer both go through `DownloadResource` — an artifact over 100 MiB can be neither constructed, digested, nor transferred.

### Regression vs. OCM v1

v1 had **no** size cap and spooled to a temp file above 4 KiB: [`CACHE_CONTENT_THRESHOLD`](https://github.com/open-component-model/ocm/blob/f7638432a166936cf4e964adf75467987ddd5ad3/api/utils/blobaccess/wget/access.go#L25) / [spool block](https://github.com/open-component-model/ocm/blob/f7638432a166936cf4e964adf75467987ddd5ad3/api/utils/blobaccess/wget/access.go#L149-L151). So a 500 MiB release tarball worked in v1 and fails in v2.

**Done Criteria**

- [x] Wget is migrated to tmp file caching instead of in-memory
- [x] Default download limit can be increased
- [x] Website docs are updated https://github.com/open-component-model/open-component-model/pull/3205

matthiasbruns:
This PR caches the wget caching / download limit internally to a temp file with no limit by default.
ocmv1 had no limit - we should not break the expected behaviour with the v2 implementation.

This PR also includes identity for the website documentation and general preparation of typed identities as in the other new bindings.

Contributes: https://github.com/open-component-model/ocm-project/issues/1234

#### Breaking change
The constructor `NewResourceRepository` gets an additional parameter

**Before**
`NewResourceRepository(opts ...Option) *ResourceRepository`

**After**
`NewResourceRepository(filesystemConfig *filesystemv1alpha1.Config, opts ...Option) *ResourceRepository`

I would have loved to do this via `Options`, but the pattern above is used in every `ResourceRepo`


matthiasbruns:
@coderabbitai review

bindings/go/wget/repository/resource_repository.go
```diff
@@ -76,18 +86,23 @@ func (r *ResourceRepository) GetResourceCredentialConsumerIdentity(ctx context.C
...
 
 // DownloadResource downloads a resource from the URL specified in the wget access spec.
+// The returned blob is backed by a file under the configured temp folder that outlives
+// this call, so the caller owns it.
 func (r *ResourceRepository) DownloadResource(ctx context.Context, resource *descriptor.Resource, credentials runtime.Typed) (blob.ReadOnlyBlob, error) {
+	return r.download(ctx, resource, credentials, r.filesystemConfig.TempFolder)
+}
+
+// download streams the resource body into tempDir and returns it as a file-backed blob.
+func (r *ResourceRepository) download(ctx context.Context, resource *descriptor.Resource, credentials runtime.Typed, tempDir string) (blob.ReadOnlyBlob, error) {
```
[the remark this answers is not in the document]
matthiasbruns: Imho the callers should clean that up - did you check the other resource repo implementations?

bindings/go/wget/internal/download/download.go
```diff
@@ -120,21 +126,39 @@ func Download(ctx context.Context, req Request, opts ...Option) (blob.ReadOnlyBl
...
+
+	written, err := io.Copy(file, respBody)
+	if closeErr := file.Close(); err == nil {
+		err = closeErr
+	}
+	if err != nil {
+		return nil, fmt.Errorf("error writing response body from %s to %s: %w", safeURL.String(), path, err)
+	}
+
+	if maxDownloadSize > 0 && written > maxDownloadSize {
```
piotrjanik, line 161: In theory, you can use `Content-Length` to return an error earlier, rather than after downloading it into a temporary file.

matthiasbruns: like here? https://github.com/open-component-model/open-component-model/pull/3212/changes/BASE..f56eb9f76744098bcfe6ca03130d0644895c51c4#diff-563400b029d31c1102bbb4b361a6276f19805c0713a6c401658e27ece6e3321dR131
its already there :)

bindings/go/wget/input/method.go
```diff
@@ -28,6 +29,11 @@ type InputMethod struct {
 	// MaxDownloadSize limits the number of bytes read from a response body. When zero,
 	// the download package default [download.DefaultMaxDownloadSize] is used. A negative value disables the limit.
 	MaxDownloadSize int64
+	// TempFolder is the directory the downloaded body is streamed into. When empty,
+	// the OS temporary directory is used. The file backing the returned blob is
+	// created here and is not removed, because it holds the content the constructor
+	// stores as a local blob.
+	TempFolder string
```
Skarlso, line 37: Can't see what sets this, but this needs to go through the filesystem tempdir config.

bindings/go/wget/internal/download/download.go
```diff
@@ -120,21 +126,39 @@ func Download(ctx context.Context, req Request, opts ...Option) (blob.ReadOnlyBl
...
 	}
 
-	var data []byte
+	// When the server announces the size up front, an oversized body is rejected
+	// before any of it is transferred. ContentLength is negative when unknown.
+	if maxDownloadSize > 0 && resp.ContentLength > maxDownloadSize {
+		return nil, fmt.Errorf("response body from %s exceeds maximum allowed size of %d bytes", safeURL.String(), maxDownloadSize)
+	}
+
+	respBody := io.Reader(resp.Body)
```
Skarlso, line 136: Since you are no longer streaming and everything is in a file, you should close the response body.

bindings/go/wget/internal/download/download.go
```diff
@@ -145,12 +169,13 @@ func Download(ctx context.Context, req Request, opts ...Option) (blob.ReadOnlyBl
 		mediaType = "application/octet-stream"
 	}
 
-	blobOpts := []inmemory.MemoryBlobOption{
-		inmemory.WithMediaType(mediaType),
-		inmemory.WithSize(int64(len(data))),
+	b, err := filesystem.GetBlobFromOSPath(path)
```
Skarlso: This file is never cleaned up only on error. We have to keep it for the read to work later, but every further call only clears specific paths. This is a regression compared to the in-memory approach. A long running process will accumulate files on each download. `ProcessResourceDigest` is the only one eventually calling RemoveAll.

bindings/go/wget/spec/identity/v1/type.go
```diff
@@ -0,0 +1,101 @@
...
+	Hostname string       `json:"hostname,omitempty"`
+	Scheme   string       `json:"scheme,omitempty"`
+	Port     string       `json:"port,omitempty"`
+	Path     string       `json:"path,omitempty"`
+}
+
+// ToIdentity converts a [WgetIdentity] into a [runtime.Identity].
+// Empty fields are omitted from the resulting map. If the type field is unset,
+// the canonical [VersionedType] is used.
+func ToIdentity(identity *WgetIdentity) runtime.Identity {
```
Skarlso: There is a discrepancy where this function sets versioned identity but `IdentityFromURL` (`identity.SetType(Type) // Type is  var Type = runtime.NewUnversionedType(WgetIdentityType)`)sets it unversioned. Is that intentional?

Skarlso (commented):
Couple of concerns I'm unsure about. :)

bindings/go/wget/input/method.go
matthiasbruns: We will do the same setup an in helm 

- https://github.com/open-component-model/open-component-model/blob/587069232109c63869417eafd81b07e885e69a41/cli/internal/plugin/builtin/input/helm/method.go#L20 
- https://github.com/open-component-model/open-component-model/blob/c5ed93a358c34a8c40efccd4ccb616f2be6d5c00/bindings/go/helm/input/method.go#L84

so you are right

bindings/go/wget/internal/download/download.go
matthiasbruns: since we are exposing it to the callers, they have to take care of this - I remember having had the same discussion in helm - how would we do this else?

Skarlso: It's probably fine yeah. :/ It just sucks a bit, because some of our code does do a RemoveAll. Ah well. I don't have a better idea about this here since, like I said, we need to keep it around. :/ Sucks. :D

bindings/go/wget/internal/download/download.go
matthiasbruns: https://github.com/open-component-model/open-component-model/blob/97f32ce725f7f2cda803e6fd97c086f46f2a10ae/bindings/go/wget/internal/download/download.go#L114

This is happening already, just not shown in the diff since it's already in main

bindings/go/wget/spec/identity/v1/type.go
matthiasbruns: So... yeah this is in all bindings right now. The identity package is just a prep - while lookup and graph uses `UNVERSIONED`, since config entries are `UNVERSIONED`, in the binding itself we use `VERSIONED` if we use the typed identities.

matthiasbruns: I removed `ToIdentity` for now from wget - we will have to add it back if and when we do typed identities later

bindings/go/wget/internal/download/download.go
Skarlso: No wait I misread this entire thing. This is the resp.Body. :D Not the request body. :DDD

Skarlso (approved):

piotrjanik:
@matthiasbruns dont you need to update `cli/internal/plugin/builtin/wget/register.go` so that it get filesystemconfig?

matthiasbruns:
> @matthiasbruns dont you need to update `cli/internal/plugin/builtin/wget/register.go` so that it get filesystemconfig?

next pr :)

piotrjanik (approved):

PR #3214, opened 2026-07-28
fix(website): bump brace-expansion to 5.0.8 for GHSA-mh99-v99m-4gvg

frewilhelm:
brace-expansion <=5.0.7 is vulnerable to an uncatchable OOM DoS (CVE-2026-14257) when expand() is called on crafted input that chains many brace groups. Adds a package.json override to force all transitive resolutions (both the direct 5.0.7 and the 1.1.16 pulled in via glob) onto the patched 5.0.8.

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -57,6 +57,7 @@
   "overrides": {
     "adm-zip": "0.6.0",
     "semver": "^7.5.4",
+    "brace-expansion": "^5.0.8",
     "eslint-config-developit": {
       "eslint-plugin-compat": {
         "eslint": "$eslint"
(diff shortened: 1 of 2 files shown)


website/package.json
```diff
@@ -57,6 +57,7 @@
   "overrides": {
     "adm-zip": "0.6.0",
     "semver": "^7.5.4",
+    "brace-expansion": "^5.0.8",
```
matthiasbruns, line 60: not really important for us, but still good to get this off our scanners (client side rendering only)

matthiasbruns (approved):

morri-son (approved):

PR #3215, opened 2026-07-28
fix: access type OCIImageLayer does not work as expected (#2873)

### LINKED ISSUE #2873: access type `OCIImageLayer` does not work as expected.
**Context**

Use the following `component-constructory.yaml`

```
components:
  - name: ocm.software/repro/oci-image-layer
    version: 1.0.0
    provider:
      name: ocm.software
    resources:
      - name: podinfo-chart-layer
        type: helmChart
        version: 6.9.1
        access:
          type: OCIImageLayer/v1
          imageReference: ghcr.io/stefanprodan/charts/podinfo@sha256:1b6a62255aef35d373d870dcd1f34aeb23ffa164e20025afc56fe4695596b53d
          mediaType: application/vnd.cncf.helm.chart.content.v1.tar+gzip
          digest: sha256:1b6a62255aef35d373d870dcd1f34aeb23ffa164e20025afc56fe4695596b53d
          size: 15040
```

Then, create the component descriptor using

```
ocm add cv
```

produces the following error

```
Error: failed to render components recursively: constructing component versions failed: failed to constructComponent components from graph: failed to process component constructor graph: failed to process value for node name=ocm.software/repro/oci-image-layer,version=1.0.0 in graph: failed processing constructor component: error constructing component "name=ocm.software/repro/oci-image-layer,version=1.0.0": error constructing component: error processing resource "name=podinfo-chart-layer,version=6.9.1" at index 0: error processing resource "name=podinfo-chart-layer,version=6.9.1" with digest processor: unsupported access type OCIImageLayer/v1: expected OCI image
```

matthiasbruns:
Resources with an `OCIImageLayer/v1` (or legacy `ociBlob/v1`) access were routed to the OCI resource repository but rejected by every read path with "expected OCI image", so `ocm add cv` failed. Digest processing, download, credential identity and ownership now handle the type; a layer addresses one existing blob, so it is verified via the blob endpoint (not manifest resolution), downloads as raw content rather than an OCI layout tar, and is rejected as an upload target.    

Contributes: https://github.com/open-component-model/open-component-model/issues/2873

##### How to test the changes

Custom sh script.

##### Verification


bindings/go/oci/repository/resource/resource_repository.go
```diff
@@ -195,14 +176,39 @@ func (p *ResourceRepository) getRepository(spec *ociv1.Repository, credentials *
...
+	var field string
+	switch access := access.(type) {
+	case *v1.OCIImage:
+		reference, field = access.ImageReference, "imageReference"
+	case *v1.OCIImageLayer:
+		reference, field = access.Reference, "ref"
+	default:
+		return "", fmt.Errorf("unsupported access type %s: expected OCI image or OCI image layer", access.GetType())
+	}
+	// Unknown keys are dropped on deserialization, so an access that spells the
```
Skarlso: What does this comment mean? What does 'Name the field it belongs in' mean? :D Name it where? How? When?

bindings/go/oci/repository/resource/resource_repository.go
```diff
@@ -195,14 +176,39 @@ func (p *ResourceRepository) getRepository(spec *ociv1.Repository, credentials *
...
-		return "", fmt.Errorf("error parsing loose image reference %q: %w", access.ImageReference, err)
+		return "", fmt.Errorf("error parsing loose image reference %q: %w", reference, err)
 	}
 	// host is the registry with sane defaulting
-	baseURL := ref.RegistryWithScheme()
-	return baseURL, nil
+	return ref.RegistryWithScheme(), nil
+}
+
+// validateUploadTarget rejects access types that can be read but not written to. An
```
Skarlso: What is this comment? :D 

>An upload pins its reference to the manifest it pushes; a layer names an existing blob.

What does this mean? How is it relevant to just plain rejecting OCIImageLayer as an uploader?

I would call this just rejectReadOnlyAccess or something. That makes more sense.

bindings/go/oci/stream/layer.go
```diff
@@ -0,0 +1,43 @@
...
+
+var _ ResourceStream = (*OCILayerResourceStream)(nil)
+
+func (s *OCILayerResourceStream) Root() ocispec.Descriptor {
+	return s.Descriptor
+}
+
+// Materialize fetches the layer and buffers it into a blob. Passing the descriptor's
+// digest along makes the blob verify the content against it as it is read.
+func (s *OCILayerResourceStream) Materialize(ctx context.Context) (blob.ReadOnlyBlob, error) {
```
Skarlso: Why is this needed and how is it relevant to the fix? 🤔

Skarlso: Also, this never closes the _in memory_ blob. Which is probably bad. 🤔

bindings/go/oci/stream/layer.go
```diff
@@ -0,0 +1,43 @@
...
+	data, err := s.Fetch(ctx, s.Descriptor)
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch layer %q: %w", s.Descriptor.Digest, err)
+	}
+	return inmemory.New(data,
+		inmemory.WithMediaType(s.Descriptor.MediaType),
```
Skarlso: Only add manifest if mediatype is not empty, otherwise you'll add `""` as mediatype _explicitly_.

bindings/go/oci/repository.go
```diff
@@ -281,11 +281,52 @@ func (repo *Repository) ProcessResourceDigest(ctx context.Context, res *descript
...
+		return nil, fmt.Errorf("invalid OCI image layer access: %w", err)
+	}
+
+	src, err := repo.resolver.StoreForReference(ctx, typed.Reference)
+	if err != nil {
+		return nil, err
+	}
+
+	// A layer is always addressed by digest, so its descriptor is fully known here.
+	// Exists checks the blob endpoint; Resolve would query the manifest one and 404.
```
Skarlso: This is only true as long as the mediatype isn't a manifest type which isn't prechecked here anywhere.

bindings/go/oci/spec/access/v1/oci_image_layer.go
```diff
@@ -43,17 +43,19 @@ type OCIImageLayer struct {
...
 
+// Validate parses the reference with looseref so that the scheme-prefixed form
+// (http://host/repo) the binding accepts for plain-HTTP registries is not rejected.
 func (t *OCIImageLayer) Validate() error {
 	if err := t.Digest.Validate(); err != nil {
-		return err
+		return fmt.Errorf("invalid digest %q: %w", t.Digest, err)
 	}
 	if t.Size < 0 {
 		return fmt.Errorf("size %d is invalid, must be greater than 0", t.Size)
```
Skarlso: Heh, this is incorrect btw. The size check is `<0` meaning it would pass with `0` as Size.

bindings/go/oci/repository/resource/resource_repository.go
```diff
@@ -245,38 +251,39 @@ func createRepository(
...
+	repo, err := p.getRepository(&ociv1.Repository{BaseUrl: baseURL}, ociCredentials)
+	if err != nil {
+		return nil, nil, err
+	}
+	return repo, obj, nil
 }
 
 func (p *ResourceRepository) DownloadResourceStream(ctx context.Context, resource *descriptor.Resource, credentials runtime.Typed) (ocistream.ResourceStream, error) {
-	repo, err := p.resolveOCIImageRepo(resource, credentials)
+	repo, _, err := p.resolveRepository(resource, credentials)
```
Skarlso, line 319: So, before this died on OCIImage. Now, we take the url and pin it to a blob digest. That will point to a manifest doesn't it? Wouldn't that then fail with a 404 which is the `Exists` check thing because of the mediatype?

Skarlso (changes requested):
I have some concerns. :D

bindings/go/oci/stream/layer.go
```diff
@@ -0,0 +1,62 @@
...
+
+	opts := []inmemory.MemoryBlobOption{
+		inmemory.WithSize(s.Descriptor.Size),
+		inmemory.WithDigest(s.Descriptor.Digest.String()),
+	}
+	// An empty media type would overwrite the default the blob applies on its own.
+	if s.Descriptor.MediaType != "" {
+		opts = append(opts, inmemory.WithMediaType(s.Descriptor.MediaType))
+	}
+	return inmemory.New(bytes.NewReader(data), opts...), nil
```
Skarlso: Can't we literally do just this?
```go
b := inmemory.New(reader, opts...)
if err := b.Load(); err != nil {
    return nil, fmt.Errorf("failed to read layer %q: %w", s.Descriptor.Digest, err)
}
return b, nil
```

Since inmemory already does size checking, and Load is lazy. 🤔

bindings/go/oci/stream/layer.go
```diff
@@ -0,0 +1,62 @@
...
+
+var _ ResourceStream = (*OCILayerResourceStream)(nil)
+
+func (s *OCILayerResourceStream) Root() ocispec.Descriptor {
+	return s.Descriptor
+}
+
+// Materialize fetches the layer and buffers it into a blob. The returned blob
+// carries the descriptor digest, so reading it verifies the content against it.
+func (s *OCILayerResourceStream) Materialize(ctx context.Context) (b blob.ReadOnlyBlob, err error) {
```
Skarlso, line 29: What's the intention here for the call site? I don't see this being used anywhere. 🤔

bindings/go/oci/spec/access/v1/oci_image_layer.go
```diff
@@ -43,17 +44,19 @@ type OCIImageLayer struct {
...
 	if t.Size < 0 {
-		return fmt.Errorf("size %d is invalid, must be greater than 0", t.Size)
+		return fmt.Errorf("size %d is invalid, must not be negative", t.Size)
 	}
 	if t.Reference == "" {
-		return fmt.Errorf("reference is empty")
+		return errors.New(`reference is empty, set it in field "ref"`)
 	}
-	ref, err := registry.ParseReference(t.Reference)
+	ref, err := looseref.ParseReference(t.Reference)
```
Skarlso, line 58: Same here. This means that registry is empty and something like `ngins:latest` is just accepted now.

bindings/go/oci/repository/resource/resource_repository.go
```diff
@@ -219,14 +200,36 @@ func (p *ResourceRepository) getRepository(spec *ociv1.Repository, credentials *
...
+		reference, field = access.ImageReference, "imageReference"
+	case *v1.OCIImageLayer:
+		reference, field = access.Reference, "ref"
+	default:
+		return "", fmt.Errorf("unsupported access type %s: expected OCI image or OCI image layer", access.GetType())
+	}
+	if reference == "" {
+		return "", fmt.Errorf("access type %s has an empty reference, set it in field %q", access.GetType(), field)
+	}
+	ref, err := looseref.ParseReference(reference)
```
Skarlso: This will optionally parse Registry. In certain places, we ignore this, in others, we make sure that registry is set. How does this need to handle that?

Skarlso (commented):
Still have some concerns. :)

bindings/go/oci/spec/access/v1/oci_image_layer.go
matthiasbruns: added to Validate()

bindings/go/oci/repository/resource/resource_repository.go
matthiasbruns: ctf does not need a registry, right? I first had in in the layer validate itself but moved it to ociImageLayerAccessToBaseURL since we only need it there

bindings/go/oci/spec/access/v1/oci_image_layer.go
matthiasbruns: as it should imho - we will check for a registry in the resource repo if needed, not in the spec validation

bindings/go/oci/stream/layer.go
matthiasbruns: called through the `ResourceStream` interface in `Repository.DownloadResource`.

I think we had the same pattern for the other oci types

bindings/go/oci/repository/resource/resource_repository.go
matthiasbruns: ora takes care of figuring that out based on the media-type. 
no media-type -> `blobs/digest`
with media-type -> `manifests/digest`

PR #3221, opened 2026-07-29
chore(doc): create an ADR for localization at transfer helm wrapper solution

### LINKED ISSUE #1162: Create an ADR for the proposed Localization Wrapper
**Description**

Convert the proposal from Fabian with regards to the wrapper located here: https://github.com/fabianburth/open-component-model/blob/ba7134bc854686d6e9b16e42ffadcd9d76a627e9/docs/adr/0025_localization_at_transfer_time/PROPOSAL-WRAPPER.md into a viable ADR.

This ADR should be proved by a POC in order to make it viable and figure out the edge-cases that it could potentially involve.

**Done Criteria**
- [ ] ADR exists that describes how the proposal could be used for localization during transport
- [ ] POC exists to prove the viability of this approach

Skarlso:

Fixes https://github.com/open-component-model/ocm-project/issues/1162.

##### Verification


morri-son (dismissed):

docs/adr/0026_localization_at_transfer.md
```diff
@@ -0,0 +1,430 @@
...
+```yaml
+podinfo:
+  image:
+    digest: sha256:abc123
+    repository: target.registry/library/podinfo
+```
+
+Design details:
+
+* **Alias equals the original chart name**, so the overrides is under the
```
piotrjanik, line 159: Hmm. I am afraid that from now on user will have to prefix each var with an alias, so for example: helm install ... --set podinfo.replicas=1 vs helm install ... --set replicas=2

I think it changes the way variables can be updated as ther ewill be a need to prefix vars with an alias.

docs/adr/0026_localization_at_transfer.md
```diff
@@ -0,0 +1,430 @@
...
+### [Option 1] Localize at Deploy Time
+
+Pros:
+
+* No transfer-time changes at all; artifact signatures untouched.
+
+Cons:
+
+* Failures surface at deploy time when everything is already packaged.
+* Every environment repeats the localization work.
```
piotrjanik, line 340: Is not that the same as with wrapper chart? Localisation will happen during helm install, right?

docs/adr/0026_localization_at_transfer.md
```diff
@@ -0,0 +1,430 @@
...
+RFC 3161 timestamp authority is used or wrappers are re-signed on re-transfer.
+
+Notary is used here instead of cosign to avoid having to set up the entire signing
+infrastructure of cosign for this. Instead, Notary perfectly integrates with Flux already
+which means that the deployment can easily verify the wrapper by providing
+the OCIRepository with the right secret.
+
+## Pros and Cons of the Options
+
+### [Option 1] Localize at Deploy Time
```
piotrjanik, line 331: That would be done by mutating webhooks?

docs/adr/0026_localization_at_transfer.md
```diff
@@ -0,0 +1,430 @@
...
+Cons:
+
+* Changes the resource digest, invalidating the component signature; requires
+  a resigning process and a transfer chain of trust.
+* Requires authored localization instructions (CEL expressions or sibling
+  resources) maintained by the component engineer.
+* Reproducing "what exactly changed" requires diffing content rather than
+  reading a small overlay.
+
+### [Option 3] Localize via Wrapper Charts
```
piotrjanik, line 359: Correct me if i am wrong, but to make chart localised we dont have to build wrapper chart but actually change default values.yaml. Usually it is values.yaml where images are defined.

If we could pack localised-values.yaml next to helm chart and use it during helm install, would not that solve the problem?

frewilhelm (commented):
I remember from the original idea some kind of admission controller who would try to rebuild the wrapper-chart based on the original one to make sure it was not changed in-between. Was this omitted?

Skarlso:
@frewilhelm yes, that was scrapped. :)

docs/adr/0026_localization_at_transfer.md
```diff
@@ -0,0 +1,430 @@
...
+* **Zero chart-author involvement.** Localization should work by convention
+  (the well-known Helm values layout also used by Renovate), with no
+  localization instructions embedded in the component version.
+* **Standard consumption.** The result _MUST_ be installable by plain Helm,
+  Flux, Argo and with no OCM-aware tooling at deploy time.
+* **Library-first integration.** The mechanism must live in the transfer graph
+  as a regular transformation node, because the Kubernetes controller consumes
+  transfer as a library.
+* **Trust for the new artifact.** The generated wrapper is new content and
+  needs its own signature, produced during the same transfer.
```
fabianburth, line 88: we should probably at least mention the options we did not go for and why (aka my validating webhook idea), especially since this is generally a question that's being asked frequently

docs/adr/0026_localization_at_transfer.md
Skarlso: That's right, yes. The alias normally matches up with the name. But there is no way around the key-ing when working with the wrapper and people would like to introduce custom values. 🤷

docs/adr/0026_localization_at_transfer.md
```diff
@@ -0,0 +1,430 @@
...
+  (`+localized.1`) was rejected because OCI distribution tags forbid `+`. A
+  prerelease suffix is OCI-tag-safe, and the prerelease-skipping semantics of
+  semver range resolution do not matter here because the wrapper is consumed
+  by exact reference. The `.1` counter leaves room for multiple wrapper
+  revisions per original chart version.
+* **The original chart archive is vendored into the wrapper byte-exact**, as
+  `charts/<name>-<version>.tgz`. Nothing in the Helm install path resolves
+  `oci://` dependencies at install time (`helm dependency build` is a dev-time
+  command; Flux installs chart archives as-is), so the dependency must be
+  physically present. Vendoring also closes a time-of-check gap: Helm
```
fabianburth, line 171: > Vendoring also closes a time-of-check gap

wouldnt that be solved by digest pinning already? Anyhow, I thought not having to deduplicate is part of the charm of the proposal 🥺

docs/adr/0026_localization_at_transfer.md
Skarlso: Wrapping happens at transfer time. If you deploy from the same environment / transferred component version, everything is already done and you don't have to do anything else.

docs/adr/0026_localization_at_transfer.md
```diff
@@ -0,0 +1,430 @@
...
+  dependencies are version-addressed and mutable, while the vendored archive
+  freezes the exact transferred bytes under the wrapper's signature.
+* **Wrapper generation is deterministic.** Same inputs produce the same chart
+  bytes, which is why provenance metadata is kept out of the chart (see
+  [Provenance](#provenance)).
+
+The wrapper is pushed as a **standalone OCI Helm artifact** next to the
+original chart, at `<chart-repository>/<chart>-localized:<version>`. It is
```
fabianburth, line 179: > `<chart-repository>/<chart>-localized:<version>`

This is effectively a different oci repository. Would tagging even be necessary? Couldn't we simply store in untagged in the same repo?

docs/adr/0026_localization_at_transfer.md
Skarlso: Webhooks aren't good for this mainly because they are super slow and time out in 500ms. They need to be fast. Our localization just isn't fast. :)

docs/adr/0026_localization_at_transfer.md
Skarlso: Packaging requires an update to the package which modifies the original chart content that you have to resign. This is in the other ADR scenario.

docs/adr/0026_localization_at_transfer.md
```diff
@@ -0,0 +1,430 @@
...
+
+* the downloaded chart archive (`chartFile` from `GetHelmChart`),
+* the chart resource after upload (`chartResource` from the add step, which
+  determines where the wrapper is pushed),
+* one `{source, target}` resource pair per transferred image, where the target
+  is the post-transfer resource that has the relocated reference and digest.
+
+Because uploads may return tag-only references, the node falls back to the
+resource's `digest` field (SHA-256) to pin the localized image when the
+reference itself carries no digest.
```
fabianburth, line 221: is that really the case?

docs/adr/0026_localization_at_transfer.md
```diff
@@ -0,0 +1,430 @@
...
+
+These live on the manifest, not in `Chart.yaml`, because component identity and
+repository are transfer-time metadata. Baking them into the chart would break
+the determinism of the wrapper bytes.
+
+An earlier design attached the wrapper to the original chart as an OCI
+referrer. That was dropped for this proposal: referrers provide forward
+discovery (original to wrapper), while this ADR here is backward
+provenance only. The referrer implementation is available in the `oci`
+bindings if forward discovery becomes a requirement.
```
fabianburth, line 298: You can discover in both directions with referrers?

docs/adr/0026_localization_at_transfer.md
fabianburth: no, that's what we are currently doing with kro

docs/adr/0026_localization_at_transfer.md
```diff
@@ -0,0 +1,430 @@
...
+* No forward discovery from the original chart to its wrapper right now, but
+  this is possible to fix with Referrers.
+
+## Validation
+
+* Unit and golden tests in `bindings/go/helm/localize` (scanner tables,
+  byte-for-byte golden wrapper output, determinism, vendoring), transformer
+  tests in `bindings/go/helm/transformation`, graph wiring and flag-guard
+  tests in `bindings/go/transfer`, and signing tests against a fake notation
+  repository.
```
fabianburth, line 387: did you link your branch somewhere?

docs/adr/0026_localization_at_transfer.md
```diff
@@ -0,0 +1,430 @@
+# Helm Localization at Transfer Time via Wrapper Charts
+
+* **Status**: proposed
+* **Deciders**: Gergely Brautigam, Fabian Burth, Jakob Moeller
```
fabianburth:
```suggestion
* **Deciders**: OCM Maintainer Team
```

fabianburth (commented):
There are quite some opportunities for improvement here. 

Currently it is less an ADR than it is an record of our PoC. 
- It goes into rather great detail about PoC implementation details that we decided on in a whim (as it is a PoC) - for example, the paragraph about the storage location of the helm chart and the paragraph about referrers.
- It pretty much completely omits the mention of scanning tools (which was a primary driver for the wrapper approach) as well as any tests that would validate that scanning of helm charts even works as we'd expect it to work and thus, works with the concept.
- It omits the alternatives that were discussed originally and were decided against during the PoC.
- It exclusively focuses on helm even though the validity of the whole approach depends on the general concept also being applicable to kustomize (and to common manifests through kustomize).

**BUT** since we decided against the wrapper approach anyway and this is considered more of a historic record, in case we might want to come back to this, I do not want to invest us anymore significant time here!

docs/adr/0026_localization_at_transfer.md
piotrjanik: IMO, it's an important issue. For users, it means they need to update all Application CRs where they set custom values.

This should be documented.

docs/adr/0026_localization_at_transfer.md
Skarlso: ❤️ talked in war room

docs/adr/0026_localization_at_transfer.md
piotrjanik: I was thinking about adding localised-values.yaml as another resource, leaving original helm chart as-is.
Though I am not sure if that could work.

Skarlso: >I was thinking about adding localised-values.yaml as another resource,

Updating the component requires resigning. :) Since you changed the component.

docs/adr/0026_localization_at_transfer.md
Skarlso: That's fair.

docs/adr/0026_localization_at_transfer.md
Skarlso: if we store it untagged, doesn't it get GC'd? 🤔

docs/adr/0026_localization_at_transfer.md
Skarlso: Ah, that's actually right, yeah.

docs/adr/0026_localization_at_transfer.md
Skarlso: Noooppee. (I will)

docs/adr/0026_localization_at_transfer.md
piotrjanik: feel free to resolve it

docs/adr/0026_localization_at_transfer.md
piotrjanik: fair point

docs/adr/0026_localization_at_transfer.md
Skarlso: 👍 That's fair. I'll add that.

jakobmoellerdev:
force merged to capture current state.

PR #3223, opened 2026-07-29
docs(agents): document the kind/* PR-label rule

morri-son:
Documents a PR-label rule that was missing and obscured.

`verify-labels` rejects any PR without a `kind/*` label. The title auto-labeler assigns one only for `feat`, `fix`, and `chore`. `docs`, `test`, and `perf` map to `area/*` only, so those PRs need a `kind/*` added by hand. The guide did not say this, and its "auto-labeling" mention implied it was handled, so `docs:`/`test:`/`perf:` PRs silently fail the check.

### DIFF
--- a/agents.md
+++ b/agents.md
@@ -74,6 +74,14 @@ chore(deps): update dependencies
 
 Types: `feat`, `fix`, `chore`, `docs`, `test`, `perf`. Breaking changes use `!`: `feat(api)!: remove deprecated method`.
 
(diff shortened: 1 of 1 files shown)


morri-son:
Closing in favor of a root-cause fix to the title auto-labeler: docs/test/perf will get a kind/* label automatically, so documenting a manual workaround is unnecessary. Follow-up PR incoming.

PR #3229, opened 2026-07-30
feat(cli): warn when digest processing is skipped but a resource declares a digest

### LINKED ISSUE #1199: Add an optional Digest in the resource spec to enable trust

morri-son:
## What

When `ocm add cv` is executed with `--skip-reference-digest-processing` AND the author declares explicit resource digest(s) in the constructor, emit a single WARN. The flag is still honored  and digest processing still skipped, nothing is blocked. The PR also updates the help text of the flag.

## Why

A declared resource digest is verified against referenced content only once, at constructor time, by the resource digest processor. `ocm sign` hashes the descriptor and `ocm verify` checks the descriptor against that signature. No re-hashing takes place. So construction is the only time that validates a declared digest. With `--skip-reference-digest-processing` this can be skipped.

When used, an author-declared digest is **not dropped**, but copied into the descriptor **unverified**  by `ConvertToDescriptorResource` and then signed over. Every later verify is done against a value that was never checked against content.

Declaring a digest ("verify this content is X") AND skipping digest processing ("do not verify") are contradictory instructions. Both are frequently authored by different people or layers: a security-conscious author pins digests in the constructor, while a build/CI layer sets the skip flag for speed. Neither sees the other.

The warning surfaces that contradiction when both are visible and fixable, without overriding the user's explicit choice. It is conditional: it fires only when the flag is set **AND** at least one resource declares a digest. Skipping with no author-declared digest is a valid intent and stays silent.

## Changes

- `cli/cmd/add/component-version/cmd.go`: conditional WARN +
  `constructorDeclaresResourceDigest` helper; updated flag help text.
- `cli/docs/reference/ocm_add_component-version.md`: regenerated.


cli/cmd/add/component-version/cmd.go
```diff
@@ -429,6 +433,28 @@ func GetComponentConstructor(file *file.Flag) (*constructorruntime.ComponentCons
...
+// is recorded as-is but never verified against the referenced content, so a stale or
+// wrong digest would go unnoticed.
+func warnOnSkippedDigestVerification(ctx context.Context, constructorSpec *constructorruntime.ComponentConstructor) {
+	if constructorSpec == nil {
+		return
+	}
+	for _, component := range constructorSpec.Components {
+		for _, resource := range component.Resources {
+			if resource.Digest == nil {
+				continue
```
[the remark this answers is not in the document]
morri-son: The analysis was based on constructor v0.0.8. This PR bumps to v0.0.12, which adds the `Digest` field to `constructorruntime.Resource` as part of #3202. The field is there, the code compiles, and CI is green.

cli/cmd/add/component-version/cmd.go
```diff
@@ -429,6 +433,28 @@ func GetComponentConstructor(file *file.Flag) (*constructorruntime.ComponentCons
...
+			}
+			slog.WarnContext(ctx,
```
matthiasbruns: this will explode on large cvs - logging this once should be more than enough

morri-son: changed to count all resources with a non-nil digest and then only emit exactly one warning.

matthiasbruns (dismissed):

cli/cmd/add/component-version/cmd.go
```diff
@@ -429,6 +433,31 @@ func GetComponentConstructor(file *file.Flag) (*constructorruntime.ComponentCons
...
+	for _, component := range constructorSpec.Components {
+		for _, resource := range component.Resources {
+			if resource.Digest != nil {
+				count++
+			}
+		}
+	}
+	if count == 0 {
+		return
+	}
```
frewilhelm: I don't think you need to loop and count every resource.Digest you find. Just loop and if you find one, display the warning and return - no need to loop over the whole thing.

cli/cmd/add/component-version/cmd.go
```diff
@@ -224,7 +224,7 @@ add component-version --%[1]s ./archive --%[2]s %[3]s.yaml
 	cmd.Flags().String(FlagBlobCacheDirectory, filepath.Join(".ocm", "cache"), "path to the blob cache directory")
 	enum.Var(cmd.Flags(), FlagComponentVersionConflictPolicy, ComponentVersionConflictPolicies(), "policy to apply when a component version already exists in the repository")
 	enum.Var(cmd.Flags(), FlagExternalComponentVersionCopyPolicy, ExternalComponentVersionCopyPolicies(), "policy to apply when a component reference to a component version outside of the constructor or target repository is encountered")
-	cmd.Flags().Bool(FlagSkipReferenceDigestProcessing, false, "skip digest processing for resources and sources. Any resource referenced via access type will not have their digest updated.")
+	cmd.Flags().Bool(FlagSkipReferenceDigestProcessing, false, "skip digest processing for resources and sources. Any resource referenced via access type will not have their digest updated, and any digest specified in the constructor will not be verified against the referenced content.")
```
fabianburth:
```suggestion
	cmd.Flags().Bool(FlagSkipReferenceDigestProcessing, false, "skip digest processing for resources . Any resource referenced via access type will not have their digest updated, and any digest specified in the constructor will not be verified against the referenced content.")
```

Sources do not even have a digest, unsure what we are skipping here?

cli/cmd/add/component-version/cmd.go
fabianburth: I would go as far as to say, it's enough to warn that digest verification is skipped and omit this whole function

cli/docs/reference/ocm_add_component-version.md
```diff
@@ -128,7 +128,7 @@ add component-version --repository ./archive --constructor component-constructor
   -o, --output enum                                   output format of the component descriptors
                                                       (must be one of [json ndjson table tree yaml]) (default table)
   -r, --repository string                             repository ref (default "transport-archive")
-      --skip-reference-digest-processing              skip digest processing for resources and sources. Any resource referenced via access type will not have their digest updated.
+      --skip-reference-digest-processing              skip digest processing for resources and sources. Any resource referenced via access type will not have their digest updated, and any digest specified in the constructor will not be verified against the referenced content.
```
fabianburth:
```suggestion
      --skip-reference-digest-processing              skip digest processing for resources. Any resource referenced via access type will not have their digest updated, and any digest specified in the constructor will not be verified against the referenced content.
```

same here

fabianburth (commented):
Mostly nits, the log might be a bit over-engineered ;)

cli/cmd/add/component-version/cmd.go
morri-son: ok, took Fabian's proposal and skipped the function completely

frewilhelm: If I think about it again it really seems off to display a warning after the user explicitly set `--skip-reference-digest-processing` :D

frewilhelm: what was the reason to display that warning in the first place?

morri-son: Ok, guess need to explain my first intention a bit more and also propose changing back to my former check :) The skip-reference-digest-processing flag before this use case was only used for performance reasons to skip downloading resources for hashing. No we also introduce it for the constructor and there it gets a security related spin. What about:

1. Restore the conditional warning. We have two cases, skip for performance (don’t download the referenced resources to hash them) vs skip verification of digest(s) the author of the constructor has given (and in addition  decided to skip it using the flag) The unconditional warning is wrong noise for the first case, ok. I would do a  check, so users of the first case will not be bothered by the warning, for the second case I would add the warning, because it touches security, first you set a digest, then you overwrite it. Worth to flag, or?

2. Rename `skip-reference-digest-processing` to `skip-resource-digest-processing`. The word "reference" was accurate when the flag only skipped hash computation resources of type access. Now that it also covers digests in the constructor that have been explicitly set, the name is a bit off. The new name covers both cases without implying a scope. I would introduce the new flag and set the other to deprerated in cobra.

frewilhelm: > first you set a digest, then you overwrite it. Worth to flag, or?

Do you mean a digest set in the constructor by the user? I don't think this should ever be overwritten.

I have the impression I am missing the bigger picture here. Maybe we can sync up about this.

morri-son:
@frewilhelm @fabianburth 

You're right that a flag a user deliberately has set should be honored. I have dropped the unconditional warning, so a plain skip with no declared digests now stays silent.

What is left is a warning that surfaces - in my point of view - a contradiction: the flag says "don't verify," but if an author  declared digests in the constructor it means "verify this content." We don't drop that digest when skipping, we copy it into the descriptor **unverified**.

One can even sign over it and when we verify it later, no re-checks against the resource happens. During construction is the only place that we do so. So we would ship a signed component and asserting a digest nobody validated. In real life these choices are often in different stages or done by different people, e.g. an author pins the digest and a CI may skip the flag. so the warning is the only place this contradiction is made visible

frewilhelm:
> During construction is the only place that we do so.

I am not sure about this. The [contract](https://github.com/open-component-model/open-component-model/blob/8aae664b345f70d62746766f9f111f265ddf1709/bindings/go/repository/interface.go#L112) for `DownloadResource` states

> DownloadResource downloads and verifies the integrity of a [descriptor.Resource] from the repository.

Hence, I assume that after downloading (or preferably while downloading with `VerifyReader`), we should check if the digest of the downloaded resource matches the digest from the descriptor.Resource. That digest is either set by the user or while constructing the component descriptor.

The controllers already do this verification in the [`Resource`](https://github.com/open-component-model/open-component-model/blob/8aae664b345f70d62746766f9f111f265ddf1709/kubernetes/controller/internal/controller/resource/resource_controller.go#L421) controller where `ProcessResourceDigest` is [called](https://github.com/open-component-model/open-component-model/blob/8aae664b345f70d62746766f9f111f265ddf1709/kubernetes/controller/internal/ocm/resource.go#L58) (if the respective plugin is available). They do that because, we do not download the resource, so we need to verify the integrity ourselves before setting the status. The `Deployer` controller, however, relies on the `DownloadResource` contract.

morri-son:
@frewilhelm 

> During construction is the only place that we do so.
>> am not sure about this. The [contract](https://github.com/open-component-model/open-component-model/blob/8aae664b345f70d62746766f9f111f265ddf1709/bindings/go/repository/interface.go#L112) for DownloadResource states

>DownloadResource downloads and verifies the integrity of a [descriptor.Resource] from the repository.

That contract is just stated in the docstring, sadly, not implemented in the CLI. In the resource controller we would find such a mismatch. But if a digest makes it into the descriptor - which can happen when using the flag - the CLI does not check `res.Digest` against the real digest of the resource. And since this is the case, you can even sign such a descriptor and a verify would just verify the signature, not the resources and their digest.

As mentioned above there is a contradiction in a flag that says "don't verify," but the author declared digests which means "verify this content.". And these stages can even be executed by different people, one adds digests and the other just skips them using the flag. I still think that we must surface this.

frewilhelm:
> the CLI does not check res.Digest against the real digest of the resource.

I don't think the CLI needs to check this but the resource download which should be aware of the resource spec and by this the digest IIRC. If the digest mismatches, we need to error out.

> author declared digests which means "verify this content."

IMO this does not mean it must be done at construction time when the user explicitly sets `FlagSkipReferenceDigestProcessing`. It can be done at a later point, for example when downloading it.

I believe I understand where you are coming from. A user sets the digests in the constructor because the user wants them to be verified. But he cannot set the skip-flag which explicitly states that they digests are not verified now and then expect them to be verified.

I think our opinions differ on when and where the digests need to be verified. I don't think it must be at construction time.  It can also be while downloading and I don't think we can skip it there (and we definitely should not!).

morri-son:
@frewilhelm 
> I think our opinions differ on when and where the digests need to be verified. I don't think it must be at construction time. It can also be while downloading and I don't think we can skip it there (and we definitely should not!).

Similar to the resource controller, the `ocm download resource` definitely requires a verification of digests that have simply been transferred from constructor to descriptor because of the flag. That is worth another PR. My point here is not about that verification, it is just about making the mismatch of intends visible, now that we can add such digests. The flag was introduced before this scenario was available.

jakobmoellerdev:
we will rework this with a better digest handling

PR #3230, opened 2026-07-30
test(wget-integration): verify constructor resource digest with wget access

morri-son:
Follow-up to #3202. Part of open-component-model/ocm-project#1199.

Adds an integration test that drives a wget access resource through the constructor with a digest processor, covering:
- a matching author-supplied digest passes and is recorded on the descriptor
- a mismatching digest causes construction to fail


matthiasbruns (approved):

morri-son:
Closing — the behaviour this test exercises is already covered by existing unit tests: `TestProcessResourceDigest` (with digest mismatch) in `wget/repository`, the constructor carrying the digest field through `ConvertToDescriptorResource` in `construct_resource_test.go`, and the field round-trip tests added in #3202. No gap left uncovered.

PR #3231, opened 2026-07-30
fix(ci): auto-apply a kind/* label for docs, test, perf, and refactor PRs

morri-son:
## Problem

Our `verify-labels` workflow requires every PR to carry a `kind/*` label. But the title auto-labeler:

- mapped `docs`/`test`/`perf` to an `area/*` label only, with no `kind/*`, and
- did not recognise `refactor` as a type at all, so `refactor:` titles failed title validation.

As a result, `docs:`/`test:`/`perf:` PRs failed `verify-labels` until someone added a `kind/*` by hand, and `refactor:` PRs were rejected outright.

## Fix

Let a Conventional Commit type map to a list of labels, then:

- `docs` → `[kind/chore, area/documentation]`
- `test` → `[kind/chore, area/quality]`
- `perf` → `[kind/chore, area/quality]`
- `refactor` → `kind/chore` (new; `refactor:` titles are now valid and yield a `kind/*`)

`feat`/`fix`/`chore` are unchanged. Every valid PR title now yields a `kind/*` automatically. `docs` keeps its `area/documentation` label; `test`/`perf` get the org-wide `area/quality` (labels updated in open-component-model/.github#103). `deriveLabels` de-duplicates so a type's list and a scope cannot double-add the same label.

## Changes

- `.github/scripts/label-from-title.js`: accept `string | string[]` per type/scope, push all, de-dup.
- `.github/workflows/pull-request.yaml`: `TYPE_TO_LABEL` updated as above.
- `.github/scripts/label-from-title.test.js`: mirrored map updated, added cases for multi-label type, `refactor`, and de-dup.

## Verification

`node --test .github/scripts/label-from-title.test.js` passes (17 tests). Ran the real workflow map through the real script: `feat/fix/chore/docs/test/perf/refactor` all produce a `kind/*` in the set `verify-labels` accepts.

Note for reviewers: this maps `docs`/`test`/`perf` and now also `refactor` to `kind/chore`, so those land under "chore" in any label-driven release notes. Say if a different kind is preferred for any of them.


a-thaler:
nice improvement! I think we should support all the conventional commit type prefixes as this is expected to work for external contributions.
However the mapping to kind and area labels is quite up to the project setup. I personally would like to condense the label set to a meaningful and concise set. I think I removed already the kind/refactor label in the past as it simply was not in use. So please validate once more if the kind labels are in a good shape and see that the mapping fits.  Maybe we can even get rid of the testing and performance area by moving it all into the area/quality?

a-thaler (changes requested):
See my general comment

morri-son:
@a-thaler 
> See my general comment

got it :) Then I propose that you clean up the labels the way you want to have them and once we locked that, we can use this PR to pick it up. Makes sense?

matthiasbruns:
@morri-son can you take care of this once you are back? this has been open for quite some time

.github/scripts/label-from-title.test.js
```diff
@@ -9,9 +9,10 @@ const maps = {
     feat: "kind/feature",
     fix: "kind/bugfix",
     chore: "kind/chore",
-    docs: "area/documentation",
-    test: "area/testing",
-    perf: "area/performance",
+    refactor: "kind/refactor",
+    docs: ["kind/chore", "area/documentation"],
+    test: ["kind/chore", "area/testing"],
+    perf: ["kind/chore", "area/performance"],
```
a-thaler:
```suggestion
    refactor: "kind/chore",
    docs: ["kind/chore", "area/documentation"],
    test: ["kind/chore", "area/quality"],
    perf: ["kind/chore", "area/quality"],
```

I updated the labels in the org, please adjust
See https://github.com/open-component-model/.github/pull/103

.github/workflows/pull-request.yaml
```diff
@@ -33,9 +33,10 @@ jobs:
               "feat":"kind/feature",
               "fix":"kind/bugfix",
               "chore":"kind/chore",
-              "docs":"area/documentation",
-              "test":"area/testing",
-              "perf":"area/performance"
+              "refactor":"kind/refactor",
+              "docs":["kind/chore","area/documentation"],
+              "test":["kind/chore","area/testing"],
+              "perf":["kind/chore","area/performance"]
```
a-thaler:
```suggestion
              "refactor":"kind/chore",
              "docs":["kind/chore","area/documentation"],
              "test":["kind/chore","area/quality"],
              "perf":["kind/chore","area/quality"]
```

a-thaler (dismissed):

.github/scripts/label-from-title.js
```diff
@@ -62,17 +62,34 @@ export function deriveLabels(prTitle, { typeToLabel, scopeToLabel, breakingLabel
...
+    if (typeof value === "string") {
+      labels.push(value);
+    } else if (Array.isArray(value)) {
+      for (const label of value) {
+        if (typeof label === "string") {
+          labels.push(label);
+        }
+      }
+    }
+  };
```
Skarlso: I'm not sure I understand why this is suddenly a list or why it is a string? Like, how is this being utilized? Why not just use a list all the time? The push doesn't matter if it has one item or several.

Skarlso: So I think this isn't right like this. :) 

```js
if (breaking) {
  labels.push(breakingLabel);
}
if (type) {
  labels.push(...(types[type] ?? []));
}
if (scope) {
  labels.push(...(scopes[scope] ?? []));
}

return { valid: true, labels: [...new Set(labels)] };
```

And then the pull-request yaml should be just:

```yaml
"feat":["kind/feature"],
"fix":["kind/bugfix"],
"chore":["kind/chore"],
"refactor":["kind/chore"],
"docs":["kind/chore","area/documentation"],
"test":["kind/chore","area/quality"],
"perf":["kind/chore","area/quality"]
```

Don't mix and match.

Skarlso (changes requested):

.github/scripts/label-from-title.js
```diff
@@ -39,16 +39,17 @@ export function nullProtoMap(source) {
  *
  * @param {string} prTitle - The pull request title.
  * @param {object} maps
- * @param {Record<string,string>} maps.typeToLabel - type -> label.
- * @param {Record<string,string>} maps.scopeToLabel - scope -> label.
+ * @param {Record<string,string[]>} maps.typeToLabel - type -> labels.
+ * @param {Record<string,string[]>} maps.scopeToLabel - scope -> labels.
```
Skarlso, line 43: It's just `string[]`. :)

.github/scripts/label-from-title.js
morri-son: jaja, totally right. I took over your proposal and removed the `add()` helper, config and tests now only use lists.

.github/scripts/label-from-title.js
morri-son: ?

a-thaler (approved):

jakobmoellerdev (approved):

.github/scripts/label-from-title.js
Skarlso: This is what I meant. :) Sorry. The record is now JUST a string[]. It's no longer a simple string. It's always a list.

Skarlso: No, sorry. :D It's the key's type. My bad. This is fine. :)

Skarlso (approved):

PR #2595, opened 2026-05-21
fix(sigstore): introduce versioned Sigstore algorithm identifier

morri-son:
Replaces the bare `"sigstore"` algorithm string with a typed, versioned identifier `"Sigstore/v1alpha1"` so future Sigstore implementation generations can coexist.

Pattern follows `bindings/go/rsa/signing/v1alpha1`:
- `SignatureAlgorithm` typed string + `AlgorithmSigstoreV1Alpha1` + `AlgorithmSigstoreDefault`.
- `SignConfig.SignatureAlgorithm` optional; default via `GetSignatureAlgorithm()`.
- Future Sigstore versions live side-by-side. When cosign ships a breaking change, we add `AlgorithmSigstoreV1Alpha2` next to `AlgorithmSigstoreV1Alpha1` and accept both on verify, so old signatures keep working. A compatible bundle MediaType bump, e.g. v0.4, doesn't need a new algorithm version, we can just extend the accepted MediaType check. Same pattern as used in rsa/signing/v1alpha1.
- Verify validates `signed.Signature.Algorithm` strictly: must be set, must be known, MediaType must be acceptable.
- `AlgorithmSigstoreLegacy = "sigstore"` is accepted on verify as a backwards-compatible alias for `AlgorithmSigstoreV1Alpha1`. Sign always emits the canonical value; legacy is canonicalized in `GetSignatureAlgorithm`.

## Test plan
- [x] `go test ./...` in `bindings/go/sigstore`
- [x] `golangci-lint run` clean
- [x] CI green

PR #3236, opened 2026-07-30
fix(sigstore): drop algorithm attribute from consumer identity

### LINKED ISSUE #3237: bug(sigstore): signing requires undocumented `algorithm` in credential config
## What

Sigstore keyless signing fails unless the `.ocmconfig` consumer identity carries `algorithm: Sigstore/v1alpha1`, which the tutorials and reference docs do not mention (and explicitly say is not needed).

## Repro

`.ocmconfig` consumer as documented (no `algorithm`):

```yaml
- identity:
    type: SigstoreSigner/v1alpha1
    signature: sigstore-signature
  credentials:
    - type: OIDCIdentityTokenProvider/v1alpha1
```

```
ocm sign cv --signature sigstore-signature --config .ocmconfig-signing --signer-spec ./sigstore-signer-spec.yaml ...
Error: signing failed: OIDC identity token required: set SIGSTORE_ID_TOKEN env var, ...
```

Adding `algorithm: Sigstore/v1alpha1` to the identity makes it work.

## Cause

The versioned `Sigstore/v1alpha1` change (#2595) started emitting an `algorithm` attribute on the signer/verifier credential consumer identity. Credential matching is exact-equality, so the documented `{type, signature}` config no longer matches and no OIDC credential is resolved — surfacing as the misleading "OIDC identity token required" error. The attribute is write-only (nothing reads it) and sigstore has a single signing method, so it only breaks matching.

## Expected

The documented minimal `{type, signature}` consumer resolves the credential; no `algorithm` field required.

morri-son:
## What

The Sigstore consumer identity no longer carries an `algorithm` attribute.

## Why

The versioned `Sigstore/v1alpha1` change (#2595) started emitting `algorithm` on the signer/verifier credential consumer identity. Credential matching is using an exact match, so this silently forced every `.ocmconfig` consumer to also carry `algorithm: Sigstore/v1alpha1`, otherwise the lookup returned no credentials and failed with a misleading `OIDC identity token required` error.

Sigstore has a single signing method (like GPG), so the attribute carries no information and only breaks matching. Before #2595 the minimal `{type, signature}` config worked, and the tutorials/reference docs still document it that way.

## Change

- Stop emitting `algorithm` on sign and verify
- Remove the now dead field and regenerate schemas
- Restore the consumer-identity test expectations #2595 had flipped, and add a table-driven test asserting the documented minimal config matches

The signature envelope still records `Algorithm=Sigstore/v1alpha1`, so verification and the legacy `sigstore` alias are unaffected. Docs need no change.

> Note: any `.ocmconfig` that added `algorithm: Sigstore/v1alpha1` as a workaround must drop it again, since the extra attribute now breaks the match.

Closes #3237


matthiasbruns:
Isn't this a breaking change?
Identities with algo will not match the new identity anymore.

morri-son:
@matthiasbruns
> Isn't this a breaking change? Identities with algo will not match the new identity anymore.
Technically yes, but it is just reverting what [#2595](https://github.com/open-component-model/open-component-model/pull/2595) wrongly introduced. The signing and verification ADR and all docs don't specify the algorithm as part of the identity. So, I would not consider this as a breaking change that we need to mention, but a bug fix to get back to the desired and documented state.

matthiasbruns (approved):

piotrjanik (approved):

PR #3241, opened 2026-07-31
fix(cli): remove downloadName label, --output wins, fallback to resource name

### LINKED ISSUE #1238: bug(cli): downloadName label silently overrides --output, --output should win
## Current behavior

`ocm download resource` checks the downloadName label before inspecting the --output flag. A resource carrying the label causes --output to be silently dropped, with only an INFO level log line as indication.

A user running:

```shell
ocm download resource ghcr.io/org/component:v1 --identity name=cli --output /tmp/ocm
```

against a resource with `downloadName: ocm.exe` will find the binary written to `./ocm.exe` in their current working directory, not `/tmp/ocm`. The only signal is an `Info`-level log line, invisible at the default log level.

## Expected behaviour

`--output`, when explicitly provided by the user, takes precedence. The `downloadName` label should be used as the default filename only when `--output` is absent. 

Proposed precedence:

1. --output (explicit user choice)
2. downloadName label on resource
3. Resource identity + media-type-derived extension (current fallback)

The `--output` flag description and the help text in `cmd.go` should be updated, too.

morri-son:
## What

`ocm download resource` checked the `downloadName` label before `--output`, silently dropping an explicit `--output` when the label was present.

The `downloadName` label is not part of the OCM spec and is removed from all OCM-owned components. This PR drops label handling entirely and fixes the `--output` precedence bug at the same time.

## Behavior

Output filename precedence:

1. `--output`, if provided (always wins)
2. Resource name (e.g. `cli`)

To download multiple variants of the same resource (e.g., different architectures), use `--output` to write each to a distinct path.

The previous fallback also attempted to append a file extension from Go's `mime` package. This was unreliable: `application/octet-stream` (the media type for OCM executables) resolved to `.bin`, and `text/plain` resolved to `.conf`. Both the extension guessing and its `init()` mime registration are removed.

## Changes

- `cli/cmd/download/resource/cmd.go`: drop label constants and label-lookup block, remove unreliable media-type extension guess and `init()` mime registration, update help text
- `cli/cmd/cmd_test.go`: replace label-centric tests with fallback naming and `--output` override tests
- `cli/docs/reference/ocm_download_resource.md`, `website/content/docs/how-to/download-resources-from-component-versions.md`, `.github/config/wordlist.txt`: remove label references, document new fallback and Windows `--output` pattern
- `.github/components/ocm-release/component-constructor.yaml`, `cli/Taskfile.yml`: strip `downloadName` labels from all CLI resources; add `--output` to `verify/ctf`

Fixes open-component-model/ocm-project#1238


cli/cmd/download/resource/cmd.go
```diff
@@ -259,7 +265,7 @@ func isTar(mediaType string) bool {
 }
 
 func processResourceOutput(output string, resource *descriptor.Resource, data blob.ReadOnlyBlob, identity string, logger *slog.Logger) (string, error) {
-	// Check for downloadName label
+	// Check for downloadName label; used as default filename when --output is not set.
```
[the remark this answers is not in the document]
morri-son: corrected

cli/cmd/download/resource/cmd.go
```diff
@@ -259,7 +264,12 @@ func isTar(mediaType string) bool {
...
 func processResourceOutput(output string, resource *descriptor.Resource, data blob.ReadOnlyBlob, identity string, logger *slog.Logger) (string, error) {
-	// Check for downloadName label
+	if output != "" {
+		logger.Info("using explicit --output", slog.String("output", output))
+		return output, nil
+	}
+
+	// Check for downloadName label; used as default filename when --output is not set.
 	for _, label := range resource.Labels {
 		if label.Name == "downloadName" {
```
fabianburth: while you are already at it - we currently have https://github.com/open-component-model/ocm-spec/pull/146

We might want to adjust that label name while we can (I did not even know this was a thing until now). You may also support both and add a deprecation warning for 1 or 2 releases.

.github/config/wordlist.txt
```diff
@@ -944,3 +940,6 @@ Neisener
 ilja
 iljaweis
 weis
+
```
fabianburth:
```suggestion
```

cli/cmd/download/resource/cmd.go
```diff
@@ -57,7 +57,12 @@ It supports optional transformation of the resource using a registered transform
 If no transformer is specified, the resource is written directly in its original format. If the media type is known,
 the appropriate file extension will be added to the output file name if no output location is given.
 
-Resources can be accessed either locally or via a plugin that supports remote fetching, with optional credential resolution.`,
+Resources can be accessed either locally or via a plugin that supports remote fetching, with optional credential resolution.
+
+The output filename is determined by the first of these that applies:
+  1. --output, if explicitly provided
+  2. A "downloadName" label on the resource, if present
+  3. The resource identity, optionally extended with a media-type-derived file extension`,
```
fabianburth: I think this is as ugly as it can get (nobody ever would keep a name with a `=` sign in it). Is downloading 2 resources with the same name really such a big concern? I'd propose just using the name and some fallback behavior.

fabianburth (dismissed):

cli/cmd/download/resource/cmd.go
morri-son: good point, didn't realize that what we did on the Hackathon also lead to a spec change. I added ocm.software/download-name as the default label, the downloadName got deprecated and emits a WARN when present. If both exist, the new one wins. We can decide if we want to enhance https://github.com/open-component-model/ocm-spec/pull/146 with that label right now or as follow-up.

cli/cmd/download/resource/cmd.go
morri-son: hehe, yes, the fallback was already there and ugly . My change just kept it and made the --output the default. I changed the fallback now to just the resource name and the extraIdentity as (sanitized) suffix if present and also dropped the  media-type-derived extension.

cli/cmd/download/resource/cmd.go
morri-son: label completely removed.

cli/cmd/download/resource/cmd.go
```diff
@@ -74,8 +72,7 @@ Resources can be accessed either locally or via a plugin that supports remote fe
 	}
 
 	cmd.Flags().String(FlagResourceIdentity, "", "resource identity to download")
-	cmd.Flags().String(FlagOutput, "", "output location to download to. If no transformer is specified, and no "+
-		"format was discovered that can be written to a directory, the resource will be written to a file.")
+	cmd.Flags().String(FlagOutput, "", "full output file path (directory + filename). Intermediate directories are created automatically.")
```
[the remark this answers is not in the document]
morri-son: done

morri-son:
@coderabbitai review

cli/cmd/download/resource/cmd.go
```diff
@@ -258,34 +246,39 @@ func isTar(mediaType string) bool {
...
+	output = fallbackFileName(resource)
+	logger.Warn("no output location specified, deriving output file name from the resource", slog.String("output", output))
 
 	return output, nil
 }
+
+// fallbackFileName derives a filesystem-safe filename from the resource name plus its
+// extra identity attribute values (sorted by key), so variants of the same name stay distinct.
+func fallbackFileName(resource *descriptor.Resource) string {
+	name := sanitizeFileName(resource.Name)
```
fabianburth: regex for our current resource names and extra identities: `^[a-z0-9]([-_+a-z0-9]*[a-z0-9])?$`
we do not need sanitize name, I believe

cli/cmd/download/resource/cmd.go
```diff
@@ -258,34 +246,39 @@ func isTar(mediaType string) bool {
...
+	// Fallback: resource name, with extra identity attributes appended when present.
+	output = fallbackFileName(resource)
+	logger.Warn("no output location specified, deriving output file name from the resource", slog.String("output", output))
 
 	return output, nil
 }
+
+// fallbackFileName derives a filesystem-safe filename from the resource name plus its
+// extra identity attribute values (sorted by key), so variants of the same name stay distinct.
+func fallbackFileName(resource *descriptor.Resource) string {
```
fabianburth: The name might still become quite complex and since you only encode the key values, we might still have collisions. We could also simply use the resource name and if you need further distinction, use `--output`.

cli/cmd/download/resource/cmd.go
morri-son: removed the sanitization

cli/cmd/download/resource/cmd.go
morri-son: kept is easy and followed your proposal: use resource name, for distiction use --output

matthiasbruns (dismissed):

cli/cmd/download/resource/cmd.go
```diff
@@ -54,10 +51,11 @@ func New() *cobra.Command {
...
 It supports optional transformation of the resource using a registered transformer plugin.
 
-If no transformer is specified, the resource is written directly in its original format. If the media type is known,
-the appropriate file extension will be added to the output file name if no output location is given.
+If no transformer is specified, the resource is written directly in its original format.
 
-Resources can be accessed either locally or via a plugin that supports remote fetching, with optional credential resolution.`,
+Resources can be accessed either locally or via a plugin that supports remote fetching, with optional credential resolution.
+
+When --output is not provided, the output filename is the resource name.`,
```
frewilhelm, line 58: The PR description says

> Resource name, with sanitized extra-identity values appended when present

which extends this one. So, will extra-identity values be appended?

cli/cmd/download/resource/cmd.go
```diff
@@ -85,15 +82,6 @@ Resources can be accessed either locally or via a plugin that supports remote fe
...
 
-func init() {
-	if err := errors.Join(
-		mime.AddExtensionType(".tar.gz", layout.MediaTypeOCIImageLayoutTarGzipV1),
-		mime.AddExtensionType(".tar", layout.MediaTypeOCIImageLayoutTarV1),
-	); err != nil {
-		panic(err)
-	}
-}
-
```
frewilhelm, line 96: what was the purpose of this?

cli/cmd/download/resource/cmd.go
```diff
@@ -258,34 +246,15 @@ func isTar(mediaType string) bool {
...
-				if extensions, err := mime.ExtensionsByType(mediaType); err == nil && len(extensions) > 0 {
-					output += extensions[0]
-				}
-			}
-		}
-		logger.Warn("no output location specified, using resource identity as output file name", slog.String("output", output))
-	}
+	// Fallback: resource name.
+	output = resource.Name
+	logger.Warn("no output location specified, using resource name as output file name", slog.String("output", output))
```
frewilhelm: I am not sure about these log-levels

The first one is `info` and just repeating what the user should know as the user just set `--output` and the second one is a warning about our default behaviour.

If you want to keep the logs, I would downgrade them to `debug`

cli/Taskfile.yml
```diff
@@ -122,9 +122,6 @@ tasks:
         - name: cli
           type: executable
           version: {{ .VERSION }}
-          labels:
-            - name: downloadName
-              value: ocm
```
frewilhelm, line 127: Do we still use this?

website/content/docs/how-to/download-resources-from-component-versions.md
```diff
@@ -365,6 +365,23 @@ tar tvf helmchart-tgz/podinfo-6.8.0.tgz
 
 {{< /details >}}
 
+## Controlling the Output Filename
+
+Use `--output` to set the destination path explicitly. This is required whenever the default name is unsuitable — for example, Windows executables must carry a `.exe` extension to be runnable:
```
frewilhelm: > Windows executables must carry a `.exe` extension to be runnable

thats not necessarily true I believe

website/content/docs/how-to/download-resources-from-component-versions.md
```diff
@@ -365,6 +365,23 @@ tar tvf helmchart-tgz/podinfo-6.8.0.tgz
...
+
+```shell
+ocm download resource <repo>//<component>:<version> \
```
frewilhelm: Why don't you use our component version? this might be a good fit

cli/cmd/download/resource/cmd.go
morri-son: my fault, description not updated. We dropped extra-identity appending and sanitization following Fabian's feedback. The fallback is now just the resource name.

cli/cmd/download/resource/cmd.go
morri-son: That's the init() function as part of the fallback. It registered OCI image layout media types with Go's mime package so the old fallback could look up file extensions by media type, e.g. .tar.gz for an OCI layout. I removed that guessing for an extension entirely.

cli/cmd/download/resource/cmd.go
morri-son: Agree. The Info for --output is noise, the user just set it. The Warn for expected default behavior signals a problem where there isn't one. Will drop both to debug level

website/content/docs/how-to/download-resources-from-component-versions.md
morri-son: Ja, you can invoke a binary on Windows by full path without .exe. The .exe is required for PATH-based discovery and Explorer. I will change the wording.

website/content/docs/how-to/download-resources-from-component-versions.md
morri-son: Good idea. Will update the examples to use it.

cli/Taskfile.yml
morri-son: No. The PR removes downloadName everywhere, from Taskfile, from component-constructor.yaml and from cmd.go label-reading logic.

frewilhelm: I meant the task `build:target`

cli/cmd/download/resource/cmd.go
frewilhelm: ah I see. should we still do this if `--output` is not specified? I think it makes kind of sense because no one will name their resource `oci-image.tar.gz`

frewilhelm: but maybe its also a bit of unexpected magic

cli/Taskfile.yml
morri-son: I don't get your point :) The build:target is still there, we just removed the `doanloadName` label: https://github.com/morri-son/open-component-model/blob/9a45211ba8621c8a744eb7216855706e009fe0a1/cli/Taskfile.yml#L106C16-L107C17. Or am I still getting something awfully wrong?

morri-son: I don't get your point :) The build:target is still there, we just removed the `doanloadName` label: https://github.com/morri-son/open-component-model/blob/9a45211ba8621c8a744eb7216855706e009fe0a1/cli/Taskfile.yml#L106C16-L107C17. Or am I still getting something awfully wrong?

jakobmoellerdev: is the build target task still used? the question was if its used in workflows

morri-son: yes, it is still used  in `build`, `build/multiarch` and `build/single` and  is what the matrix build job in  `.github/workflows/cli.yml` invokes. This PR only removed the downloadName label line from the resource yaml

matthiasbruns (approved):
Tested with my magic `.sh` file - works

cli/Taskfile.yml
frewilhelm: The build target task is still used but the resource template is not used imo

jakobmoellerdev (approved):

PR #3270, opened 2026-08-03
chore(cli): integration tests for the gitHub access type in cli

### LINKED ISSUE #1070: The GitHub access method and implementation are integrated with the transformer
**Description**

Wire the GitHub get transformer (from #1069) into the transformation DAG and add the blob and add transformers. Depends on #1069.

- **Blob transformer**: converts the downloaded tarball → OCI layout for OCI registry targets (follow [`bindings/go/oci/transformer/`](../bindings/go/oci/transformer/)) - **We don't think we use blob transformers for this, but to be verified**
- **Add transformer**: uploads the processed blob to the target repository
- Wire together in a transformer chain via [`bindings/go/transform/graph/builder/`](../bindings/go/transform/graph/builder/)

**Done Criteria**
- [ ] Blob transformer: tarball → OCI layout conversion
- [ ] Add transformer: upload to target repository
- [ ] Full pipeline wired in the transform DAG: get → blob → add
- [ ] Integration test: transfer a GitHub resource end-to-end to an OCI registry
- [ ] End-user documentation for `gitHub` access type added

piotrjanik:

Adds an end-to-end test for the gitHub access type: a CTF to OCI registry transfer against live github.com.

It pins that copying embeds the archive as a local blob keeping the digest the source CTF recorded, that the bytes are the repository at the pinned commit, that the default copy mode keeps the access external, and that a ref-only resource is refused. The same archive is declared as a source, which stays remote either way.

Set GITHUB_TOKEN to lift GitHub's anonymous cap of 60 requests per hour.

Fixes: https://github.com/open-component-model/ocm-project/issues/1070


cli/integration/transfer_github_integration_test.go
```diff
@@ -0,0 +1,284 @@
...
+
+	// addToCTF adds the constructor to a CTF in its own directory and returns the
+	// stored component version's reference and the CTF path.
+	addToCTF := func(t *testing.T, constructor string, extraArgs ...string) (ref, ctf string) {
+		t.Helper()
+		r := require.New(t)
+
+		dir := t.TempDir()
+		constructorPath := filepath.Join(dir, "constructor.yaml")
+		r.NoError(os.WriteFile(constructorPath, []byte(constructor), os.ModePerm))
```
morri-son: In a temp folder not really security relevant, but maybe for consistency with how you do it e.g. for the config file, use 0o600?

cli/integration/transfer_github_integration_test.go
```diff
@@ -0,0 +1,284 @@
...
+}
+
+// Test_Integration_Transfer_GitHub transfers a resource with a
+// GitHub access from a CTF to an OCI registry.
+//
+// A source archive has no OCI-artifact representation and no digest handler for
+// its media type, so — unlike a helm chart — it can only travel as a local
+// blob, and its digest is the digest of the exact archive bytes.
+//
+// The component version declares the same archive as an OCM source as well.
```
fabianburth: what?

cli/integration/transfer_github_integration_test.go
```diff
@@ -0,0 +1,284 @@
...
+		// pinning it to a commit is exactly that processor's job.
+		refOnlyRef, _ := addToCTF(t,
+			githubConstructor(internal.GitHubRepoURL, "ref", internal.GitHubRef),
+			"--skip-reference-digest-processing")
+
+		_, err := runTransfer(t, refOnlyRef, "github-transfer-refonly", "--copy-resources")
+		r.Error(err, "transferring a ref-only github resource must fail")
+		r.ErrorContains(err, "no pinned commit",
+			"the failure must name the missing commit, since the embedded bytes would not be reproducible")
+	})
```
fabianburth, line 283: is this really desired behaviour? generally looks weird that we accept skipping digest but then dont allow you to work with the artifact any more.

cli/integration/transfer_github_integration_test.go
```diff
@@ -0,0 +1,284 @@
...
+		// archive that hashes consistently.
+		archivePath := filepath.Join(t.TempDir(), "archive.tar.gz")
+		r.NoError(os.WriteFile(archivePath, stored, 0o600))
+		internal.AssertGitHubArchiveAtCommit(t, archivePath)
+
+		// --copy-resources has no counterpart for sources, so the source keeps
+		// pointing at github even though the resource beside it was embedded.
+		// Sources carry no digest either, so nothing about them is verifiable.
+		r.Equal("GitHub/v1", desc.Component.Sources[0].Access.GetType().String(),
+			"--copy-resources must not embed a source")
```
fabianburth, line 246: should we use the same pattern as here? 

```go
		var localBlobAccess v2.LocalBlob
		r.Error(v2.Scheme.Convert(res.Access, &localBlobAccess),
			"a github access must not convert to a local blob")
```

fabianburth (commented):
maybe a bit too many function closures, but looks good generally!

fabianburth (dismissed):

cli/integration/transfer_github_integration_test.go
piotrjanik: I assumed that transferring CVs without a commit shouldn't be permitted, as this makes the resource non-deterministic.

cli/integration/transfer_github_integration_test.go
piotrjanik: @fabianburth done

cli/integration/transfer_github_integration_test.go
piotrjanik: done

fabianburth (approved):

Skarlso (approved):

PR #3275, opened 2026-08-04
fix(sigstore): update integration test setup for scaffolding v0.7.37 (Rekor v2)

morri-son:
Fixes the integration test failure in #3272 and bumps sigstore/scaffolding to **v0.7.37** (Rekor v1/Trillian → Rekor v2/tiles).

v0.7.36 shipped a stale setup script referencing removed assets (the 404 in #3272); v0.7.37 fixed it (its release asset is byte-identical to upstream `main`). We keep downloading the script from the release rather than vendoring a fork.

### Changes

- **`.env`**: `SCAFFOLDING_VERSION` bumped to `v0.7.37`
- **`Taskfile.yml`**: the tiles path is untested upstream and aborts under `set -e`. Use `sed` to patch the downloaded script (same way sigstore's own `actions/setup` does) for two bugs:
  - the redundant first Rekor apply fails base64 on placeholder Secrets -> made non-fatal
  - rekor/ctlog lost their provisioning Jobs, so `wait ... jobs --all` errors "no matching resources" -> apply upstream's own no-jobs guard (which they added only to fulcio 🧐)
- **`hack/extract-sigstore-env.sh`**: Rekor v2 (tiles) has no `/api/v1` endpoint, so read the pubkey from the `rekor-pub-key` secret and set `api-version=2`. `cosign trusted-root create` derives the tlog `logId` as `SHA256(pubkey)`, which is wrong for a tiles log (keyed by its checkpoint note keyhash) -> verification fails with "not enough verified log entries". So we splice the `tlogs` entry from the scaffolding's TUF-published trusted root into ours (Fulcio/CTLog/TSA material stays)
  - the spliced root is now located by **deriving** cosign's per-mirror cache path (reproducing sigstore-go's `URLToPath`) instead of `find ... | head -1`, so it can't pick an unrelated cached root (e.g. a public-good root or a stale run with a different bridge port) and fails loudly if the cache is missing
- **`ci.yml`**: every image in the scaffolding stack has a native `linux/arm64` variant, so drop the `ubuntu-latest` override and use the cheaper `ubuntu-24.04-arm`

### Validation

Full stack + entire suite pass locally on my colima (using kind and only arm64) and in CI on the arm64 runner, including the previously failing `Test_Integration_VerifyWithExplicitTrustedRoot`.

The `Taskfile.yml` `sed` patches are marked to be removed once upstream fixes the tiles path.


bindings/go/sigstore/integration/.env
```diff
@@ -1,2 +1,2 @@
 # renovate: datasource=github-releases depName=sigstore/scaffolding
-SCAFFOLDING_VERSION=v0.7.31
+SCAFFOLDING_VERSION=v0.7.37
```
Skarlso, line 2: This will just be bumped to 38, right?

Skarlso (commented):
This looks okay, but unsure about that giant `sed`. There is a comment says sigstore is fixing it the same way. Where does that happen? Can you link to it?

bindings/go/sigstore/integration/.env
morri-son: v.0.7.37 is the latest version, so this should be fine. Renovate just created #3283 to bump it to that version.

morri-son:
@Skarlso 
> This looks okay, but unsure about that giant `sed`. There is a comment says sigstore is fixing it the same way. Where does that happen? Can you link to it?

Tried to explain all sed commands in details and also added the link to sigstore upstream repo in the last [commit](fa8be0294bf75580cf854273f8e1d82d5c03fbd6).

frewilhelm (commented):
I can't run it locally. Can you help me find out what is going wrong? (I replaced the values for the placeholders with `...`)

```
...
[bindings/go/sigstore/integration:scaffolding:setup] ++ base64 -w0
[bindings/go/sigstore/integration:scaffolding:setup] + sed -i -e 's/<private-placeholder>/.../' -e 's/<password-placeholder>/.../' /var/folders/my/cxvsk48d6_z_ncvslvkqc_r00000gn/T/fulcioBM1
[bindings/go/sigstore/integration:scaffolding:setup] sed: -e: No such file or directory
[bindings/go/sigstore/integration:scaffolding:setup] + cleanup
[bindings/go/sigstore/integration:scaffolding:setup] + eval 'cleanup_rekor ; cleanup_fulcio'
[bindings/go/sigstore/integration:scaffolding:setup] ++ cleanup_rekor
[bindings/go/sigstore/integration:scaffolding:setup] ++ rm /var/folders/my/cxvsk48d6_z_ncvslvkqc_r00000gn/T/tmp.3LsrRmy8Kk/pub.pem /var/folders/my/cxvsk48d6_z_ncvslvkqc_r00000gn/T/tmp.3LsrRmy8Kk/key.pem
[bindings/go/sigstore/integration:scaffolding:setup] ++ cleanup_fulcio
[bindings/go/sigstore/integration:scaffolding:setup] ++ rm /var/folders/my/cxvsk48d6_z_ncvslvkqc_r00000gn/T/tmp.opvOw0jIpr/cert.pem /var/folders/my/cxvsk48d6_z_ncvslvkqc_r00000gn/T/tmp.opvOw0jIpr/key.pem
task: [bindings/go/sigstore/integration:scaffolding:bridge] kubectl -n kourier-system port-forward service/kourier-internal 8080:80 &
for i in $(seq 1 20); do
  lsof -ti :8080 >/dev/null 2>&1 && break
  sleep 0.5
done
lsof -ti :8080 >/dev/null 2>&1 || { echo "port-forward failed to start" >&2; exit 1; }

[bindings/go/sigstore/integration:scaffolding:bridge] Forwarding from 127.0.0.1:8080 -> 8081
[bindings/go/sigstore/integration:scaffolding:bridge] Forwarding from [::1]:8080 -> 8081
task: Failed to run task "test/integration": task: Failed to run task "bindings/go/sigstore/integration:test/integration": task: Failed to run task "bindings/go/sigstore/integration:scaffolding:env": task: Failed to run task "bindings/go/sigstore/integration:scaffolding:setup": exit status 1
```

morri-son:
@frewilhelm classic MacOS issue. Some commands, like sed, behave differently under GNU and BSD. So, the easiest thing is a `brew install gnu-sed`and it should work. I added a precondition now to warn a user on MacOS (or other <> GNU using OS) to do that.

frewilhelm:
> @frewilhelm classic MacOS issue. Some commands, like sed, behave differently under GNU and BSD. So, the easiest thing is a `brew install gnu-sed`and it should work. I added a precondition now to warn a user on MacOS (or other <> GNU using OS) to do that.

Yeah, now it works locally. Thanks a lot!

frewilhelm (approved):

Skarlso (approved):

PR #3288, opened 2026-08-05
feat(github)!: require a pinned commit in the gitHub access

### LINKED ISSUE #1070: The GitHub access method and implementation are integrated with the transformer

piotrjanik:

A gitHub access could name only a ref. Refs move, so what a component version pointed at moved with them and a digest computed today would not verify tomorrow.

The commit is now required. Validate and the JSON schema reject an access without one, so a ref-only access fails at authoring time instead of mid-download. The ref stays optional and never selects what gets downloaded; a download resolves it only to warn that it has moved off the commit. Pinning a ref is the caller's job, via ResolveCommit.

Fixes: https://github.com/open-component-model/ocm-project/issues/1070


bindings/go/github/integration/integration_test.go
```diff
@@ -22,10 +22,9 @@ import (
 	"ocm.software/open-component-model/bindings/go/runtime"
 )
 
-// This repository, pinned at a published release tag. The ref-only cases below
-// assert that resolving the ref yields exactly ocmCommit, so the ref must never
-// move: a release tag of a published version is the strongest such guarantee a
-// live repository offers (branches advance, release tags do not).
+// This repository, pinned at a published release tag. The ResolveCommit case
+// below asserts the ref resolves to exactly ocmCommit, so the ref must never
+// move — a release tag never does, a branch would.
```
frewilhelm, line 27: > a release tag never does, a branch would.

It should never move, but it definitely can. This is just to test `ResolveCommit` right?

bindings/go/github/integration/integration_test.go
```diff
@@ -47,6 +46,13 @@ func testCredentials() runtime.Typed {
 	}
 }
 
+// testGitHubCredentials is testCredentials for the calls that take the concrete
+// type. The assertion yields nil for an anonymous run.
+func testGitHubCredentials() *credsv1.GitHubCredentials {
+	creds, _ := testCredentials().(*credsv1.GitHubCredentials)
+	return creds
+}
+
```
frewilhelm, line 55: Where is `testGitHubCredentials` used?

bindings/go/github/digest/digest_test.go
```diff
@@ -148,29 +148,15 @@ func TestDigestProcessor(t *testing.T) {
...
-			assert.Equal(t, testCommit, pinned.Commit, "commit must be pinned to the resolved sha")
-			require.NotNil(t, processed.Digest)
-			assert.Equal(t, payloadDigest, processed.Digest.Value)
-
-			orig := res.Access.(*v1.GitHub)
-			assert.Empty(t, orig.Commit, "input resource access must not be mutated")
-		})
-
-		t.Run("resolves the ref once, not again for the download", func(t *testing.T) {
+		t.Run("rejects a ref-only resource", func(t *testing.T) {
```
frewilhelm, line 151: I believe this test case is re-testing `gitHub.Validate()`. I don't think you need to do that again.

bindings/go/github/integration/integration_test.go
piotrjanik: well, if you create a cv with provided ref and commit in constructor, this ensures that at creation time reg pointed to a commit. Otherwise, you could provide `main` for a commit which has never been `HEAD` of main.

PR #3303, opened 2026-08-07
chore: Create release workflow for monolithic binding library 

### LINKED ISSUE #1207: Create release workflow for monolithic binding library
**Description**
Write a new release workflow that tags `bindings/go/vX.Y.Z` as a single version for the entire binding library. This replaces the per-module `release-go-submodule.yaml`.

Should be developed and tested on a fork against the monolithic structure before it lands upstream. The per-module workflow continues to exist for CLI and controller releases, only binding usage is replaced.

We should start with patch-releases `0.0.1` until we are ready to consolidate all the release workflows (bindings, cli, controller). We should guard against anything that is not a patch release

**Done Criteria**
  - [x] Workflow accepts bump type (major/minor/patch) and optional suffix
  - [x] Computes next version from latest bindings/go/v* tag (when we consolidate the release workflows we should derive it from the release branch)
  - [x] Generates changelog from commits since last tag
  - [x] Creates signed/annotated tag and pushes
  - [x] Dry-run mode works without side effects
  - [x] Tested on a fork

jneisener:

Add release flow for the monolithic bindings release. New flow is closely related to https://github.com/open-component-model/open-component-model/blob/main/.github/workflows/release-go-submodule.yaml, it mainly removes logic that was only necessary for the individual releases.

Some test runs in combination with https://github.com/jneisener/open-component-model/tree/chore/merge_go_mods_1208 are here: https://github.com/jneisener/open-component-model/actions/workflows/release-go-bindings.yaml

Here is a tagged commit https://github.com/jneisener/open-component-model/commit/18e9ff4e188a52b03436cf9888307809186cd01f (commit will not survive next force push, so here is a screenshot as well)
image
Release action (non dry-run): https://github.com/jneisener/open-component-model/actions/runs/31574007693/job/94041973813

Fixes https://github.com/open-component-model/ocm-project/issues/1207

##### Verification


.github/workflows/release-go-bindings.yaml
```diff
@@ -0,0 +1,174 @@
...
+          NEW_TAG: ${{ needs.version.outputs.new_tag }}
+          CHANGELOG_B64: ${{ needs.version.outputs.changelog_b64 }}
+        run: |
+          tag="${NEW_TAG}"
+          msg=$(echo "${CHANGELOG_B64}" | base64 -d)
+
+          printf "%s\n" "$msg" > .tagmsg
+          git tag -a "$tag" -F .tagmsg
+          git remote set-url origin "https://x-access-token:${GITHUB_TOKEN}@github.com/${GITHUB_REPOSITORY}.git"
+          git push origin "$tag"
```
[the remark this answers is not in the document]
jneisener: This is the exact state of the current release flow and thus changing it is out of scope for this PR

.github/workflows/release-go-bindings.yaml
```diff
@@ -0,0 +1,174 @@
...
+on:
+  workflow_dispatch:
+    inputs:
+      bump:
+        description: "Bump Mode (major, minor, patch, none)"
+        required: false
+        type: string
+        default: "patch"
+        #options: [patch, minor, major, none] # temporary guard, remove when confident in the release process
+        options: [patch]
```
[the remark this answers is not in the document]
jneisener: This is technically correct, but it works with `type: string` anyway as Github UI does not seem to mind

frewilhelm (commented):
I believe we will use this workflow only for bumping patch versions. When we want to integrate this into the lockstep-release-model this will change completely. Thus, I would propose to make it very simple here, but make sure that no major or minor versions are bumped

.github/workflows/release-go-bindings.yaml
```diff
@@ -0,0 +1,174 @@
...
+          NEW_TAG: ${{ steps.version.outputs.new_tag }}
+          CHANGELOG_B64: ${{ steps.changelog.outputs.changelog_b64 }}
+        run: |
+          echo "## Dry run – would create tag:" >> "$GITHUB_STEP_SUMMARY"
+          echo "${NEW_TAG}" >> "$GITHUB_STEP_SUMMARY"
+          echo "" >> "$GITHUB_STEP_SUMMARY"
+          echo "### Changelog" >> "$GITHUB_STEP_SUMMARY"
+          echo "${CHANGELOG_B64}" | base64 -d >> "$GITHUB_STEP_SUMMARY"
+
+      - name: Guard version ceiling
```
frewilhelm, line 123: This can be done earlier i believe

.github/workflows/release-go-bindings.yaml
frewilhelm, line 1: maybe we should add a comment that this is only a temporary solution until the lockstep release exists

Skarlso (commented):
Looks okay. :)

Skarlso (dismissed):

frewilhelm (commented):
overall it looks okay. it feels a bit off because, I believe, you just used a lot from the previous sub-module releases which is not needed for our use-case here. I think it is okay to merge it, knowing it will change in the near future

.github/workflows/release-go-bindings.yaml
jneisener: It already is done earlier via the choice options only having `patch`. This one is meant as a last stop before the actual release is triggered

jneisener:
> overall it looks okay. it feels a bit off because, I believe, you just used a lot from the previous sub-module releases which is not needed for our use-case here. I think it is okay to merge it, knowing it will change in the near future

Correct, thats what I meant with 

> Add release flow for the monolithic bindings release. New flow is closely related to [main/.github/workflows/release-go-submodule.yaml](https://github.com/open-component-model/open-component-model/blob/main/.github/workflows/release-go-submodule.yaml?rgh-link-date=2026-08-07T16%3A35%3A56Z), it mainly removes logic that was only necessary for the individual releases.

frewilhelm (approved):

Skarlso (approved):

PR #3274, opened 2026-08-04
docs(website): explain OCM_VERSION and side-by-side installs in CLI guide

morri-son:
The guide shows `OCM_VERSION` via the `site-version` shortcode on versioned pages but never explains the variable or how to control the install location. This PR documents both and updates `install-cli.sh` to support side-by-side installs via a new `OCM_BIN_NAME` environment variable (default: `ocm`), following the established Helm `get-helm-3` model (`HELM_INSTALL_DIR` + `BINARY_NAME`).

The positional argument stays an install directory; `OCM_BIN_NAME` overrides the installed filename, so multiple versions can live side by side in the same directory without any dir-vs-file path heuristic:

```shell
curl -sfL https://ocm.software/install-cli.sh | OCM_VERSION=0.12 OCM_BIN_NAME=ocm-v0.12 bash -s -- ~/.local/bin
```

Suggested by @fabianburth in #3246. Supersedes that draft.

- Closes #3260

PR #3328, opened 2026-08-10
fix(website): Fix PATH warning for macOS and expand BIN_DIR in install script

alexbass01:
**Note:** This is my first PR on this project — I've been here about a week. Happy to receive guidance on process, conventions, or anything I've missed!

Improves two small things in the setup script `website/static/install-cli.sh` that affect new macOS users:

1. **`BIN_DIR` printed literally** — the `ensure_path` warning used single quotes,
     preventing variable expansion. `${BIN_DIR}` printed as a literal string instead
     of the actual install path.

2. **Wrong shell config on macOS** — the warning pointed to `~/.profile`, but macOS
     defaults to zsh, which reads `~/.zshrc`. Users following the instructions would
     add the `PATH` entry to a file their shell never sources.

Before PR:
```
> wget -qO- https://ocm.software/install-cli.sh | bash
[INFO]  Downloading metadata https://api.github.com/repos/open-component-model/open-component-model/releases?per_page=100
[INFO]  Using 0.12.0 as release
[INFO]  Downloading binary https://github.com/open-component-model/open-component-model/releases/download/v0.12.0/ocm-darwin-arm64
[INFO]  Verifying binary attestation...
[INFO]  Attestation verification successful
[INFO]  Installing ocm to /Users/XXX/.local/bin/ocm
[WARN]  /Users/XXX/.local/bin is not in your PATH.
[WARN]  Add it by running:
[WARN]
[WARN]    echo "export PATH=${BIN_DIR}:$PATH" >> ~/.profile && source ~/.profile
[WARN]
[INFO]  OCM CLI v0.12.0 installed successfully

> echo $BIN_DIR # is empty

```

  N/A

  #### Testing

  ##### How to test the changes
Remove `~/.local/bin` from your path, make sure the `ocm` command cannot be found and run

```bash
 bash website/static/install-cli.sh
```

  The warning should now print the resolved install path and reference ~/.zshrc on macOS.

```bash
> bash install-cli.sh
[INFO]  Downloading metadata https://api.github.com/repos/open-component-model/open-component-model/releases?per_page=100
[INFO]  Using 0.12.0 as release
[INFO]  Downloading binary https://github.com/open-component-model/open-component-model/releases/download/v0.12.0/ocm-darwin-arm64
[INFO]  Verifying binary attestation...
[INFO]  Attestation verification successful
[INFO]  Installing ocm to /Users/XXX/.local/bin/ocm
[WARN]  /Users/XXX/.local/bin is not in your PATH.
[WARN]  Add it by running:
[WARN]
[WARN]    echo 'export PATH="$PATH:/Users/XXX/.local/bin"' >> "/Users/XXX/.zshrc" && source "/Users/XXX/.zshrc"
[WARN]
[INFO]  OCM CLI v0.12.0 installed successfully
```

  Verification

  - [ ] I have added/updated tests for my changes (see Test Requirements)
  - [ ] Tests pass locally (task test and task test/integration if applicable)
  - [ ] If touching multiple modules, go work is enabled (see go.work) — N/A, shell script only
  - [ ] My changes do not decrease test coverage — N/A, no tests exist for the install script
  - [ ] I have tested the changes locally by running ocm


website/static/install-cli.sh
```diff
@@ -102,10 +102,17 @@ ensure_path() {
...
+        shell_rc="${HOME}/.zshrc"
+    else
+        shell_rc="${HOME}/.profile"
+    fi
+
     warn "${BIN_DIR} is not in your PATH."
     warn "Add it by running:"
     warn ""
-    warn '  echo "export PATH=${BIN_DIR}:$PATH" >> ~/.profile && source ~/.profile'
+    warn "  echo 'export PATH=\"\$PATH:${BIN_DIR}\"' >> ${shell_rc} && source ${shell_rc}"
```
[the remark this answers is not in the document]
alexbass01: Implemented that

morri-son (dismissed):

website/static/install-cli.sh
```diff
@@ -102,10 +102,17 @@ ensure_path() {
             ;;
     esac
 
+    local shell_rc
+    if [[ "${OS}" == "darwin" ]]; then
```
frewilhelm: I am unsure if `OS` is always set. If I open a `zsh` on my macOS, it is not. Might something like `SHELL` be a better option? (be aware that I don't use `zsh` and there might be a better way)

alexbass01: Thanks. I assumed it was set in the function above so I just reused it.
image
https://github.com/alexbass01/open-component-model/blob/0e45d356136ff8e4d1b898e877c101e3e23322c7/website/static/install-cli.sh#L51
What do you think?

frewilhelm: Oh, I overlooked that step, sorry!

Nevertheless, if I use `bash` on my macOS I get now the info for `zsh`. Maybe it makes sense to check `OS` and `SHELL`

alexbass01: Good thinking.
How do you feel about not checking the OS but just checking if the `$SHELL` variable is ending with `zsh`?

```
local shell_rc
if [[ "${SHELL}" == */zsh ]]; then
    shell_rc="${HOME}/.zshrc"
else
    shell_rc="${HOME}/.profile"
fi
 ```
 
 On mac, this works both directly in zsh and inside bash:
 ```bash
 ❯ echo $SHELL
/bin/zsh

❯ bash
The default interactive shell is now zsh.
To update your account to use zsh, please run `chsh -s /bin/zsh`.
For more details, please visit https://support.apple.com/kb/HT208050.
bash-3.2$ echo $SHELL
/bin/zsh
bash-3.2$  
```

I think the `$SHELL` variable is always set: https://unix.stackexchange.com/questions/277944/what-sets-the-shell-environment-variable

alexbass01: I don't really think we need to consider bash with ~/.profile on mac because zsh is the standard since 2019 https://www.reddit.com/r/programming/comments/bwjny2/zsh_is_now_the_default_shell_for_macos/
Otherwise not sure now to solve it if that's really important.

frewilhelm: > I don't really think we need to consider bash with ~/.profile on mac because zsh is the standard since 2019 https://www.reddit.com/r/programming/comments/bwjny2/zsh_is_now_the_default_shell_for_macos/

There is at least one developer here using `bash` on macOS :D and I strongly assume that there are others too (otherwise we would have noticed this earlier). Nevertheless, you got a point.

What do you think about using `echo $0` or `ps -p $$ -o comm=` and check for `zsh`. If true, use `. zshrc`, if not, default to `.profile`. WDYT?

alexbass01: Haha, I am afraid `ps -p $$ -o comm=` just moves the problem. Concrete case: on my Mac my interactive shell is zsh, but if I install by piping the script into bash (e.g. curl ... | bash, or just running it inside a bash subshell), then `ps -p $$` reports `bash` — so it'd point me at `.profile` even though the PATH needs to go in `.zshrc`, which is where I actually type commands. It detects the shell running the script, not the shell I use.

It seems this is a more complex topic than I anticipated with my first proposal 🙈 . How about just leaving it to the user and printing this:
```
warn "${BIN_DIR} is not in your PATH."
warn "Add it by adding this line to your shell profile"
warn "(e.g. ~/.zshrc for zsh, or ~/.bashrc / ~/.profile for bash / sh):"
warn ""
warn "  export PATH=\"\$PATH:${BIN_DIR}\""
warn ""
```
Would that be an acceptable way forward?
Thanks a lot for your feedback!

frewilhelm: Yes, I believe this is the best way forward! Thanks a lot for contributing :)

morri-son: pls also check #3274 which overlaps with this PR and Fabian also flagged this. #3274 now contains a warning. By the way: would be good if we can get #3274 approved and merged soon 😊

alexbass01: Hi @morri-son! I've pushed the shell-agnostic PATH warning we agreed on with @frewilhelm. I think the overlap in #3274 came from an earlier version of my branch that you pulled in before our review discussion concluded (see [Aug 10 commits in #3274](https://github.com/open-component-model/open-component-model/pull/3274/commits)). Would it make sense to pull in the latest version of my branch into #3274 and then close this PR?

frewilhelm (approved):

jakobmoellerdev (approved):

PR #3332, opened 2026-08-11
fix: add merge function to the extract config

### LINKED ISSUE #1218: Extract configuration is a no-op
**Description**

The Merge function creates a fresh Config and returns it without looking at inputs. The Rules field is never merged. The [code comment](https://github.com/open-component-model/open-component-model/blob/4dfe7b59/bindings/go/configuration/extract/v1alpha1/spec/config.go?rgh-link-date=2026-07-15T14%3A33%3A45.000Z#L80) says `// Update later with values to configure.`, so probably a TODO that was never done.

Simple merge example:

```go
// Merge merges the provided configs into a single config.
func Merge(configs ...*Config) *Config {
	if len(configs) == 0 {
		return nil
	}

	merged := new(Config)
	_, _ = Scheme.DefaultType(merged)

	for _, config := range configs {
		if config.TempFolder != merged.TempFolder {
			merged.TempFolder = config.TempFolder
		}
		if config.WorkingDirectory != merged.WorkingDirectory {
			merged.WorkingDirectory = config.WorkingDirectory
		}
	}

	return merged
}

```

In the extract we didn't do any of the merging:

```go

// Merge merges the provided configs into a single config.
func Merge(configs ...*Config) *Config {
	if len(configs) == 0 {
		return nil
	}

	merged := new(Config)
	_, _ = Scheme.DefaultType(merged)

	return merged
}
```

Should deal with the values here:
```go
type Config struct {
	// +ocm:jsonschema-gen:enum=extract.oci.artifact.ocm.software/v1alpha1
	// +ocm:jsonschema-gen:enum:deprecated=extract.oci.artifact.ocm.software
	// Rules defines rules for extracting layers to specific files.
	Rules []Rule `json:"rules,omitempty"`
}

```

**Done Criteria**
- [ ] Fix the TODO.

Skarlso:

Extract merge function added that merges in order of last declared wins.

Alternatively, we could give it the order of the Resolvers. Which is in order as declared.

Fixes https://github.com/open-component-model/ocm-project/issues/1218

##### Verification


bindings/go/configuration/extract/v1alpha1/spec/config.go
```diff
@@ -91,5 +91,12 @@ func Merge(configs ...*Config) *Config {
 	merged := new(Config)
 	_, _ = Scheme.DefaultType(merged)
 
+	for _, config := range configs {
+		if config == nil || len(config.Rules) == 0 {
+			continue
+		}
+		merged.Rules = config.DeepCopy().Rules
```
frewilhelm, line 98: If we have several rules like

```go
A := &Config{Rules: []Rule{
    {Filename: "hello.tgz", LayerSelectors: [..]},
    {Filename: "world.tgz", LayerSelectors: [..]},
}}

B := &Config{Rules: []Rule{
    {Filename: "world.tgz", LayerSelectors: [..]},
}}
````

then, only `B` would survive. Is this intended?

Comment (Skarlso) [reply bindings/go/configuration/extract/v1alpha1/spec/config.go:98]: I believe so, yes. That's what I would expect intuitively. 🤔 

Like here:

```yaml
configurations:
  - type: extract.oci.artifact.ocm.software/v1alpha1
    rules:
      - filename: chart.tgz
        layerSelectors:
          - matchExpressions:
              - key: layer.mediaType
                operator: In
                values:
                  - application/vnd.cncf.helm.chart.content.v1.tar+gzip
  - type: extract.oci.artifact.ocm.software
    rules:
      - filename: chart.tgz
        layerSelectors:
          - matchExpressions:
              - key: layer.mediaType
                operator: In
                values:
                  - application/tar+gzip
      - filename: chart.prov
        layerSelectors:
          - matchProperties:
              layer.mediaType: application/vnd.cncf.helm.chart.provenance.v1.prov
```

I would expect the last declared extract to be the one surviving.

I could also imagine a scenario where we define multiple configs with a `--config` flag and you'd expect the last config to overwrite whatever the first one had in it either by accident or intentionally.

But feel free to challenge my on that assumption.

I can also see how a merge would happen where only existing values are overridden and different ones are taken and merged together. Somewhat similar like resolvers. What is your intuition?

frewilhelm: I am a little bit undecided. On the one hand the current behaviour does not feel like a merge but rather like a replace-with-the-last-one. On the other hand, what will users expect and is the complexity thats coming with a real merge worth it. After thinking about that, I believe your approach is right. Lets do it like that and see if any user has a problem with it.

frewilhelm (approved):

jneisener (approved):

PR #3352, opened 2026-08-11
chore: add missing signature to the signing config

Skarlso:

Missing `signature` field addition that is used to select a specific signature with the `--signature` flag. 

Towards https://github.com/open-component-model/ocm-project/issues/1240.

##### Verification


frewilhelm (dismissed):

bindings/go/signing/v1alpha1/spec/config.go
```diff
@@ -52,6 +53,11 @@ type Config struct {
 	// +ocm:jsonschema-gen:enum:deprecated=signing.config.ocm.software
 	Type runtime.Type `json:"type"`
 
+	// Signature is the name of the signature this entry applies to, the same
+	// name that defines the credentials used to produce it. If empty, the entry
```
morri-son:
```suggestion
	// Signature is the name of the signature this config applies to, the same
	// name that defines the credentials used to produce it. If empty, the config
```

Skarlso: What's wrong with `entry`? 🤔

bindings/go/signing/v1alpha1/spec/config.go
```diff
@@ -109,7 +122,11 @@ func LookupConfig(cfg *genericv1.Config) (*Config, error) {
 	if err != nil {
 		return nil, fmt.Errorf("failed to filter config: %w", err)
 	}
-	cfgs := make([]*Config, 0, len(filtered.Configurations))
+
+	// we need configurations first that are without signature and then those that are with
+	// signature.
```
morri-son:
```suggestion
	// we first need the configurations without a signature, then those which have a
	// signature.
```

bindings/go/signing/v1alpha1/spec/config.go
morri-son: nothing wrong, it is a suggestion :) `config entry` would also match.

morri-son (approved):

frewilhelm (approved):

PR #3395, opened 2026-08-17
fix: Handle relative paths for e.g. `--repository`

ntnn:

I tried to pass a relative path to `--repository` which ocm interpreted as an http uri.
My first iteration checked all path segments until I realized that `filepath.Clean` folds those away, hence the additional tests, but I think they might be worth keeping around.

##### How to test the changes

```bash
ocm add cv --repository ../ctf/somewhere/else --constructor test/component-constructor.yaml 
```

##### Verification


jakobmoellerdev (dismissed):

bindings/go/oci/compref/compref.go
```diff
@@ -382,6 +383,20 @@ func guessType(repository string) (string, error) {
 		return runtime.NewVersionedType(ctfv1.Type, ctfv1.Version).String(), nil
 	}
 
+	// Relative filesystem path -> assume CTF
+	if split := strings.Split(cleaned, string(filepath.Separator)); len(split) > 1 {
```
matthiasbruns, line 387: Can't you use something like 
https://pkg.go.dev/path/filepath#IsAbs
to check if its relative or not?

frewilhelm: `IsAbs` is called above but you cannot assume that `!isAbs(...)` is automatically a relative path. It could be an URL (see below)

frewilhelm (approved):

matthiasbruns (approved):

PR #3397, opened 2026-08-18
fix: credential type registration [2/2] (#1146)

### LINKED ISSUE #1146: Incorrect Credential Type Registrations
**Description**

The typed credentials migration had several faults in its plugin use and type registration when first introduced.

Firstly, typed credentials are currently hardcoded in the consumer instead of being reported by the Plugin that supports them. This means that one has to explicitly configure which typed credentials are available separate to the plugin registry. This happens through a `Register` method thats only available on the credential repository registry: https://github.com/open-component-model/open-component-model/blob/main/bindings/go/plugin/manager/registries/credentialrepository/registry.go#L62

Instead of this we should:
1. Open up a new registry that only contains typed credentials
2. Have plugins that implement either [Credential Repository](https://github.com/open-component-model/open-component-model/tree/main/bindings/go/plugin/manager/registries/credentialrepository) or [Credential Plugin](https://github.com/open-component-model/open-component-model/tree/main/bindings/go/plugin/manager/registries/credentialplugin) Contracts return their supported credential Types as a [`CredentialTypeSchemeProvider`](https://github.com/open-component-model/open-component-model/blob/main/bindings/go/credentials/interface.go#L30)
3. Then use the types returned here and register them as a central registry
4. Then switch all occurences of hardcoded credentials registration to type information offered by the plugins themselves. As an example: [RSA](https://github.com/open-component-model/open-component-model/blob/main/cli/internal/plugin/builtin/rsa/register.go#L31) should not need a separate hardcoded registration. Instead there should be a method on  the repository plugin that can automatically be called during plugin registration.

**Done Criteria**

matthiasbruns:
Refactors the way we can register custom credential types in ocm.
Before this change the cli and controller had to manually access `scheme` and add it to its credential type registry.
Now we inverse the call chain and expect the implementation of `credentials.CredentialTypeSchemeProvider` through the `BuiltinCredentialTypeSchemeProviderPlugin` for built-in bindings/types-

For external plugins, use `credentialpluginv1.CapabilitySpec`'s `CustomCredentialTypes` field, which is being consumes by the same registry.

Contributes: https://github.com/open-component-model/ocm-project/issues/1146

Followup of  https://github.com/open-component-model/open-component-model/pull/3398

##### How to test the changes

All the existing tests execute as is. I added new tests to test the new registry and credential conflict handling.

##### Verification

### DIFF
--- a/bindings/go/github/digest/digest.go
+++ b/bindings/go/github/digest/digest.go
@@ -18,6 +18,7 @@ import (
 	githubinternal "ocm.software/open-component-model/bindings/go/github/internal"
 	"ocm.software/open-component-model/bindings/go/github/repository/resource"
 	"ocm.software/open-component-model/bindings/go/github/spec/access"
+	ghcreds "ocm.software/open-component-model/bindings/go/github/spec/credentials"
 	credsv1 "ocm.software/open-component-model/bindings/go/github/spec/credentials/v1"
 	"ocm.software/open-component-model/bindings/go/plugin/manager/registries/digestprocessor"
 	"ocm.software/open-component-model/bindings/go/runtime"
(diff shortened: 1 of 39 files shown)


matthiasbruns:
will be merged into one pr after the one go.mod change

PR #3398, opened 2026-08-18
fix: credential type registration (#1146)

### LINKED ISSUE #1146: Incorrect Credential Type Registrations

matthiasbruns:
Refactors the way we can register custom credential types in ocm.
Before this change the cli and controller had to manually access `scheme` and add it to its credential type registry.
Now we inverse the call chain and expect the implementation of `credentials.CredentialTypeSchemeProvider` through the `BuiltinCredentialTypeSchemeProviderPlugin` for built-in bindings/types-

For external plugins, use `credentialpluginv1.CapabilitySpec`'s `CustomCredentialTypes` field, which is being consumes by the same registry.

Contributes: https://github.com/open-component-model/ocm-project/issues/1146

##### How to test the changes

All the existing tests execute as is. I added new tests to test the new registry and credential conflict handling.

##### Verification


bindings/go/plugin/manager/registries/credentialrepository/registry.go
```diff
@@ -26,18 +25,16 @@ func NewCredentialRepositoryRegistry(ctx context.Context) *RepositoryRegistry {
...
-	mu                   sync.Mutex
-	capabilities         map[string]credentialsv1.CapabilitySpec
-	registry             map[runtime.Type]mtypes.Plugin
-	scheme               *runtime.Scheme
-	credentialTypeScheme *runtime.Scheme
+	ctx          context.Context
+	mu           sync.Mutex
+	capabilities map[string]credentialsv1.CapabilitySpec
+	registry     map[runtime.Type]mtypes.Plugin
+	scheme       *runtime.Scheme
```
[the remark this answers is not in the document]
matthiasbruns: this is what the 2nd PR is for - we cannot update bindings & cli in one pr

bindings/go/plugin/manager/registries/credentialtyperepository/credentialtype_test.go
```diff
@@ -0,0 +1,268 @@
...
+	r.True(reg.GetCredentialTypeScheme().IsRegistered(typeB), "non-conflicting type must still be registered")
+}
+
+// TestRegisterCustomTypes_MultipleTypesDoNotConflictWithRaw verifies that registering several plugin
+// credential types does not cause them to alias each other through *runtime.Raw.
+func TestRegisterCustomTypes_MultipleTypesDoNotConflictWithRaw(t *testing.T) {
+	typeA := runtime.NewVersionedType("PluginCredA", "v1")
+	typeB := runtime.NewVersionedType("PluginCredB", "v1")
+	typeC := runtime.NewVersionedType("PluginCredC", "v2")
+
```
piotrjanik, line 156: It looks like all unversioned types would resolve to the same canonical type.
Try adding this:

```
	t.Run("each type is its own canonical type", func(t *testing.T) {
		r := require.New(t)
		for _, typ := range []runtime.Type{typeA, typeB, typeC} {
			canonical, ok := scheme.ResolveCanonicalType(typ)
			r.True(ok)
			r.Equal(typ, canonical, "%s must not be an alias of anotherl type", typ)
		}
	})
```

piotrjanik: I also tried to resolve canonical type by adding aliases to each versioned type and then resolving to canonical.

matthiasbruns: This feels like something we need to fix in runtime 🤔

matthiasbruns: and this bug would correctly affect non-built-in plugins
I would postpone this until we actually are doing custom plugins and revisit this problem, wdyt?

piotrjanik: sure, just create a follow-up task

matthiasbruns: https://github.com/open-component-model/ocm-project/issues/1274

bindings/go/plugin/manager/registries/credentialtyperepository/registry.go
```diff
@@ -0,0 +1,168 @@
...
+			}
+			// Already known: only the plugin that declared it may declare it again, e.g. from a
+			// second capability spec of the same plugin binary.
+			if declaredBy, ok := r.registry[alias]; ok && declaredBy.ID == plugin.ID {
+				continue
+			}
+			errs = append(errs, fmt.Errorf("credential type %s already registered", alias))
+			conflict = true
+		}
+		if conflict {
```
piotrjanik, line 132: nit: any debug/warn log here?

matthiasbruns: This happens anyway

`errs = append(errs, fmt.Errorf("credential type %s already registered", alias))`

piotrjanik (dismissed):

bindings/go/plugin/manager/registries/credentialtyperepository/registry.go
```diff
@@ -0,0 +1,168 @@
...
+			}
+		}
+
+		if len(missing) == 0 {
+			continue
+		}
+
+		if err := r.scheme.RegisterWithAlias(prototype, missing...); err != nil {
+			return fmt.Errorf("failed to register credential type %q: %w", typ, err)
+		}
```
Skarlso, line 81: Missing will contain the type as well. So this might be something like `OCI, OCI` where the alias will register the same type as an alias. This _might_ not be a problem but I think it might be. 🤔

matthiasbruns: just tested this, and it does not happen - I dont think I will checkin the tests though - but we are fine here
but bimap in runtime might be flawed anyway - we should investigate this in a follow up https://github.com/open-component-model/ocm-project/issues/1277

matthiasbruns:
let me also bring in cli and controller in here - draft until I am done

jakobmoellerdev (approved):

Skarlso (approved):

PR #3286, opened 2026-08-05
chore(dep): add depguard rules for all the packages

jneisener:
#### What this PR does / why we need it

As planed in the [ADR](https://github.com/open-component-model/open-component-model/blob/main/docs/adr/0025_bindings_ci_and_release_strategy.md) this PR persists the dependency tree currently managed inside of `go.mod` files in `depguard` rules in  preparation for the removal of the individual `go.mod` files.

In order to make the rules easier to reason about, a table (similar to the one in the ADR) is created from them and kept up-to-date via a new Github action.

#### Which issue(s) this PR fixes
Fixes: https://github.com/open-component-model/ocm-project/issues/1206

##### Verification

- [ ] I have added/updated tests for my changes (see [Test Requirements](../CONTRIBUTING.md#test-requirements))
- [ ] Tests pass locally (`task test` and `task test/integration` if applicable)
- [ ] If touching multiple modules, `go work` is enabled (see `go.work`)
- [ ] My changes do not decrease test coverage
- [ ] I have tested the changes locally by running `ocm`

PR #3409, opened 2026-08-19
chore: add dependency check to completion

frewilhelm:

https://github.com/open-component-model/open-component-model/pull/3286 introduced the workflow step `dependency_table_verify` to make sure the dependency table is updated if any dependency changes. However, this step is not required yet. We cannot add the step as required as the step is only executed when `${{ needs.discover_modules.outputs.dependency_table_changed == 'true' }}`.

### DIFF
--- a/tools.Taskfile.yml
+++ b/tools.Taskfile.yml
@@ -115,9 +115,19 @@ tasks:
 
   dependency-table/generate:
     desc: "Regenerate docs/dependency-table.md from golangci.yml depguard rules"
+    deps: [dependency-table/install]
     cmd: node {{ .ROOT_DIR }}/.github/scripts/dependency-table.js generate > {{ .ROOT_DIR }}/docs/dependency-table.md
 
   dependency-table/verify:
     desc: "Verify docs/dependency-table.md is in sync with golangci.yml depguard rules"
+    deps: [dependency-table/install]
     cmd: |
       node {{ .ROOT_DIR }}/.github/scripts/dependency-table.js generate | diff - {{ .ROOT_DIR }}/docs/dependency-table.md
+
+  dependency-table/install:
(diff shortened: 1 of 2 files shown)


jneisener (dismissed):

jneisener (dismissed):

tools.Taskfile.yml
```diff
@@ -115,9 +115,19 @@ tasks:
...
\ No newline at end of file
+      node {{ .ROOT_DIR }}/.github/scripts/dependency-table.js generate | diff - {{ .ROOT_DIR }}/docs/dependency-table.md
+
+  dependency-table/install:
+    internal: true
+    dir: '{{ .ROOT_DIR }}/.github/scripts'
+    sources:
+      - package.json
+      - package-lock.json
+    cmd: npm ci --omit=dev
```
jneisener, line 133: https://github.com/open-component-model/open-component-model/blob/996a8dc8d504a18ab8aad2eceac5bc4ebb9a6119/.github/workflows/ci.yml#L277

can then be removed (to avoid installing twice)

jneisener: Oh - skipped writing down the preceding thought. https://github.com/open-component-model/open-component-model/blob/996a8dc8d504a18ab8aad2eceac5bc4ebb9a6119/.github/workflows/ci.yml#L280 should then also call the verify task

frewilhelm: I believe, we would then need to install `task` and run the `target` itself

frewilhelm: adjusted it :)

jneisener (dismissed):

jakobmoellerdev (approved):

jneisener (approved):

PR #3423, opened 2026-08-20
fix: unresolved relative path in `GetBlobFromPath`

### LINKED ISSUE #3420: Inconsistent behavior between File/v1 and Dir/v1 for relative paths
**Context**

**Version**

> ocm version
{"major":"0","minor":"13","patch":"0","gitVersion":"0.13.0","goVersion":"go1.26.4","compiler":"gc","platform":"darwin/arm64"}

**To Reproduce**

```sh
mkdir my-components
mkdir my-components/files
touch my-components/files/test.txt
```

Create `my-components/component-constructor.yaml`:

```yaml
# yaml-language-server: $schema=https://ocm.software/0.13/schemas/bindings/go/constructor/schema-2020-12.json
components:
- name: github.com/acme.org/helloworld # Must at least be a DNS domain as per RFC 1123.
  version: 1.0.0 # Version conforming to SemVer 2.0.
  provider:
    name: acme.org # You must declare a provider, who serves as the owner of the component.
  resources:
    - name: mylocalfile
      type: blob
      input: # Embed by value
        type: File/v1
        path: ./files/test.txt
    - name: mylocaldir
      type: blob
      input: # Embed by value
        type: Dir/v1
        path: ./files/
```

```
│── my-components
│   ├── component-constructor.yaml
│   └── files
│       └── test.txt
```

Run from root dir:

> ocm add cv --constructor my-components/component-constructor.yaml

Run from my-components dir:

> cd my-components
> ocm add cv --constructor component-constructor.yaml

**Actual behavior**

* Run from root dir fails for `Dir/v1` works for `File/v1`:

```
Error: failed to render components recursively: constructing component versions failed: failed to constructComponent components from graph: failed to process component constructor graph: failed to process value for node name=github.com/acme.org/helloworld,version=1.0.0 in graph: failed processing constructor component: error constructing component "name=github.com/acme.org/helloworld,version=1.0.0": error constructing component: error processing resource "name=mylocaldir" at index 1: error processing resource "name=mylocaldir": error getting blob from input method: error getting dir blob based on resource input specification: error accessing path "./files/": stat ./files/: no such file or directory
```

* Run from components-dir: Works for both `Dir/v1` and `File/v1`

**Expected behavior**

* Same behavior between `Dir/v1` and `File/v1`

**Additional info**

* Running from root dir and chaning paths to include `./my-components/`:
  * `File/v1` fails with `failed to ensure path in working directory: failed to open path "./my-components/files/test.txt" in root "/Users/xxxx/ocm-test/my-components": openat ./my-components/files/test.txt: no such file or directory`
  * `Dir/v1` fails with: `failed to open path "./my-components/files/" in root "/Users/xxx/ocm-test/my-components": openat ./my-components/files/: no such file or directory`

jneisener:
`GetBlobFromPath` discards the result of `ensurePathInWorkingDirectory` and thus does not work correctly when passed a working directory and a relative path. Fix is to actually use the result of `ensurePathInWorkingDirectory`.

While working on that found and fixed two more issues:
* `ensurePathInWorkingDirectory` would leak because the handle returned by `os.OpenInRoot` was never closed
* `mimetype.DetectFile` used the raw path instead of resolving it from the working directory

Fixes https://github.com/open-component-model/open-component-model/issues/3420
##### Verification

### DIFF
--- a/bindings/go/blob/filesystem/blob_from_path_test.go
+++ b/bindings/go/blob/filesystem/blob_from_path_test.go
@@ -97,6 +97,10 @@ func TestGetBlobFromPath_RelativeDirectory(t *testing.T) {
 	b, err := filesystem.GetBlobFromPath(context.Background(), "testFolder", opt)
 	r.NoError(err)
 	r.NotNil(b)
+
+	reader, err := b.ReadCloser()
+	r.NoError(err)
+	defer func() { r.NoError(reader.Close()) }()
 }
 
 // PATTERN FILTERING: Include/exclude pattern functionality
--- a/bindings/go/blob/filesystem/blob_from_path_test.go
+++ b/bindings/go/blob/filesystem/blob_from_path_test.go
@@ -80,6 +80,25 @@ func TestGetBlobFromPath_SimpleDirectory(t *testing.T) {
 	r.Equal("content2", foundFiles["file2.txt"])
 }
 
+func TestGetBlobFromPath_RelativeDirectory(t *testing.T) {
+	r := require.New(t)
+
+	// Change working directory to explicitly not be in the working directory we pass via options
+	wd := t.TempDir()
+	t.Chdir(wd)
+
+	// Setup: create directory with multiple files
+	tmpDir := t.TempDir()
+	createTestFile(t, filepath.Join(tmpDir, "testFolder"), "file.txt", "content")
+
+	// Test: create blob from directory (should be TAR archive)
+	opt := filesystem.DirOptions{Reproducible: true}
+	opt.WorkingDir = tmpDir
+	b, err := filesystem.GetBlobFromPath(context.Background(), "testFolder", opt)
+	r.NoError(err)
+	r.NotNil(b)
+}
+
 // PATTERN FILTERING: Include/exclude pattern functionality
 func TestGetBlobFromPath_PatternSemantics(t *testing.T) {
 	r := require.New(t)
--- a/bindings/go/blob/filesystem/blob_from_path_test.go
+++ b/bindings/go/blob/filesystem/blob_from_path_test.go
@@ -2,7 +2,6 @@ package filesystem_test
 
 import (
 	"archive/tar"
-	"context"
 	"io"
 	"os"
 	"path/filepath"
@@ -26,7 +25,7 @@ func TestGetBlobFromPath_SingleFile(t *testing.T) {
 
 	// Test: create blob from single file (should be raw, not TAR)
 	opt := filesystem.DirOptions{Reproducible: true}
-	b, err := filesystem.GetBlobFromPath(context.Background(), testFile, opt)
+	b, err := filesystem.GetBlobFromPath(t.Context(), testFile, opt)
 	r.NoError(err)
 	r.NotNil(b)
 
@@ -50,7 +49,7 @@ func TestGetBlobFromPath_SimpleDirectory(t *testing.T) {
 
 	// Test: create blob from directory (should be TAR archive)
 	opt := filesystem.DirOptions{Reproducible: true}
-	b, err := filesystem.GetBlobFromPath(context.Background(), tmpDir, opt)
+	b, err := filesystem.GetBlobFromPath(t.Context(), tmpDir, opt)
 	r.NoError(err)
 	r.NotNil(b)
 
@@ -94,7 +93,7 @@ func TestGetBlobFromPath_RelativeDirectory(t *testing.T) {
 	// Test: create blob from directory (should be TAR archive)
 	opt := filesystem.DirOptions{Reproducible: true}
 	opt.WorkingDir = tmpDir
-	b, err := filesystem.GetBlobFromPath(context.Background(), "testFolder", opt)
+	b, err := filesystem.GetBlobFromPath(t.Context(), "testFolder", opt)
 	r.NoError(err)
 	r.NotNil(b)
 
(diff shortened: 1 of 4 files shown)


jneisener:
@coderabbitai review

bindings/go/input/file/blob.go
```diff
@@ -56,9 +56,13 @@ func GetV1FileBlob(file v1.File, workingDirectory string) (blob.ReadOnlyBlob, er
 
 	mediaType := file.MediaType
 	if mediaType == "" {
-		// see https://github.com/gabriel-vasile/mimetype/blob/master/supported_mimes.md for supported types
-		mime, _ := mimetype.DetectFile(file.Path)
-		mediaType = mime.String()
+		readCloser, _ := b.ReadCloser()
+		if readCloser != nil {
+			// see https://github.com/gabriel-vasile/mimetype/blob/master/supported_mimes.md for supported types
+			mime, _ := mimetype.DetectReader(readCloser)
```
jakobmoellerdev: I think this is incorrect, we should not start reading from the file to discover a mime type but only the path should be enough

bindings/go/input/file/blob.go
```diff
@@ -56,9 +56,13 @@ func GetV1FileBlob(file v1.File, workingDirectory string) (blob.ReadOnlyBlob, er
...
+			mediaType = mime.String()
+			_ = readCloser.Close()
```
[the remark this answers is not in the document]
jneisener: Media type discovery is best effort and already ignored the returned error previously: `mime, _ := mimetype.DetectFile(file.Path)`
We don't want to fail the processing of the blob because the media type discovery didn't work, so I think this should stay as is.

If `readCloser` is not `nil`, `Close()` is called on it, if its `nil` there is no point. I don't see the path where its not closed

bindings/go/input/file/blob.go
jneisener: `DetectFile` is just a wrapper around `DetectReader` so the file is read either way

jneisener: Though only the first `4096` bytes by default

bindings/go/blob/filesystem/file.go
```diff
@@ -91,17 +91,22 @@ func GetBlobInWorkingDirectory(path, workingDir string) (*Blob, error) {
...
 		}
 	}
 
-	_, err = os.OpenInRoot(workingDirectory, path)
+	resolved, err := os.OpenInRoot(workingDirectory, path)
 	if err != nil {
 		return "", fmt.Errorf("failed to open path %q in root %q: %w", path, workingDirectory, err)
 	}
+	defer func() {
+		if cErr := resolved.Close(); retErr == nil && cErr != nil {
```
jakobmoellerdev: does this need to be deferred?

bindings/go/input/file/blob.go
jakobmoellerdev: I think in this case maybe we are missing here the mime type detection based on file extension right?

matthiasbruns (commented):
oh this brings back memories 

good fix - waiting for the comments to be resolved

bindings/go/input/file/blob.go
jneisener: Valid concern, we are not using the file extension for the detection. The [`mimetype`](https://pkg.go.dev/github.com/gabriel-vasile/mimetype#section-readme) package always reads the file, so if we wanted to go by extension alone we'd have to use e.g. [`TypeByExtension`](https://pkg.go.dev/mime#TypeByExtension).

As discussed in the daily, this if something to evaluate in a follow-up

bindings/go/input/file/blob.go
```diff
@@ -56,9 +56,13 @@ func GetV1FileBlob(file v1.File, workingDirectory string) (blob.ReadOnlyBlob, er
 
 	mediaType := file.MediaType
 	if mediaType == "" {
-		// see https://github.com/gabriel-vasile/mimetype/blob/master/supported_mimes.md for supported types
-		mime, _ := mimetype.DetectFile(file.Path)
-		mediaType = mime.String()
+		readCloser, _ := b.ReadCloser()
```
jakobmoellerdev: why dont you return an error if the read closer call fails?

jneisener: My thinking was that media type discovery is best effort and already ignored the returned error previously: `mime, _ := mimetype.DetectFile(file.Path)` (the error returned by `DetectFile` would have contained the read error).
I didn't think we want to fail the processing of the blob because the media type discovery didn't work - it will probably fail later in the stack if it can't be opened, but I thought I'd keep it close to the established behavior

bindings/go/blob/filesystem/blob_from_path_test.go
```diff
@@ -80,6 +80,29 @@ func TestGetBlobFromPath_SimpleDirectory(t *testing.T) {
...
+	t.Chdir(wd)
+
+	// Setup: create directory with multiple files
+	tmpDir := t.TempDir()
+	createTestFile(t, filepath.Join(tmpDir, "testFolder"), "file.txt", "content")
+
+	// Test: create blob from directory (should be TAR archive)
+	opt := filesystem.DirOptions{Reproducible: true}
+	opt.WorkingDir = tmpDir
+	b, err := filesystem.GetBlobFromPath(context.Background(), "testFolder", opt)
```
jakobmoellerdev: t.Context()

bindings/go/input/file/blob.go
jakobmoellerdev: well if you cant open a blob reader isnt it almost guaranteed that we will fail later on?

jakobmoellerdev (dismissed):

jakobmoellerdev (approved):

matthiasbruns (approved):

PR #3429, opened 2026-08-21
feat: get sbom for a given resource [1/2]

matthiasbruns:

This change adds tracking SBOMs to a component version. The strategy to discover SBOMs is as follows:

First, check if the component contains any artefactReference labels pointing to the resource for which you are trying to fetch an SBOM for. If none exist, then we check any OCI Artefact types for attached SBOMs.

Right now, the only technology that we support is via docker buildx SBOMs `--sbom` described here: https://docs.docker.com/build/metadata/attestations/sbom.

Further, once we find any SBOMs attached to the resource, we will combine them. Combination is handled via this library https://github.com/protobom/protobom.

The end-result then can be piped into a tool for verification.

This PR contains both the library implementation, and the CLI integration for ease of review.
The library, further, contains implementation for the artefactReference ocm-spec addition.

Part of https://github.com/open-component-model/ocm-project/issues/1247.

##### Verification


bindings/go/descriptor/runtime/labels/artefactref/v1alpha1/artefactref.go
```diff
@@ -0,0 +1,119 @@
...
+//	    type: ociImage
+//	    extraIdentity:
+//	      foo: bar
+//	  - name: my-image-sbom
+//	    version: 1.2.3
+//	    type: sbom
+//	    extraIdentity:
+//	      architecture: amd64
+//	    labels:
+//	      - name: ocm.software/artefact-references
```
piotrjanik: Maybe I missed that part. Was there a reason we did not consider using Referer API for SBOMs?

bindings/go/descriptor/runtime/labels/artefactref/v1/artefactref.go
```diff
@@ -0,0 +1,119 @@
...
+		if key == descriptor.IdentityAttributeName || key == descriptor.IdentityAttributeVersion {
+			continue
+		}
+		extra[key] = value
+	}
+	return extra
+}
+
+// FromLabels decodes the artefact reference label out of labels.
+func FromLabels(labels []descriptor.Label) (Reference, bool, error) {
```
piotrjanik: I think this can be private

bindings/go/descriptor/runtime/labels/artefactref/v1alpha1/artefactref.go
matthiasbruns: https://github.com/open-component-model/ocm-project/issues/1247
pretty much designed in the issue

bindings/go/oci/attestation/convert.go
```diff
@@ -0,0 +1,36 @@
+package attestation
+
+import (
+	ociImageSpecV1 "github.com/opencontainers/image-spec/specs-go/v1"
+
+	"ocm.software/open-component-model/bindings/go/repository"
+)
+
+// ToRepositoryPlatform converts an OCI platform to its technology agnostic equivalent.
+func ToRepositoryPlatform(platform ociImageSpecV1.Platform) repository.Platform {
```
piotrjanik: private?

matthiasbruns: no - used in separate packages

bindings/go/descriptor/runtime/labels/artefactref/v1/artefactref.go
matthiasbruns: I would keep this public - this could be useful for others later on

bindings/go/oci/attestation/sbom.go
```diff
@@ -0,0 +1,280 @@
...
+	switch {
+	case want.Architecture != "" && want.Architecture != have.Architecture:
+		return false
+	case want.OS != "" && want.OS != have.OS:
+		return false
+	case want.Variant != "" && want.Variant != have.Variant:
+		return false
+	case want.OSVersion != "" && want.OSVersion != have.OSVersion:
+		return false
+	case len(want.OSFeatures) > 0 && !slices.Equal(want.OSFeatures, have.OSFeatures):
```
piotrjanik: OSFeatures should be sorted

matthiasbruns: done - good catch

piotrjanik (dismissed):

bindings/go/descriptor/runtime/labels/artifactref/v1alpha1/artifactref.go
```diff
@@ -0,0 +1,119 @@
...
+func extraIdentityOf(identity runtime.Identity) runtime.Identity {
+	extra := make(runtime.Identity, len(identity))
+	for key, value := range identity {
+		if key == descriptor.IdentityAttributeName || key == descriptor.IdentityAttributeVersion {
+			continue
+		}
+		extra[key] = value
+	}
+	return extra
+}
```
fabianburth, line 91: dont we have existing matchers for something like this?

bindings/go/oci/internal/attestation/sbom.go
```diff
@@ -0,0 +1,292 @@
...
+//	└── manifests[1]  platform: unknown/unknown              the attestation
+//	      annotations:
+//	        vnd.docker.reference.type:   attestation-manifest
+//	        vnd.docker.reference.digest: <digest of manifests[0]>
+//	      layers[*]  mediaType: application/vnd.in-toto+json
+//	        annotations:
+//	          in-toto.io/predicate-type: https://spdx.dev/Document
+//
+// Each in-toto layer wraps the actual document in a statement envelope. The SBOM is the
+// predicate of that envelope.
```
fabianburth, line 18: should be separate doc.go

bindings/go/descriptor/runtime/labels/artefactref/v1/artefactref.go
```diff
@@ -0,0 +1,119 @@
...
+//	    extraIdentity:
+//	      architecture: amd64
+//	    labels:
+//	      - name: ocm.software/artefact-references
+//	        version: v1
+//	        value:
+//	          identity:
+//	            name: my-image
+//	            version: 1.2.3
+//	            foo: bar
```
fabianburth: should be separate doc.go

bindings/go/descriptor/runtime/labels/artefactref/v1/artefactref.go
```diff
@@ -0,0 +1,119 @@
...
+)
+
+// ErrNotFound used to signal that there are no references.
+var ErrNotFound = errors.New("no resource references the artefact")
+
+const (
+	// LabelName is the name under which artefact references are stored on an element.
+	LabelName = "ocm.software/artefact-references"
+	// Version is the specification version of the label value implemented here.
+	Version = "v1"
```
morri-son: I tried the cited example from the spec against this and think it doesn't match? Spec uses version: v1alpha1 (in Lookup Algorithm, step 3), but the code uses v1. So FromLabels just skips a v1alpha1 label, is that intended? And in the spec the label value is a list but Reference decodes a single object, so even a version v1 errors with"cannot unmarshal array". Your doc comment on top also has just one object as identity. Do we want to allow that new label to reference multiple resources? At least its name artefact-references, plural, sounds like. Anyway, something is off here between code and spec :)

from the ocm spec

 ```yaml
  resources:
  - name: my-image
    version: 1.2.3
    type: ociImage

  - name: my-image-sbom
    version: 1.2.3
    type: sbom
    labels:
      - name: ocm.software/artefact-references
        version: v1alpha1
        value:
          - identity:
              name: my-image
 ```

bindings/go/oci/attestation/sbom.go
```diff
@@ -0,0 +1,292 @@
...
+
+// DiscoverSBOMs resolves reference in store and returns every SBOM attestation attached
+// to the images matching the requested platform, or to every platform in the index when
+// WithAllSBOMPlatforms is given.
+//
+// Returns the following set of errors:
+// - ErrNotAnIndex if the reference is a plain manifest
+// - ErrPlatformNotFound if there was no platform for the given value
+// - ErrNoAttestation if there was no SBOM.
+func DiscoverSBOMs(ctx context.Context, store spec.Store, reference string, opts ...repository.SBOMOption) ([]SBOM, error) {
```
fabianburth: this should not be public API

fabianburth (dismissed):

bindings/go/descriptor/runtime/labels/artefactref/v1/artefactref.go
Skarlso: I documented all these in the issue already. :)

There is a mismatch between what's in the example and what's in the blog post and what's in the spec. Neither of them are actually correct.

The version mismatch is known and v1alpha1 was never something that was possible because of the version regex that exists since ocmv1. So that was just plain wrong in both cases.

The label structure is also a mismatch between what exists and what's possible currently according to the spec. I also documented this in the issue.

TL;DR: Nothing that we can fix right now, but needs a followup to change the examples in the doc and the blog post.

bindings/go/descriptor/runtime/labels/artifactref/v1alpha1/artifactref.go
Skarlso: Nothing that I could find that is this specific and/or doesn't require a rewrite. But tell me if you find something. :D

matthiasbruns: A wild @Skarlso appeared :D 

assnaadam-bndyz

Skarlso: OH GODS that gif. :D Jesus :D

bindings/go/oci/attestation/sbom.go
Skarlso: The purpose here was separation of the code. This is called from oci/sbom.go. Of course, I could have put everything in here in there, but that would have been messy because the conversion the constants and all that jazz live in the oci/attestation package.

Now, there is also a bit of a discrapency here because I wanted a resource repo that implements Discovery interface, and the OCI repo has a DiscoverSBOM function. But it's slightly different because that's without credentials. We have other patterns like these around the codebase with the repository since creds have already been resolved at that point, right?

I'll see what I can do here to restructure the code a bit better, but I also don't really want to break existing patterns and don't really want to put everything into one package of course.

Skarlso: Actually, I think I found a bit of a residue from a previous refactor I did. Let me see if I can remove some of this fluff with the conversion and the options.

Skarlso:
I'm doing some restructuring here. :) Hold on to your butts. :D

bindings/go/oci/internal/attestation/sbom.go
```diff
@@ -0,0 +1,269 @@
...
+
+// DiscoverSBOMs resolves reference in store and returns every SBOM attestation attached
+// to the images matching the requested platform, or to every platform in the index when
+// WithAllSBOMPlatforms is given.
+//
+// Returns the following set of errors:
+// - ErrNotAnIndex if the reference is a plain manifest
+// - ErrPlatformNotFound if there was no platform for the given value
+// - ErrNoAttestation if there was no SBOM.
+func DiscoverSBOMs(ctx context.Context, store spec.Store, reference string, opts ...repository.SBOMOption) ([]repository.SBOM, error) {
```
Skarlso, line 58: Moved this into the internal package. Now it should be only used by us. :) The separation still stands.

bindings/go/oci/internal/attestation/sbom.go
```diff
@@ -0,0 +1,269 @@
...
+
+		sboms = append(sboms, repository.SBOM{
+			Platform: platform,
+			// the in-toto layer digest is the only value unique across the SBOMs of
+			// one resource, the image digest is shared by all of them.
+			ID:            layer.Digest.String(),
+			Digest:        image.Digest.String(),
+			PredicateType: predicateType,
+			Name:          named.Name,
+			Data:          envelope.Predicate,
```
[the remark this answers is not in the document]
Skarlso: >An annotated SPDX layer can contain a SLSA predicate, or an attestation linked to image A can contain a statement for image B. The returned repository.SBOM then reports metadata that does not describe Data.

I doubt that this is possible. 🤔

bindings/go/oci/internal/identity/descriptor.go
```diff
@@ -49,10 +51,29 @@ var mappings = []platformAttributeMapper{
...
+	}
+	return platform
+}
+
 // Adopt modifies the provided OCI descriptor to represent an artifact.
 // It sets the platform fields based on the resource's extra identity attributes
-// and adds a annotations.ArtifactOCIAnnotation to indicate that the descriptor
-// is a annotations.ArtifactKindResource.
+// and adds an annotations.ArtifactOCIAnnotation to indicate that the descriptor
+// is an annotations.ArtifactKindResource.
```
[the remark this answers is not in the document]
Skarlso: I'm not touching this, because I just fixed the `a` to an `an`. So I'll leave the validity of the comment as is.

Skarlso (dismissed):
LGTM 😆

Skarlso (approved):
Still LGTM :D :D

bindings/go/oci/internal/attestation/sbom.go
```diff
@@ -0,0 +1,269 @@
...
+		}
+	}
+
+	if attested == 0 {
+		return nil, fmt.Errorf("%w: %q", ErrNoAttestation, reference)
+	}
+
+	if len(sboms) == 0 {
+		return nil, fmt.Errorf("%w: %q has no attestation of predicate types %v", ErrNoAttestation, reference, o.PredicateTypes)
+	}
```
jneisener, line 127: Might be worth distinguishing these two if the first case is "no sbom" but the second might be "sbom in unexpected format"?

website/content/blog/2026-07-28-shipping-sboms-with-your-components.md
```diff
@@ -21,13 +21,13 @@ The idea explored here: make SBOMs travel *with* the component, related to the a
 
 To explore the idea, we built a proof of concept across three areas. None of this is a shipped feature yet: it is a spike to see what the workflow could feel like, and one we are now actively driving forward.
 
-**Linking an SBOM to what it describes.** A resource of `type: sbom` carries an `ocm.software/artefactReference` label whose `identitySelector` points at the resource the SBOM is about. Because it is an ordinary, signed resource, the link travels and is verified with the component, with no side channel.
+**Linking an SBOM to what it describes.** A resource of `type: sbom` carries an `ocm.software/artifactReference` label whose `identitySelector` points at the resource the SBOM is about. Because it is an ordinary, signed resource, the link travels and is verified with the component, with no side channel.
```
jneisener, line 24: Isn't it `identity` instead of `identitySelector`?

bindings/go/oci/internal/attestation/sbom.go
```diff
@@ -0,0 +1,269 @@
...
+		return nil, fmt.Errorf("reading image index of %q failed: %w", reference, err)
+	}
+
+	var images, attestations []ociImageSpecV1.Descriptor
+	for _, entry := range index.Manifests {
+		switch {
+		case entry.Annotations[AnnotationDockerReferenceType] != ReferenceTypeAttestationManifest:
+			images = append(images, entry)
+		case entry.Annotations[AnnotationDockerReferenceDigest] != "":
+			attestations = append(attestations, entry)
```
jneisener, line 86: So this silenty drops any `vnd.docker.reference.type: attestation-manifest` that doesn't have a digest. I assume because it would be invalid? Does this warrant a warning or is it fine to ignore?

jneisener (commented):
Some small questions / nits, looks good otherwise

bindings/go/oci/internal/attestation/sbom.go
Skarlso: How could the second one be `"sbom in unexpected format"`?

website/content/blog/2026-07-28-shipping-sboms-with-your-components.md
Skarlso: True. :) i'll rephrase it.

bindings/go/oci/internal/attestation/sbom.go
Skarlso: The buildkit code:
```go
	return &ocispecs.Descriptor{
		Digest:    mfstDigest,
		Size:      int64(len(mfstJSON)),
		MediaType: manifestType,
		Annotations: map[string]string{
			attestationTypes.DockerAnnotationReferenceType:   attestationTypes.DockerAnnotationReferenceTypeDefault,
			attestationTypes.DockerAnnotationReferenceDigest: string(target.Digest),
		},
	}, nil
```

So yeah, a manifest without digest should not appear.

website/content/blog/2026-07-28-shipping-sboms-with-your-components.md
Skarlso: Although this is the original blog post. :D So you have to ask @a-thaler :)

bindings/go/oci/internal/attestation/sbom.go
jneisener: Okay, I think unexpected format was badly phrased - or maybe entirely incorrect.

I meant that in this case `attested` was `>0` but `len(sboms` is `0`. So there was an attestation found, but no predicate matched. Yet the error type is `ErrNoAttestation`when it could be `ErrNoMatchingPredicate`.

But while writing this I realize that thats even to nitty for my taste :D

jneisener (approved):

website/content/blog/2026-07-28-shipping-sboms-with-your-components.md
jneisener: Fair point, not introduced here, so doesn't have to be fixed here

PR #3400, opened 2026-08-18
fix: toc tab linking

matthiasbruns:
#### What this PR does / why we need it
Since @Skarlso complained :P - ToC deeplinking fix

#### Which issue(s) this PR fixes
Fixes: https://github.com/open-component-model/ocm-project/issues/1152

PR #3435, opened 2026-08-21
chore: Add meta action to serve as completion guard

jneisener:
On-behalf-of: @SAP <johannes.neisener@sap.com>

##### Verification


jakobmoellerdev (commented):
I really dont think you need this. We now have much less dynamic actions we generated so you can just use inbuilt rulesets for now

jneisener:
> I really dont think you need this. We now have much less dynamic actions we generated so you can just use inbuilt rulesets for now

So my reasoning for that was 
image
in #3400 

image
in #3393


And PRs like #3409.

I.e. I think there is a general pattern of required actions not being required and that our default should instead be that any action that runs is required. Since Github actions doesn't have that capability I started sketching a solution here

PR #3463, opened 2026-08-24
fix: double close on mimetype

matthiasbruns:
fixes: https://github.com/open-component-model/open-component-model/actions/runs/32707899492/job/97372905840?pr=3460

### DIFF
--- a/bindings/go/input/file/blob.go
+++ b/bindings/go/input/file/blob.go
@@ -1,6 +1,7 @@
 package file
 
 import (
+	"errors"
 	"fmt"
 
 	"github.com/gabriel-vasile/mimetype"
@@ -44,7 +45,7 @@ var _ interface {
 //  3. Detects the media type using mimetype.DetectFile if not explicitly provided
 //  4. Wraps the blob with InputFileBlob to provide media type awareness
 //  5. Applies gzip compression with [compression.Compress] if the Compress flag is set
-func GetV1FileBlob(file v1.File, workingDirectory string) (data blob.ReadOnlyBlob, err error) {
+func GetV1FileBlob(file v1.File, workingDirectory string) (_ blob.ReadOnlyBlob, err error) {
 	if file.Path == "" {
 		return nil, fmt.Errorf("file path must not be empty")
 	}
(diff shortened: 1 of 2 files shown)


bindings/go/input/file/blob.go
```diff
@@ -55,25 +55,27 @@ func GetV1FileBlob(file v1.File, workingDirectory string) (blob.ReadOnlyBlob, er
...
+
 		// see https://github.com/gabriel-vasile/mimetype/blob/master/supported_mimes.md for supported types
 		mime, err := mimetype.DetectReader(readCloser)
-		cErr := readCloser.Close()
 		if err != nil {
-			return nil, err
-		}
-		if cErr != nil {
-			return nil, cErr
+			return nil, fmt.Errorf("error detecting media type for blob '%s': %v", file.Path, err)
```
matthiasbruns: should `GetV1FileBlob` really fail if we cannot detect the `mimetype`?

jakobmoellerdev: I questioned this to Johannes earlier, I think so. Why? because if we cant get the readcloser here we likely cannot use it from something else anyways

matthiasbruns: ack

jakobmoellerdev (dismissed):

jakobmoellerdev (dismissed):

bindings/go/input/file/blob.go
```diff
@@ -55,28 +55,28 @@ func GetV1FileBlob(file v1.File, workingDirectory string) (blob.ReadOnlyBlob, er
...
-			return nil, err
+		readCloser, rErr := b.ReadCloser()
+		if rErr != nil {
+			return nil, fmt.Errorf("error opening reader for blob '%s': %w", file.Path, rErr)
 		}
+		defer func() {
+			if cErr := readCloser.Close(); cErr != nil && err == nil {
+				data, err = nil, fmt.Errorf("error closing reader for blob '%s': %w", file.Path, cErr)
+			}
+		}()
```
Skarlso, line 69: Holy spaghetti code Batman. :D 

Just do:
```go
defer func() {
    if cErr := readCloser.Close(); cErr != nil {
        err = errors.Join(err, fmt.Errorf("error closing reader for blob %q: %w", file.Path, cErr))
    }
}()
```

Like a normal person. :D 

So the entire thing would be:
```go
func GetV1FileBlob(file v1.File, workingDirectory string) (_ blob.ReadOnlyBlob, err error) {
	if file.Path == "" {
		return nil, fmt.Errorf("file path must not be empty")
	}

	b, err := filesystem.GetBlobInWorkingDirectory(file.Path, workingDirectory)
	if err != nil {
		return nil, err
	}

	mediaType := file.MediaType

	if mediaType == "" {
		readCloser, rErr := b.ReadCloser()
		if rErr != nil {
			return nil, fmt.Errorf("error opening reader for blob %q: %w", file.Path, rErr)
		}
		defer func() {
			if cErr := readCloser.Close(); cErr != nil {
				err = errors.Join(err, fmt.Errorf("error closing reader for blob %q: %w", file.Path, cErr))
			}
		}()

		// see https://github.com/gabriel-vasile/mimetype/blob/master/supported_mimes.md for supported types
		mime, dErr := mimetype.DetectReader(readCloser)
		if dErr != nil {
			return nil, fmt.Errorf("error detecting media type for blob %q: %w", file.Path, dErr)
		}

		mediaType = mime.String()
	}

	data := blob.ReadOnlyBlob(&InputFileBlob{b, mediaType})

	if file.Compress {
		data = compression.Compress(data)
	}

	return data, nil
}
```

you don't need the `err == nil` check since errors.Join works nicely that way.

Skarlso: OR, and here me out. You could literally just put that entire internal if statement into a function and call it detectMinetype and then, you can keep it even tighter. But I'll accept what I pasted in there. :P

matthiasbruns: I shouldn't have coded myself :D 
Kidding . but that's why I usually prefer warning logs over this mangled error handling in defer

Skarlso (approved):

jakobmoellerdev (approved):

PR #3465, opened 2026-08-24
fix(constructor): allow arbitrary JSON values for labels

### LINKED ISSUE #3431: Array values for OCM component labels fail schema validation
**Context**

* I want to construct an OCM component with `cloud.gardener.cnudie/responsibles` labels and array as value (a common pattern, see [GitHub search](https://github.com/search?q=cloud.gardener.cnudie%2Fresponsibles+language%3AYAML&type=code&l=YAML))
* This is allowed by OCM spec: https://github.com/open-component-model/ocm-spec/blob/main/doc/01-model/03-elements-sub.md#labels

```
The label value may be an arbitrary JSON compatible YAML value
```

* However https://ocm.software/0.13/schemas/bindings/go/constructor/schema-2020-12.json limits labels to:

```json
        "value": {
          "oneOf": [
            {
              "description": "Value of the label",
              "$ref": "#/$defs/nonEmptyString"
            },
            {
              "type": "object",
              "description": "an arbitrarily complex structured value"
            }
          ]
        },
```

**Version**

> ocm version
{"major":"0","minor":"13","patch":"0","gitVersion":"0.13.0","goVersion":"go1.26.4","compiler":"gc","platform":"darwin/arm64"}

**To Reproduce**

`ocm add cv` the following component constructor:

```yaml
name: chrisschneider.dev/ocm-examples/temp-array-label-test
version: 1.0.0
provider:
  name: acme
labels:
  - name: cloud.gardener.cnudie/responsibles
    value:
      - github_hostname: github.com
        teamname: open-component-model/odg-maintainers
        type: githubTeam
resources:
  - name: nginx
    type: ociImage
    version: 1.25.4
    relation: external
    access:
      type: ociArtifact/v1
      imageReference: docker.io/library/nginx:1.25.4
```

**Actual behavior**

```
Error: getting component constructor failed: unmarshalling component constructor "examples/temp-array-label-test.yml" failed: error unmarshaling JSON: while decoding JSON: component constructor validation failed: jsonschema validation failed with 'file:///Users/xxx/Projects/ocm-examples/resources/schema-2020-12.json#'
- at '': 'oneOf' failed, none matched
  - at '': missing property 'components'
  - at '/labels/0/value': 'oneOf' failed, none matched
    - at '/labels/0/value': got array, want string
    - at '/labels/0/value': got array, want object
```

**Expected behavior**

The component gets constructed correctly

jakobmoellerdev:
## What
Broadens the constructor JSON schema so a label `value` may be an arbitrary JSON-compatible value (object, array, number, boolean, null) in addition to a non-empty string.

## Why
The OCM spec states a label value "may be an arbitrary JSON compatible YAML value", but the constructor schema only allowed a non-empty string or an object. This rejected common array-valued labels such as `cloud.gardener.cnudie/responsibles`. The fix also aligns the constructor schema with the descriptor v2 schema (which uses `"value": true`).

## Changes
- Expand the label `value` `oneOf` structured branch to accept `object`, `array`, `number`, `boolean`, `null`.
- Update/extend tests: array-valued and scalar label values now validate; corrected the previously-wrong "integer label value is invalid" case.

Fixes #3431


matthiasbruns (approved):
Can we somehow migrate this scheme to be generated from a struct later on? I was looking for a struct, but we just embed the scheme in the code

jakobmoellerdev:
> Can we somehow migrate this scheme to be generated from a struct later on? I was looking for a struct, but we just embed the scheme in the code

in theory yes I think the ability is there, but we might wanna patch the jsonschema gen a few times before that lands perfectly. I can take a look at that

Skarlso (approved):

PR #3468, opened 2026-08-25
fix: changelog version generation

matthiasbruns:

Promoting a changelog does not update the `versions` in our install script snippets in the changelog.
https://github.com/open-component-model/open-component-model/releases/tag/v0.13.0

This PR fixes the versions in the promote step.


.github/workflows/publish-components.yml
```diff
@@ -173,7 +176,6 @@ jobs:
               "$CLI_IMAGE_REF" \
                 sign component-version \
                 --config "/.ocmconfig" \
-                --signer-spec "${WORKSPACE}/sigstore-sign.yaml" \
```
Skarlso, line 176: 🫂

Skarlso (approved):

jakobmoellerdev (approved):

PR #3470, opened 2026-08-25
fix(ci): keep the testcontainers reaper alive for the whole integration run

matthiasbruns:

`Test_Integration_HTTPClient` fails on nearly every integration run, e.g. https://github.com/open-component-model/open-component-model/actions/runs/32829398458/job/97744547154

```
reaper: from container "a9be570b": wait for reaper a9be570b: context deadline exceeded: failed to create network
```

Raising `RYUK_RECONNECTION_TIMEOUT` (default `10s`) beyond the length of the suite keeps one reaper up for the whole run, so that gap never opens. The suite takes ~92s normally and ~8m at worst, so 30m has plenty of headroom.

##### How to test the changes

```bash
cd bindings/go
RYUK_RECONNECTION_TIMEOUT=30m go test -run Integration -count=1 ./http/integration/...
# reaper stays up after the test exits, instead of vanishing ~10s later:
docker inspect -f '{{json .Config.Env}}' $(docker ps -q --filter label=org.testcontainers.ryuk=true) | tr ',' '\n' | grep -i ryuk
```

##### Verification


jakobmoellerdev:
i dont think this is correct. we should have a normal reaper running because otherwise we can have orphaned containers

matthiasbruns:
> i dont think this is correct. we should have a normal reaper running because otherwise we can have orphaned containers

yeah I will have a look later

matthiasbruns:
Thinking about this - would that matter? The runner will be killed anyway once the workflows are done, wouldn't they?

bindings/go/http/integration/timeout_integration_test.go
```diff
@@ -167,8 +167,8 @@ func newIntegrationEnv(t *testing.T, ctx context.Context) *integrationEnv {
 	r := require.New(t)
 
 	nw, err := network.New(ctx)
+	testcontainers.CleanupNetwork(t, nw)
 	r.NoError(err, "create docker network")
-	t.Cleanup(func() { _ = nw.Remove(ctx) })
```
matthiasbruns, line 171: Every integration test already terminates its own containers via `t.Cleanup` + `TerminateContainer` 

The one exception is fixed here

matthiasbruns:
For the love of ...
https://github.com/open-component-model/open-component-model/actions/runs/32855856710/job/97827628128?pr=3447#step:6:110
:D

.github/workflows/ci.yml
```diff
@@ -358,8 +363,6 @@ jobs:
         with:
           go-version-file: '${{ github.workspace }}/${{ matrix.module }}/go.mod'
           cache-dependency-path: '${{ github.workspace }}/${{ matrix.module }}/go.sum'
-      - name: Create Docker network for running Testcontainers
-        run: docker network create ${{ env.TESTCONTAINER_DOCKER_NETWORK }}
```
Skarlso, line 362: Why is this no longer needed? 🤔

matthiasbruns: it is never used - I can keep it, but what's the benefit?
I've seen it in some python wrapper setup back in the days, but here it's useless

Skarlso: It's fine if it was never used. :)

Skarlso (dismissed):

bindings/go/http/integration/timeout_integration_test.go
matthiasbruns: rethink context handling here - use t.context

jakobmoellerdev (changes requested):
lets recheck the testing setup / cleanup please

bindings/go/http/integration/timeout_integration_test.go
matthiasbruns: https://pkg.go.dev/testing#T.Context

> Context returns a context that is canceled just before Cleanup-registered functions are called.

jakobmoellerdev (approved):

Skarlso (approved):

Skarlso:
So we went from killing it to lengthening the timeout muhahahahha :D :D :D The other end of the graph of fixing arbitrary timeouts. :D

PR #3472, opened 2026-08-25
feat: harden caching, fix strategy and add tests

matthiasbruns:
This PR is a result of the failing integration test while wiring the cli
https://github.com/open-component-model/open-component-model/pull/3433

Contributes: https://github.com/open-component-model/ocm-project/issues/1242


bindings/go/oci/cache/reference_cache.go
```diff
@@ -243,7 +243,8 @@ func (c *ReferenceCache) Resolve(ctx context.Context, upstream Resolver, ref reg
 		return upstream.Resolve(ctx, ref.Reference)
 	}
 	k := c.buildKey(ref)
-	if c.opts.RemotePolicy == RemotePolicyIfNotPresent {
+	servableFromCache := c.opts.RemotePolicy == RemotePolicyIfNotPresent && isImmutableReference(ref)
```
matthiasbruns: This should be discussed - is it ok to do a request for tags, even if they might be stored in the cache? I am not sure

Skarlso: What does v1 do in this case?

Skarlso: I mean, v1 has an object cache. If that caches tags as well, I think this should be okay.

matthiasbruns: No idea. Can check. But in this pr we would not cache tags

Skarlso: Okay.

bindings/go/oci/cache/references_test.go
```diff
@@ -50,13 +50,13 @@ func TestReferenceCache_Resolve_HitAfterFirstCall(t *testing.T) {
 		Size:      int64(len(manifest)),
 	}
 
-	var calls atomic.Int64
```
Skarlso, line 53: Why is this no longer an atomic? 🤔

matthiasbruns: that's on me - bringing it back

bindings/go/oci/cache/reference_cache.go
matthiasbruns: so v1 does get the digest for a tag and checks the digest against the cache - which makes sense - let me check our implementation

Skarlso (approved):

jakobmoellerdev (approved):

Skarlso (approved):

PR #3480, opened 2026-08-26
fix(oci): resolve resources by exact identity before falling back to subset match

akafazov:

`localArtifact` in the OCI repository used `IdentitySubset` matching when looking up resources by identity. A lookup key of `{name, version}` is a subset of any resource that has at least those two keys, so when a component version contains multiple resources sharing the same name+version but distinguished only by `extraIdentity`, all of them matched — causing "found N candidates, expected exactly one".

Fix by doing a two-pass candidate selection: exact equality (`IdentityEqual`) first, falling back to subset matching only when no exact match exists. This correctly resolves resources that share name+version but differ by `extraIdentity`, while preserving partial-identity lookups (e.g. `{name, os, architecture}` without `version`) used by existing callers.

##### How to test the changes

Save the repro script below and run it with a recent OCM binary:

```bash
OCM_BIN=/path/to/ocm bash ocm-bug-repro.sh
```

The script creates a component with 3 resources sharing `name=my-resource, version=v1.0.0` but different `extraIdentity`, transfers it into a CTF with `--copy-resources` (producing local blobs), then transfers from that CTF to a second CTF. Before this fix step 3 fails with "found 3 candidates"; after it completes successfully.

##### Verification

### DIFF
--- a/bindings/go/oci/repository.go
+++ b/bindings/go/oci/repository.go
@@ -523,34 +523,35 @@ func (repo *Repository) localArtifact(ctx context.Context, component, version st
 		return nil, nil, fmt.Errorf("failed to get component version: %w", err)
 	}
 
-	var exactCandidates, subsetCandidates []descriptor.Artifact
+	findCandidates := func(artifacts []descriptor.Artifact) []descriptor.Artifact {
+		var exact, subset []descriptor.Artifact
+		for _, a := range artifacts {
+			meta := a.GetElementMeta()
+			id := meta.ToIdentity()
+			if identity.Match(id, runtime.IdentityMatchingChainFn(runtime.IdentityEqual)) {
+				exact = append(exact, a)
+			} else if identity.Match(id, runtime.IdentityMatchingChainFn(runtime.IdentitySubset)) {
+				subset = append(subset, a)
+			}
+		}
+		if len(exact) > 0 {
+			return exact
+		}
+		return subset
+	}
+
+	var artifacts []descriptor.Artifact
 	switch kind {
 	case annotations.ArtifactKindResource:
-		for _, res := range desc.Component.Resources {
-			resIdentity := res.ToIdentity()
-			if identity.Match(resIdentity, runtime.IdentityMatchingChainFn(runtime.IdentityEqual)) {
-				exactCandidates = append(exactCandidates, &res)
-			} else if identity.Match(resIdentity, runtime.IdentityMatchingChainFn(runtime.IdentitySubset)) {
-				subsetCandidates = append(subsetCandidates, &res)
-			}
+		for i := range desc.Component.Resources {
+			artifacts = append(artifacts, &desc.Component.Resources[i])
 		}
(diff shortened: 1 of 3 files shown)


bindings/go/oci/repository.go
```diff
@@ -523,21 +523,34 @@ func (repo *Repository) localArtifact(ctx context.Context, component, version st
...
 		}
 	case annotations.ArtifactKindSource:
 		for _, src := range desc.Component.Sources {
-			if identity.Match(src.ToIdentity(), runtime.IdentityMatchingChainFn(runtime.IdentitySubset)) {
-				candidates = append(candidates, &src)
+			srcIdentity := src.ToIdentity()
+			if identity.Match(srcIdentity, runtime.IdentityMatchingChainFn(runtime.IdentityEqual)) {
+				exactCandidates = append(exactCandidates, &src)
+			} else if identity.Match(srcIdentity, runtime.IdentityMatchingChainFn(runtime.IdentitySubset)) {
+				subsetCandidates = append(subsetCandidates, &src)
```
jneisener: Minor duplication here, could be extracted into a helper/closure

jneisener (dismissed):
Thanks for the report and proposed fix, its looking good to me. One minor observation inline.

For this fix to land in the cli we'll need a follow-up, because similar logic lives [here](https://github.com/open-component-model/open-component-model/blob/c547e0579f647251fc3cfe6138de20ed5e35c444/cli/cmd/download/resource/cmd.go#L137-L143) and [here](https://github.com/open-component-model/open-component-model/blob/c547e0579f647251fc3cfe6138de20ed5e35c444/cli/cmd/download/plugin/cmd.go#L191-L194)

bindings/go/oci/repository.go
akafazov: thanks, extracted the matching in a closure func. This fixes the `transfer` command. Will open separate PR for the download commands.

jneisener (dismissed):

bindings/go/oci/repository.go
```diff
@@ -523,21 +523,35 @@ func (repo *Repository) localArtifact(ctx context.Context, component, version st
 		return nil, nil, fmt.Errorf("failed to get component version: %w", err)
 	}
 
-	var candidates []descriptor.Artifact
+	findCandidates := func(artifacts []descriptor.Artifact) []descriptor.Artifact {
```
matthiasbruns: Would you make this public? That we can reuse them in the cli?

akafazov: Sure, exposed via the public helper FindArtifactsByIdentity.

matthiasbruns (approved):

jneisener (approved):

jneisener:
@akafazov One more snag: 
grafik
See https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification

(I think you'll need to force-push over the previous commits so that every commit is signed)

PR #3482, opened 2026-08-27
fix(cli): use exact-identity-first matching for resource lookups

akafazov:
Replace the subset-only identity matching in `ocm download resource` and `ocm download plugin` with the two-pass `FindArtifactsByIdentity` helper introduced in the bindings package. Exact equality is now preferred over subset matching, so components with multiple resources sharing name+version but differing by extraIdentity are resolved correctly instead of returning ambiguous candidates.

This is a follow-up of https://github.com/open-component-model/open-component-model/pull/3480

Fixes remaining issues it subset-only identity matching in the 'ocm download ...' cli commands.

#### Testing

Unit testing.

##### How to test the changes

Unit/integration tests cover the two changed commands end-to-end using a real CTF archive:

**`ocm download resource`** (`cli/cmd/download/resource/cmd_test.go`):
- `TestDownloadResource_ExactMatchWinsOverSubset` — creates a component with two resources sharing `name=my-resource,version=1.0.0` but differing by `extraIdentity` (`architecture=amd64` vs. none). Verifies that an exact-identity lookup (`name=my-resource,version=1.0.0,architecture=amd64`) returns the amd64 variant, and that a lookup without `architecture` returns the base variant.
- `TestDownloadResource_AmbiguousSubsetMatchFails` — creates a component with two resources both requiring `extraIdentity` (`architecture=amd64` and `architecture=arm64`). Verifies that a partial lookup (`name=my-resource`) that matches both as subsets returns an error rather than silently picking one.

**`ocm download plugin`** (`cli/cmd/download/plugin/cmd_test.go`):
- `TestDownloadPlugin_ExactMatchWinsOverSubset` — creates a component with two plugin resources for `linux/amd64` and `linux/arm64`. Verifies that passing `--extra-identity os=linux --extra-identity architecture=amd64` downloads the correct binary, and same for arm64.

All tests run via `task test` and pass locally.

##### Verification


cli/cmd/download/plugin/cmd.go
```diff
@@ -188,10 +188,12 @@ func DownloadPlugin(cmd *cobra.Command, args []string) error {
 		// if the type matches we have our resource; we set the name for the identity match.
 		resourceIdentity["name"] = resource.Name
 
-		resourceIdent := resource.ToIdentity()
-		if resourceIdentity.Match(resourceIdent, ocmruntime.IdentityMatchingChainFn(ocmruntime.IdentitySubset)) {
-			toDownload = append(toDownload, resource)
-		}
+		typedArtifacts = append(typedArtifacts, &desc.Component.Resources[i])
+	}
+	candidates := descriptor.FindArtifactsByIdentity(resourceIdentity, typedArtifacts)
```
jneisener, line 189: This function will only be available after the bindings have been released. I triggered the release process, function should become available shortly in `bindings/go/v0.0.9`

akafazov: Thanks, I had a replacement in go.mod locally. I guess I'll have to rebase after the release is ready.

akafazov: Seems like the release requires approval

jneisener: Its available now: https://github.com/open-component-model/open-component-model/releases/tag/bindings/go/v0.0.9

cli/cmd/download/plugin/cmd.go
```diff
@@ -188,10 +188,12 @@ func DownloadPlugin(cmd *cobra.Command, args []string) error {
...
+	toDownload := make([]descriptor.Resource, 0, len(candidates))
+	for _, c := range candidates {
+		toDownload = append(toDownload, *c.(*descriptor.Resource))
```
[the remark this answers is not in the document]
akafazov: I am hesitant changing this, as not sure if it wasn't the intended behaviour already.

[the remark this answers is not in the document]
akafazov: Now returning error instead of the first match. Not sure if this brakes existing logic elsewhere.

jakobmoellerdev (approved):

jneisener (approved):

PR #3492, opened 2026-08-28
fix: Adjust CLI to `TempFolder` & `WorkingDiretory` being pointers

### LINKED ISSUE #1219: Filesystem config zero value is overwritten
**Description**

[At Location](https://github.com/open-component-model/open-component-model/blob/4dfe7b59/bindings/go/configuration/filesystem/v1alpha1/spec/config.go#L103-L108):

If a later config only sets WorkingDirectory, it also clears TempFolder. Similarly to transfer we should use a pointer type to be able to distinguish.

Because of the type change this is a breaking change from the code base's perspective. Meaning, release it with `!`. From the user's perspective it is not breaking.

**Done Criteria**
- [ ] Fix the overwrite

jneisener:
For context see https://github.com/open-component-model/ocm-project/issues/1169#issuecomment-4981779255

Fixes https://github.com/open-component-model/ocm-project/issues/1219 

Depends on #3491
##### Verification

### DIFF
--- a/cli/internal/plugin/builtin/input/helm/method.go
+++ b/cli/internal/plugin/builtin/input/helm/method.go
@@ -3,8 +3,6 @@ package helm
 import (
 	"fmt"
 
-	"k8s.io/utils/ptr"
-
 	filesystemv1alpha1 "ocm.software/open-component-model/bindings/go/configuration/filesystem/v1alpha1/spec"
 	helminput "ocm.software/open-component-model/bindings/go/helm/input"
 	helm "ocm.software/open-component-model/bindings/go/helm/spec/credentials"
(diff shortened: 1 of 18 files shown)


cli/internal/plugin/builtin/input/helm/method.go
```diff
@@ -3,6 +3,8 @@ package helm
 import (
 	"fmt"
 
+	"k8s.io/utils/ptr"
```
jakobmoellerdev: please dont use this its a k8s dependency in the cli

jakobmoellerdev: you can use new("") instead

jneisener: The dependency is already there because of helm, this just makes it direct. The `utils` package is also pretty lean. 

> you can use new("") instead

Okay, that only solves `ptr.To`, not `ptr.Deref`

jakobmoellerdev (approved):

matthiasbruns (approved):

PR #3495, opened 2026-08-28
feat: add constructor validation [2/2] (#1185)

matthiasbruns:
#### What this PR does / why we need it
This PR implements the constructor validation.
For this we are tacking advantage of the already existing `Validate()` methods in several specs and add it to missing ones.

The CLI takes care of the required `scheme` params in the constructor options.

Follow-up of https://github.com/open-component-model/open-component-model/pull/3496

#### Which issue(s) this PR fixes
Fixes: https://github.com/open-component-model/ocm-project/issues/1185

#### Testing

- Adds cli integration tests that verify that constructions fails on incorrect resources.

##### Verification

- [x] I have added/updated tests for my changes (see [Test Requirements](../CONTRIBUTING.md#test-requirements))
- [x] Tests pass locally (`task test` and `task test/integration` if applicable)
- [x] My changes do not decrease test coverage
- [ ] I have tested the changes locally by running `ocm`

PR #3496, opened 2026-08-28
feat: add constructor validation [1/2] (#1185)

### LINKED ISSUE #1185: introduce schema validation into the constructor
**Description**
While discussing the github access type, we noticed that we currently do not perform any kind of json schema validation during `ocm add cv`. 

For unknown access types, this is alright, as we want users to be able to add their custom access types. But for known access types, this should be validated. 

**Done Criteria**

matthiasbruns:
This PR implements the constructor validation.
For this we are tacking advantage of the already existing `Validate()` methods in several specs and add it to missing ones.

The CLI takes care of the required `scheme` params in the constructor options.

Fixes: https://github.com/open-component-model/ocm-project/issues/1185

##### Verification


bindings/go/wget/spec/input/v1/wget.go
```diff
@@ -43,3 +47,18 @@ type Wget struct {
...
+	if t.URL == "" {
+		return errors.New("url is required")
+	}
+	parsed, err := url.Parse(t.URL)
+	if err != nil {
+		return fmt.Errorf("invalid url %q: %w", t.URL, err)
+	}
+	if parsed.Scheme != "http" && parsed.Scheme != "https" {
+		return fmt.Errorf("url must use the http or https scheme, got %q", parsed.Scheme)
+	}
```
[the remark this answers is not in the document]
matthiasbruns: wont do - thats not something I would care about at this point

bindings/go/helm/spec/access/v1/helm.go
```diff
@@ -43,6 +44,21 @@ type Helm struct {
...
 }
 
+// Validate verifies that the required fields of the Helm access are set and that the helm
+// repository is a well formed URL.
+func (h *Helm) Validate() error {
+	if h.HelmRepository == "" {
+		return errors.New("helmRepository is required")
+	}
+	if h.HelmChart == "" {
+		return errors.New("helmChart is required")
```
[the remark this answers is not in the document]
matthiasbruns: just moved the logic . wont touch it

bindings/go/constructor/construct.go
```diff
@@ -105,6 +105,10 @@ func (c *DefaultConstructor) Construct(ctx context.Context) error {
 		return nil
 	}
 
+	if err := validateSpecifications(c.constructor, c.opts); err != nil {
+		return fmt.Errorf("failed to validate component constructor specifications: %w", err)
```
piotrjanik: nit: Wouldn't "component constructor contains invalid access or input specifications" be more informative?

bindings/go/constructor/validation.go
```diff
@@ -0,0 +1,89 @@
...
+	if err != nil {
+		return fmt.Errorf("cannot create an object of type %q: %w", typ, err)
+	}
+	if err := scheme.Convert(spec, obj); err != nil {
+		return fmt.Errorf("type %q cannot be decoded: %w", typ, err)
+	}
+
+	validatable, ok := obj.(ocmruntime.Validatable)
+	if !ok {
+		return nil
```
piotrjanik: nit: Should we log a warning or debug message indicating that this type cannot be validated?

bindings/go/constructor/validation.go
```diff
@@ -0,0 +1,89 @@
...
+		}
+	}
+
+	return nil
+}
+
+// validateSpecification decodes spec into the type registered for it and, if that type implements
+// ocmruntime.Validatable, checks it with Validate. A specification of a type that is not registered
+// is not validated.
+func validateSpecification(spec ocmruntime.Typed, scheme *ocmruntime.Scheme) error {
```
jakobmoellerdev: the constructor package should already have a scheme based validation that I was looking for here

matthiasbruns: I kept the original Validate method in `AccessOrInput`, added a new one - hope that was, what you meant

jakobmoellerdev: this is not what I meant. the entire constructor has a JSON Schema that we should validate against and this is the authoritive source for validation

jakobmoellerdev: https://github.com/open-component-model/open-component-model/blob/main/bindings/go/constructor/spec/v1/resources/schema-2020-12.json

matthiasbruns: ok, but this would be the general constructor scheme - this PR takes care of access and input type validation, which is not covered fully by the scheme - I can add the constructor scheme validation as a first step as well, if not already happening

jakobmoellerdev: Hmmm, you have 2 things here. we need a static validation against the schema and also a runtime validation i guess when the plugins are known

bindings/go/constructor/validation_test.go
```diff
@@ -0,0 +1,297 @@
...
+    access:
+      type: validatedAccess/v1
+      url: https://ocm.software
+  sources:
+  - name: by-input
+    type: blob
+    version: 1.0.0
+    input:
+      type: validatedInput/v1
+      path: ./some/path
```
piotrjanik, line 122: I think the validation logic ignores unknown fields in a known type.
So, if this `validatedInput` type had an optional field like `compress` and the user typed `compres`, it would be accepted as a valid input type, but no compression would occur.

So I would say that for known types we should check for unknown fields.

wdyt?

matthiasbruns: Yeah thats true . I wasn't sure hard _hard_ I should go on validation

matthiasbruns: I will think about this more when I am back

bindings/go/constructor/validation.go
matthiasbruns: So
if you check https://github.com/open-component-model/open-component-model/pull/3495
both are implemented

the constructor check was already there, I added access and input checks on top of that. `cli` takes care of providing known schemas through the plugin registries and sets them during construction. Imho this is done - I cannot think of better design as well

bindings/go/constructor/validation_test.go
matthiasbruns: decided to hard error out

matthiasbruns: no wait -- we do not have support for this in runtime at all - thus every existing type convert allows other fields, maybe we should not break it during construction at all

piotrjanik: But shouldn't we validate each versioned input/output spec for unknown fields in known types? I still see a value in it, like typical typo. We can have a follow-up task for it.

matthiasbruns: Yeah I feel this is a design discussion for the whole stack :) I will create an issue

matthiasbruns: https://github.com/open-component-model/ocm-project/issues/1273

piotrjanik (approved):

Skarlso (approved):

PR #3518, opened 2026-08-31
chore: lock website to the same 28 days gate as the renovate lockfile

Skarlso:

https://github.com/open-component-model/open-component-model/pull/3458

Fixes this behavior. Note. If we merge this, the website updates will go quite for a while because it will wait for the actual 28 day interval to occur.

##### Verification


jneisener (dismissed):

.github/renovate.json5
```diff
@@ -236,6 +236,16 @@
         '* * * * 0' // At every minute on Sunday it is allowed to create branches
       ],
     },
+    {
+      // Restrict the website update to 28 days as well.
+      matchDatasources: ['npm'],
+      minimumReleaseAge: '28 days',
+      internalChecksFilter: 'strict',
```
jakobmoellerdev: why dont you just group them together?

Skarlso: because I didn't think of that. :D :D

Skarlso: Wait, group them together how? They are separate matchUpdateTypes setting the same value for different Types and this is AND-ed.. How would I group this together? 🤔

.github/renovate.json5
```diff
@@ -236,6 +236,16 @@
         '* * * * 0' // At every minute on Sunday it is allowed to create branches
       ],
     },
+    {
+      // Restrict the website update to 28 days as well.
+      matchDatasources: ['npm'],
```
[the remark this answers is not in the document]
Skarlso: The point of this thing was to fix all of the npm updates including the scripts. Doing this would mitigate that part of the fix.

.github/renovate.json5
Skarlso: Ah, you mean `internalChecksFilter`? Actually, I can remove this. This is the default. :) So this has no effect here.

jakobmoellerdev: i mean grouping the updates of both managers, you can do that with the config by aggregating them to one rule set

Skarlso: @jakobmoellerdev Sorry, somehow missed this comment. 😞 

But wouldn't that mean that it would break the fix that this supports which is that they are separate deliberately? If I put them into one group an update failure in one, kills the entire update.

So if I get this right, I can do something like this:

```json5
{
   matchDatasources: ['npm'],
   matchUpdateTypes: ['minor', 'patch'],
   groupName: 'NPM dependencies (non-major)',
   groupSlug: 'npm-deps',
   separateMinorPatch: false,  -> I think I need this because the above config also sets it to separate the minors, right?
}
```

.github/renovate.json5
```diff
@@ -236,6 +236,15 @@
...
       ],
     },
+    {
+      // Restrict NPM updates to 28 days old packages at least.
+      matchDatasources: ['npm'],
+      minimumReleaseAge: '28 days',
+    },
+    {
+      matchUpdateTypes: ['lockFileMaintenance'],
+      minimumReleaseAge: null,
```
piotrjanik, line 253: wouldn't this mean that Renovate from now on would regenerate package-lock.json without grace period?

Skarlso: Yes, that's the intent. The maintenance is supposed to be only refreshing the lockfile.

Skarlso: Ah no wait. This would mean transitive dependencies get refreshed with no age minimum when the lockfile resets. 🤔

Skarlso: This was a good catch! :)

jneisener (approved):

matthiasbruns (approved):
let'se go

PR #3526, opened 2026-09-02
fix: rendering errors in `ocm add cv` for unregistered types

### LINKED ISSUE #3279: `ocm add cv -o json|yaml|ndjson|tree` fails with "cannot encode from unregistered type: ociArtifact"
**Context**

`ocm add component-version` errors out for every non-`table` output format as soon as the constructor contains a resource with a typed access spec (`ociArtifact`, `helm`, `gitHub`, `wget`, …). The `transport-archive/` is still written correctly; only the terminal rendering after the write fails, so the command exits non-zero and looks like a full failure to the caller. Default `-o table` works. `ocm get cv -o json|yaml` on the same archive afterwards also works.

**Version**

`ocm v0.12.0` (also reproduces on `main`).

```
$ ocm version
{"major":"0","minor":"12","patch":"0","gitVersion":"0.12.0","goVersion":"go1.26.4","compiler":"gc","platform":"darwin/arm64"}
```

**To Reproduce**

Environment: macOS, bash. No `~/.ocmconfig` needed.

```bash
mkdir /tmp/ocm-repro && cd /tmp/ocm-repro

cat > component-constructor.yaml <<'EOF'
components:
  - name: ocm.software/repro/oci-access
    version: 1.0.0
    provider:
      name: ocm.software
    resources:
      - name: image
        type: ociArtifact
        version: 1.0.0
        access:
          type: ociArtifact
          imageReference: ghcr.io/stefanprodan/charts/podinfo:6.9.1
EOF

ocm add cv -o json
```

Same failure with `-o yaml`, `-o ndjson`, `-o tree`. Only `-o table` (default) succeeds.

**Actual behavior**

```
Error: failed to render components recursively: failed to render graph: failed to
traverse and render graph in read lock: failed to render objects: failed to encode
vertices: failed to serialize vertex name=ocm.software/repro/oci-access,version=1.0.0:
converting descriptor to v2 failed: could not convert resources: could not convert
resource "image": could not convert access "image:1.0.0":
cannot encode from unregistered type: ociArtifact
```

Exit code `1`. The default `transport-archive/` is written correctly regardless.

**Expected behavior**

Either

1. `add cv -o json|yaml|ndjson|tree` prints the descriptor of the just-added component version(s) in the requested format, or
2. `add cv` does not offer these output flags in the first place (the flag is registered but the render path cannot handle typed access specs; if descriptor output is not a goal for a write command, removing the flag values is fine).

jneisener:
`add cv` render path converts the constructed descriptor with `descriptorv2.Scheme` (only LocalBlob registered). Constructor produces typed access specs (e.g. ociArtifact), so Typed->Raw encoding fails registration check. `get cv` works because its access is already Raw.

Proposed fix: render with a scheme built via `runtime.NewScheme(runtime.WithAllowUnknown())` +`descriptorv2.MustAddToScheme`.

Fixes: https://github.com/open-component-model/open-component-model/issues/3279

#### Testing
Example provided in bug report works locally:
```
mkdir /tmp/ocm-repro && cd /tmp/ocm-repro

cat > component-constructor.yaml <<'EOF'
components:
  - name: ocm.software/repro/oci-access
    version: 1.0.0
    provider:
      name: ocm.software
    resources:
      - name: image
        type: ociArtifact
        version: 1.0.0
        access:
          type: ociArtifact
          imageReference: ghcr.io/stefanprodan/charts/podinfo:6.9.1
EOF

ocm add cv -o json
```
=>
```
[
  {
    "meta": {
      "schemaVersion": "v2"
    },
    "component": {
      "name": "ocm.software/repro/oci-access",
      "version": "1.0.0",
      "repositoryContexts": null,
      "provider": "ocm.software",
      "resources": [
        {
          "name": "image",
          "version": "1.0.0",
          "type": "ociArtifact",
          "relation": "external",
          "access": {
            "imageReference": "ghcr.io/stefanprodan/charts/podinfo:6.9.1@sha256:565d310746f1fa4be7f93ba7965bb393153a2d57a15cfe5befc909b790a73f8a",
            "type": "ociArtifact"
          },
          "digest": {
            "hashAlgorithm": "SHA-256",
            "normalisationAlgorithm": "genericBlobDigest/v1",
            "value": "565d310746f1fa4be7f93ba7965bb393153a2d57a15cfe5befc909b790a73f8a"
          }
        }
      ],
      "sources": null,
      "componentReferences": null
    }
  }
]
```

##### Verification


bindings/go/cli/cmd/add/component-version/cmd.go
```diff
@@ -599,6 +599,17 @@ func buildRenderer(ctx context.Context, graph *syncdag.SyncedDirectedAcyclicGrap
 	}
 }
 
+// renderScheme converts internal descriptors to their v2 representation for
+// output rendering. Unlike descriptorv2.Scheme it allows unknown (typed but
+// unregistered) access and input specs. A freshly constructed component version
+// can still carry typed access specs (for example ociArtifact) that are not
+// registered in the descriptor scheme.
+var renderScheme = func() *runtime.Scheme {
```
matthiasbruns, line 609: wouldnt it make more sense to actually add all known type here? we should have them ready in the plugins

jneisener: But what would the benefit be?

Seems to me like this would create a list that has the risk to grow stale.

matthiasbruns: no - it is already registered in the plugins itself - e.g. check how describe works

jneisener: Ok, good clarification - but I'm still wondering why thats better than the current approach?

matthiasbruns: is... probably ... isnt D:

matthiasbruns (approved):

fabianburth:
Can you explain why we need the `allow unknown` for add cv but not for get cv?

jneisener:
> Can you explain why we need the `allow unknown` for add cv but not for get cv?

Admittedly, I looked at `ocm add` because that was in the issue description and I didn't check how `ocm get` is different.
So, `ocm get` ends up calling [`GetComponentVersionRepositoryForComponent`](https://github.com/open-component-model/open-component-model/blob/208356020be05b6f9f8a5e5e1e72a8861a1fced5/bindings/go/cli/cmd/get/component-version/cmd.go#L362) and through `GetComponentVersion` we call[`SingleFileDecodeDescriptor`](https://github.com/open-component-model/open-component-model/blob/208356020be05b6f9f8a5e5e1e72a8861a1fced5/bindings/go/oci/spec/descriptor/decode.go#L49-L52), there its unmarshalled and passed to `ConvertFromV2Resource` and then the raw access is just deepCopied in [`ConvertFromV2Resource`](
https://github.com/open-component-model/open-component-model/blob/208356020be05b6f9f8a5e5e1e72a8861a1fced5/bindings/go/descriptor/runtime/convert_v2.go#L182-L184).
So to me it looks like the `get` path is raw and stays raw, while `add` is actually processed e.g. `ProcessResourceDigest`, because `add` needs to be able to write things like the digetst, while `get` only needs to read. 

Thats how it looks like to me at least, does that make sense to you?

jneisener:
@fabianburth So where are we on this? Is this approach okay or does something need to be changed?

frewilhelm (approved):
lgtm - thank you!

PR #3528, opened 2026-09-02
fix: handle special characters in tranfer IDs; ensure uniqueness

### LINKED ISSUE #3520: ocm transfer: resource version with SemVer build metadata (x.y.z+meta) produces an invalid transformation id
## What

`ocm transfer cv` fails when a resource's `version` contains SemVer **build metadata** (`x.y.z+meta`). The `+meta` segment leaks into the derived transformation id, which then fails OCM's own lower-camelCase id validation.

`ocm add cv` accepts the same version, so the failure only surfaces at transfer time. SemVer pre-release versions (`x.y.z-foo`) are unaffected — only `+` build metadata breaks.

- `ocm` CLI: `0.14.0`

## Repro

Resource with a build-metadata version:

```yaml
resources:
  - name: operator-image
    type: ociImage
    version: "0.2.1+a0b6f97"          # x.y.z+meta
    access:
      type: ociArtifact
      imageReference: ghcr.io/example/img:a0b6f97@sha256:7d1f68...
```

```sh
ocm add cv --repository ./ctf --constructor component-constructor.yaml   # ok
ocm transfer cv --copy-resources --upload-as ociArtifact --recursive \
  ctf::./ctf//ocm.software/concourse-operator:0.5.2 <registry>           # fails
```

```
Error: naming convention violation: id transform...TransfertransformOperatorImage021+a0b6f97
  is not a valid OCM transformations id: must be lower camelCase
```

(`0.2.1` → `021`, but `+a0b6f97` passes through unchanged into the id.)

## Expected

Either normalize the version when deriving the transformation id (strip build metadata), or reject such versions in `ocm add cv` instead of failing later at transfer.

## Workaround

Drop the build metadata (`0.2.1`).

jneisener:
`var toWordRunes = []rune{',', '.', '/', '-'}` was incomplete and missed e.g. `+`. Instead we should be able to split on any non-alphanumeric character thats still allowed in an ID

As the bot [mentioned](https://github.com/open-component-model/open-component-model/pull/3528#discussion_r3915426340) this increases the already existing potential for name collisions, so we add disambiguation via postfix to the ID generation process.

Fixes https://github.com/open-component-model/open-component-model/issues/3520

##### Verification


bindings/go/transfer/internal/discovery.go
```diff
@@ -248,9 +254,7 @@ func identityToTransformationID(id runtime.Identity) string {
 	}
 	sort.Strings(keys)
 	for _, k := range keys {
-		words = append(words, strings.FieldsFunc(id[k], func(r rune) bool {
-			return slices.Contains(toWordRunes, r)
-		})...)
+		words = append(words, strings.FieldsFunc(id[k], isTransformationIDWordBoundary)...)
```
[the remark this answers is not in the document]
jneisener: Technically true, but already possible before with e.g. `.` and `-`. Now we have the same potential conflict with `+` as well.

Should metadata even be allowed then? Because with it a component could contain two resources that only differ in the metadata, like `1.0.0+build1` and `1.0.0+build2`

bindings/go/transfer/internal/graph.go
```diff
@@ -288,23 +293,21 @@ func logSkippedResource(ctx context.Context, component, version string, resource
 // target repository. wget and s3 resources are always downloaded and embedded as local blobs.
 // It returns CEL spec-field expressions for the file buffers produced, referencing consumer spec
 // fields (not producer outputs) so the DAG edge points from consumer to the cleanup node.
-func processResource(resource descriptorv2.Resource, access runtime.Typed, id string, val *discoveryValue, tgd *transformv1alpha1.TransformationGraphDefinition, toSpec runtime.Typed, resourceTransformIDs map[int]string, i int, uploadType transferv1alpha1.UploadType) ([]string, error) {
+func processResource(resource descriptorv2.Resource, access runtime.Typed, id string, resourceID string, val *discoveryValue, tgd *transformv1alpha1.TransformationGraphDefinition, toSpec runtime.Typed, resourceTransformIDs map[int]string, i int, uploadType transferv1alpha1.UploadType) ([]string, error) {
```
matthiasbruns, line 302: this param list is getting out of hand

jneisener: Would you prefer a parameter object?

jneisener: e.g. 
```
type resourceProcessingContext struct {                                                                                                                                                 
      id         string                                                                                                                            
      tgd        *transformv1alpha1.TransformationGraphDefinition                                                                                                                          
      toSpec     runtime.Typed                                                                                                                                                             
      uploadType transferv1alpha1.UploadType                                                                                                                                               
   }     
```
and `resource, access, resourceID, i` stay as parameters

matthiasbruns: Maybe I would refactor the id to a real type and move the id logic into there and the callers can either pick the resource id or the id

bindings/go/transfer/internal/graph_test.go
```diff
@@ -215,8 +215,10 @@ func TestBuildGraphDefinition_CollidingResourceVersionsGetUniqueIDs(t *testing.T
 	for _, tr := range tgd.Transformations {
 		ids = append(ids, tr.ID)
 	}
-	r.Contains(ids, "transformOcmSoftwareTest100GettransformOperatorImage021Meta")
-	r.Contains(ids, "transformOcmSoftwareTest100GettransformOperatorImage021MetaR1")
+	// the escaping is injective: build metadata and pre-release versions produce
+	// different IDs without any disambiguation suffix
+	r.Contains(ids, "transform_slash_ocm_dot_software_slash_test_slash_1_dot_0_dot_0Gettransform_slash_operator_dash_image_slash_0_dot_2_dot_1_plus_meta")
```
matthiasbruns: holy moly

bindings/go/transfer/internal/discovery.go
```diff
@@ -229,35 +229,54 @@ func (d *discoverer) Discover(ctx context.Context, parent *discoveryValue) ([]st
...
-		})...)
+		words = append(words, id[k])
 	}
-	result := strings.ToLower(words[0])
-	for i := 1; i < len(words); i++ {
-		w := strings.ToLower(words[i])
-		if len(w) > 0 {
-			result += strings.ToUpper(w[:1]) + w[1:]
+	return escapeTransformationID(strings.Join(words, "/"))
+}
```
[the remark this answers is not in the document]
jneisener: Fair finding, but not related to the changes in this PR. This is a pre-existing gap

bindings/go/transfer/internal/discovery.go
```diff
@@ -229,35 +229,54 @@ func (d *discoverer) Discover(ctx context.Context, parent *discoveryValue) ([]st
...
+		case "+":
+			return "_plus_"
+		case "/":
+			return "_slash_"
 		}
-	}
-	return result
+		return s
+	})
+	return ident
```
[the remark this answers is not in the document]
jneisener: Again pre-existing gap

bindings/go/transform/graph/builder/validation.go
```diff
@@ -94,7 +89,7 @@ func isOCMReservedWord(word string) bool {
 	return false
 }
 
-// isValidResourceID checks if the given id is a valid OCM resource id (loawercase)
+// isValidResourceID checks if the given id is a valid CEL identifier
 func isValidResourceID(id string) bool {
-	return lowerCamelCaseRegex.MatchString(id)
+	return celIdentifierRegex.MatchString(id)
```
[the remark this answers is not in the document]
jneisener: Also pre-existing

bindings/go/transfer/internal/graph_test.go
matthiasbruns: how did we end up here? I remember the discussion - why didnt we follow the "increase a counter" - idea and keep the simple naming + a incrementing number? :D

jneisener: Well you didn't like the number of parameters and refactoring to a type made the changeset huge because it affected all the places :D

But with https://github.com/open-component-model/open-component-model/pull/3528#discussion_r3965885399 and https://github.com/open-component-model/open-component-model/pull/3528#discussion_r3965885395 I might want to reconsider.

I'm still not convinced the id type is worth it though, I think I prefer another parameter

jneisener: Another issue with the counting approach is also that IDs are not guaranteed to be stable. I'm not sure thats an actual problem; it would only be if TGDs are diffed by people.

And the lack of a deterministic backwards translation of the transfer IDs might be annoying when debugging

bindings/go/transfer/internal/graph.go
```diff
@@ -173,15 +175,22 @@ func fillGraphDefinitionWithPrefetchedComponents(
...
 
-	for key, v := range d.Vertices {
+	// allocate component base IDs and target-suffixed IDs from one allocator so that a
+	// collision between a target suffix and an identity-folded base ID (for example
+	// "a" with two targets vs. component "a-t0") is disambiguated as well.
+	componentAllocator := newTransformationIDAllocator()
+	vertexKeys := slices.Sorted(maps.Keys(d.Vertices))
+
+	for _, key := range vertexKeys {
+		v := d.Vertices[key]
```
jneisener, line 185: Technically there can still be collisions between different components, e.g.
```
 - component a + resource x-gettransform-y → transformAGettransformXGettransformY                                                                                                          
 - component a-gettransform-x + resource y → transformAGettransformXGettransformY   
```

But I think at this point the naming scheme is just unrealistic to happen accidentally

jneisener:
@coderabbitai review

jakobmoellerdev:
how realistic is the actual collision? i feel like postfixing is maybe not even needed?

jneisener:
> how realistic is the actual collision? i feel like postfixing is maybe not even needed?

Depends on the case I think, some cases are very constructed, others might actually happen by accident.

Mixing `-`, `_`, and `.` for example is not unheard of:
```yaml
component:
  name: ocm.software/example
  version: 1.4.0
  resources:
  - name: manifests_base
    version: 1.4.0
  - name: manifests-base
    version: 1.4.0
```
(without postfixing both get `...ManifestsBase140`).

We don't own `T{x}` postfix, so that might be in a name (e.g. referring to tiers):
```yaml
component:
  name: ocm.software/example-t0
...
component:
  name: ocm.software/example
```
(these would clash on the preexisting `T` postfix)


Mixing prerelease and meta seems pretty unlikely:
```yaml
component:
  name: ocm.software/example/sample-app
  version: 1.0.0-rc.1
  provider: ocm.software
...
component:
  name: ocm.software/example/sample-app
  version: 1.0.0+rc.1
```
(both get `...SampleApp100Rc1`)

jakobmoellerdev (commented):
I thought about it a lot and reject the current proposal as the original proposition of human readable identifiers is no longer possible. Please consider how to simplify this and improve the UX (i.e. by no longer using these ids as readable identifiers during transfer like in https://github.com/open-component-model/open-component-model/blob/main/docs/adr/0023_human_readable_transformation_labels.md)

jneisener:
Closed in favor of #3612

PR #3531, opened 2026-09-02
fix(controller): requeue with backoff when a referenced object is not ready

jakobmoellerdev:
## What

The `resource`, `component`, `replication`, and `deployer` controllers all gate reconciliation on a referenced object being `Ready` via `util.GetReadyObject`. On the "not ready" (or "being deleted") branch they returned `ctrl.Result{}, nil`.

This changes that branch to **return the error** instead, so controller-runtime requeues the dependent with the exponential backoff each controller already configures (`5ms→5min`).

## Why — when the old behaviour bites

`return ctrl.Result{}, nil` is controller-runtime's *"reconcile succeeded, done"* outcome. It sends the request to `Queue.Forget` and **enqueues nothing**. Recovery then depends entirely on a watch event on the referenced object firing later — and that event can be missed:

- **Coalesced watch events.** The Component→Resource watch (and the analogous ones) use `ComponentInfoChangedPredicate`, which only passes an `Update` when `Status.Component`, `EffectiveOCMConfig`, or the `Ready` `True↔False` boolean differs between the cached old and new objects. If a referenced object flips `Ready` `True→False→True` inside one informer resync window — e.g. a `Component` re-reconciling on its periodic interval and briefly rewriting status — the informer can deliver a **single coalesced event** whose old and new both observe `Ready=True`. The predicate drops it, and the dependent is never re-enqueued.
- **No periodic safety net.** `resource`, `replication`, and `deployer` have no `spec.interval` of their own, so there is no periodic resync to paper over a missed event. The object is stranded `NotReady` **indefinitely**, even though its dependency became ready long ago. (`component` *does* have an interval and self-heals on the next tick — but that's accidental cover; it still reports a stale `NotReady` for up to a full interval.)

### Observed failure

In an air-gapped workload cluster, a Concourse `Resource` sat `Ready=False` (`"object is not ready: default/concourse-component"`) for 20+ minutes after its `Component` had returned to `Ready=True` **at the same second** — the recovery event was coalesced away and nothing re-triggered the Resource. That in turn pinned kro's aggregate `Instance` state at `IN_PROGRESS`, even though the underlying HelmRelease and pods were healthy. Only the next hourly `Component` resync would have unstuck it.

## The fix

On the not-ready / being-deleted branch, return the error rather than a bare `nil` result. That routes the request into controller-runtime's `err != nil` path (`AddWithOpts{RateLimited: true}`), giving the intended exponential backoff. The dependent now retries until its dependency is ready **regardless of whether the watch event was delivered**, while still marking itself `NotReady` for observability.

The **status** the controllers report is unchanged — only the requeue behaviour differs.

## Controllers touched

| Controller | Dependency gated on | Had a periodic interval? |
|---|---|---|
| `resource` | `Component` ready | no |
| `replication` | `Component` + target `Repository` ready | no |
| `deployer` | `Resource` ready | no |
| `component` | `Repository` ready | yes (masked, still fixed) |

## Tests

- Component unit tests that asserted the old no-error/no-requeue behaviour are updated to assert the error-driven backoff and renamed accordingly (`…ReturnsNoRequeue` → `…RequeuesWithBackoff`).
- envtest specs that assert the object reaches `NotReady` are unaffected (the reported status is identical).
- Full `internal/controller/...` suite passes locally against this branch's base (`upstream/main`).

## Notes

- Opened as a **draft** for maintainer discussion on the chosen idiom (return the error vs. a fixed `RequeueAfter` on the not-ready branch). Returning the error reuses the already-configured rate limiter and needs no new API field; a fixed `RequeueAfter` would be quieter in logs/metrics but is a separate design choice.
- Scope intentionally excludes any not-yet-merged controllers on feature branches.


bindings/go/kubernetes/controller/internal/controller/component/component_controller.go
```diff
@@ -225,7 +225,12 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (_ ctrl.Re
 		if errors.As(err, &notReadyErr) || errors.As(err, &deletionErr) {
 			logger.Info("repository is not available", "error", err)
 
-			return ctrl.Result{}, nil
+			// Return the error so controller-runtime requeues with the configured
+			// exponential backoff. Returning ctrl.Result{}, nil would Forget the
+			// item and leave recovery solely to a Repository watch event, which can
+			// be missed (e.g. a coalesced Ready flip) — stranding the Component
+			// NotReady until the next periodic interval resync papers over it.
+			return ctrl.Result{}, err
```
matthiasbruns, line 228: I am pretty sure I was sitting next to @Skarlso (virtually) when errors for reconciliation were removed but this makes sense. I wonder if I am tripping and if not, what the reasoning was :D

Skarlso: I remember this discussion.

TL;DR:

The problem with erroring here was that in a GitOps flow, when the repo was not there yet, this would backoff into minutes of waiting time even though the repository was there in seconds after, reconciled. The other idea we had here is to just return a constant 1s or 2s check instead of a potential backoff loop wait.

So all in all, we might want to do that here, I don't believe the error response here is beneficial. 🤔 A flat value might be better honestly. Something like, "repository is not yet available, retrying in 5s".

Skarlso: Tested this in the warroom and this worked fine :)

Skarlso (dismissed):

fabianburth (dismissed):

jneisener (approved):
Didn't even have to rebase, merged cleanly:
```
git pull upstream main
From https://github.com/open-component-model/open-component-model
 * branch                main       -> FETCH_HEAD
Auto-merging bindings/go/kubernetes/controller/internal/controller/component/component_controller.go
Auto-merging bindings/go/kubernetes/controller/internal/controller/component/component_controller_unit_test.go
Auto-merging bindings/go/kubernetes/controller/internal/controller/deployer/deployer_controller.go
Auto-merging bindings/go/kubernetes/controller/internal/controller/replication/replication_controller.go
Auto-merging bindings/go/kubernetes/controller/internal/controller/resource/resource_controller.go
Auto-merging bindings/go/kubernetes/controller/internal/controller/resource/resource_controller_test.go
Merge made by the 'ort' strategy.
```

Skarlso (approved):

PR #3541, opened 2026-09-04
chore: Generate dependency table & readme overview from linting rules

jneisener:
Table in `bindings/go/README.md` was stale, it should be generated just like the https://github.com/open-component-model/open-component-model/blob/main/docs/dependency-table.md

##### Verification


.github/scripts/dependency-list.js
```diff
@@ -0,0 +1,271 @@
...
+        const allows = (rule.allow ?? [])
+            .filter((/** @type {string} */ a) => a.startsWith(BINDING_PREFIX))
+            .map((/** @type {string} */ a) => a.slice(BINDING_PREFIX.length))
+            .filter((/** @type {string} */ a) => a !== binding);
+
+        deps.set(binding, allows.sort());
+    }
+
+    return deps;
+}
```
jneisener, line 54: File was renamed and modified, so the diff is deceptive. This part is unchanged from before

.github/scripts/dependency-list.js
```diff
@@ -0,0 +1,271 @@
...
+        for (const [depsStr, mods] of groups) {
+            const modCell = mods.map((m) => `\`${m}\``).join(", ");
+            const depsCell = depsStr ? depsStr.split(", ").map((d) => `\`${d}\``).join(", ") : "—";
+            lines.push(`| ${layer} | ${modCell} | ${depsCell} |`);
+        }
+    }
+
+    lines.push("");
+    return lines.join("\n");
+}
```
jneisener, line 236: Unchanged

.github/scripts/dependency-list.js
```diff
@@ -0,0 +1,301 @@
...
+        const allows = (rule.allow ?? [])
+            .filter((/** @type {string} */ a) => a.startsWith(BINDING_PREFIX))
+            .map((/** @type {string} */ a) => a.slice(BINDING_PREFIX.length))
+            .filter((/** @type {string} */ a) => a !== binding);
+
+        deps.set(binding, allows.sort());
+    }
+
+    return deps;
+}
```
jneisener, line 51: File was renamed and modified, so the diff is deceptive. This part is unchanged from before

.github/scripts/dependency-list.js
```diff
@@ -0,0 +1,301 @@
...
+        for (const [depsStr, mods] of groups) {
+            const modCell = mods.map((m) => `\`${m}\``).join(", ");
+            const depsCell = depsStr ? depsStr.split(", ").map((d) => `\`${d}\``).join(", ") : "—";
+            lines.push(`| ${layer} | ${modCell} | ${depsCell} |`);
+        }
+    }
+
+    lines.push("");
+    return lines.join("\n");
+}
```
jneisener, line 233: Block is also unchanged

.github/scripts/dependency-list.js
```diff
@@ -0,0 +1,301 @@
...
+ * @returns {string}
+ */
+export function generatePackagesTable(deps, bindingsDir) {
+    const names = [...deps.keys()].sort();
+    /** @type {string[]} */
+    const missing = [];
+    const rows = [];
+    for (const name of names) {
+        try {
+            rows.push(`| **${name}** | ${extractPackageSummary(join(bindingsDir, name, "doc.go"))} |`);
```
matthiasbruns, line 104: what happens if we push a row with "|" in it?

jneisener: Package name can't contain a pipe, the first sentence of the docstring would have to.
But even then the linter would complain about too many cells in the generated readme

.github/scripts/dependency-list.js
```diff
@@ -0,0 +1,301 @@
...
+    }
+    // Only the first paragraph of the doc comment is considered.
+    const paragraph = [];
+    for (const line of comment) {
+        const text = line.replace(/^\/\/ ?/, "");
+        if (text === "") break;
+        paragraph.push(text);
+    }
+    // The first sentence ends at the first period followed by whitespace or the end of the paragraph.
+    const sentence = /^.*?\.(?=\s|$)/s.exec(paragraph.join(" "))?.[0] ?? paragraph.join(" ");
```
matthiasbruns, line 83:
```
// Package s3 provides access to OCM resources stored as objects in
// S3-compatible object stores, e.g. AWS S3, MinIO or Ceph.
```

would stop at e.g. right?

jneisener: Yep, just like the synopsis would https://pkg.go.dev/go/doc#Package.Synopsis

matthiasbruns: What about

`^.*?\.(?=\s|$)\X+`

jneisener: `\X` isn't implemented in Javascript https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Regular_expressions/Character_classes

Also, what exactly is the usecase of supporting a docstring that go itself would not render?

matthiasbruns (approved):

Skarlso (approved):

PR #3577, opened 2026-09-08
feat: git access

### LINKED ISSUE #1257: Git access type specification, credentials and consumer identity
**Description**

Create the `Git/v1` access spec in `bindings/go/git/spec/access/v1/`, together with the typed credentials and the consumer identity that the rest of the epic builds on.

Follow the pattern from [`bindings/go/github/spec/access/v1/github.go`](https://github.com/open-component-model/open-component-model/blob/main/bindings/go/github/spec/access/v1/github.go) and [`bindings/go/wget/spec/access/v1/wget.go`](https://github.com/open-component-model/open-component-model/blob/main/bindings/go/wget/spec/access/v1/wget.go).

The spec must be backwards-compatible with the [OCM v1 `git` access type](https://github.com/open-component-model/ocm/blob/main/api/ocm/extensions/accessmethods/git/method.go) (type names `git`, `git/v1alpha1`, `Git`, `Git/v1alpha1`; fields `repository`, `ref`, `commit`).

```go
// +k8s:deepcopy-gen:interfaces=ocm.software/open-component-model/bindings/go/runtime.Typed
// +k8s:deepcopy-gen=true
// +ocm:typegen=true
// +ocm:jsonschema-gen=true
type Git struct {
	// +ocm:jsonschema-gen:enum=Git/v1
	// +ocm:jsonschema-gen:enum:deprecated=git,git/v1alpha1,Git/v1alpha1

	// Repository is the Git repository URL (https, ssh, or scp-style, e.g.
	// git@github.com:open-component-model/ocm.git).
	Repository string `json:"repository"`

	// Ref is the git reference used to resolve the commit (e.g. refs/heads/main).
	Ref string `json:"ref,omitempty"`

	// Commit is the 40-hex-character SHA of the commit.
	Commit string `json:"commit,omitempty"`
}

const (
	Version    = "v1"
	Type       = "Git"
	LegacyType = "git"
)
```

`ref` and `commit` follow the `gitHub` model: commit is authoritative, ref is informational and pinned to a commit by the digest processor. Documents carrying both (as OCM v1 allowed) must still decode.

Folder structure:

```
bindings/go/git/
└── spec/
    ├── access/
    │   ├── scheme.go
    │   └── v1/
    │       ├── git.go
    │       ├── group_version.go
    │       ├── zz_generated.deepcopy.go
    │       ├── zz_generated.ocm_type.go
    │       ├── zz_generated.ocm_jsonschema.go
    │       └── schemas/Git.schema.json
    ├── credentials/
    │   ├── scheme.go
    │   └── v1/git_credentials.go, convert.go, group_version.go, schemas/
    └── identity/
        └── v1/type.go, register.go, schemas/
```

Credentials and identity:

- `GitCredentials/v1` supporting username/password, token, and SSH private key — follow [`bindings/go/github/spec/credentials/v1/github_credentials.go`](https://github.com/open-component-model/open-component-model/blob/main/bindings/go/github/spec/credentials/v1/github_credentials.go) (`MustRegisterCredentialType` plus a `ConvertTo…` helper).
- `Git` consumer identity type in `spec/identity/v1/`, hostpath-based (scheme / hostname / port / pathPrefix), compatible with OCM v1's `Git` consumer type. Model: [`bindings/go/s3/spec/identity/v1/register.go`](https://github.com/open-component-model/open-component-model/blob/main/bindings/go/s3/spec/identity/v1/register.go) and [`bindings/go/wget/spec/identity/v1/type.go`](https://github.com/open-component-model/open-component-model/blob/main/bindings/go/wget/spec/identity/v1/type.go) (`IdentityFromURL`). The `gitHub` binding builds its identity ad hoc in `internal/` — do not copy that.
- Identity derivation must be stable across `https://`, `ssh://` and scp-style (`git@github.com:org/repo.git`) URLs pointing at the same repository.

**Done Criteria**

- [ ] `Git` struct with deepcopy / typegen / jsonschema-gen annotations and `Validate()`
- [ ] `group_version.go` with `Version = "v1"`, `Type = "Git"`, `LegacyType = "git"`
- [ ] Scheme registration including the legacy aliases `git`, `git/v1alpha1`, `Git/v1alpha1`
- [ ] `GitCredentials/v1` (username/password, token, SSH private key) registered as credential type
- [ ] `Git` consumer identity type with URL → identity derivation for https / ssh / scp-style URLs
- [ ] Code generation run (deepcopy, ocm type, JSON schema)
- [ ] Unit tests for spec validation, decoding of real OCM v1 `git` access documents, and identity derivation per URL form

matthiasbruns:
On-behalf-of: SAP <matthias.bruns@sap.com>

Adds git access resource repo and typing.

Contributes: https://github.com/open-component-model/ocm-project/issues/1257


matthiasbruns:
@coderabbitai - compare with s3 and wget and do a review

bindings/go/git/internal/endpoint/endpoint.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+
+	ep.Protocol = strings.ToLower(ep.Protocol)
+	switch ep.Protocol {
+	case protocolHTTP, protocolHTTPS, protocolSSH, protocolGit:
+		if ep.Host == "" || ep.Path == "" || ep.Path == "/" {
+			return nil, fmt.Errorf("git repository URL requires a hostname and path")
+		}
+	case protocolFile:
+		if ep.Host != "" && ep.Host != "localhost" {
```
frewilhelm, line 36: What is the reason that you check for `ep.Host != ""` first? Can't you check directly for `localhost`?

bindings/go/git/internal/endpoint/endpoint.go
```diff
@@ -0,0 +1,75 @@
...
+		}
+
+		if ep.Path == "" {
+			return nil, fmt.Errorf("file repository path must not be empty")
+		}
+	default:
+		return nil, fmt.Errorf("unsupported git transport %q", ep.Protocol)
+	}
+
+	if ep.Port < 0 || ep.Port > 65535 {
```
frewilhelm, line 47: Just wondering - when its a `protocolFile`, then `ep.Port` might not be set, right?

bindings/go/git/internal/endpoint/endpoint.go
frewilhelm, line 1: I am missing unit tests and documentation for `Parse()`. For example, I am unsure what the behaviour and expectation for `file` would be

bindings/go/git/repository/resource_repository.go
```diff
@@ -0,0 +1,175 @@
...
+		return nil, err
+	}
+	defer func() { err = errors.Join(err, b.Close()) }()
+	result := res.DeepCopy()
+	spec, err := accessFrom(result.Access)
+	if err != nil {
+		return nil, err
+	}
+
+	spec.Commit = commit
```
frewilhelm, line 119: This might overwrite `spec.Commit` without checking if it was set and if `spec.Commit` equals `commit`. Is this intended?

bindings/go/git/repository/resource_repository.go
```diff
@@ -0,0 +1,175 @@
...
+	pinned := &runtime.Raw{}
+	if err := access.Scheme.Convert(spec, pinned); err != nil {
+		return nil, fmt.Errorf("cannot encode pinned git access: %w", err)
+	}
+
+	result.Access = pinned
+	raw, _ := b.Digest()
+	result.Digest = &descriptor.Digest{
```
frewilhelm, line 127: I believe we need to check if the passed `res *descriptor.Resource.Digest` is set and equals `raw` before overwriting it.

bindings/go/git/internal/download/archive.go
```diff
@@ -0,0 +1,145 @@
...
+	defer func() {
+		if err != nil {
+			_ = os.Remove(file.Name())
+		}
+	}()
+
+	digester := digest.SHA256.Digester()
+	writer := &limitedWriter{Writer: io.MultiWriter(file, digester.Hash()), limit: opts.MaxDownloadSize}
+	// The archive stays uncompressed: its digest is pinned into the descriptor and
+	// verified on other machines, and compress/flate output is not stable across Go releases.
```
Skarlso: I do not understand this comment in this context. 🤔

bindings/go/git/internal/download/auth.go
```diff
@@ -0,0 +1,57 @@
...
+		username := creds.Username
+		if username == "" {
+			username = ep.User
+		}
+
+		if username == "" {
+			username = "git"
+		}
+
+		auth, err := gitssh.NewPublicKeysFromFile(username, creds.PrivateKey, creds.Password)
```
frewilhelm, line 33: Are you sure the `PrivateKey` must be a file path? From ocm v1s doc I can only see ""Private Key authentication certificate".

bindings/go/git/internal/download/download.go
```diff
@@ -0,0 +1,186 @@
...
+		repo, err = git.PlainCloneContext(ctx, dir, false, &git.CloneOptions{
+			URL:        access.Repository,
+			Auth:       auth,
+			NoCheckout: true,
+			Tags:       git.AllTags,
+			CABundle:   opts.CABundle,
+		})
+	} else {
+		// A pinned commit must not depend on a valid remote HEAD.
+		repo, err = git.PlainInit(dir, false)
```
frewilhelm, line 66: Why don't we handle a possible error from `git.PlainInit(..)`?

frewilhelm: just saw this is checked on L102 - this is confusing :D any chance we can make that simpler?

frewilhelm: Hm, but we checking for `ErrRepositoryAlreadyExists` which is returned if the path is not empty.

bindings/go/git/internal/download/blob.go
```diff
@@ -0,0 +1,104 @@
...
+	_ blob.DigestAware           = (*Blob)(nil)
+	_ blob.MediaTypeAware        = (*Blob)(nil)
+	_ blob.MediaTypeOverrideable = (*Blob)(nil)
+	_ io.Closer                  = (*Blob)(nil)
+)
+
+// removeTempFile deletes the file at path. A file that is already gone is not an
+// error, which makes repeated calls (Close plus the cleanup) idempotent without
+// tracking any removal state.
+func removeTempFile(path string) error {
```
frewilhelm, line 39: Why is this function just calling another function (+ error checking)? Do it needs its own function?

frewilhelm (commented):
preliminary review

bindings/go/git/internal/endpoint/endpoint.go
matthiasbruns: `file:///path` and plain paths have no host, so empty host must be allowed
only a set host that isn't localhost gets rejected

bindings/go/git/internal/download/blob.go
matthiasbruns: same as in wget
used in two places

bindings/go/git/internal/download/download.go
matthiasbruns: we do all of this now in `fetchCommit` :)

bindings/go/git/internal/download/auth.go
matthiasbruns: that's what ocvm1 does. but I will add inline support and host key fallback

bindings/go/git/internal/download/download.go
```diff
@@ -0,0 +1,197 @@
...
+	if err := ctx.Err(); err != nil {
+		return nil, "", err
+	}
+
+	worktree, err := repo.Worktree()
+	if err != nil {
+		return nil, "", fmt.Errorf("cannot open git worktree: %w", err)
+	}
+
+	if err := worktree.Checkout(&git.CheckoutOptions{Hash: selected.Hash}); err != nil {
```
piotrjanik, line 113: Have you considered writing the tar straight from commit.Tree()? Then no local checkout would be needed, and the host file system couldn't change the tar: file modes and file names containing some special characters.

matthiasbruns:
Fully tested including @piotrjanik idea 
here is our git https://github.com/users/matthiasbruns/packages/container/package/ocm-git-e2e%2Fcomponent-descriptors%2Focm.software%2Fgit-e2e pushed via ocm

bindings/go/cli/internal/plugin/builtin/git/register.go
```diff
@@ -0,0 +1,37 @@
...
+	"ocm.software/open-component-model/bindings/go/plugin/manager/registries/digestprocessor"
+	"ocm.software/open-component-model/bindings/go/plugin/manager/registries/resource"
+)
+
+// Register wires the Git resource repository, digest processor and credential
+// scheme into the CLI plugin registries.
+func Register(resourcePluginRegistry *resource.ResourceRegistry,
+	digestProcessorRegistry *digestprocessor.RepositoryRegistry,
+	credentialRepository *credentialrepository.RepositoryRegistry,
+	filesystemConfig *filesystemv1alpha1.Config,
```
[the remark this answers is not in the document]
matthiasbruns: not supported by httpconfig right now

bindings/go/git/spec/credentials/v1/convert.go
```diff
@@ -0,0 +1,66 @@
...
+		Username:      properties[credentialKeyUsername],
+		Password:      properties[credentialKeyPassword],
+		Token:         properties[credentialKeyToken],
+		PrivateKey:    properties[credentialKeyPrivateKey],
+		PrivateKeyPEM: properties[credentialKeyPrivateKeyPEM],
+	}
+}
+
+func ConvertToGitCredentials(creds runtime.Typed) (*GitCredentials, error) {
+	if creds == nil {
```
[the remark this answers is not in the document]
matthiasbruns: nope - we are not doing nil-guards in any convert - the caller must take care of this

bindings/go/git/internal/download/auth.go
```diff
@@ -0,0 +1,72 @@
...
+		return auth, nil
+	case creds.Token != "":
+		if ep.Protocol != "https" && ep.Protocol != "http" {
+			return nil, fmt.Errorf("tokens require an HTTP or HTTPS repository")
+		}
+
+		return &githttp.TokenAuth{Token: creds.Token}, nil
+	case creds.Username != "":
+		if ep.Protocol != "https" && ep.Protocol != "http" {
+			return nil, fmt.Errorf("username/password authentication requires an HTTP or HTTPS repository")
```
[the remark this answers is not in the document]
matthiasbruns: should be fixed already

bindings/go/git/internal/download/download.go
```diff
@@ -0,0 +1,187 @@
...
+		}
+	}()
+
+	var repo *git.Repository
+	if access.Commit == "" {
+		repo, err = git.PlainCloneContext(ctx, dir, true, &git.CloneOptions{
+			URL:      access.Repository,
+			Auth:     auth,
+			Tags:     git.AllTags,
+			CABundle: opts.CABundle,
```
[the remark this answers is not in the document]
matthiasbruns: wont do, already bounded and the same pattern as the other access types

bindings/go/git/repository/resource_repository.go
matthiasbruns: https://github.com/open-component-model/open-component-model/pull/3577/changes#diff-58c408db54e08ec44b197145c9ab38b8afa52aa46e928c27a65c6c20d8981260R93

bindings/go/git/internal/download/archive.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/opencontainers/go-digest"
+)
+
+// archive writes the files of the commit tree as tar, so names, modes and symlinks
+// come from git and not from a checkout on the host file system. The archive stays
+// uncompressed: its digest is pinned into the descriptor and verified on other
+// machines, and compress/flate output is not stable across Go releases.
+// Directories are implied by the file paths, submodules are not part of the tree content.
+func archive(ctx context.Context, commit *object.Commit, opts Options) (_ *Blob, err error) {
+	tree, err := commit.Tree()
```
piotrjanik, line 30: nice +1

bindings/go/git/internal/download/archive.go
```diff
@@ -0,0 +1,100 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("cannot read git tree: %w", err)
+	}
+
+	file, err := os.CreateTemp(opts.TempDir, "ocm-git-archive-*.tar")
+	if err != nil {
+		return nil, err
```
piotrjanik: maybe wrap with some context?

bindings/go/git/repository/resource_repository.go
```diff
@@ -0,0 +1,181 @@
...
+	}
+
+	spec, err := accessFrom(res.Access)
+	if err != nil {
+		return nil, "", err
+	}
+
+	var typed *credsv1.GitCredentials
+	if creds != nil {
+		if typed, err = credsv1.ConvertToGitCredentials(creds); err != nil {
```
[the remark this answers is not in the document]
matthiasbruns: nothing to do here

bindings/go/git/spec/access/v1/schemas/Git.schema.json
```diff
@@ -0,0 +1,57 @@
...
+          "deprecated": true,
+          "const": "Git/v1alpha1"
+        }
+      ]
+    }
+  },
+  "required": [
+    "type",
+    "repository"
+  ],
```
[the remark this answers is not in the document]
matthiasbruns: we cannot have enums right now

bindings/go/git/internal/download/auth.go
```diff
@@ -0,0 +1,77 @@
...
+
+func authMethod(ep *transport.Endpoint, creds *credsv1.GitCredentials, opts Options) (transport.AuthMethod, error) {
+	if creds == nil {
+		creds = &credsv1.GitCredentials{}
+	}
+
+	// go-git turns URL userinfo into basic auth, which plain HTTP would send in clear text.
+	if ep.Protocol == "http" && (ep.User != "" || ep.Password != "") {
+		return nil, fmt.Errorf("credentials require an HTTPS repository")
+	}
```
frewilhelm, line 21: First I was a bit confused because I thought that check is covered by the `switch` below. But that switch is checking `creds` while this on is checking the endpoint. So, what you are checking here, is if the user specified userinfo credentials in the passed URL that gets parsed to `ep`, right? If so, I would recommend to adjust the error message to make it a bit clearer

bindings/go/git/internal/download/download.go
```diff
@@ -0,0 +1,187 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("cannot create git repository: %w", err)
+	}
+
+	if _, err := repo.CreateRemote(&config.RemoteConfig{Name: "origin", URLs: []string{access.Repository}}); err != nil {
+		return repo, fmt.Errorf("cannot configure git remote: %w", err)
+	}
+
+	// Asking for the pinned commit alone avoids transferring every ref and its
+	// history. Servers without the capability reject it before any transfer.
```
frewilhelm: > Servers without the capability reject it before any transfer.

This sounds worrying. Are there any known servers? Do we have a fallback?

bindings/go/git/internal/download/download.go
```diff
@@ -0,0 +1,187 @@
...
+		CABundle: opts.CABundle,
+		Tags:     git.NoTags,
+		RefSpecs: []config.RefSpec{config.RefSpec("+" + access.Commit + ":refs/ocm/commit")},
+	})
+	if errors.Is(err, git.ErrExactSHA1NotSupported) {
+		err = repo.FetchContext(ctx, &git.FetchOptions{
+			Auth:     auth,
+			CABundle: opts.CABundle,
+			Tags:     git.AllTags,
+			RefSpecs: []config.RefSpec{"+refs/*:refs/*"},
```
frewilhelm, line 170: Is this the fallback?

bindings/go/git/internal/download/download.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+
+	if err != nil && !errors.Is(err, git.NoErrAlreadyUpToDate) {
+		return repo, transportError(ctx, "cannot fetch git repository", err)
+	}
+
+	return repo, nil
+}
+
+func peelCommit(repo *git.Repository, hash plumbing.Hash) (*object.Commit, error) {
```
frewilhelm, line 183: a small little comment what this is and what its needed for would be nice :D

bindings/go/cli/internal/plugin/builtin/builtin.go
```diff
@@ -79,6 +80,13 @@ func Register(manager *manager.PluginManager, filesystemConfig *filesystemv1alph
 		return fmt.Errorf("could not register s3 inbuilt plugin: %w", err)
 	}
 
+	if err := git.Register(manager.ResourcePluginRegistry,
+		manager.DigestProcessorRegistry,
+		manager.CredentialRepositoryRegistry,
+		filesystemConfig); err != nil {
```
frewilhelm: Why don't we add the `httpConfig`?

bindings/go/git/integration/repository_fixture_test.go
frewilhelm, line 1: Do I have a deja vu? Is there any better way for the fixture_test setups :D

bindings/go/git/internal/download/download.go
```diff
@@ -0,0 +1,187 @@
...
+	var result *Blob
+	defer func() {
+		err = errors.Join(err, os.RemoveAll(dir))
+		if err != nil && result != nil {
+			err = errors.Join(err, result.Close())
+		}
+	}()
+
+	var repo *git.Repository
+	if access.Commit == "" {
```
piotrjanik, line 78: hmm, in this case we would do something similar to git clone and few lines below we would resolve branch name into a commit. How about resolving ref into a commit first and then fetching this commit only?

matthiasbruns: `ls-remote` costs an extra round trip and doesn't remove the second code path

bindings/go/cli/internal/plugin/builtin/git/register.go
```diff
@@ -0,0 +1,37 @@
...
+) error {
+	var opts []gitrepository.Option
+	if filesystemConfig.TempFolder != nil {
+		opts = append(opts, gitrepository.WithTempDir(*filesystemConfig.TempFolder))
+	}
```
fabianburth: our coding patterns usually suggest a different option handling, I believe (dedicated options struct)

bindings/go/git/internal/download/blob.go
```diff
@@ -0,0 +1,104 @@
...
+// temporary download artifact owned by the blob: [Blob.Close] removes it, and a
+// cleanup attached to the blob removes it once the blob becomes unreachable.
+// A caller that drops the blob without closing it therefore does not keep the
+// file around for the lifetime of the process, which matches the reclamation an
+// in-memory blob got for free.
+type Blob struct {
+	*filesystem.Blob
+	path   string
+	digest string
+}
```
fabianburth: don't we have an existing blob implementation that we can re-use?

bindings/go/cli/internal/plugin/builtin/git/register.go
matthiasbruns: yeah we started both, going back to the struct

bindings/go/git/integration/repository_fixture_test.go
matthiasbruns: dropped them, back to slightly duplicated test-data

bindings/go/git/internal/download/blob.go
matthiasbruns: we really need to get rid of the wget blob implementation and maybe move it into our filesystem binding - that's the auto-cleaning one

removing custom blob.go for now

bindings/go/git/internal/download/download.go
matthiasbruns: https://github.com/open-component-model/open-component-model/pull/3577/changes#diff-21f0f64051ad14df8a81784a543bcee6ab698d1d2181e68dd214d65465b9301fR164

bindings/go/git/internal/download/download.go
matthiasbruns: There are `allow-tip-sha1-in-want` / `allow-reachable-sha1-in-want`
github and gitlab support both, but others might not

`reject it before any transfer` happens on the client, nothing will be lost, so all good :)

bindings/go/git/internal/download/download.go
matthiasbruns: yep :) 
that's the `ErrExactSHA1NotSupported` branch

bindings/go/git/internal/download/auth.go
matthiasbruns: updated

bindings/go/git/repository/resource_repository.go
```diff
@@ -0,0 +1,250 @@
...
+			return nil, "", err
+		}
+	}
+
+	b, commit, err := download.Download(ctx, spec, typed, r.downloadOptions(tempDir))
+	if err != nil {
+		return nil, "", err
+	}
+
+	raw, _ := b.Digest()
```
Skarlso: This is happening twice. 🤔  Once here and then on the same thing in ProcessResourceDigest. That will result reading that thing two times into memory.

bindings/go/git/repository/resource_repository.go
```diff
@@ -0,0 +1,250 @@
...
+		spec.Commit = commit
+	}
+	pinned := &runtime.Raw{}
+	if err := access.Scheme.Convert(spec, pinned); err != nil {
+		return nil, fmt.Errorf("cannot encode pinned git access: %w", err)
+	}
+
+	result.Access = pinned
+	// r.download already rejected a set digest that does not match this archive.
+	raw, _ := b.Digest()
```
Skarlso: Are you sure it's okay to ignore the `ok`? It will result in an empty `raw` that then gets passed along.

bindings/go/git/repository/options.go
```diff
@@ -0,0 +1,69 @@
...
+
+import (
+	"golang.org/x/crypto/ssh"
+
+	httpv1alpha1 "ocm.software/open-component-model/bindings/go/http/spec/config/v1alpha1"
+)
+
+// Options holds configuration for the Git resource repository.
+type Options struct {
+	// MaxDownloadSize caps the archive bytes a single download may produce. Nil
```
Skarlso: Shouldn't this then be called MaxArchiveSize instead?

bindings/go/git/internal/download/download.go
```diff
@@ -0,0 +1,211 @@
...
+	case errors.Is(err, git.ErrRepositoryNotExists):
+		return fmt.Errorf("%s: repository not found", operation)
+	case errors.Is(err, transport.ErrAuthenticationRequired):
+		return fmt.Errorf("%s: authentication required", operation)
+	case errors.Is(err, transport.ErrAuthorizationFailed):
+		return fmt.Errorf("%s: authorization failed", operation)
+	default:
+		return fmt.Errorf("%s: transport failed; check repository access and server trust", operation)
+	}
+}
```
Skarlso, line 247: Shouldn't you also include the original error's text?

bindings/go/git/internal/download/download.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+
+	// The archive file is removed only when this call fails; on success it belongs
+	// to the caller.
+	var archivePath string
+	defer func() {
+		err = errors.Join(err, os.RemoveAll(dir))
+		if err != nil && archivePath != "" {
+			err = errors.Join(err, removeIgnoringMissing(archivePath))
+		}
```
Skarlso: This will not mean what you think it does and will result in an otherwise clean error being overwritten.

You should check err nil separately and THEN decide if you want to join. Or alternatively, log the RemoveAll error instead of returning it.

```go
defer func() {
      oErrFailed := err != nil
      err = errors.Join(err, os.RemoveAll(dir))
      if oErrFailed && archivePath != "" {
              err = errors.Join(err, removeIgnoringMissing(archivePath))
      }
}()
```

Skarlso (commented):
Initial rundown :)

bindings/go/git/repository/options.go
matthiasbruns: why? in `ResourceRepository` it's capping the download

bindings/go/git/repository/resource_repository.go
matthiasbruns: added the checks

bindings/go/git/repository/options.go
matthiasbruns: yeah maybe, wanted to keep this in sync with out other bindings, but I'll rename it

bindings/go/git/internal/download/archive.go
```diff
@@ -0,0 +1,105 @@
...
+// The digest is taken while writing, so no caller has to read the archive back.
+func archive(ctx context.Context, commit *object.Commit, file *os.File, opts Options) (_ *filesystem.Blob, _ digest.Digest, err error) {
+	tree, err := commit.Tree()
+	if err != nil {
+		return nil, "", fmt.Errorf("cannot read git tree: %w", err)
+	}
+
+	digester := digest.Canonical.Digester()
+	limited := &limitedWriter{Writer: file, limit: opts.MaxArchiveSize}
+	tw := tar.NewWriter(io.MultiWriter(limited, digester.Hash()))
```
piotrjanik, line 37: I think v1 compresses tar into tgz and then calcualtes digest, so existing CVs might not work Please check https://github.com/open-component-model/ocm/blob/bab58e1c9ef8ad0642291842efda385ef9dd1da4/api/utils/blobaccess/git/access.go#L93

matthiasbruns: yeah I saw that gap - I swear it was a discussion I made but cant remember why and now I am questioning myself... - give me a bit time

matthiasbruns: now I remember why and it's actually a v1 bug

v1 includes uname and group into the tar content, thus digest will never be stable with v2 in that case

```
===== v1 (application/x-tgz, sha256:14b94490…) =====
.gitignore      type=0 mode=0644 uid=501 gid=0 uname="mbruns" gname="wheel" mtime=0
CHANGELOG       type=0 mode=0644 uid=501 gid=0 uname="mbruns" gname="wheel" mtime=0
binary.jpg      type=0 mode=0644 uid=501 gid=0 uname="mbruns" gname="wheel" mtime=0
go              type=5 mode=0755 uid=501 gid=0 uname="mbruns" gname="wheel" mtime=0
go/example.go   type=0 mode=0644 uid=501 gid=0 uname="mbruns" gname="wheel" mtime=0
...

===== v2, this PR (application/x-tar, sha256:05eaa92c…) =====
.gitignore      type=0 mode=0644 uid=0   gid=0 uname=""       gname=""      mtime=0
CHANGELOG       type=0 mode=0644 uid=0   gid=0 uname=""       gname=""      mtime=0
binary.jpg      type=0 mode=0644 uid=0   gid=0 uname=""       gname=""      mtime=0
go/example.go   type=0 mode=0644 uid=0   gid=0 uname=""       gname=""      mtime=0
...
```


ran a few tests again to remember why - so since we cannot get byte-identical with v1 anyway, we can also ditch compression.

the only thing that we should decide on is - drop support for the ocmv1 git spec?

matthiasbruns: So what I will most likely do
lift the git version here to `v1` and drop support for `alpha1` git types (from ocmv1)

jakobmoellerdev: woah stop, this is off. for dir input type we have a reproducible flag, and also in v2 youre not adding directory nodes right?

matthiasbruns: The problem is the following

ocmv1 does not nullify `Uid/Gid/Uname/Gname` (ownership) in the tar, so every file has meta infos that ocmv2 resets.
I have fiddles with our current `filesystem` implementation, to at least add `symlink` support, but adding back "who packed that tar" into v2 feels utterly wrong to me.

```text
$ /usr/bin/env bash /Users/mbruns/dev/sap/open-component-model/bindings/go/git/hack/ocmv1-compat.sh
packing 6ecf0ef2c2df with OCM v0.50.0 and with this binding

files
  identical
archives
  v1 mbruns/staff  ccfc31db742b3676c439fe5b93584cc85029e8fc9e1ac05b7299ba429a208cc1
  v2 0/0  05eaa92c15e5cb2bcb18439ff6d8a9a27e373fb78086adfa5270eea261712db9
  differ, as expected: v1 writes the packing user into every tar header
```

I have written a script that tries to pack git with `ocmv1` and consume it with `ocmv2`.

One thing is important here: we start to verify digests on Download in git. This issue would have come up only after we fully rolled our verification-on-download.

So imho we only have four real options here

- reimplement tar'ing in git to match ocmv1 - do not add this to filsystem
- bake ownership into the tar optionally in v2 in our filesystem package
- be able to disable verification on download
- drop ocmv1 support

matthiasbruns: also this means that digest only are verifiable on the same host 🤔

matthiasbruns: another one

```
OCM v0.50.0 signature verification
  signed under umask 022, verified under umask 077: FAILS
    recomputed SHA-256:17e0049136f51e5a8135e3802c35af70ee05411ec32420d87cf6bac8f6f3f55a
    recorded   SHA-256:ccfc31db742b3676c439fe5b93584cc85029e8fc9e1ac05b7299ba429a208cc1
  a v1 git digest is reproducible only by the process that wrote it
```
  
 the tar implementation in ocmv1 is broken - it only does not pop up because there is no verification on the download side as well

matthiasbruns: new findings
ocmv1 calculated the digest on the compressed tar, not the tar itself

matthiasbruns: so the failing path:

**ocmv1 access path**

- ocmv1 calculates the digest based on the compressed tar. the tar itself contains uname etc which are bound to the host it ran on
- ocmv2 tries to download the git commit and verifies the digest
  - it will tar the folders, the tar does not hold uname etc, and even if it would, this would instantly fail when another user tries to verify
  - compress might be instable between go versions, so this might become problematic itself



| v1 behaviour                                                               | Can v2 copy it?                                     |
|----------------------------------------------------------------------------|-----------------------------------------------------|
| tar of the checked-out files, no `.git`, v1's entry order                  | yes                                                 |
| mtime 0                                                                    | yes                                                 |
| owner name/group of the packing user (e.g. `mbruns staff`) in every header | only as "current user"; the original isn't recorded |
| modes from the packing process's umask                                     | only as "current umask"                             |
| Go default gzip                                                            | yes, for now                                        |

matthiasbruns: So my result of all of this

- we won't be able to support git access from ocmv1 - once we introduce download resource verification, each digest check will fail
- input should be fine, since we just calculate the digest based on the blob we get

jakobmoellerdev: to me this sounds like you need to fix v1 as there is quite some unintended behavior in there X)

jakobmoellerdev: > it will tar the folders, the tar does not hold uname etc, and even if it would, this would instantly fail when another user tries to verify

dont understand how this is not a v1 problem

> compress might be instable between go versions, so this might become problematic itself

this is normal even for v1

matthiasbruns: sure I can fix v1 - but old generated CVs wont work anyway

matthiasbruns: I created the issue https://github.com/open-component-model/ocm-project/issues/1337
I will finish git access here with that in mind.

matthiasbruns: @jakobmoellerdev  the whole issue is a v1 problem :) I just wanted to try to make it work in v2 without touching v1 at all - so the goal is to fix v1 - this implementation here is as good as it can get for now

matthiasbruns:
validated end to end - I tried to make the diff as small as possible but had to add a few changes to filesystem to enable that our tar building can work the same as in ocmv1 (content-wise, not file-flags).

```bash
#!/bin/sh
# Build the CLI and verify one ref-only Git access against a tiny public repository.
set -eu

cli_dir=$(CDPATH= cd -- "$(dirname -- "$0")/.." && pwd)
work=$(mktemp -d)
trap 'rm -rf "$work"' EXIT

repository=https://github.com/octocat/Hello-World.git
ref=refs/heads/master
component=ocm.software/git-e2e:1.0.0

echo "== Building CLI"
(cd "$cli_dir" && go build -o "$work/ocm" .)
echo "== Resolving $ref in $repository"
git ls-remote --exit-code "$repository" "$ref" > "$work/ref"
commit=$(awk '{print $1}' "$work/ref")

cat > "$work/constructor.yaml" <<EOF
components:
- name: ocm.software/git-e2e
  version: 1.0.0
  provider:
    name: ocm.software
  resources:
  - name: hello-world
    version: 1.0.0
    type: blob
    relation: external
    access:
      type: Git/v1
      repository: $repository
      ref: $ref
EOF

echo "== Creating component version with Git access"
"$work/ocm" add component-version --repository "$work/ctf" --constructor "$work/constructor.yaml"
echo "== Checking pinned commit and digest metadata"
"$work/ocm" get component-version "$work/ctf//$component" -o json > "$work/descriptor.json"
jq -e '(if type == "array" then .[0] else . end) | .component.resources |
  select(length == 1) | .[0]' "$work/descriptor.json" > "$work/resource.json"
jq -e --arg repository "$repository" --arg ref "$ref" --arg commit "$commit" '
  .access.type == "Git/v1" and .access.repository == $repository and
  .access.ref == $ref and .access.commit == $commit and
  .digest.hashAlgorithm == "SHA-256" and
  .digest.normalisationAlgorithm == "genericBlobDigest/v1"
' "$work/resource.json" > /dev/null

echo "== Downloading archive at $commit"
"$work/ocm" download resource "$work/ctf//$component" --identity name=hello-world \
  --output "$work/archive.tar.gz" --extraction-policy disable
echo "== Verifying compressed archive digest and README"
checksum=$(shasum -a 256 "$work/archive.tar.gz")
if ! jq -e --arg digest "${checksum%% *}" '.digest.value == $digest' "$work/resource.json" > /dev/null; then
  echo "FAIL: downloaded archive digest does not match the descriptor" >&2
  exit 1
fi
tar -tzf "$work/archive.tar.gz" > "$work/entries"
grep -qx README "$work/entries"

echo "PASS: Git/v1 resolved $ref to $commit; archive digest and README verified"


```

bindings/go/cli/integration/add_component_version_integration_test.go
```diff
@@ -1287,6 +1292,137 @@ components:
...
+	content, err := io.ReadAll(tr)
+	r.NoError(err)
+	r.Equal("hello from git access\n", string(content))
+	r.Equal(int64(len(content)), header.Size)
+	_, err = tr.Next()
+	r.ErrorIs(err, io.EOF)
+
+}
+
+func createGitRepository(t *testing.T) (dir, commit string) {
```
piotrjanik: nit: move it to internal/git.go. same with other helper func.

bindings/go/git/internal/download/http.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"fmt"
+	"net/http"
+
+	gitclient "github.com/go-git/go-git/v5/plumbing/transport/client"
+	githttp "github.com/go-git/go-git/v5/plumbing/transport/http"
+)
+
+func init() {
+	InstallHTTPClient(nil)
```
piotrjanik: Repository already calls InstallHTTPClient. Do we need that? It would be called twice

PR #3583, opened 2026-09-08
fix(release): fix pr generator workflow pointing to the correct tag

frewilhelm:
https://github.com/open-component-model/open-component-model/pull/3582 moved the job to create a PR for the release documentation. However, it looked for a tag `bindings/go/v${VERSION}` which does not exist for `0.15.0` because it gets introduced starting with `0.16.0`. Fortunately, we can just use the tag `v${VERSION}` which does basically the same for the workflow and is even more backwards compatible.

### DIFF
--- a/.github/workflows/website-v2-release-docs.yml
+++ b/.github/workflows/website-v2-release-docs.yml
@@ -41,13 +41,13 @@ jobs:
           ref: main
           persist-credentials: false
 
-      - name: Verify bindings/go tag exists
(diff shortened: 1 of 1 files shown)


jneisener (dismissed):

frewilhelm:
Tested via PR trigger
image

jneisener (approved):

matthiasbruns (approved):

PR #3585, opened 2026-09-08
chore(docs): manually release documentation for 0.15.0

frewilhelm:
Release ocm v2 `0.15.0` [failed](https://github.com/open-component-model/open-component-model/actions/runs/32859801161/job/101986803880) in its last step "Open website update PR". The reason for this is that it checks out the `main` branch and wants to install go based on `cli/go.mod`. However, that file does not exist anymore since the `cli` module was moved to `bindings/go`.

We first tried to fix this by moving that doc-job from the release workflow to its own workflow with an input-parameter (https://github.com/open-component-model/open-component-model/pull/3582, https://github.com/open-component-model/open-component-model/pull/3583; especially the fact that we could not easily re-run the docs generation was the reason for this).

However, this did not solve another problem where we want to import `bindings/go` with the passed version. But for the release `0.15.0` we did not started with the lock-step release for `bindings/go`. Thus, the used version for `bindings/go` was `0.0.8` (we decided to only push patch-versions until the lock-step release is ready).

Instead of fixing the generator script, we decided to fix it manually. This is ugly, and if we need to patch `0.15.0`, we will need to do it again. But implementing such a edge-case into the workflow and/or script feels a bit too much.

### DIFF
--- a/website/config/_default/hugo.yaml
+++ b/website/config/_default/hugo.yaml
@@ -20,7 +20,7 @@ defaultContentLanguageInSubdir: false
 editPage: false
 lastMod: false
 defaultContentVersionInSubdir: false
-defaultContentVersion: '0.14'
+defaultContentVersion: '0.15'
 pagination:
   pagerSize: 10
 params:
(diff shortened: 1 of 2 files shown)


jneisener (approved):
Thanks for taking care of this, hopefully with 0.16 this part is smooth 🤞

matthiasbruns (approved):

PR #3588, opened 2026-09-08
fix(oci): name the requested artifact in digest-pinned resource layouts

### LINKED ISSUE #3587: transfer fails on resources pointing at OCI referrers (etcd-druid)
**Context**

To test Platform Mesh in air-gapped environments we need a CTF which is then transferred into our internal registry via the OCM CLI.

**Version**

```
$ ocm version
{"major":"0","minor":"13","patch":"0","gitVersion":"0.13.0", ...}
```

Reproduced on 0.13.0, 0.14.0, 0.15.0 and current `main`. (v1) 0.41.0 transfers the same component successfully.

**To Reproduce**

macOS 26.5.2 (arm64), zsh. No `.ocmconfig` involved cause the component is public, no credentials.

```bash
ocm transfer component-version \
  "europe-docker.pkg.dev/gardener-project/releases//github.com/gardener/etcd-druid:v0.36.4" \
  "ctf::./out" --copy-resources
```

The same failure occurs on the real workload, where etcd-druid is pulled in as a component reference:

```bash
ocm transfer component-version --recursive --copy-resources \
  "ghcr.io/platform-mesh//github.com/platform-mesh/platform-mesh:0.5.1-build.82" \
  "ctf::./out"
```

**Actual behavior**

The transfer aborts:

```
failed adding local resource "etcd-druid-spdx-ref" ... failed to pack resource blob:
failed to copy OCI layout: multiple manifests found in oci store, but no manifest
could be identified as the top level parent
```

**Expected behavior**

The resource is copied by value into the CTF, like every other resource and like (v1) 0.41.0 does. 

**Additional Comments**

*Workaround:* Generate the spec with `--dry-run -o yaml`, drop the offending transformations plus every reference to them, replay via `--transfer-spec`. Costs the SBOMs. 
In use in `platform-mesh/helm-charts` but it only strips etcd-druid's nodes and we'd like to get rid of it :) 

I'd like to give it a try an create an PR myself, if I got something wrong... don't hesitate to leave comments^^

Perseus985:

`ocm transfer component-version --copy-resources` aborts on any resource whose access points at an OCI referrer (e.g. an SBOM attestation) and is pinned by digest:

```
failed adding local resource "etcd-druid-spdx-ref" ... failed to pack resource blob:
failed to copy OCI layout: multiple manifests found in oci store, but no manifest
could be identified as the top level parent
```

This blocks air-gapped Platform Mesh transfers, since `etcd-druid` is pulled in as a component reference.

Fixes #3587 

##### How to test the changes

No `.ocmconfig` needed. The component is public.

```bash
ocm transfer component-version \
  "europe-docker.pkg.dev/gardener-project/releases//github.com/gardener/etcd-druid:v0.36.4" \
  "ctf::./out" --copy-resources
  
# the SBOM resources must land on exactly the digest their access asked for
ocm get component-version "ctf::./out//github.com/gardener/etcd-druid:v0.36.4" -o json \
  | jq -r '.[0].component.resources[] | select(.name=="etcd-druid-spdx-ref")
           | "\(.extraIdentity.architecture)  \(.access.localReference)"'

# expected:
#   arm64  sha256:05f5eccfd31e6efde77898a0fbbbf1b0f57f8b1f4e0435a10f9926dcc07b1fd0
#   amd64  sha256:9c2bb7e314c3b42d013e2ec38a6688f37d30aceb7bab23d0ef41f926d46743fd
```

##### Verification


Perseus985:
**Update**

After the kind review from @Skarlso I dug a bit deeper into the code trying to fix: "Recording the root during download and have the packiging read it back correctly" <- Thats what I tried to solve in the update :)

_What did I do now?_

`copyToOCILayoutInMemoryAsync` now marks `base` with a new annotation `software.ocm.layout.root`. Because `base`should be the requested resource if I understood it correctly. 

Then `pickTopLevelDescriptor` also checks for the new annotation instead of just checking for one manifest OR ref.name. So our case doesnt error out anymore. 

`TopLevelArtifacts` was edited so that he can see the marker aswell. 

`repostory.go` is now untouched because of https://github.com/open-component-model/open-component-model/pull/2495#discussion_r3266119692 as you mentioned :) 

#### Testing
The commands from before didnt change.

#### New Tests
- `TestRepository_DownloadResourceStream_DigestPinnedWithReferrer` for the reported case 
- `TestRepository_DownloadResourceStream_ReferrerIsTheRequestedArtifact`where the access points to a referrer instead of its subject 
- `TestRepository_DownloadResourceStream_MultiArchByDigest`for a digest pinned multi-arch index with no referrer involved at all... which you mentioned in the review and it seems that the new marker helps with that too. 
- `TestCloseableReadOnlyStore_MainArtifacts` has a new "A marked referrer is the main artifact" case. 

#### Addition
Thanks again for the digging which helped me to create a more reasonable fix and the opportunity! That was actualy my first time, taking a look into the OCI Image-spec 🙈 
If there is anything to change pls let me know :) 

Next up.. CodeRabbit findings^^

bindings/go/oci/tar/blob_io.go
```diff
@@ -70,9 +71,17 @@ func copyToOCILayoutInMemoryAsync(ctx context.Context, src content.ReadOnlyGraph
...
+	// base is the only thing that knows which manifest the caller wanted.
+	// Tag is what carries a descriptor's annotations into index.json.
+	root := base
+	root.Annotations = maps.Clone(base.Annotations)
+	if root.Annotations == nil {
+		root.Annotations = make(map[string]string, 1)
+	}
+	root.Annotations[AnnotationLayoutRoot] = "true"
+
+	for _, ref := range append([]string{base.Digest.String()}, opts.Tags...) {
```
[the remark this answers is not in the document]
Perseus985: `markedRoot` now counts distinct digests instead of entries. `updateIndex`
lists a descriptor once per reference name it carries, so a root that is both
addressed by digest and tagged arrives as several entries with the same
digest.

Regression test added as `TestCloseableReadOnlyStore_MainArtifacts/a root
tagged several times is still one root`, covering the digest plus two tags.
It fails before the change and passes after. 

And I've run Coderabbit locally and there were no more findings :)

Skarlso:
@Perseus985 Hello! :) I assume you did a manual test run as well and see if this is now working as intended? :) ( I will do that same, just asking now. :) )

Perseus985:
> @Perseus985 Hello! :) I assume you did a manual test run as well and see if this is now working as intended? :) ( I will do that same, just asking now. :) )

Yes :) I ran it manually a few times. With etcd-druid into a CTF and I've spun up a local zot registry to also make an OCI to OCI transfer. And because it was all part of platform-mesh I also tried it with that. (0.5.1-build.82)

Skarlso:
Nice. This looks now a lot better. Sorry, I didn't have the time to fully go through this today, but it's on my Monday morning todo list. :D

bindings/go/oci/tar/blob_io.go
```diff
@@ -70,9 +71,17 @@ func copyToOCILayoutInMemoryAsync(ctx context.Context, src content.ReadOnlyGraph
...
-	for _, tag := range opts.Tags {
-		if err = errors.Join(err, target.Tag(ctx, base, tag)); err != nil {
+	// base is the only thing that knows which manifest the caller wanted.
+	// Tag is what carries a descriptor's annotations into index.json.
+	root := base
+	root.Annotations = maps.Clone(base.Annotations)
+	if root.Annotations == nil {
+		root.Annotations = make(map[string]string, 1)
+	}
+	root.Annotations[AnnotationLayoutRoot] = "true"
```
jakobmoellerdev, line 81: If you transfer from a tar into any other repo, this annotation will persist right? im a bit concerned about this, the annotation change. also this would change the digest of the descriptor from what i understand. for digest pinned resources, the ref.Name would be the digest so im wondering if thats the right way to resolve it?

Im considering this lookup logic:

1. if there is only one candidate, use it
2. if there is ref.name set, use that
3. error

For adding into the CTF, ref.name could be set to the digest if there is no tag provided.

WDYT?

Perseus985: Hey, sorry for the delay^^

I've imported the CTF into a local Harbor and had a look at the annotations and `software.ocm.layout.root` doesnt travel with it, wheter images, charts or the descriptor. It only lives in the index.json of the OCI layout tar..  So I also didnt noticed any change in the digest of the descriptor or the resources that are being requested. 

But your lookup logic works too. And its the one I've tried at first(https://github.com/open-component-model/open-component-model/pull/3588/changes/69f09956c6607edf1ce16d24e41a0da8e4bab22a) passing the whole ImageReference, so `registry/repo@sha256:…` ends up in ref.name.

The feedback was https://github.com/open-component-model/open-component-model/issues/3587#issuecomment-5599249998

jakobmoellerdev (dismissed):

bindings/go/oci/tar/top_level_artifacts.go
```diff
@@ -9,11 +9,17 @@ import (
 	"oras.land/oras-go/v2/content"
 )
 
-// TopLevelArtifacts returns the main top-level artifacts from a list of
-// candidates. A candidate is excluded when it is a referrer (declares a
-// subject) or when another candidate contains it as a successor. The remaining
-// candidates are returned in input order.
+// AnnotationLayoutRoot names the artifact a layout was built for, on that
+// artifact's index.json entry, with the value "true".
+const AnnotationLayoutRoot = "software.ocm.layout.root"
```
Skarlso: Nit, but kinda important though :D

All our annotations live under `bindings/go/oci/spec/annotations`. Could you please put this  one there as well? :) Thanks! 🙇‍♀️

bindings/go/oci/tar/blob_io.go
```diff
@@ -150,15 +159,19 @@ func CopyOCILayoutWithIndex(ctx context.Context, dst content.Storage, src blob.R
...
-// layout's index.json. With one manifest in the index it returns that
-// manifest; with many it returns the one tagged via
-// `org.opencontainers.image.ref.name`. Returns an error if neither rule
-// uniquely identifies a top-level descriptor.
+// pickTopLevelDescriptor works out which manifest in the layout was the one
+// requested, trying three things in order: if there is only one, it wins; else
+// the entry marked with [AnnotationLayoutRoot]; else the entry named by
+// `org.opencontainers.image.ref.name`, which is only ever set for tag-based
+// references. Returns an error when none of the three settles it, because a
+// wrong guess here silently packs the wrong artifact.
```
Skarlso: Okay, here's the thing. :D I think this could be improved further.

With your fix right now, we correctly list the correct root yes. And you prove this for Multi-arch builds as well with your new test `MultiArchByDigest`. BUT. That goes through the `DownloadResourceStream` part in which the layer ends up being marked. This does NOT prove that currently existing UNMARKED images will work. And in fact, after writing a little test, they do not work.

But, this is easy to fix. :) We have the other function you already edited a tiny bit. `TopLevelArtifacts`.

Call that from here as a fourth fall-back option! :)

So add something like this:

```go
    // Note: You'll have to modify pick's parameter to include the context.
	if descs := TopLevelArtifacts(ctx, ociStore, ociStore.Index.Manifests); len(descs) == 1 {
		return descs[0], nil
	}

	return ociImageSpecV1.Descriptor{}, fmt.Errorf(... // this probably needs to be updated now ( the error message )
```

And then here is the test you can put into blob_io_test.go that should go green:

```go
func TestCopyOCILayoutWithIndex_UnmarkedMultiArch(t *testing.T) {
	r := require.New(t)

	var buf bytes.Buffer
	w, err := NewOCILayoutWriterWithTempFile(&buf, t.TempDir())
	r.NoError(err)
	idx := packIndex(t, w, pack(t, w, "amd64", "", nil), pack(t, w, "arm64", "", nil))
	r.NoError(w.Close())
	layoutBytes := buf.Bytes()

	store, err := ReadOCILayout(t.Context(), &testReadOnlyBlob{data: layoutBytes})
	r.NoError(err)
	r.Len(store.Index.Manifests, 3, "index and both children must be listed, otherwise this test proves nothing")
	for _, manifest := range store.Index.Manifests {
		r.Empty(manifest.Annotations[AnnotationLayoutRoot], "the layout must be unmarked")
		r.Empty(manifest.Annotations[ociImageSpecV1.AnnotationRefName], "the layout must be untagged")
	}
	r.NoError(store.Close())

	top, err := CopyOCILayoutWithIndex(t.Context(), memory.New(), &testReadOnlyBlob{data: layoutBytes}, CopyOCILayoutWithIndexOptions{})
	r.NoError(err)
	r.Equal(idx.Digest, top.Digest, "the children are contained by the index, so the index is the only top level")
}
```

WDYT?

Skarlso (commented):
Two additions. :) Otherwise, great work! :) 👏

bindings/go/oci/tar/top_level_artifacts.go
Perseus985: Done 👨🏼‍💻

bindings/go/oci/tar/blob_io.go
Perseus985: Done ^^ Also included your test :) Hope this is good now 🚀

Skarlso (approved):

jneisener (approved):

jakobmoellerdev (approved):

Skarlso:
Thank you so much @Perseus985 for working on this! Nice! :) :) 🎉

PR #3591, opened 2026-09-09
fix(oci): allow digest-only uploads in the materialized upload path

fabianburth:
uploadOCIImage (used by UploadResource/UploadSource) hard-errored when the target ImageReference carried no tag, while UploadResourceStream tolerated digest-only references, pinned the pushed digest, and tagged only when one was present. This made the two upload paths behave inconsistently for the same access (e.g. ctf->oci transfers with --upload-as ociArtifact failed where oci->oci streaming succeeded).

Drop the tag gate and mirror the streaming behavior: tag only when the reference carries one, pin the pushed digest into the access when it carries neither tag nor digest, and log a warning when the upload goes out untagged since retention then depends on the target registry's garbage collection policy.

This behavior was discovered during the investigation of https://github.com/open-component-model/open-component-model/pull/3588.

##### Verification

### DIFF
--- a/bindings/go/oci/repository.go
+++ b/bindings/go/oci/repository.go
@@ -620,10 +620,10 @@ func (repo *Repository) getStore(ctx context.Context, component string, version 
 
 // UploadResource uploads a [*descriptor.Resource] to the repository.
 // The target access image reference may carry a tag, a digest, or neither:
-// a tag is applied on upload, otherwise the access is pinned to the pushed
-// digest. Uploads without a tag are logged as a warning because retention
-// of the artifact then depends on the target registry's garbage collection
-// policy.
+// a tag is applied on upload, a digest-only reference is preserved as-is,
+// and a reference carrying neither is pinned to the pushed digest. Uploads
+// without a tag are logged as a warning because retention of the artifact
+// then depends on the target registry's garbage collection policy.
 func (repo *Repository) UploadResource(ctx context.Context, res *descriptor.Resource, b blob.ReadOnlyBlob) (newRes *descriptor.Resource, err error) {
 	ctx = slogcontext.NewCtx(ctx, repo.logger)
 	done := log.Operation(ctx, "upload resource", log.IdentityLogAttr("resource", res.ToIdentity()))
--- a/bindings/go/oci/repository.go
+++ b/bindings/go/oci/repository.go
@@ -1153,6 +1153,9 @@ func (repo *Repository) UploadResourceStream(ctx context.Context, res *descripto
 		if err := store.Tag(ctx, rs.Root(), ref.Tag); err != nil {
 			return nil, fmt.Errorf("failed to tag artifact with tag %q: %w", ref.Tag, err)
 		}
+	} else {
+		slogcontext.Warn(ctx, "uploading OCI artifact without a tag, retention depends on the target registry's garbage collection policy",
+			"imageReference", access.ImageReference)
 	}
 
 	res = res.DeepCopy()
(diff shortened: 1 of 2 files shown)


fabianburth:
Add a warning in the constructor already that image references with a tag are dangerous because they might be garbage collected on re-upload to another repository because we don't tag it (as we don't know the tag)

bindings/go/oci/repository.go
```diff
@@ -296,6 +296,11 @@ func (repo *Repository) processOCIImageDigest(ctx context.Context, res *descript
 	if err != nil {
 		return nil, fmt.Errorf("error parsing image reference %q: %w", typed.ImageReference, err)
 	}
+	if resolved.Tag == "" {
+		slogcontext.Warn(ctx, "resource access references an image without a tag; if it is transferred and uploaded as oci image, it will be untagged, retention depends on the target registry's garbage collection policy",
+			"imageReference", typed.ImageReference,
+			log.IdentityLogAttr("resource", res.ToIdentity()))
+	}
```
fabianburth, line 303: this will cause a warning during construction

bindings/go/oci/repository.go
```diff
@@ -619,6 +624,11 @@ func (repo *Repository) getStore(ctx context.Context, component string, version
 }
 
 // UploadResource uploads a [*descriptor.Resource] to the repository.
+// The target access image reference may carry a tag, a digest, or neither:
+// a tag is applied on upload, a digest-only reference is preserved as-is,
+// and a reference carrying neither is pinned to the pushed digest. Uploads
```
Skarlso, line 629: >and a reference carrying neither is pinned to the pushed digest

What digest? Didn't it just say there is no digest?

bindings/go/oci/repository.go
frewilhelm: We don't call the digest processor if `--skip-reference-digest-processing` is set on construction. Do you think this might be a problem?

bindings/go/oci/repository.go
frewilhelm: I believe we calculate a digest while constructing. Hence, the digest is not present in the constructor but in the resulting descriptor.

fabianburth: also if it is not in an existing descriptor (e.g. because you skipped digest generation), it will be pinned here (I think the identical behaviour already existed in `UploadResourceStream`)

Skarlso: Can we write that instead in this comment please? 🙇‍♀️

Skarlso (approved):

bindings/go/oci/repository.go
```diff
@@ -692,9 +704,6 @@ func (repo *Repository) uploadOCIImage(ctx context.Context, newAccess runtime.Ty
 	if err != nil {
 		return ociImageSpecV1.Descriptor{}, nil, fmt.Errorf("failed to parse target access image reference %q: %w", access.ImageReference, err)
 	}
-	if err := ref.ValidateReferenceAsTag(); err != nil {
```
piotrjanik, line 695: nit: instead of removing this validation, we could alternatively run `ref.ValidateReferenceAsDigest()`, since it's either tag or digest

fabianburth: I'm not sure that is true. I believe we can currently upload a reference without either of those. Thus, we have 
```go
	// if we don't have a pinned access we can pin it now.
	if ref.Reference.Reference == "" {
		ref.Reference.Reference = main.Digest.String()
		access.ImageReference = ref.String()
	}
```
in line 725

jakobmoellerdev (approved):

bindings/go/oci/repository.go
fabianburth: already covered by `looseref.ParseReference`

PR #3592, opened 2026-09-09
feat(go): use go toolchain directive

### LINKED ISSUE #3295: Please do not needlessly "bump" Go via go.mod
**Description**
In my opinion, PRs like https://github.com/open-component-model/open-component-model/pull/3014 are misguided. The `go ` directive in a go.mod file declares the _minimum_ supported Go version, not the "recommended" version. By constantly bumping it, OCM is forcing all downstream consumers to also bump Go constantly, even when it may be inconvenient at the time. This is causing unnecessary churn for no real gain.

There is the `toolchain` directive that can be used to instruct what Go version to use when compiling a module directly, but the `go` directive should IMHO practically never be touched by a human, but only be updated by `go mod tidy`.

Thank you.

Skarlso:

- [ ] TODO: I'll fill this out after some amount of testing.

Fixes https://github.com/open-component-model/open-component-model/issues/3295

##### Verification

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -73,6 +73,25 @@ task init/go.work
 task tidy
 ```
 
+## Go Versions
+
+`bindings/go/go.mod` has two directives:
+
+| Directive   | Value      | Meaning                                                                        |
+|-------------|------------|--------------------------------------------------------------------------------|
+| `go`        | `1.26.0`   | The **minimum** Go a consumer needs to build against this module.              |
+| `toolchain` | `go1.26.5` | The Go version **open-component-model** uses for builds, testing and scanning. |
+
+Renovate bumps `toolchain` on every Go release, along with the `golang` container images, etc. Since `setup-go` already
+references the go.mod files, it should automatically get the right version. It never bumps the `go` directive in the go
+mod files. That is only increased if we require or would like to use some new go version ability.
+
+This directive CANNOT be below the HIGHEST directive compared to all dependencies:
+
+```bash
+cd bindings/go && go list -m -f '{{.GoVersion}} {{.Path}}' all | sort -V | tail -5
+```
+
 ## Linting
 
 A single `golangci.yml` at the repository root configures linting for all Go modules. The `task tools:lint` command
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -77,10 +77,10 @@ task tidy
 
 `bindings/go/go.mod` has two directives:
 
-| Directive   | Value      | Meaning                                                                        |
-|-------------|------------|--------------------------------------------------------------------------------|
-| `go`        | `1.26.0`   | The **minimum** Go a consumer needs to build against this module.              |
-| `toolchain` | `go1.26.5` | The Go version **open-component-model** uses for builds, testing and scanning. |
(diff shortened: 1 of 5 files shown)


CONTRIBUTING.md
```diff
@@ -73,6 +73,25 @@ task init/go.work
...
+## Go Versions
+
+`bindings/go/go.mod` has two directives:
+
+| Directive   | Value      | Meaning                                                                        |
+|-------------|------------|--------------------------------------------------------------------------------|
+| `go`        | `1.26.0`   | The **minimum** Go a consumer needs to build against this module.              |
+| `toolchain` | `go1.26.5` | The Go version **open-component-model** uses for builds, testing and scanning. |
+
+Renovate bumps `toolchain` on every Go release, along with the `golang` container images and `setup-go` steps. It never
```
frewilhelm: > `setup-go` steps

I believe (hope) all of our `setup-go` steps use the `go.mod`

CONTRIBUTING.md
```diff
@@ -73,6 +73,25 @@ task init/go.work
...
 ```
 
+## Go Versions
+
+`bindings/go/go.mod` has two directives:
+
+| Directive   | Value      | Meaning                                                                        |
+|-------------|------------|--------------------------------------------------------------------------------|
+| `go`        | `1.26.0`   | The **minimum** Go a consumer needs to build against this module.              |
+| `toolchain` | `go1.26.5` | The Go version **open-component-model** uses for builds, testing and scanning. |
```
frewilhelm: Do we need to update those go versions manually? A renovate directive might look ugly.

Skarlso: You mean the toolchain version? That is updated by renovate. Go version is only ever allowed to be updated manually IF needed. That's the whole plan here. :)

CONTRIBUTING.md
Skarlso: My intention here was saying the setup-go steps are bumped because they already reference the go.mod files. But I can re-formulate this to make it clearer. :)

CONTRIBUTING.md
frewilhelm: No, I mean the `CONTRIBUTING.md` document :D If we specify the version here, we might want to update it too

frewilhelm: But I don't want to update it manually and I don't think configuring renovate to update it in this document will look nice. That's why I hope we can maybe just omit the versions and tell people to look at the `go.mod` directly

Skarlso: Ah, got it. This could be something like x.y.z I guess.

website/go.mod
```diff
@@ -1,3 +1,5 @@
 module ocm.software/open-component-model/website
 
-go 1.26.5
+// Leave this lower so we don't force website devs to download the latest version
+// and they might be okay with what their environment has locally already.
+go 1.24.0
```
jakobmoellerdev: imho: do not deviate or go lower now that you already bumped it

Skarlso: Okay.

Skarlso: Done 👍

frewilhelm (approved):

matthiasbruns (approved):
lgtm

PR #3595, opened 2026-09-09
docs(adr): add ADR 0028 for packaging and shipping VM images and configuration

jakobmoellerdev:
## What

Adds **ADR 0025 — Packaging and Shipping Virtual Machine Images and Configuration** (`docs/adr/0025_vm_image_packaging.md`), status `proposed`.

Opened as a **draft** for TSC discussion.

## Normative baseline (decisions 1–8)

- VM images are ordinary OCM `resources` — **no component-descriptor schema change**.
- Resource types `virtualMachineImage` / `virtualMachineConfig` (free-form strings, matching `ociImage`/`helmChart` convention).
- Namespaced media-type registry `application/vnd.ocm.software.vm.disk.{raw,qcow2,vmdk,vhd,iso}.v1` + `.vm.ova.v1+tar`, patterned on the existing `application/vnd.ocm.software.oci.layout.v1+tar`.
- Three packaging flows with runnable `component-constructor.yaml`: embedded `file` input (air-gap default), external `wget` access, external `ociArtifact` access.
- Configuration shipped as a sibling resource (`utf8`/`dir`/`file` input), bound to the image via **signing-relevant** labels `software.ocm.vm/format` and `software.ocm.vm/config-for`.
- Large-blob transport via the streaming path from ADR-0022 (`ResourceStream` / `oras.CopyGraph`); `Materialize()` (tar) is the anti-pattern.
- Integrity/signing reuses ADR-0008/0023; consumption via `ocm download resource` + label-driven format/config resolution (ADR-0007). Booting/importing is out of OCM scope.

## Open discussion (decision 9 — deferred to EPIC)

Large-blob splitting for 40 GiB+ images (registry layer caps): three candidate wire representations (split-manifest indirection, digest-list in `localReference`, OCI Image Index of chunks) and two triggers (explicit `input.split.maxLayerSize` vs. implicit ocmconfig/registry auto-detect), presented with tradeoffs and **no choice made** — plus the invariants any solution must preserve.

Feeds the ADR deliverable of EPIC open-component-model/ocm-project#1290 (predecessor open-component-model/ocm-project#12).

## Notes

- No code changes — documentation-level conventions reusing existing binding machinery.
- References ADR-0007, ADR-0008, ADR-0013, ADR-0022, ADR-0023.


docs/adr/0028_vm_image_packaging.md
```diff
@@ -0,0 +1,348 @@
...
+    input:
+      type: dir
+      path: ./config
+      mediaType: application/vnd.ocm.software.vm.config.v1+tar
+```
+
+**Association (image ↔ config binding).** OCM has no typed resource→resource link; the only schema-native relation is `srcRefs` (source→resource). Resource→resource association is therefore expressed with labels, and this ADR mandates two **signing-relevant** labels (`Label` struct in `bindings/go/constructor/spec/v1/constructor.go` — fields `name`, `value`, `signing`, `version`):
+
+* `software.ocm.vm/format` on the image resource — value is the disk-format enum: `raw | qcow2 | vmdk | vhd | iso | ova`.
+* `software.ocm.vm/config-for` on the config resource — value is the bound image resource identity `{ name, version }`.
```
Skarlso, line 209: Why not `artifactReference`? Like an SBOM the config would reference resource that it belongs to. And that's already in the spec and there are functions for discovery implemented that will find the resource mentioned with given identity selectors.

docs/adr/0025_vm_image_packaging.md
```diff
@@ -0,0 +1,348 @@
...
+    DL -->|"config-for label resolves config"| PLAT
+```
+
+### 9. Large-Blob Splitting (Open Discussion — deferred to EPIC #1290)
+
+> **Non-normative.** Decisions 1–8 are the normative baseline. This subsection records the problem and candidate solutions for images that exceed a registry's push/layer size limit; **it makes no choice.** The representation and trigger will be settled by EPIC [#1290](https://github.com/open-component-model/ocm-project/issues/1290).
+
+**Problem.** OCI registries impose per-push / per-layer size limits (e.g. Google's ~10 GiB layer cap). A VM disk larger than that cannot be pushed as a single blob layer. OCM v1 transparently split large blobs into multiple chunk layers on upload and reassembled them on download; OCM v2 lacks this. This blocks the 40 GiB+ artefact use case that motivates [#1290](https://github.com/open-component-model/ocm-project/issues/1290) and the predecessor discussion in [#12](https://github.com/open-component-model/ocm-project/issues/12).
+
+**Candidate representations (no selection made).**
```
Skarlso: Just one remark, I think we should avoid using OCI specific mentions here. For example, option 1's chunking and indexing and ordering with digests, can work with any technology as long as we get packing right.

We might even store and index together with the chunks that tells us something like `{digest, chunkindex, size}` and then we can use this information later for assembly and verification of the chunk's integrity.

Right now, if I get this right, this information will live in the manifest / descriptor / layer information, right?

Otherwise, option 1 looks good. :)

Skarlso: Ah I see the config is already there in the issue
```yaml
type: blobLimits.ocireg.ocm.config.ocm.software
blobLimits:
  ghcr.io: 10000000000
```

So that point is mute.

Skarlso: This should be solved for OCI. So this ADR should/is focusing on OCI layer size limitations instead of trying to generally solve chunking that OCM would be doing. That is kinda a different beast all together.

S3's multipart ends up with one item so it doesn't apply as a technology where we would transfer split entities.

Therefore the argument that we would want to transfer from OCI to (somethingelse) S3 and we would want to have the same endformat, doesn't really hold true. Because the chunk limit comes from OCI really. S3 or other technologies might not have this limitation so no need for chunking.

docs/adr/0028_vm_image_packaging.md
```diff
@@ -0,0 +1,348 @@
...
+   * *Pro:* keeps `LocalReference` single-valued (matches today's `v2.LocalBlob.LocalReference string` in `bindings/go/descriptor/v2/local_access.go`); minimal reader impact; whole-blob digest lives in the descriptor. This is #12's "blob descriptor with a dedicated media type" variant.
+   * *Con:* one extra fetch to read the split descriptor; introduces a new media type.
+2. **Digest list in `localReference`.** Overload `LocalReference` to carry an ordered list of chunk-layer digests directly (the variant mandelsoft favored in #12).
+   * *Pro:* no extra descriptor fetch.
+   * *Con:* reinterprets the `LocalReference string` field semantics; touches every `localBlob` reader; least backward-compatible.
+3. **OCI Image Index of chunks.** Reuse the [ADR-0013](0013_oci_format_compatibility.md) index path: `localReference` points at an OCI Image Index (`application/vnd.oci.image.index.v1+json`) whose child manifests/layers are the chunks.
+   * *Pro:* maximally OCI-native; reuses ADR-0013 machinery.
+   * *Con:* overloads "index" for pure byte-chunking; risks confusing OCI-native consumers that interpret an index as a multi-arch image; complicates non-OCI/CTF handling.
```
Skarlso, line 273: Since our images / descriptors already really look like multi-arch images, we feel like this is a mute `con`.

Skarlso:
- [ ] TODO: Validate the assumption about size limit.

docs/adr/0028_vm_image_packaging.md
```diff
@@ -0,0 +1,348 @@
...
+   * *Pro:* keeps `LocalReference` single-valued (matches today's `v2.LocalBlob.LocalReference string` in `bindings/go/descriptor/v2/local_access.go`); minimal reader impact; whole-blob digest lives in the descriptor. This is #12's "blob descriptor with a dedicated media type" variant.
+   * *Con:* one extra fetch to read the split descriptor; introduces a new media type.
+2. **Digest list in `localReference`.** Overload `LocalReference` to carry an ordered list of chunk-layer digests directly (the variant mandelsoft favored in #12).
+   * *Pro:* no extra descriptor fetch.
+   * *Con:* reinterprets the `LocalReference string` field semantics; touches every `localBlob` reader; least backward-compatible.
+3. **OCI Image Index of chunks.** Reuse the [ADR-0013](0013_oci_format_compatibility.md) index path: `localReference` points at an OCI Image Index (`application/vnd.oci.image.index.v1+json`) whose child manifests/layers are the chunks.
```
matthiasbruns, line 271: tl;dr option 3

i tried to figure out if we can provide a solution that does not require ocm to "merge" the blobs back together. Unfortunately not - this would have been my first rule (no ocm lock in after using our chunking).

So I would probably go for option 3. We are talking only about oci here, we discussed the concept for other access types, and maybe they will have their own logic and implementation.

matthiasbruns: but as an alternative - maybe use a storage that supports huge blobs instead of pushing everything into oci?

Skarlso: With the question of _should we even allow this_ pending a discussion, we are selecting number 3 as a possible solution for this ADR's purpose.

_Should we even allow this_: We need to discuss if we even think that this is a good thing to allow users to do. This is pending.

Skarlso:
Func fact:

```json
{
  "schemaVersion": 2,
  "mediaType": "application/vnd.oci.image.manifest.v1+json",
  "config": {
    "mediaType": "application/vnd.oci.image.config.v1+json",
    "digest": "sha256:2f07c38b83d3cfc31d8a13bea098bb5b2132992cde6778a45c90a24131eb4aa4",
    "size": 432
  },
  "layers": [
    {
      "mediaType": "application/vnd.oci.image.layer.v1.tar+gzip",
      "digest": "sha256:b5c6aa97e7335526359723caa6db0dd1e8b60c3f100299c92ee60003965b61e6",
      "size": 16111042825
    }
  ]
}
```

I tested an upload to GHCR of a 15GB image (single layer) and it worked. :D

matthiasbruns:
> Func fact:
> 
> ```json
> {
>   "schemaVersion": 2,
>   "mediaType": "application/vnd.oci.image.manifest.v1+json",
>   "config": {
>     "mediaType": "application/vnd.oci.image.config.v1+json",
>     "digest": "sha256:2f07c38b83d3cfc31d8a13bea098bb5b2132992cde6778a45c90a24131eb4aa4",
>     "size": 432
>   },
>   "layers": [
>     {
>       "mediaType": "application/vnd.oci.image.layer.v1.tar+gzip",
>       "digest": "sha256:b5c6aa97e7335526359723caa6db0dd1e8b60c3f100299c92ee60003965b61e6",
>       "size": 16111042825
>     }
>   ]
> }
> ```
> 
> I tested an upload to GHCR of a 15GB image (single layer) and it worked. :D

image

jneisener:
Outside the scope of the 1 vs 3 discussion is the question of how we chunk. `maxLayerSize` should be used as an upper bound for content defined chunking (like borg etc do it) so that a minimal change to a VM image does not lead to a complete re-upload.

docs/adr/0030_vm_image_packaging.md
```diff
@@ -0,0 +1,357 @@
...
+A VM image is a component-version `resource`, not a new first-class descriptor kind. There is **no change to the component descriptor structure**. This keeps every existing OCM tool — transfer, signing, download, transport — working unchanged.
+
+### 2. Resource type strings
+
+`blob` remains the generic fallback. This ADR defines two domain-specific artifact `type` strings:
+
+| Resource `type`        | Meaning                                                                           |
+|------------------------|-----------------------------------------------------------------------------------|
+| `virtualMachineImage`  | A bootable VM disk image.                                                         |
+| `virtualMachineConfig` | Machine-readable VM deployment configuration (cloud-init, ignition, OVF, sizing). |
```
fabianburth: For access types, we standardized to Pascal Case. We should probably do the same here.

docs/adr/0030_vm_image_packaging.md
```diff
@@ -0,0 +1,357 @@
...
+The media type identifies the disk format. For an embedded image it is carried in `localBlob.mediaType`; for an external image it is carried in the access spec (`wget.mediaType`) or implied by the referenced OCI artifact. The normative mapping:
+
+| Disk format            | mediaType                                          |
+|------------------------|----------------------------------------------------|
+| `raw`                    | `application/vnd.ocm.software.vm.disk.raw.v1`      |
+| `qcow2`                  | `application/vnd.ocm.software.vm.disk.qcow2.v1`    |
+| `vmdk`                   | `application/vnd.ocm.software.vm.disk.vmdk.v1`     |
+| `vhd` / `vhdx`           | `application/vnd.ocm.software.vm.disk.vhd.v1`      |
+| `ISO` (installer / live) | `application/vnd.ocm.software.vm.disk.iso.v1`      |
+| `OVA` (tar of OVF+disks) | `application/vnd.ocm.software.vm.ova.v1+tar`       |
```
fabianburth: I think for some of those exist official media types

docs/adr/0030_vm_image_packaging.md
```diff
@@ -0,0 +1,357 @@
...
+          name: disk
+          version: 24.04.1
+        signing: true
+    input:
+      type: dir
+      path: ./config
+      mediaType: application/vnd.ocm.software.vm.config.v1+tar
+```
+
+**Association (image ↔ config binding).** OCM has no typed resource→resource link; the only schema-native relation is `srcRefs` (source→resource). Resource→resource association is therefore expressed with labels, and this ADR mandates two **signing-relevant** labels (`Label` struct in `bindings/go/constructor/spec/v1/constructor.go` — fields `name`, `value`, `signing`, `version`):
```
fabianburth: https://github.com/open-component-model/ocm-spec/blob/main/doc/01-model/06-conventions.md#artifact-linking-label

docs/adr/0030_vm_image_packaging.md
```diff
@@ -0,0 +1,357 @@
...
+
+* `software.ocm.vm/format` on the image resource — value is the disk-format enum: `raw | qcow2 | vmdk | vhd | iso | ova`.
+* `software.ocm.vm/config-for` on the config resource — value is the bound image resource identity `{ name, version }`.
+
+Both labels set `signing: true`, so the pairing and the declared format are covered by the component signature and cannot be tampered with independently of the signed component version.
+
+### 6. Transport of large images
+
+VM images are routinely 2–40 GiB. Transport MUST use the streaming path from [ADR-0022](0022_streaming_resource_content.md) (`bindings/go/oci/stream`, `ResourceStream`, `oras.CopyGraph`), which moves content blob-by-blob with existence checks (layer deduplication) and never fully materializes the image in memory or `/tmp`. Both OCI-registry and CTF backends benefit — CTF stores blobs as individual files.
```
fabianburth: I don't think streaming is currently implemented for CTF.

docs/adr/0030_vm_image_packaging.md
```diff
@@ -0,0 +1,357 @@
...
+   * *Pro:* keeps `LocalReference` single-valued (matches today's `v2.LocalBlob.LocalReference string` in `bindings/go/descriptor/v2/local_access.go`); minimal reader impact; whole-blob digest lives in the descriptor. This is #12's "blob descriptor with a dedicated media type" variant.
+   * *Con:* one extra fetch to read the split descriptor; introduces a new media type.
+2. **Digest list in `localReference`.** Overload `LocalReference` to carry an ordered list of chunk-layer digests directly (the variant mandelsoft favored in #12).
+   * *Pro:* no extra descriptor fetch.
+   * *Con:* reinterprets the `LocalReference string` field semantics; touches every `localBlob` reader; least backward-compatible.
+3. **OCI Image Index of chunks.** Reuse the [ADR-0013](0013_oci_format_compatibility.md) index path: `localReference` points at an OCI Image Index (`application/vnd.oci.image.index.v1+json`) whose child manifests/layers are the chunks.
```
fabianburth: Why would we need an index manifest? The whole thing can be respresented by a manifest with layers.

docs/adr/0030_vm_image_packaging.md
```diff
@@ -0,0 +1,357 @@
...
+
+Its con is mitigated rather than accepted: the index carries an OCM-specific artifact type / annotation so a consumer can tell a chunk index from a platform index, and the CTF path stores the chunk layers as individual files exactly as it already does for any other blob.
+
+**Open — should splitting be offered at all?** #1290 raises this explicitly. The alternative is to reject over-limit pushes with a clear error and require the publisher to use a backend without the limit. Settled in the EPIC.
+
+**Open — the split trigger.** #1290 requires sharding to be configurable, with `maxLayerSize` as the value. Where that configuration lives is undecided:
+
+* **(a) Explicit per-resource directive** on the `file` input, e.g. `input: { type: File/v1, split: { maxLayerSize: 1Gi } }` — would add a `Split` field to `bindings/go/input/file/spec/v1.File`. Deterministic and self-documenting.
```
fabianburth: Above, we decided that we go for an OCI specific solution as it is an oci limitation. This essentially bundles the `File/v1` with the assumption that we always store as OCI representation.
I don't think that is a good idea.

jakobmoellerdev: i disagree. we do the same with helm already and we only really store in OCI layouts on disk anyhow, even for S3 we are thinking of that no?

docs/adr/0030_vm_image_packaging.md
jakobmoellerdev: agreed on this

docs/adr/0030_vm_image_packaging.md
jakobmoellerdev: separate issue, but true

docs/adr/0030_vm_image_packaging.md
jakobmoellerdev: Done in 0ce47b76d — resource types are now `VirtualMachineImage`/`VirtualMachineConfig` and access types `Wget`/`OCIImage`, matching the PascalCase access-type convention. Decision 2 now states the rationale explicitly.

docs/adr/0030_vm_image_packaging.md
jakobmoellerdev: Addressed in 0ce47b76d. The media-type table now carries a "Known de-facto / recognized type" column citing the recognized type per format (`application/x-qemu-disk`, `application/x-vmdk`, `application/x-vhd`, `application/x-iso9660-image` from the freedesktop shared MIME database; DMTF OVF for OVA). The prose clarifies none is IANA standards-tree registered, so the `application/vnd.ocm.software.vm.*` namespace stays the canonical dispatch key while tooling MAY additionally record the de-facto type as a label.

docs/adr/0030_vm_image_packaging.md
jakobmoellerdev: Adopted in 0ce47b76d. The config→image binding now uses the spec's predefined `ocm.software/artifact-references` linking label (version `v1alpha1`) with its documented lookup algorithm and identity selectors, replacing the invented `software.ocm.vm/config-for` key. Linked the conventions doc in decision 5.

docs/adr/0030_vm_image_packaging.md
jakobmoellerdev: Fixed in 0ce47b76d — decision 6 now states CTF streaming is not yet implemented and may still buffer large images; the ADR mandates the streaming contract and flags closing the CTF gap as follow-up work.

docs/adr/0030_vm_image_packaging.md
jakobmoellerdev: Agreed and changed in 0ce47b76d. Decision 9 now selects representation 1 — a single OCI image manifest whose ordered layers are the chunks — and demotes the Image Index to a rejected option, noting an index is unnecessary since a plain manifest with layers already represents the whole blob.

docs/adr/0030_vm_image_packaging.md
```diff
@@ -0,0 +1,357 @@
...
+    type: virtualMachineImage
+    version: 24.04.1
+    labels:
+      - name: software.ocm.vm/format
+        value: qcow2
+        signing: true
+    access:
+      type: wget/v1
+      url: https://cloud-images.acme.org/noble/noble-server-cloudimg-amd64.qcow2
+      mediaType: application/vnd.ocm.software.vm.disk.qcow2.v1
```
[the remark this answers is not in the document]
jakobmoellerdev: Documented in 0ce47b76d. Flow B now states `Wget` URLs are mutable and that the recorded digest MUST be re-verified on every lazy download (the consumption path passes the expected digest to the downloader and rejects mismatches). Flow C documents that OCI digest processing resolves the tag and rewrites the access to the descriptor digest, so digest-qualified input is not required.

PR #3600, opened 2026-09-11
feat: discovery controller

### LINKED ISSUE #1279: Discovery API types and CRD
**Description**

Add the Discovery kind to `delivery.ocm.software/v1alpha1` as specified in ADR 0027. API surface only: types, generated CRD, deepcopy, RBAC, samples. No reconciler and no evaluation logic in this one..

New types in `bindings/go/kubernetes/controller/api/v1alpha1/discovery_types.go`:

- The API Object with the right Spec from the ADR (maybe slightly differently if warranted)
- Selector: matchIdentity map[string]string, matchLabels map[string]string, expression string. All three ANDed, empty selector matches everything.
- Extract: byResources map[string]string, byComponents map[string]string, expression string, mutually exclusive.
- UnresolvedReference: component, version, reason, message.
- Create the Status(this is crucial to work sync this with (https://github.com/openmcp-project/opencontrolplane-runtime/issues/43)) components []apiextensionsv1.JSON, extracted []map[string]apiextensionsv1.JSON, unresolved []UnresolvedReference, plus the shared observedGeneration / effectiveOCMConfig / conditions, implementing the same api/v1alpha1/interfaces.go contracts as Component and Resource.

New condition reasons in api/v1alpha1/condition_types.go: SelectorFailed, ExtractFailed, ResolutionFailed, PartiallyResolved, PayloadTooLarge, NoReferencesMatched, NoComponentsMatched.

Validation to encode as kubebuilder markers

The ADR restricts the short-circuit trigger to a matchIdentity of exactly {componentName, version} with no matchLabels / expression. This needs to be decided if we want a kuberbuilder marker or programatic check.

**Done Criteria**

- [ ] Types added
- [ ] config/samples/ example matching the ADR's YAML potentially
- [ ] Validation rules and tied in reconciler
- [ ] Reviewed by another team member
- [ ] Documentation

fabianburth:
Implements the discovery API described in https://github.com/open-component-model/ocm-project/issues/1153. 

I intentionally omitted the early exit path, if a concrete component is found. This introduces quite a bit of complexity as the resolution might not be deterministic anymore but dependent on which resolution branch is faster. 

A whole lot of them, ideally! 

##### Verification


Skarlso:
Hold on to your butts. A couple of fixes incoming here. :)

bindings/go/kubernetes/controller/internal/discovery/traverse.go
```diff
@@ -0,0 +1,136 @@
...
+	}.String()
+
+	rd := &resolverAndDiscoverer{resolver: resolver}
+	discoverer := syncdag.NewGraphDiscoverer(&syncdag.GraphDiscovererOptions[string, *descriptor.Descriptor]{
+		Roots:      []string{rootKey},
+		Resolver:   syncdag.ResolverFunc[string, *descriptor.Descriptor](rd.Resolve),
+		Discoverer: syncdag.DiscovererFunc[string, *descriptor.Descriptor](rd.Discover),
+	})
+
+	if err := discoverer.Discover(ctx); err != nil {
```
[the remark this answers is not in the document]
Skarlso: Yes, we know. Will be fixed separately in another PR I will open in that package.

bindings/go/kubernetes/controller/internal/controller/discovery/discovery_controller.go
```diff
@@ -0,0 +1,415 @@
...
+		logger.Info("discovery is suspended, skipping reconciliation")
+		return ctrl.Result{}, nil
+	}
+
+	old := discovery.DeepCopy()
+	result, rerr := r.reconcile(ctx, discovery)
+	if perr := r.publish(ctx, old, discovery); perr != nil {
+		// Status write failures are re-tried.
+		return ctrl.Result{}, errors.Join(rerr, perr)
+	}
```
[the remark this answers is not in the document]
Skarlso: >Every non-nil error returned by reconcile is wrapped with reconcile.TerminalError.

Where?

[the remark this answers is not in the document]
Skarlso: This is true, but. If we do need to surface the error that isn't just a log but in the actual status, if the publish failed, that point becomes mute because we couldn't update the object anyways. We can't separate the Terminal error from the publish error if its important to return and surface the error.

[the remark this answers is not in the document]
Skarlso: Yes. And do you see the problem? If we just log the error we will never return a Terminal error.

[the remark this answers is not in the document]
Skarlso: That's a fair point.

bindings/go/kubernetes/controller/api/v1alpha1/discovery_types.go
```diff
@@ -0,0 +1,237 @@
...
+	corev1 "k8s.io/api/core/v1"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const KindDiscovery = "Discovery"
+
+// Selector filters elements of the transitive component graph of a Discovery.
+// All specified clauses are ANDed. A nil or empty selector matches everything.
+type Selector struct {
```
frewilhelm, line 16: We might want to use that `Selector` for the `Resource` too. What do you think about putting the `Selector` into `common_types.go` instead?

bindings/go/kubernetes/controller/internal/controller/discovery/discovery_controller.go
```diff
@@ -0,0 +1,417 @@
...
+	// references stay observable without an extra round trip.
+	discovery.Status.EffectiveOCMConfig = configs
+
+	if upToDate(discovery, info) {
+		logger.V(1).Info("root component digest unchanged, skipping re-discovery",
+			"digest", discovery.Status.ObservedComponentDigest)
+
+		// Purely watch-driven: no periodic requeue.
+		return ctrl.Result{}, nil
+	}
```
frewilhelm, line 200: What if the component digest did not change but the selector on the discovery? If I get this correctly, we would skip another filtering

bindings/go/kubernetes/controller/internal/discovery/query.go
```diff
@@ -0,0 +1,238 @@
...
+// celInterruptCheckFrequency enables context cancellation checks during CEL
+// evaluation. cel-go only observes the evaluation context when the interrupt
+// check frequency is at least 1.
+const celInterruptCheckFrequency = 100
+
+// celCostLimit bounds the runtime cost of a single evaluation
+//
+// Roughly measured with unit tests. A 10_000 component graph would eat
+// approximately half of this budget.
+const celCostLimit = 1_000_000
```
frewilhelm, line 26: Should this be configurable by the user?

bindings/go/kubernetes/controller/internal/ocm/ocm.go
```diff
@@ -59,6 +59,10 @@ func GetEffectiveConfig(ctx context.Context, client ctrl.Client, obj v1alpha1.Co
 				resource = &v1alpha1.Component{}
 			case v1alpha1.KindResource:
 				resource = &v1alpha1.Resource{}
+			case v1alpha1.KindReplication:
+				resource = &v1alpha1.Replication{}
+			case v1alpha1.KindDiscovery:
+				resource = &v1alpha1.Discovery{}
```
frewilhelm, line 65: Hm, I am unsure if we want to allow this.

I think we can, because its the decision of the user how he wants to f up, but it feels weird that I can now reference a configuration from a `Discovery` in my `Resource` although these CRs are not connected. We don't block CR deletion for a resource where its configuration gets referenced. But it is the responsibility of the user I guess. Also in a sane setup such cases should not happen.

But for completeness, we should add the `Deployer` too :D

bindings/go/kubernetes/controller/README.md
```diff
@@ -53,6 +53,36 @@ Take a look at our [installation guide](https://ocm.software/docs/getting-starte
...
+### Discovery evaluation
+
+Selector filtering keeps runtime descriptors and never serializes them. v2
+conversion, marshalling, and generic-map decoding happen only during projection,
+and only for resources that survive selection: a resource removed by the
+resource selector is never serialized, so a bad access on a discarded resource
+cannot fail the discovery. Conversion or serialization failures on retained
+data are retryable and surface under `ExtractFailedReason` (not
+`SelectorFailedReason`), so the controller retries instead of stalling.
+
```
frewilhelm, line 85: Those sections feel a bit out-of-place. There is no mention of the discovery explaining what it is, it just starts with the resolver precedence.

docs/coding-patterns.md
```diff
@@ -908,7 +908,15 @@ For(&v1alpha1.Component{}, builder.WithPredicates(predicate.GenerationChangedPre
 
 ### Field Indexing and Cross-Resource Watches
 
-Field indexes are registered at controller setup for efficient cross-resource lookups. Watch handlers use `handler.EnqueueRequestsFromMapFunc` with `client.MatchingFields{}` to find related objects:
+Field indexes support efficient cross-resource lookups. Controller-private indexes are registered by the owning controller during setup. Indexes shared by multiple controllers are registered exactly once by manager bootstrap, before controller setup; standalone manager-backed test harnesses own the same prerequisite. For example, Component and Discovery share the Discovery component-reference index:
```
frewilhelm, line 911: > For example, Component and Discovery share the Discovery component-reference index:

Did I overlooked this somewhere or is this not up-to-date?

website/content/docs/concepts/ocm-controllers.md
```diff
@@ -115,6 +117,23 @@ A few more things about replication:
...
+
+A `Discovery` also sits alongside the chain. It references a `Component` and, once that `Component` is `Ready`, resolves
+the entire transitive component graph from the `Component`'s resolved repository, filters it with reference, component,
+and resource selectors, and publishes either the filtered raw v2 descriptors (`status.components`) or CEL-projected
+free-form records (`status.extracted`). It is a read-only, query-style resource: it downloads no artifacts, creates no
+external resources, and provides no signature-verification guarantees for the descriptors it filters.
+
+A few things about discovery:
+
+- Traversal is full and fail-fast: the whole reachable graph is resolved before filtering, and the first resolution failure retains the last successful payload and reports `Ready=False` with reason `ResolutionFailed`. There is no partial resolution or identity short-circuiting.
```
frewilhelm, line 130: > There is no [...] identity short-circuiting

What was the reason omitting this? I believe we should update the ADR then

website/content/docs/concepts/ocm-controllers.md
```diff
@@ -115,6 +117,23 @@ A few more things about replication:
...
+- Results are gated on freshness: consumers must check `Ready=True` **and** `status.observedGeneration == metadata.generation`, because failures retain the last successful (possibly stale) payload.
+- A controller-wide safety interval (`--discovery-controller-safety-interval`, Helm `manager.discovery.safetyInterval`, default `30m`) re-queues successful Discoveries as insurance against missed watch events. It is not a per-object interval.
```
frewilhelm, line 132: I believe this is deprecated

bindings/go/kubernetes/controller/internal/controller/discovery/suite_test.go
```diff
@@ -0,0 +1,57 @@
...
+	Expect(err).NotTo(HaveOccurred())
+	Expect(cfg).NotTo(BeNil())
+
+	scheme := runtime.NewScheme()
+	Expect(corev1.AddToScheme(scheme)).To(Succeed())
+	Expect(v1alpha1.AddToScheme(scheme)).To(Succeed())
+
+	k8sClient, err = client.New(cfg, client.Options{Scheme: scheme})
+	Expect(err).NotTo(HaveOccurred())
+	Expect(k8sClient).NotTo(BeNil())
```
frewilhelm, line 51: Is this needed if the reconciler is created in `discovery_controller_test.go`?

bindings/go/kubernetes/controller/internal/controller/discovery/discovery_controller_test.go
```diff
@@ -0,0 +1,441 @@
...
+			Client:           fakeClient,
+			Scheme:           scheme,
+			EventRecorder:    record.NewFakeRecorder(100),
+			NewPluginManager: test.StaticPluginManager(nil),
+		},
+	}
+	return r, fakeClient
+}
+
+func readyComponent(name, namespace string) *v1alpha1.Component {
```
frewilhelm, line 70: You can also use `MockComponent` although `readyComponent` works too

bindings/go/kubernetes/controller/internal/controller/discovery/discovery_controller_test.go
frewilhelm, line 1: These test-cases are structured differently from the test-cases in our other controllers. Was there any specific reason for that?

frewilhelm (commented):
preliminary code-review. I will test the controller and the how-to tomorrow. Overall it looks solid!

bindings/go/kubernetes/controller/internal/controller/discovery/discovery_controller.go
matthiasbruns: what about the other way
digest unchanged + previous run failed ends in skip

bindings/go/kubernetes/controller/internal/discovery/traverse.go
```diff
@@ -0,0 +1,136 @@
...
+// themselves remain untouched.
+//
+// The traversal is fail-fast: the first resolution failure cancels the
+// remaining traversal and is returned wrapped with the failing component
+// identity. The graph is only consumed after the entire traversal succeeded;
+// on error no partial graph is returned.
+//
+// Cycle detection is not implemented here; it belongs to the shared DAG
+// package (open-component-model/ocm-project#705).
+func Traverse(ctx context.Context, root ComponentKey, resolver resolvers.ComponentVersionRepositoryResolver) (*Graph, error) {
```
matthiasbruns, line 97: I know we do not have cycle detection fully in our DAG but would we want a deadlock in our controllers? What about a safeguard / timeout here?

bindings/go/kubernetes/controller/internal/discovery/query.go
Skarlso: It's an upper limit that has no consequence technically. This will not be able to be hit and there is no like.. memory that will exhaust this. So I believe, this is inconsequential from the user's or the controller's perspective. Like, it's not like this amount of thing is pre-allocated so it doesn't really matter. And a million you'll have to have an enormous amount of graph data to hit that at once.

bindings/go/kubernetes/controller/internal/controller/discovery/discovery_controller.go
Skarlso: >What if the component digest did not change but the selector on the discovery?

Any change in the discovery triggers a full run.

>digest unchanged + previous run failed ends in skip

This field is only ever set in case of a success.

Any other update would result in an observed generation update of the same object which should result in a reconcile again. But I will verify this claim. :)

frewilhelm: > Any change in the discovery triggers a full run.

Ah, so in that case this would return `false` right? 

```go
	if discovery.Status.ObservedGeneration != discovery.GetGeneration() {
		return false
	}
```

bindings/go/kubernetes/controller/config/samples/delivery_v1alpha1_discovery.yaml
frewilhelm, line 1: we do not have any other samples from the other CRs in here. I believe we removed them when we started using helm instead of kustomize. do we still want this single sample?

bindings/go/kubernetes/controller/api/v1alpha1/discovery_types.go
```diff
@@ -0,0 +1,237 @@
...
+	// +optional
+	ReferenceSelector *Selector `json:"referenceSelector,omitempty"`
+
+	// ComponentSelector filters the components of the filtered graph.
+	// If unset, all components are kept.
+	// +optional
+	ComponentSelector *Selector `json:"componentSelector,omitempty"`
+
+	// ResourceSelector filters the resources of each surviving component.
+	// Components with zero surviving resources are kept. If unset, all
```
frewilhelm, line 78: Why would we keep component with zero surviving resources? If I want to have a list of components that contain a specific resource (by any identity), then I would set the resource selector and expect only components that have matching resources.

bindings/go/kubernetes/controller/internal/controller/discovery/discovery_controller.go
Skarlso: Correct.

bindings/go/kubernetes/controller/internal/discovery/traverse.go
Skarlso: It's technically not a deadlock as it times out eventually with a context cancelled error. 🤔 

But we opted for not fixing this in this PR, because it will require SOME amount of fiddling.

bindings/go/kubernetes/controller/config/samples/delivery_v1alpha1_discovery.yaml
Skarlso: Oh good point. I can remove it.

bindings/go/kubernetes/controller/internal/ocm/ocm.go
Skarlso: Nah, you're right I think. Nothing will reference a Discovery and inherit configs from it. I'm removing it.

website/content/docs/concepts/ocm-controllers.md
Skarlso: It was complicating error propagation way too much. It requires specific CEL error value parsing BY string, because CEL doesn't expose most of its internal error types. It added like... ~200-300 lines of error parsing and logic bullshittery which was ugly as hell and unmaintainable.

On top of that, it didn't _actually_ provide and value. Because you'd have to fix whatever the hell is wrong anyways immediately. So we decided against partial results. :)

docs/coding-patterns.md
Skarlso: Not up to date. I fixed this.

bindings/go/kubernetes/controller/api/v1alpha1/discovery_types.go
```diff
@@ -0,0 +1,237 @@
...
+	// +optional
+	// +kubebuilder:validation:items:Type=object
+	// +kubebuilder:validation:items:XPreserveUnknownFields
+	Components []apiextensionsv1.JSON `json:"components,omitzero"`
+
+	// Extracted contains the records projected from the filtered components
+	// by spec.extract. It is only set when spec.extract is set. A selected but
+	// empty result is an empty list; an uncomputed result is absent.
+	// +optional
+	Extracted []ExtractedRecord `json:"extracted,omitzero"`
```
fabianburth, line 135: What's the reasoning behind having different fields for extracted and components?

bindings/go/kubernetes/controller/api/v1alpha1/discovery_types.go
```diff
@@ -0,0 +1,237 @@
...
+
+	// ObservedComponentDigest is the digest of the root Component
+	// It is recorded only when every reference in the discovered graph has a
+	// digest. An empty value forces a full re-discovery.
+	//
+	// This value is used to determine if a full re-discovery is required or not.
+	// If the root component version's digest didn't change, there is no need
+	// to re-walk the entire component version. Everything should be still the same.
+	// Force-pushing over a version is not supported.
```
fabianburth, line 144: why not? wouldn't that change the digest?

bindings/go/kubernetes/controller/internal/controller/discovery/discovery_controller_test.go
Skarlso: Not really, I think.

bindings/go/kubernetes/controller/api/v1alpha1/condition_types.go
```diff
@@ -83,3 +83,24 @@ const (
...
+	ResolutionFailedReason = "ResolutionFailed"
+
+	// PayloadTooLargeReason is used when the Discovery status payload exceeds the API server size limit.
+	PayloadTooLargeReason = "PayloadTooLarge"
+
+	// NoReferencesMatchedReason is used when the Discovery reference selector matched no references.
+	NoReferencesMatchedReason = "NoReferencesMatched"
+
+	// NoComponentsMatchedReason is used when the Discovery component selector matched no components.
+	NoComponentsMatchedReason = "NoComponentsMatched"
```
fabianburth: Couldn't those two share the same condition?

bindings/go/kubernetes/controller/internal/controller/discovery/discovery_controller_test.go
Skarlso: MockComponent actually doesn't really work here. the discovery test doesn't register the same runners like the others for ginkgo.

bindings/go/kubernetes/controller/chart/templates/rbac/manager-role.yaml
```diff
@@ -57,10 +57,19 @@ rules:
...
       - patch
       - update
+  - apiGroups:
+      - delivery.ocm.software
+    resources:
+      - discoveries
+    verbs:
+      - get
+      - list
+      - watch
```
fabianburth, line 75: Why does `discoveries` only get `get,list,watch` here? Shouldn't it be added to the list in line 32?

bindings/go/kubernetes/controller/api/v1alpha1/discovery_types.go
Skarlso: Good point. I think this is a bit broken now. I'll fix it in a sec.

bindings/go/kubernetes/controller/internal/cel/functions/semver.go
fabianburth, line 1: There is a kubernetes semver library https://kubernetes.io/docs/reference/using-api/cel/#kubernetes-semver-library. I think we should prefer using that, WDYT?

bindings/go/kubernetes/controller/internal/controller/component/component_controller.go
```diff
@@ -83,6 +86,18 @@ func (r *Reconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager) err
...
 	}
+	// This index is required to get all discoveries that reference a component. This is required to make sure that when
+	// deleting the component, no discovery exists anymore that references that component.
+	if err := mgr.GetFieldIndexer().IndexField(ctx, &v1alpha1.Discovery{}, discoveryIndex, func(obj client.Object) []string {
+		discovery, ok := obj.(*v1alpha1.Discovery)
+		if !ok {
+			return nil
+		}
+
+		return []string{discovery.Spec.ComponentRef.Name}
```
fabianburth, line 97: Don't we need to add the discoveries namespace to the key here? Isn't the index shared by discoveries in different namespaces?

bindings/go/kubernetes/controller/internal/controller/component/component_controller.go
```diff
@@ -375,6 +399,36 @@ func (r *Reconciler) reconcileDelete(ctx context.Context, component *v1alpha1.Co
...
+			names = append(names, fmt.Sprintf("%s/%s", d.Namespace, d.Name))
+		}
+
+		msg := fmt.Sprintf(
+			"component cannot be removed as discoveries are still referencing it: %s",
+			strings.Join(names, ","),
+		)
+		status.MarkNotReady(r.EventRecorder, component, v1alpha1.DeletionFailedReason, msg)
+
+		return errors.New(msg)
```
fabianburth, line 429: nit: could be a normal `fmt.Errorf`, couldn't it?

bindings/go/kubernetes/controller/internal/controller/discovery/status.go
```diff
@@ -0,0 +1,20 @@
...
+
+// publish writes the mutated status to the API server with an
+// optimistic-lock merge patch.
+func (r *Reconciler) publish(ctx context.Context, old, discovery *v1alpha1.Discovery) error {
+	if equality.Semantic.DeepEqual(old.Status, discovery.Status) {
+		return nil
+	}
+	return r.GetClient().Status().Patch(ctx, discovery,
+		client.MergeFromWithOptions(old, client.MergeFromWithOptimisticLock{}))
+}
```
fabianburth, line 20: Are we fine with the discovery controller having a completely different status handling than the other controllers?

bindings/go/kubernetes/controller/internal/discovery/errors.go
```diff
@@ -0,0 +1,149 @@
...
+		return false
+	}
+	msg := err.Error()
+	for _, prefix := range missingAccessPrefixes {
+		if strings.HasPrefix(msg, prefix) {
+			return true
+		}
+	}
+	return false
+}
```
fabianburth, line 98: didn't we agree we wanted to get rid of this error handling?

bindings/go/kubernetes/controller/internal/discovery/identity.go
```diff
@@ -0,0 +1,73 @@
...
+// "componentName" attribute holding the referenced component's name.
+// Reference.ToComponentIdentity alone would lose the local name and extras.
+func referenceIdentity(ref *descriptor.Reference) runtime.Identity {
+	identity := ref.ToIdentity()
+	if identity == nil {
+		identity = runtime.Identity{}
+	}
+	identity["componentName"] = ref.Component
+	return identity
+}
```
fabianburth, line 19: I feel like we already have such a function

bindings/go/kubernetes/controller/internal/discovery/identity.go
```diff
@@ -0,0 +1,73 @@
...
+
+// componentIdentity returns the selector identity of a component.
+func componentIdentity(component *descriptor.Component) runtime.Identity {
+	return component.ToIdentity()
+}
+
+// resourceIdentity returns the selector identity of a resource.
+func resourceIdentity(res *descriptor.Resource) runtime.Identity {
+	return res.ToIdentity()
+}
```
fabianburth: this looks like useless AI slop too (I know it might be my slop, so no blame here ;)) we should probably remove that

bindings/go/kubernetes/controller/internal/discovery/project.go
```diff
@@ -0,0 +1,265 @@
...
+// output fields is selected; the selected field is always non-nil, even when
+// the result is empty.
+type Payload struct {
+	// Components carries the filtered raw v2 descriptors, sorted
+	// lexicographically by (component.name, component.version). Selected when
+	// the Discovery has no extract configuration.
+	Components []json.RawMessage
+	// Extracted carries the projected records. Selected when the Discovery
+	// has an extract configuration.
+	Extracted []map[string]any
```
fabianburth, line 23: I asked this in the API already. Do we really need separate fields here?

bindings/go/kubernetes/controller/internal/discovery/project.go
```diff
@@ -0,0 +1,265 @@
...
+			return nil, err
+		}
+		var generic map[string]any
+		if err := json.Unmarshal(raw, &generic); err != nil {
+			return nil, fmt.Errorf("failed to unmarshal descriptor %s into generic map: %w", d.Component.String(), err)
+		}
+		maps = append(maps, generic)
+	}
+	return maps, nil
+}
```
fabianburth: didn't @matthiasbruns implement a to unstructured at some point?

bindings/go/kubernetes/controller/api/v1alpha1/discovery_types.go
Skarlso: It is described in the ADR https://github.com/open-component-model/open-component-model/pull/3068#discussion_r3845750234

bindings/go/kubernetes/controller/api/v1alpha1/discovery_types.go
Skarlso: If you don't regenerate your component descriptor I would think not. But hopefully, that would break different checks down the line in OCM... :)

bindings/go/kubernetes/controller/api/v1alpha1/condition_types.go
Skarlso: Sure.

bindings/go/kubernetes/controller/internal/controller/component/component_controller.go
Skarlso: Ugh, I think you're right here.

bindings/go/kubernetes/controller/internal/controller/discovery/status.go
Skarlso: I tried to do it the same way, but it actually was breaking things... I'll pull up my notes on why I choose to leave this as is in a bit.

bindings/go/kubernetes/controller/internal/discovery/errors.go
Skarlso: Please read the description if this function. :D

bindings/go/kubernetes/controller/internal/discovery/identity.go
Skarlso: Agreed. :D

bindings/go/kubernetes/controller/internal/discovery/project.go
Skarlso: Ah NewUnstructured. Let's see if I can use that here.

bindings/go/kubernetes/controller/chart/templates/rbac/manager-role.yaml
Skarlso: No clue. I'll investigate. :D

bindings/go/kubernetes/controller/internal/discovery/project.go
matthiasbruns: yeah we now have support for `Unstructured` `scheme.Convert` or `UnstructuredFromMixedData`

bindings/go/kubernetes/controller/chart/templates/rbac/manager-role.yaml
Skarlso: The answer is: Because it actually doesn't really require anything else as far as I can see.

So what else would you expect here?

bindings/go/kubernetes/controller/internal/cel/functions/semver.go
Skarlso: This isn't a nice constraint checker though, right? Like, I don't see how `^1.2.3, ~1.2, 1.x` would work at all with this library. Sure, you can do comparison, but not selection.

bindings/go/kubernetes/controller/internal/controller/component/component_controller.go
Skarlso: Ah actually no, I don't think so. In none of our watches we do that and that's mostly because a) the controller runtime already adds the namespace[^1]; b) actually there is a bug in the repository controller I think. :)

[^1]:

```go
vals[i] = internal.KeyToNamespacedKey(ns, rawVal)   // "<ns>/<value>"
if ns != "" {
    vals[i+len(rawVals)] = internal.KeyToNamespacedKey("", rawVal)  // all-namespaces variant
}
```

bindings/go/kubernetes/controller/internal/controller/component/component_controller.go
Skarlso: `msg` is reused in the MarkNotReady condition above, that's why it's a variable.

bindings/go/kubernetes/controller/internal/controller/component/component_controller.go
Skarlso: Yah, the repository's delete operation was missing a namespace.... ups... :D

bindings/go/kubernetes/controller/internal/discovery/identity.go
Skarlso: I couldn't find one. Do you have it?

bindings/go/kubernetes/controller/internal/discovery/project.go
Skarlso: @matthiasbruns Any reason why `toUnstructuredMap` is not exported? That's what we need here.

Skarlso: I did the change, but honestly unsure if this was worth it here. 🤷‍♀️ 

CEL wants a map[string]any. Changing this to unstructured resulting in having to call .Data everywhere instead.

Skarlso: I think we could replace the entire function call but for that I really need ToUnstructuredMap. :) And I don't know if I can export that. :) @matthiasbruns ++

Skarlso: Oh not even that. It's using Typed so I can't push in the descriptor to it unless I create a wrapper. I don't know how this could work. :) Please discuss.

bindings/go/kubernetes/controller/internal/controller/discovery/discovery_controller.go
```diff
@@ -0,0 +1,415 @@
...
```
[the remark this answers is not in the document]
Skarlso: Any refresh would cause a generation change on the object, no? Please verify.

bindings/go/kubernetes/controller/README.md
```diff
@@ -61,6 +61,17 @@ which tool reconciles the final workload. Both **FluxCD** and **ArgoCD** are sup
...
+## Discovery
+
+The `Discovery` controller resolves a component graph and puts the resulting component
+descriptor into the `status` of the `Discovery` object. It selects a repository per component
+identity, using the effective OCM configuration (including inherited config and credentials) and
+following the same resolver precedence as the `ocm` CLI.
+
+To read more about how the `Discovery` works and how it could be used to fetch available
+versions of certain components, please read the [Discovery Controller Documentation](https://ocm.software/docs/how-to/discover-component-graphs).
```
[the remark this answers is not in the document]
Skarlso: Since the document is IN THIS PR, I won't be able to fix that. 😏

bindings/go/kubernetes/controller/internal/controller/discovery/discovery_controller.go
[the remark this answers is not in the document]
Skarlso: Fair. @matthiasbruns was right. :D

bindings/go/kubernetes/controller/internal/ocm/ocm.go
frewilhelm: What do you think about the replication as its still there. Does it make sense to add it?

website/content/docs/concepts/ocm-controllers.md
frewilhelm: How is this related to CEL? If the unique identity (`name` + `version`) is found on a component version, we exit the DAG traversal (see [adr](https://github.com/open-component-model/open-component-model/blob/2e80662c63253308e3304f68f8f9999ace9821eb/docs/adr/0027_ocm_k8s_toolkit_discover.md?plain=1#L189)).

I can agree that this might be an optimisation we can do later, but I don't get the error parsing etc :D

bindings/go/kubernetes/controller/internal/controller/discovery/discovery_controller_test.go
frewilhelm: What runners do you mean?

bindings/go/kubernetes/controller/internal/controller/discovery/discovery_controller_test.go
frewilhelm: We create a new client and reconciler for nearly every test-tun instead of setting it up once in the `suite_test.go`. I am not sure if that makes sense, but also I am unsure if that really hits performance.

bindings/go/kubernetes/controller/internal/ocm/ocm.go
Skarlso: I... don't know. :D I didn't want to touch unrelated things in here. 🤔

website/content/docs/concepts/ocm-controllers.md
Skarlso: Basically, partial results would have also needed to check CEL parsing errors. Because in certain cases, a CEL parser if it didn't match, would return some obscure black magic error. So we needed to check that.

And on top of that... I don't remember. :D Let me check my notes.

bindings/go/kubernetes/controller/internal/controller/discovery/discovery_controller_test.go
Skarlso: MockComponent is a ginkgo helper. This is not a ginkgo suite.

website/content/docs/concepts/ocm-controllers.md
Skarlso: AH I REMEMBER!! :D It was the Graph traversal and the potential circular deps and things like, the graph doesn't know which node failed. So like, in the CLI when you get a partial result from the graph, it would give you ALL things that didn't resolve and not just the thing that failed because it doesn't know which failed since it's inherently parallel. 

So you would get a bunch of things and you'd have to somehow decide what to display. In CLI that's fine, because it would just error, done. Here, we need to construct objects out of them, and it would be confusing as hell in the status to get partial objects or null entries.

So yeah, no partial result matching. That's just ugly.

bindings/go/kubernetes/controller/internal/controller/discovery/discovery_controller_test.go
Skarlso: I'll try to see if I can consolidate these tests. There are a rather large number of them in there. :D

Skarlso: Ah we actually do this for other controllers too I see.

bindings/go/kubernetes/controller/internal/controller/discovery/suite_test.go
Skarlso: This client is used in discovery_api_test.go. And also, this is a different kind of client. :)

We could move it, but that would break the pattern of how the rest of the controllers are all doing it. So it stayed this way.

bindings/go/kubernetes/controller/internal/controller/discovery/discovery_controller_test.go
Skarlso: Refactored some helpers to make this a bit better looking. :)

bindings/go/kubernetes/controller/api/v1alpha1/discovery_types.go
Skarlso: I don't think my link works quite right:

TL;DR:

```
[ntnn](https://github.com/ntnn)
[3 weeks ago](https://github.com/open-component-model/open-component-model/pull/3068#discussion_r3924847058)
[Contributor](https://github.com/open-component-model/open-component-model/commits?author=ntnn)
How about this?

   Components []v2.ComponentDescriptor           `json:"components,omitempty"`
   Extracted  []map[string]apiextensionsv1.JSON  `json:"extracted,omitempty"`
Then discovery with no cel extraction stil gets a fully typed status and with cel extraction a list of results per discovered component

@[Skarlso](https://github.com/Skarlso)
Skarlso
[3 weeks ago](https://github.com/open-component-model/open-component-model/pull/3068#discussion_r3925569869)
Member
Hmmm, I feel like, two entries would make this ambiguous, no? Like, this means that you'll have to check both of them all the time, not?

@[ntnn](https://github.com/ntnn)
ntnn
[3 weeks ago](https://github.com/open-component-model/open-component-model/pull/3068#discussion_r3925962559)
[Contributor](https://github.com/open-component-model/open-component-model/commits?author=ntnn)
If you request discovery with extraction I'd expect you check .extracted, if you didn't you check .components for the full typed spec.
```

Skarlso: Also, thinking about this and trying to explain it in the review I think this is a bit flawed. :D 

I mean I see how it's beneficial, but in CEL you'd have to still do something like `${if spec.extract.... if spec.component` because you don't know which of them has the result if it failed.

And you don't always control the CEL that led you here so the argument
>If you request discovery with extraction I'd expect you check .extracted, if you didn't you check .components for the full typed spec.

Isn't always true, I think 🤔

frewilhelm (commented):
I did some tests with a custom service catalog and it works fine so far! Well done

<details><summary>`service-catalog.yaml` (component descriptor; is publicly available in my ghcr.io)</summary>

```yaml
# yaml-language-server: $schema=https://ocm.software/schemas/configuration-schema.yaml
components:
  # Flux components
  - name: ocm.software/service-catalog/flux
    version: "2.8.3"
    provider:
      name: ocm.software
    resources:
      - name: flux
        version: 2.18.2
        type: helmChart
        relation: external
        access:
          type: ociArtifact
          imageReference: ghcr.io/fluxcd-community/charts/flux2:2.18.2
      - name: image-helm-controller
        version: v1.5.3
        type: ociImage
        relation: external
        access:
          type: ociArtifact
          imageReference: ghcr.io/fluxcd/helm-controller:v1.5.3
      - name: image-automation-controller
        version: v1.1.1
        type: ociImage
        relation: external
        access:
          type: ociArtifact
          imageReference: ghcr.io/fluxcd/image-automation-controller:v1.1.1
      - name: image-reflector-controller
        version: v1.1.1
        type: ociImage
        relation: external
        access:
          type: ociArtifact
          imageReference: ghcr.io/fluxcd/image-reflector-controller:v1.1.1
      - name: image-kustomize-controller
        version: v1.8.2
        type: ociImage
        relation: external
        access:
          type: ociArtifact
          imageReference: ghcr.io/fluxcd/kustomize-controller:v1.8.2
      - name: image-notification-controller
        version: v1.8.2
        type: ociImage
        relation: external
        access:
          type: ociArtifact
          imageReference: ghcr.io/fluxcd/notification-controller:v1.8.2
      - name: image-source-controller
        version: v1.8.1
        type: ociImage
        relation: external
        access:
          type: ociArtifact
          imageReference: ghcr.io/fluxcd/source-controller:v1.8.1
  - name: ocm.software/service-catalog/flux
    version: "2.18.2"
    provider:
      name: ocm.software
    resources:
      - name: flux
        version: 2.18.2
        type: helmChart
        relation: external
        access:
          type: ociArtifact
          imageReference: ghcr.io/fluxcd-community/charts/flux2:2.18.2
      - name: image-helm-controller
        version: v1.5.3
        type: ociImage
        relation: external
        access:
          type: ociArtifact
          imageReference: ghcr.io/fluxcd/helm-controller:v1.5.3
      - name: image-automation-controller
        version: v1.1.1
        type: ociImage
        relation: external
        access:
          type: ociArtifact
          imageReference: ghcr.io/fluxcd/image-automation-controller:v1.1.1
      - name: image-reflector-controller
        version: v1.1.1
        type: ociImage
        relation: external
        access:
          type: ociArtifact
          imageReference: ghcr.io/fluxcd/image-reflector-controller:v1.1.1
      - name: image-kustomize-controller
        version: v1.8.2
        type: ociImage
        relation: external
        access:
          type: ociArtifact
          imageReference: ghcr.io/fluxcd/kustomize-controller:v1.8.2
      - name: image-notification-controller
        version: v1.8.2
        type: ociImage
        relation: external
        access:
          type: ociArtifact
          imageReference: ghcr.io/fluxcd/notification-controller:v1.8.2
      - name: image-source-controller
        version: v1.8.1
        type: ociImage
        relation: external
        access:
          type: ociArtifact
          imageReference: ghcr.io/fluxcd/source-controller:v1.8.1
  # OCM
  - name: ocm.software/kubernetes/controller
    version: "0.9.0"
    provider:
      name: ocm.software
    resources:
      - name: chart
        version: 0.9.0
        type: helmChart
        relation: local
        access:
          type: ociArtifact
          imageReference: ghcr.io/open-component-model/kubernetes/controller/chart:0.9.0
      - name: image
        version: 0.9.0
        type: ociImage
        relation: local
        access:
          type: ociArtifact
          imageReference: ghcr.io/open-component-model/kubernetes/controller:0.9.0
  - name: ocm.software/kubernetes/controller
    version: "0.10.0"
    provider:
      name: ocm.software
    resources:
      - name: chart
        version: 0.10.0
        type: helmChart
        relation: local
        access:
          type: ociArtifact
          imageReference: ghcr.io/open-component-model/kubernetes/controller/chart:0.10.0
      - name: image
        version: 0.10.0
        type: ociImage
        relation: local
        access:
          type: ociArtifact
          imageReference: ghcr.io/open-component-model/kubernetes/controller:0.10.0
  - name: ocm.software/kubernetes/controller
    version: "0.11.0"
    provider:
      name: ocm.software
    resources:
      - name: chart
        version: 0.11.0
        type: helmChart
        relation: local
        access:
          type: ociArtifact
          imageReference: ghcr.io/open-component-model/kubernetes/controller/chart:0.11.0
      - name: image
        version: 0.11.0
        type: ociImage
        relation: local
        access:
          type: ociArtifact
          imageReference: ghcr.io/open-component-model/kubernetes/controller:0.11.0
  # Kro
  - name: ocm.software/service-catalog/kro
    version: v0.9.0
    provider:
      name: ocm.software
    labels:
      - name: experimental
        value: "true"
    resources:
      - name: kro
        version: v0.9.0
        type: helmChart
        relation: external
        access:
          type: ociArtifact
          imageReference: registry.k8s.io/kro/charts/kro:0.9.0
      - name: image-kro
        version: v0.9.0
        type: ociImage
        relation: external
        access:
          type: ociArtifact
          imageReference: registry.k8s.io/kro/kro:v0.9.0
  - name: ocm.software/service-catalog/kro
    version: v0.9.1
    provider:
      name: ocm.software
    labels:
      - name: experimental
        value: "true"
    resources:
      - name: kro
        version: v0.9.1
        type: helmChart
        relation: external
        access:
          type: ociArtifact
          imageReference: registry.k8s.io/kro/charts/kro:0.9.1
      - name: image-kro
        version: v0.9.1
        type: ociImage
        relation: external
        access:
          type: ociArtifact
          imageReference: registry.k8s.io/kro/kro:v0.9.1
  - name: ocm.software/service-catalog/kro
    version: v0.9.2
    provider:
      name: ocm.software
    labels:
      - name: experimental
        value: "true"
    resources:
      - name: kro
        version: v0.9.2
        type: helmChart
        relation: external
        access:
          type: ociArtifact
          imageReference: registry.k8s.io/kro/charts/kro:0.9.2
      - name: image-kro
        version: v0.9.2
        type: ociImage
        relation: external
        access:
          type: ociArtifact
          imageReference: registry.k8s.io/kro/kro:v0.9.2
  # Service catalog
  - name: ocm.software/service-catalog
    version: "1.0.0"
    provider:
      name: ocm.software
    componentReferences:
      - name: flux
        componentName: ocm.software/service-catalog/flux
        version: "2.8.3"
        labels:
          - name: landscape
            value: ["prod"]
      - name: flux
        componentName: ocm.software/service-catalog/flux
        version: "2.18.2"
        labels:
          - name: landscape
            value: ["dev", "canary"]
      - name: ocm
        componentName: ocm.software/kubernetes/controller
        version: "0.9.0"
        labels:
          - name: landscape
            value: ["prod"]
      - name: ocm
        componentName: ocm.software/kubernetes/controller
        version: "0.10.0"
        labels:
          - name: landscape
            value: ["canary"]
      - name: ocm
        componentName: ocm.software/kubernetes/controller
        version: "0.11.0"
        labels:
          - name: landscape
            value: ["dev"]
      - name: kro
        componentName: ocm.software/service-catalog/kro
        version: v0.9.0
        labels:
          - name: landscape
            value: ["prod"]
      - name: kro
        componentName: ocm.software/service-catalog/kro
        version: v0.9.1
        labels:
          - name: landscape
            value: ["canary"]
      - name: kro
        componentName: ocm.software/service-catalog/kro
        version: v0.9.2
        labels:
          - name: landscape
            value: ["dev"]
```

</details>

<details><summary>`discovery.yaml`</summary>

```yaml
apiVersion: delivery.ocm.software/v1alpha1
kind: Repository
metadata:
  name: repository
spec:
  repositorySpec:
    # Adjust to your OCM repository
    baseUrl: ghcr.io/frewilhelm
    type: OCIRegistry
  interval: 10m
---
apiVersion: delivery.ocm.software/v1alpha1
kind: Component
metadata:
  name: component
spec:
  component: ocm.software/service-catalog
  repositoryRef:
    name: repository
  semver: 1.0.0
  interval: 10m
---
apiVersion: delivery.ocm.software/v1alpha1
kind: Discovery
metadata:
  name: simple
spec:
  componentRef:
    name: component
  resourceSelector:
    expression: identity.name in ["kro"]
  extract:
    byResources:
      imageRef: resource.access.imageReference
      resourceName: resource.name
      componentName: component.name
      componentVersion: component.version
---
apiVersion: delivery.ocm.software/v1alpha1
kind: Discovery
metadata:
  name: semver
spec:
  componentRef:
    name: component
  resourceSelector:
    matchIdentity:
      name: flux
    expression: semverCheck(identity.version, ">=2.10")
  extract:
    byComponents:
      name: component.name
      version: component.version
---
apiVersion: delivery.ocm.software/v1alpha1
kind: Discovery
metadata:
  name: references
spec:
  componentRef:
    name: component
  referenceSelector:
    matchIdentity:
      name: ocm
  extract:
    byResources:
      name: resource.name
      imageReference: resource.access.imageReference
---
apiVersion: delivery.ocm.software/v1alpha1
kind: Discovery
metadata:
  name: labels
spec:
  componentRef:
    name: component
  componentSelector:
    expression: 'labels.?experimental.orValue("false") != "true"'
  referenceSelector:
    expression: '["prod", "canary"].exists(e, e in labels.landscape)'
  extract:
    byComponents:
      name: component.name
      version: component.version
---
apiVersion: delivery.ocm.software/v1alpha1
kind: Discovery
metadata:
  name: flux-helm-values
spec:
  componentRef:
    name: component
  referenceSelector:
    matchIdentity:
      name: flux
  extract:
    byComponents:
      version: component.version
      chartVersion: component.resources.filter(r, r.name == "flux")[0].version
      chartUrl: '"oci://" + component.resources.filter(r, r.name == "flux")[0].access.imageReference.split("@")[0].split(":")[0]'
      values.helmController.image: component.resources.filter(r, r.name == "image-helm-controller")[0].access.imageReference.split("@")[0].split(":")[0]
      values.helmController.tag: component.resources.filter(r, r.name == "image-helm-controller")[0].access.imageReference.split("@")[0].split(":")[1]
      values.helmController.digest: component.resources.filter(r, r.name == "image-helm-controller")[0].access.imageReference.split("@")[1]
      values.sourceController.image: component.resources.filter(r, r.name == "image-source-controller")[0].access.imageReference.split("@")[0].split(":")[0]
      values.sourceController.tag: component.resources.filter(r, r.name == "image-source-controller")[0].access.imageReference.split("@")[0].split(":")[1]
      values.sourceController.digest: component.resources.filter(r, r.name == "image-source-controller")[0].access.imageReference.split("@")[1]
```
</details>

I will do another code review later

bindings/go/kubernetes/controller/internal/ocm/ocm.go
frewilhelm: But you add it in this PR :D 
image

bindings/go/kubernetes/controller/internal/discovery/query.go
frewilhelm: okay, lets see how it goes

website/content/docs/concepts/ocm-controllers.md
frewilhelm: I don't understand that completely. Can't we just, while traversing, check if a unique identity matches the current node and if so, cancel the remaining traversal?

bindings/go/kubernetes/controller/internal/controller/discovery/suite_test.go
frewilhelm: So sorry, I just saw `k8sClient` is used in another test. I was mixing up the test structures with the other controllers

bindings/go/kubernetes/controller/internal/controller/discovery/digest_shortcircuit_test.go
```diff
@@ -0,0 +1,154 @@
...
+		ObjectMeta: metav1.ObjectMeta{Name: "discovery", Namespace: namespace},
+		Spec: v1alpha1.DiscoverySpec{
+			ComponentRef: corev1.LocalObjectReference{Name: componentName},
+		},
+	}
+}
+
+// TestReconcile_UnchangedRootDigestSkipsTraversal proves the skip by deleting
+// the repository between reconciles: a second reconcile that still succeeds
+// cannot have fetched anything.
```
frewilhelm: Just wondering - the descriptor is cached, is it? If so, removing the original `dir` should not make any difference.

bindings/go/kubernetes/controller/internal/controller/discovery/discovery_api_test.go
```diff
@@ -0,0 +1,289 @@
...
+					Namespace:    namespace.Name,
+				},
+				Spec: v1alpha1.DiscoverySpec{
+					ComponentRef: corev1.LocalObjectReference{Name: "releasechannel"},
+				},
+			}
+			Expect(k8sClient.Create(ctx, discovery)).To(Succeed())
+		})
+
+		It("distinguishes absent from empty payload fields", func(ctx SpecContext) {
```
frewilhelm: I am having trouble understanding the purpose of this test. The `discovery` created in `BeforeEach` would fail on reconciliation, because the `Component` is missing. Nevertheless, the object is there of course. Now, we get that object and check if some status fields are `nil` which they are, because its failing. Then, we manually set one status field with an empty payload and check afterwards if it is set. What do we test with that?

docs/coding-patterns.md
```diff
@@ -906,19 +906,6 @@ All controllers use `GenerationChangedPredicate` to only reconcile on spec chang
 For(&v1alpha1.Component{}, builder.WithPredicates(predicate.GenerationChangedPredicate{}))
 ```
 
-### Field Indexing and Cross-Resource Watches
```
frewilhelm: I don't think we need to remove that. My previous comment was challenging a change in this file. But I believe this pattern is still correct.

website/content/docs/how-to/discover-component-graphs.md
frewilhelm: This How-To definitely needs more love. First of all, its not a How-To as we expect it in our [`CONTRIBUTING.md`](https://github.com/open-component-model/open-component-model/blob/main/website/CONTRIBUTING.md#how-to-guides-contentdocshow-to), it's a Tutorial with parts of a concept. Its way too long and describe a lot of stuff instead of focusing on one task. For a How-To I would rather have expected something like `How To discover resources by name and label` or the like and, then, a step-by-step guide on how to set it up.

Additionally, we should have a real-life example to guide the users through. A non-existent example is super hard to follow and reproduce. Maybe we can work with the service-catalog.yaml I posted earlier.

Nevertheless, I don't want this to be a blocker. Lets move this to a tutorial and rework it later

bindings/go/kubernetes/controller/internal/controller/discovery/discovery_api_test.go
Skarlso: yah, I agree. This seems to be testing the api and the k8s client at this point, which is weird and unnecessary. I take a good look at this.

Skarlso: Sorry, envtest rather.

Skarlso: So this thing is testing the CEL validations rather. And it doesn't fail yes, because there are no reconcilers here of course.

That said, there are a bunch of redundant bullshittery here. :D So I'll trim this, and leave the rest. Like, three tests or so still make some sense. :D

Skarlso: Trimmed.

docs/coding-patterns.md
Skarlso: Mmm, I don't like the added complexity and possible different behavior in this huge PR already. I saved as a TODO to revisit this as we might want to pick it up again later. And it did lead to a few places where namespace logic was fucked up anyways. :) Here, it's fine now. But we can do this in a separate PR.

website/content/docs/how-to/discover-component-graphs.md
Skarlso: Agreed. I will rename it though as a Tutorial because it is NOT a how-to... :)

Skarlso: >Additionally, we should have a real-life example to guide the users through. A non-existent example is super hard to follow and reproduce. Maybe we can work with the service-catalog.yaml I posted earlier.

Actually, don't we do that mostly for Tutorials?

Skarlso: Let's put this into the Concept instead?

Skarlso: @frewilhelm I refactored most of the documentation now into Concept + a dedicated focused How to. :)

bindings/go/kubernetes/controller/api/v1alpha1/discovery_types.go
Skarlso: We talked about this a lot. It's easier to figure out which one to parse based on which one you get. If it's one field it's less easy to figure out what type the data is that it holds. It's nicer ux.

website/content/docs/concepts/ocm-controllers.md
Skarlso: Nope, that's not how the parallel processing of the graph works.

bindings/go/kubernetes/controller/internal/discovery/project.go
matthiasbruns: Checked again - we can't since Descriptor is not `Typed` and Unstructured relies on scheme.Convert

matthiasbruns: What I do not understand is
1. why is this not typed with a proper schema
2. the json Marshal/Unmarshal is not implemented and throws errors

Skarlso: Thanks for checking. I consider this resolved. :) 🙇‍♀️

bindings/go/kubernetes/controller/internal/ocm/ocm.go
Skarlso: Fuck :D

bindings/go/kubernetes/controller/api/v1alpha1/discovery_types.go
Skarlso: The optimistic locking on the patch prevents this one from failing to update because we actively use the status for the next reconcile. We don't want potential stale status information in discovery object. So it's a bit stricter in the status update. The others, not so much.

PR #3611, opened 2026-09-15
fix(website): fix website pages that should be unversioned.

frewilhelm:
Some of our website pages should be unversioned, e.g. `community`, `governance`, or `blog`. It does not make sense that the meeting schedule in `community/how-we-work` differs between `0.12.0` and `0.15.0`. 

This PR proposes this change by mounting the current pages to present versions. This is required because, for example, in the How-To `website/content/docs/how-to/container-image-usage.md` we refer `[Community Support]({{< relref "community/_index.md" >}})` which is a relative link that only works in that version. Those relative links are also changed in this PR. Hence, if all those versions are retired we do not need to mount the unversioned pages to the module anymore because they are absolute links now, e.g. `/community`.

Additionally, this PR proposes to hide the version dropdown menu on those unversioned pages as it makes no sense there. Its only hidden (not removed) to contain the websites structure. Otherwise the items on the navigation bar would shift to the right.

Additionally, we had a little bug in our scripts where we used `!blog/**` instead of `! blog/**`. So, the blog page is currently versioned as well (see [discussion](https://discourse.gohugo.io/t/why-does-glob-negation-require-a-space-after/56651) or [docs](https://gohugo.io/configuration/module/#example)).

At last, I decided to remove the `community` page we used for ocm v1 as it is outdated.


frewilhelm:
@coderabbitai review

website/scripts/register-docs-version.test.js
```diff
@@ -143,7 +143,7 @@ test('buildModuleBlocks: website import has correct tag format', () => {
     // project owns the module graph (ignoreImports) and config (ignoreConfig).
     assert.equal(website.ignoreImports, true);
     assert.equal(website.ignoreConfig, true);
-    assert.deepEqual(website.mounts[0].files, ['**', '!blog/**']);
+    assert.deepEqual(website.mounts[0].files, ['! blog/**', '! community/**', '! governance/**']);
     assert.equal(website.mounts[0].source, 'content/');
     assert.equal(website.mounts[0].target, 'content');
     assert.deepEqual(website.mounts[0].sites.matrix.versions, ['0.3']);
```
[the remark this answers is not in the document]
frewilhelm: addressed in https://github.com/open-component-model/open-component-model/pull/3611/commits/d77d5c80422363e764594ef10c9377da219fefba

website/config/_default/menus/menus.en.yaml
```diff
@@ -3,16 +3,16 @@ main:
     url: docs/
     weight: 10
   - name: Community
-    url: community/
+    url: /community/
```
morri-son, line 6: why do we need leading slashes?

morri-son: ah, absolute vs relative

morri-son:
lgtm in general. There are two missing places where we need absolute paths: community/how-we-work.md:85 and community/onboarding.md:95.

morri-son (approved):

jakobmoellerdev (approved):

PR #3616, opened 2026-09-15
fix(website): fix register-docs script for `cli` and `kubernetes/controller`

frewilhelm:
https://github.com/open-component-model/open-component-model/pull/3613 failed the page building because it still tried to access to modules `bindings/go/cli` and `bindings/go/kubernetes/controller`. However, since `0.16.0` both modules are packages inside of `bindings/go`.

This PR proposes a fix, so the packages are added as mounts of `bindings/go` instead of its own module.

After this PR is merged, the action for #3613 can be retriggered and it should work (i tested it locally beforehand)


morri-son (dismissed):

website/scripts/register-docs-version.js
jneisener: I think this can now be removed

jneisener: This too

jneisener, line 490: Preexisting issue I think, but this check is no longer correct

website/scripts/register-docs-version.js
```diff
@@ -381,28 +401,40 @@ function buildModuleBlocks(version, fullVersion, deps) {
...
+                mounts: [{
+                    source: 'docs/reference',
+                    target: 'content/docs/reference/ocm-cli',
+                    sites: { matrix: { versions: [version] } }
+                }]
+            },
+        );
+    }
+
+    imports.push(...bindingSchemaImports(version, deps))
```
jneisener: Trailing `;` is missing, did you run the linter?

website/scripts/register-docs-version.js
frewilhelm: ah yes, i forgot about this

website/scripts/register-docs-version.js
frewilhelm: linter is not complaining :D

jneisener: Yep, you're right. I only added the bare minimum last time, I should enable some sane defaults.

`"semi": ["error", "always"],` is missing for this particular issue, but that doesn't have to be enabled here

website/scripts/register-docs-version.js
frewilhelm: discussed async that it does not matter for now

morri-son (dismissed):

jneisener (dismissed):

jneisener (approved):

matthiasbruns (approved):

PR #3635, opened 2026-09-17
feat(versioning): pluggable, config-driven versioning schemes

jakobmoellerdev:

OCM assumed **loose semantic versioning** everywhere component/resource/source/reference versions are validated, compared, sorted, or constrained. Arbitrary schemes such as calendar versioning (CalVer, e.g. `2024.03.15`) or monotonic build numbers were rejected on write and dropped or mis-ordered on read.

This PR makes versioning **pluggable and configuration-driven**:

- New leaf package `bindings/go/runtime/versioning` with an ordered, regex-based `Scheme`/`Registry` abstraction (`Compare`, `Valid`, `SortDescending`, `Filter`, `Satisfies`). `regexScheme` orders by named capture groups (numeric groups compared as integers, otherwise lexical).
- New typed config `versioning.config.ocm.software/v1alpha1` (`bindings/go/configuration/versioning/v1alpha1/spec`). The built-in loose-semver scheme is always appended as the final fallback, so **behavior is unchanged unless a scheme is configured** and adding schemes is purely additive.
- Every version-handling site now consults the registry instead of calling `semver` directly: CLI listing (`get`/`transfer`), OCI lister (optional comparator hook, semver default preserved), resolver `versionConstraint` (strict `Satisfies` gate), controller version resolution, and constructor write-side validation.
- Write-side: the embedded descriptor/constructor JSON schema no longer pins versions to a semver regex (validity is now scheme-driven); `descriptor/v2.ValidateVersions` and the constructor validate versions against the active registry.
- **OCI tag safety:** the version→tag rewrite (`LooseSemverToOCITag` → `VersionToOCITag`) now validates against the OCI tag grammar `^[\w][\w.-]{0,127}$`, warns loudly, and fails `add cv` at publish when a version cannot map to a valid tag — instead of silently storing a mangled tag. The `+`→`.build-` round-trip caveat is documented.

Version *constraints* (`>=1.0.0`) remain a semver concept: they filter semver-versioned entries and leave non-semver histories (e.g. CalVer) intact.

Docs: new reference page (`docs/reference/versioning-configuration.md`) with a copy-paste scheme catalog (CalVer variants, Ubuntu `YY.MM`, AWS dates, build numbers), an "author your own scheme" guide, and OCI tag constraints; plus a hands-on tutorial (`docs/tutorials/configure-versioning.md`).

##### How to test the changes

`versioning.ocmconfig`:
```yaml
configurations:
  - type: versioning.config.ocm.software/v1alpha1
    schemes:
      - name: calver-date
        pattern: '^(?P<year>\d{4})\.(?P<month>\d{2})\.(?P<day>\d{2})$'
        comparisonGroups: [year, month, day]
```

```bash
# add two CalVer versions (rejected today without the config)
ocm --config versioning.ocmconfig add cv --repository ./ctf --constructor cc.yaml   # version 2024.03.15
ocm --config versioning.ocmconfig add cv --repository ./ctf --constructor cc.yaml   # version 2024.10.01

# newest-first ordering: 2024.10.01 before 2024.03.15
ocm --config versioning.ocmconfig get cv ./ctf//acme.org/service

# confirm the active scheme
ocm --config versioning.ocmconfig get config
```

Verified end-to-end with the built CLI: CalVer add + newest-first listing; semver unchanged without config; genuinely non-semver rejected with a clear error; invalid OCI tag (`1:2.3`) warned and failed at publish.

##### Verification


website/content/docs/reference/versioning-configuration.md
```diff
@@ -0,0 +1,201 @@
...
+
+```yaml
+- name: release-train
+  pattern: '^(?P<train>\d{4}Q[1-4])\.(?P<hotfix>\d+)$'
+  comparisonGroups: [train, hotfix]
+```
+
+## Comparison and Ordering
+
+For any pair of versions, OCM selects the **first** scheme (in list order) whose `pattern` matches **both** versions and
```
matthiasbruns: This is pretty smart. Good job

matthiasbruns (commented):
So there is one thing I found

`ListComponentVersions` is unchanged, so once we accept non `semver` versions, the `repo.ListComponentVersions` does not return them anymore

I wrote a unit-test on repo level verifying it, as well as some testing util scripts to test this e2e 

https://github.com/matthiasbruns/open-component-model/blob/b27064f7f/review/pr-3635/repro.sh
https://github.com/matthiasbruns/open-component-model/blob/b27064f7f/review/pr-3635/README.md
https://github.com/matthiasbruns/open-component-model/blob/b27064f7f/bindings/go/oci/pr3635_roundtrip_test.go
https://github.com/matthiasbruns/open-component-model/blob/b27064f7f/bindings/go/runtime/versioning/pr3635_filter_test.go

This might be completely wrong but it feels like a gap

bindings/go/cli/internal/repository/ocm/repository.go
```diff
@@ -3,17 +3,29 @@ package ocm
...
+// configuration. When the config carries no versioning entry, the loose-semver
+// default is returned.
+func RegistryFromConfig(config *genericv1.Config) (*versioning.Registry, error) {
+	cfg, err := versioningspec.Lookup(config)
+	if err != nil {
+		return nil, fmt.Errorf("failed to look up versioning config: %w", err)
+	}
+	return cfg.Registry()
+}
+
```
fabianburth: Considering we have a lot of registries, this name is to generic in this package. Maybe we should move this function to `configuration/versioning/v1alpha1`

bindings/go/configuration/versioning/v1alpha1/spec/config.go
```diff
@@ -0,0 +1,206 @@
...
+	Pattern string `json:"pattern,omitempty"`
+
+	// ComparisonGroups lists the named capture groups from Pattern used to order
+	// versions, most significant first. Numeric groups compare as integers;
+	// non-numeric groups compare lexically. When empty, whole matched strings
+	// compare lexically. Must be empty when Builtin is set.
+	ComparisonGroups []string `json:"comparisonGroups,omitempty"`
+}
+
+// Lookup creates a new Config from a central V1 config.
```
fabianburth:
```suggestion
// Lookup creates a new Config from a central generic config.
```

bindings/go/cli/internal/repository/ocm/list.go
```diff
@@ -60,6 +70,13 @@ func WithSort() ComponentVersionsFilterOption {
 	}
 }
 
+// WithRegistry sets the versioning schemes used to filter and sort versions.
+func WithRegistry(registry *versioning.Registry) ComponentVersionsFilterOption {
+	return func(o *ComponentVersionsFilterOptions) {
+		o.registry = registry
+	}
+}
+
```
fabianburth: too broad of an option name
```suggestion
// WithRegistry sets the versioning schemes used to filter and sort versions.
func WithVersioningRegistry(registry *versioning.Registry) ComponentVersionsFilterOption {
	return func(o *ComponentVersionsFilterOptions) {
		o.registry = registry
	}
}

```

bindings/go/constructor/construct.go
```diff
@@ -513,6 +548,12 @@ func (c *DefaultConstructor) processSource(ctx context.Context, targetRepo Targe
 		res, err = c.processSourceWithInput(ctx, targetRepo, src, component, version)
 	} else {
 		logger.Debug("processing source with existing access")
+		// Sources with existing access may omit their version; default it to the
+		// component version so validateVersions does not reject an otherwise
+		// schema-valid source for an empty version. Explicit versions are kept.
+		if src.Version == "" {
+			src.Version = version
+		}
```
fabianburth, line 556: why did we add this now? is this defaulting valid?

bindings/go/kubernetes/controller/internal/controller/component/component_controller.go
```diff
@@ -398,11 +398,15 @@ func (r *Reconciler) reconcileDelete(ctx context.Context, component *v1alpha1.Co
 func (r *Reconciler) DetermineEffectiveVersionFromRepo(ctx context.Context, component *v1alpha1.Component,
 	repo repository.ComponentVersionRepository,
 ) (string, error) {
+	// The controller resolves versions with the loose-semver default; Spec.Semver
+	// is a semver constraint by CRD contract.
+	registry := versioning.Default()
```
fabianburth, line 403: are you planning to do a follow up PR for this? this feels like something we have to change for controller ga

bindings/go/oci/spec/transformation/v1alpha1/schemas/OCIGetComponentVersionOutput.schema.json
```diff
@@ -481,13 +481,13 @@
...
-          "pattern": "^[v]?(0|[1-9]\\d*)(?:\\.(0|[1-9]\\d*))?(?:\\.(0|[1-9]\\d*))?(?:-((?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\\.(?:0|[1-9]\\d*|\\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\\+([0-9a-zA-Z-]+(?:\\.[0-9a-zA-Z-]+)*))?$",
+          "description": "Component version string. Structural validation only requires a non-empty string; the concrete versioning scheme (semver by default, or a configured scheme such as calver) is validated separately at runtime.",
+          "pattern": "^.+$",
           "examples": [
             "v1.0.0",
-            "v1.0",
             "1.0.0",
-            "1.0"
+            "2024.03.15",
+            "22.04"
```
fabianburth: this still being called relaxed semver feels off

bindings/go/oci/repository.go
```diff
@@ -164,8 +171,15 @@ func (repo *Repository) ListComponentVersions(ctx context.Context, component str
 		return nil, fmt.Errorf("failed to create lister: %w", err)
 	}
 
+	// The OCI binding is versioning-config-agnostic. We use the loose-semver
+	// default comparator (via the lister's Comparator hook) rather than
+	// SortPolicyLooseSemverDescending: the comparator preserves versions that are
+	// not loose semver (e.g. calver or build numbers) instead of dropping them,
+	// so a configured, non-semver history survives listing. The CLI layer then
+	// re-sorts the returned versions with the configured versioning registry.
+	defaultRegistry := versioning.Default()
```
fabianburth, line 180: why is oci version config agnostic?

bindings/go/runtime/versioning/versioning.go
```diff
@@ -0,0 +1,559 @@
...
+//
+// The constraint is interpreted by each version's authoritative scheme, not
+// globally as semver: loose-semver versions use semver range syntax (">=1.0.0",
+// "^1.2"), while a regex scheme uses its own relational grammar (">=2024.03.15
+// <2024.10.01"). A version no scheme claims is retained unchanged, so a
+// constraint never silently discards an unrecognized history.
+//
+// An empty constraint returns the input unchanged. A constraint no registered
+// scheme can parse is malformed and returns an error.
+func (r *Registry) Filter(versions []string, constraint string) ([]string, error) {
```
fabianburth, line 235: I believe the lenient semantics of 

> A version no scheme claims is retained unchanged, so a constraint never silently discards an unrecognized history.

might not always be desired. For example during
`ocm transfer cv --semver-constraint`. We might need a strict mode here.

bindings/go/repository/component/pathmatcher/v1alpha1/spec_provider.go
```diff
@@ -6,23 +6,23 @@ import (
...
 // compiledResolver holds a resolver together with its pre-compiled glob pattern
-// and optional semver constraint, so they are validated once at construction time
-// instead of on every call.
+// and optional version constraint, so they are validated once at construction
+// time instead of on every call.
 type compiledResolver struct {
 	resolver             *resolverspec.Resolver
 	componentNamePattern glob.Glob
-	versionConstraint    *semver.Constraints // nil when no version constraint is set
+	hasConstraint        bool // true when a non-empty version constraint is set
```
fabianburth: This flag feels off. Since `ValidateConstraint` is a no-op with an empty constraint, couldn't we simply omit it?

bindings/go/constructor/construct.go
```diff
@@ -230,13 +231,47 @@ func (c *DefaultConstructor) constructComponent(ctx context.Context, component *
 		return nil, fmt.Errorf("component %q failed validation: %w", component.Name, err)
 	}
 
+	if err := c.validateVersions(desc); err != nil {
+		return nil, fmt.Errorf("component %q failed version validation: %w", component.Name, err)
+	}
+
```
fabianburth, line 237: shouldnt `c.validateVersions` (and probably also `descriptor.Validate`) happen before `c.processDescriptor`?

bindings/go/runtime/versioning/versioning.go
```diff
@@ -0,0 +1,559 @@
...
+
+// Valid reports whether any registered scheme considers the version well-formed.
+func (r *Registry) Valid(version string) bool {
+	for _, s := range r.schemes {
+		if s.Matches(version) && s.Valid(version) {
+			return true
+		}
+	}
+	return false
+}
```
fabianburth, line 203: I believe this being inconsistent with `schemeFor` and `rankFor` is probably undesired. I guess we should either
 ```go
   for _, scheme := range r.schemes {
      if scheme.Matches(version) {
         return scheme.Valid(version)
      }
   }
```

or check valid in the other functions too

File: bindings/go/cli/internal/repository/ocm/repository.go [hunk 13]
```diff
@@ -63,19 +75,11 @@ func GetComponentVersions(ctx context.Context, opts GetComponentVersionsOptions,
...
-			return 0
-		}
-		semverVersionB, err := semver.NewVersion(b.Component.Version)
-		if err != nil {
-			slog.ErrorContext(ctx, "failed parsing version, this may result in wrong ordering", "version", b.Component.Version, "error", err)
-			return 0
-		}
-		return semverVersionB.Compare(semverVersionA)
+		c, _ := reg.Compare(b.Component.Version, a.Component.Version)
+		return c
```
fabianburth: We should probably surface this error.

bindings/go/cli/internal/repository/ocm/list.go
```diff
@@ -133,54 +149,12 @@ func ListComponentVersions(ctx context.Context, repo repository.ComponentVersion
 
 	// Ensure deterministic global ordering across components and versions.
 	if options.sort {
-		return sortDescriptorsBySemver(result)
+		reg := options.reg()
+		slices.SortFunc(result, func(a, b *descriptor.Descriptor) int {
+			c, _ := reg.Compare(b.Component.Version, a.Component.Version)
+			return c
+		})
```
fabianburth, line 164: We should probably surface this error.

bindings/go/runtime/versioning/versioning.go
```diff
@@ -0,0 +1,559 @@
...
+			return i, s
+		}
+	}
+	return len(r.schemes), nil
+}
+
+// looseSemverScheme is the default scheme, wrapping github.com/Masterminds/semver/v3.
+type looseSemverScheme struct{}
+
+func newLooseSemverScheme() *looseSemverScheme { return &looseSemverScheme{} }
```
fabianburth: We wrap and expose exactly this as a public function too. Can be deduplicated.

bindings/go/runtime/versioning/versioning.go
```diff
@@ -0,0 +1,559 @@
...
+
+func (looseSemverScheme) Name() string { return "loose-semver" }
+
+func (looseSemverScheme) Matches(version string) bool {
+	_, err := semver.NewVersion(version)
+	return err == nil
+}
```
fabianburth, line 353: Since we now removed the semver hardcoded in the descriptor scheme, I believe this becomes more relevant:

https://github.com/masterminds/semver
> CoerceNewVersion is true by default. When set to true it coerces non-compliant versions into SemVer. For example, allowing a leading 0 in a major, minor, or patch part. This enables the use of CalVer in versions even when not compliant with SemVer. When set to false less coercion work is done.
DetailedNewVersionErrors provides more detailed errors. It only has an affect when CoerceNewVersion is set to false. When DetailedNewVersionErrors is set to true it can provide some more insight into why a version is invalid. Setting DetailedNewVersionErrors to false is faster on performance but provides less detailed error messages if a version fails to parse.

I believe for us, that could mean that our evaluations say
   01.2.3  → 1.2.3
   1.02.3  → 1.2.3
but they create distinct OCI tags.

bindings/go/cli/cmd/get/component-version/cmd.go
fabianburth: I believe we have to rename this flag
```suggestion
	FlagConstraint = "constraint"
```

(I think I read somewhere in the documentation that you called it `--constraint`, otherwise, I would have probably suggested `versioning-constraint`)

bindings/go/configuration/versioning/v1alpha1/spec/config.go
```diff
@@ -0,0 +1,206 @@
...
+	// Builtin selects a named built-in scheme instead of a regular expression.
+	// Supported values: "loose-semver" (the historical default), "calver-full"
+	// (YYYY.MM.DD), "calver-month" (YYYY.MM), "calver-ubuntu" (YY.MM),
+	// "calver-micro" (YYYY.M(M).PATCH), "aws-date" (YYYY-MM-DD), and
+	// "build-number" (monotonic integer). Each is exactly the equivalent regex
+	// pattern plus comparison groups, documented in the versioning configuration
+	// reference. When set, Pattern and ComparisonGroups must be empty. Add an
+	// entry with builtin: loose-semver (typically last) to keep recognizing semver
+	// versions alongside custom schemes.
+	Builtin string `json:"builtin,omitempty"`
```
fabianburth, line 86: our json schema generator supports enums, right? we could probably use that here

bindings/go/configuration/versioning/v1alpha1/spec/config.go
```diff
@@ -0,0 +1,206 @@
...
+	// Schemes defines an ordered list of version schemes.
+	//
+	// The list is evaluated in order: for any version the first scheme that
+	// claims it wins. Unlike the historical default, the built-in loose-semver
+	// scheme is NOT appended automatically once schemes are configured: only the
+	// listed schemes apply. To keep recognizing semver versions, add an explicit
+	// entry with builtin: loose-semver (typically last, as a fallback). List
+	// more specific schemes before broader ones. When no config is present at
+	// all, OCM still uses loose semver (see [versioning.Default]).
+	Schemes []*VersionScheme `json:"schemes,omitempty"`
```
fabianburth:
```suggestion
	Schemes []VersionScheme `json:"schemes,omitempty"`
```
nit: not sure whether there is an advantage having it as pointers and it would improve the accuracy of the json schema validation as it wouldnt allow null values

bindings/go/constructor/construct.go
jakobmoellerdev: we did this in ocmv1 and we also do it for resources i believe

bindings/go/kubernetes/controller/internal/controller/component/component_controller.go
jakobmoellerdev: I believe we have problems in the APIs and will get breaking changes here...

bindings/go/oci/repository.go
jakobmoellerdev: sorry dont understand that question, can you elaborate?

bindings/go/runtime/versioning/versioning.go
jakobmoellerdev: hmmm im wondering about this, will think about a fallback matcher one can always use

bindings/go/repository/component/pathmatcher/v1alpha1/spec_provider.go
jakobmoellerdev: +1, like that idea

bindings/go/constructor/construct.go
jakobmoellerdev: now that i look at this, should the same then be the case for descriptor.Validate?

bindings/go/configuration/versioning/v1alpha1/spec/config.go
jakobmoellerdev: +1 for builtin

bindings/go/configuration/versioning/v1alpha1/spec/config.go
jakobmoellerdev: Done in 21ab634 — Schemes is now []VersionScheme (value slice); the JSON schema no longer permits null entries and the nil-guard in Registry() was removed.

bindings/go/cli/internal/repository/ocm/repository.go
jakobmoellerdev: Done in 21ab634 — moved this to configuration/versioning/v1alpha1/spec as spec.RegistryFromConfig and migrated all callers, so it no longer lives under the generic cli repository package.

bindings/go/configuration/versioning/v1alpha1/spec/config.go
jakobmoellerdev: Done in 21ab634 — updated the doc comment to "Lookup creates a new Config from a central generic config."

bindings/go/cli/internal/repository/ocm/list.go
jakobmoellerdev: Done in 21ab634 — renamed to WithVersioningRegistry (with the suggested doc comment) and updated the caller.

bindings/go/oci/spec/transformation/v1alpha1/schemas/OCIGetComponentVersionOutput.schema.json
jakobmoellerdev: Done in 21ab634 — renamed the schema def relaxedSemver to version (validity is now scheme-driven, not semver) across the descriptor v2 and constructor source schemas, and regenerated the OCI transformation schemas.

bindings/go/repository/component/pathmatcher/v1alpha1/spec_provider.go
jakobmoellerdev: Done in 21ab634 — dropped compiledResolver.hasConstraint and gate directly on the resolver's own VersionConstraint != "". The construction-time ValidateConstraint guard is kept only to avoid wrapping a no-op error.

bindings/go/runtime/versioning/versioning.go
jakobmoellerdev: Done in 21ab634 — Registry.Valid now resolves the first matching scheme and returns its Valid(), consistent with schemeFor/rankFor first-match semantics.

bindings/go/cli/internal/repository/ocm/repository.go
jakobmoellerdev: Done in 21ab634 — GetComponentVersions now captures the reg.Compare error from the sort callback and returns it instead of discarding it.

bindings/go/cli/internal/repository/ocm/list.go
jakobmoellerdev: Done in 21ab634 — ListComponentVersions now captures the reg.Compare error from the final sort callback and returns it instead of discarding it.

bindings/go/runtime/versioning/versioning.go
jakobmoellerdev: Done in 21ab634 — collapsed the redundant private newLooseSemverScheme into the exported NewLooseSemverScheme.

bindings/go/cli/cmd/get/component-version/cmd.go
jakobmoellerdev: Done in 21ab634 — renamed the flag to --constraint (FlagConstraint) for both get and transfer, updated tests, integration test, and regenerated the CLI reference docs.

bindings/go/configuration/versioning/v1alpha1/spec/config.go
jakobmoellerdev: Done in 21ab634 — added +ocm:jsonschema-gen:enum with the seven builtin names; the generated VersionScheme schema now enumerates builtin values.

website/content/docs/reference/versioning-configuration.md
```diff
@@ -0,0 +1,201 @@
...
+For OCI registries, a component version is stored as an **OCI tag**. OCI tags follow the grammar
+`^[\w][\w.-]{0,127}$`: they may contain only alphanumerics, `_`, `.`, and `-`, must start with an alphanumeric or
+underscore, and may be at most 128 characters.
+
+Most calendar and build-number schemes are already valid tags (`2024.03.15`, `22.04`, `2024-03-15`, `1837`). Choose a
+scheme whose versions satisfy the tag grammar. Versions containing characters such as `+`, `:`, `/`, `~`, or spaces are
+not valid tags:
+
+- A `+` (semver build metadata) is rewritten to `.build-` for the tag. This rewrite is not reversed when listing, so the
+  version read back differs from the original — avoid `+` in versions destined for OCI registries.
```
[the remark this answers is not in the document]
jakobmoellerdev: Both pages already state the current behavior consistently: VersionToOCITag rewrites the first + to .build- (lossy, not reversed on read), and any other invalid character fails add cv at publish. Verified against oci/semver_tag.go. Resolving.

bindings/go/constructor/construct.go
jakobmoellerdev: As noted above: this defaulting mirrors what we already do in ocmv1 and for resources, so it is intentional and valid here. Resolving.

bindings/go/kubernetes/controller/internal/controller/component/component_controller.go
jakobmoellerdev: As discussed: the controller APIs will need breaking changes for GA and this belongs in that follow-up, not this PR. Tracking it there; resolving here.

bindings/go/constructor/construct.go
jakobmoellerdev: It has to stay after processDescriptor: processDescriptor defaults omitted resource/source versions to the component version (the fix from the earlier review comment), and descriptor.Validate needs the populated access/blobs. Validating before processing would reject the schema-valid elements whose versions are only defaulted during processing. So both validations intentionally run on the processed descriptor. Resolving.

bindings/go/oci/repository.go
jakobmoellerdev: The rationale is in the comment just above (lines 174-179): the OCI binding stays config-agnostic and uses the loose-semver default comparator via the lister hook, which preserves non-semver tags (calver, build numbers) instead of dropping them; the CLI layer then re-sorts with the configured registry. Keeping the versioning config out of the OCI binding avoids threading it through every store/repository constructor. Resolving; happy to reopen if you want the config pushed down.

bindings/go/runtime/versioning/versioning.go
jakobmoellerdev: Agreed the lenient retain-unknown semantics is not always desired (e.g. transfer cv --constraint should be able to hard-fail unrecognized histories). This needs a dedicated strict/fallback-matcher mode rather than changing the default retain behavior in this PR. Tracking as a follow-up so the additive, behavior-preserving default here stays intact. Resolving this thread.

bindings/go/runtime/versioning/versioning.go
jakobmoellerdev: Good catch — with the semver regex removed from the descriptor scheme, Masterminds' CoerceNewVersion means 01.2.3 and 1.2.3 compare equal yet map to distinct OCI tags. Flipping CoerceNewVersion=false changes matching semantics globally, so it deserves its own change with test coverage rather than riding along here. Tracking as a follow-up; resolving this thread.

bindings/go/runtime/versioning/versioning.go
```diff
@@ -0,0 +1,560 @@
...
+// constraintOperators lists the relational operators a regex scheme constraint
+// term may start with, longest first so the two-character operators win over
+// their single-character prefixes.
+var constraintOperators = []string{">=", "<=", "!=", "==", "=", ">", "<"}
+
+// splitConstraintTerms splits a constraint expression into conjunctive terms on
+// commas and runs of whitespace, dropping empty terms.
+func splitConstraintTerms(constraint string) []string {
+	return strings.FieldsFunc(constraint, func(r rune) bool {
+		return r == ',' || r == ' ' || r == '\t' || r == '\n' || r == '\r'
```
Skarlso: `r == ' '` -> `>= 2024.03.15` fails on this because of the space between the >= and the version.

bindings/go/constructor/construct.go
```diff
@@ -230,13 +231,47 @@ func (c *DefaultConstructor) constructComponent(ctx context.Context, component *
 		return nil, fmt.Errorf("component %q failed validation: %w", component.Name, err)
 	}
 
+	if err := c.validateVersions(desc); err != nil {
```
Skarlso: This is too late in a step, no? The push already happened at this point. 🤔 The validation should happen earlier.

bindings/go/descriptor/v2/validate.go
```diff
@@ -90,3 +93,34 @@ func ValidateRawYAML(raw []byte) error {
 
 	return schema.Validate(mm)
 }
+
+// ValidateVersions checks the component version and every resource, source, and
+// reference version against the given versioning registry. It is separate from
+// [Validate] (which only performs structural JSON-schema validation) because
+// version-scheme validity is configuration-driven and cannot be expressed in the
+// static schema. It returns a joined error naming each offending field.
+func ValidateVersions(desc *Descriptor, registry *versioning.Registry) error {
```
Skarlso: Err... I don't see any callers for this? Am I blind? :D

fabianburth (dismissed):

Skarlso (commented):
Partial review.. will continue later on.

fabianburth (dismissed):

matthiasbruns (commented):
.

bindings/go/runtime/versioning/versioning.go
```diff
@@ -0,0 +1,590 @@
...
+// decimal integers and lexically otherwise. Numeric comparison uses arbitrary
+// precision so arbitrarily large groups (e.g. long build numbers) never overflow
+// a machine integer and fall back to lexical ordering.
+func compareGroup(a, b string) int {
+	na, aok := new(big.Int).SetString(a, 10)
+	nb, bok := new(big.Int).SetString(b, 10)
+	if aok && bok {
+		return na.Cmp(nb)
+	}
+	return strings.Compare(a, b)
```
matthiasbruns: **Comparison is not transitive for mixed numeric/text captures.**

The pair-dependent switch between numeric and lexical comparison breaks transitivity within a single configured scheme. With `^(?P<value>[0-9a-z]+)$` and `comparisonGroups: [value]`, we get `2 < 10 < 1a < 2`.

This conflicts with the documented total, deterministic ordering. The CLI selects `1a` from `[10, 1a]`, but selects `10` after adding `2`. Could we define a consistent ordering between numeric and nonnumeric captures, or a fixed comparison mode per group? This is unrelated to DAG discovery.

[API reproducer](https://github.com/matthiasbruns/open-component-model/blob/6bad2375f/review-pr3635-comparison-cycle.sh) · [CLI reproducer](https://github.com/matthiasbruns/open-component-model/blob/6bad2375f/review-pr3635-cli-ordering.sh)

bindings/go/runtime/versioning/versioning.go
```diff
@@ -0,0 +1,590 @@
...
+			// Skip the operator (if any) and the whitespace padding after it.
+			rest = strings.TrimLeft(rest[len(op):], " \t\n\r")
+			// The operand runs up to the next whitespace-separated token.
+			operand := rest
+			if i := strings.IndexAny(rest, " \t\n\r"); i >= 0 {
+				operand, rest = rest[:i], strings.TrimLeft(rest[i:], " \t\n\r")
+			} else {
+				rest = ""
+			}
+			if operand == "" {
```
matthiasbruns: **Missing constraint operands are silently discarded.**

An operator without an operand is dropped here. Consequently, `>=` and `<` match every version, while `>=2024.03.15 <` behaves like `>=2024.03.15`.

I reproduced this in version listings and transfer dry-run graphs. Complete bounds filter correctly; incomplete expressions select versions instead of failing. This is distinct from intentional foreign-history retention: an incomplete predicate in the scheme's own relational grammar is discarded. Could we reject missing operands rather than treating the remaining expression as satisfied?

[Go tests](https://github.com/matthiasbruns/open-component-model/blob/6bad2375f/bindings/go/cli/internal/repository/ocm/review_pr3635_test.go) · [CLI reproducer](https://github.com/matthiasbruns/open-component-model/blob/6bad2375f/review-pr3635-malformed-constraints.sh)

bindings/go/runtime/versioning/versioning.go
```diff
@@ -0,0 +1,590 @@
...
+// Satisfies evaluates a semver range constraint (e.g. ">=1.0.0 <2.0.0", "^1.2")
+// against a semver version. A version that does not parse as semver never
+// satisfies a constraint. A malformed constraint returns an error.
+func (looseSemverScheme) Satisfies(version, constraint string) (bool, error) {
+	if constraint == "" {
+		return true, nil
+	}
+	constraints, err := semver.NewConstraint(constraint)
+	if err != nil {
+		return false, fmt.Errorf("parsing semantic version constraint failed: %w", err)
```
matthiasbruns: **Foreign-constraint handling is asymmetric.**

With a `build-N` scheme and an explicit `loose-semver` fallback, `>=build-100` passes registry validation and works on custom-only history. Adding a semver entry makes the same query fail with `parsing semantic version constraint failed: improper constraint: ">=build-100"`.

The documented policy retains versions whose schemes cannot interpret a constraint. The reverse direction already works: semver constraints retain custom-version history. Could we distinguish a valid foreign-scheme constraint from a genuinely malformed constraint? Under the intended lenient policy, this query should filter the build versions and retain semver entries, rather than abort.

[Go tests](https://github.com/matthiasbruns/open-component-model/blob/6bad2375f/bindings/go/cli/internal/repository/ocm/review_pr3635_test.go) · [CLI reproducer](https://github.com/matthiasbruns/open-component-model/blob/6bad2375f/review-pr3635-mixed-constraints.sh)

bindings/go/configuration/versioning/v1alpha1/spec/config.go
```diff
@@ -0,0 +1,214 @@
...
+			continue
+		}
+		pattern, err := regexp.Compile(s.Pattern)
+		if err != nil {
+			return nil, fmt.Errorf("versioning scheme %q (index %d): invalid pattern: %w", s.Name, i, err)
+		}
+		names := make(map[string]struct{}, len(pattern.SubexpNames()))
+		for _, name := range pattern.SubexpNames() {
+			if name != "" {
+				names[name] = struct{}{}
```
matthiasbruns: **Accepted duplicate captures silently lose their values.**

The configuration `pattern: '^(?:v(?P<n>\d+)|r(?P<n>\d+))$'` with `comparisonGroups: [n]` is accepted. During extraction, the unmatched second capture overwrites the first with an empty string. As a result, `v2` and `v10` compare equal, and the CLI includes `v2` when filtering with `>=v10`.

Could we reject duplicate comparison-group names during configuration validation, unless their resolution is explicitly supported? Rejecting this ambiguous configuration is sufficient; duplicate-name support does not need to be added.

[Go test](https://github.com/matthiasbruns/open-component-model/blob/6bad2375f/bindings/go/configuration/versioning/v1alpha1/spec/review_pr3635_5d3e672f_test.go) · [CLI reproducer](https://github.com/matthiasbruns/open-component-model/blob/6bad2375f/review-pr3635-duplicate-captures.sh)

website/content/docs/tutorials/configure-versioning.md
```diff
@@ -0,0 +1,161 @@
...
+
+### Confirm the active scheme
+
+Print the effective merged configuration to verify the CalVer scheme is loaded:
+
+```bash
+ocm --config versioning.ocmconfig get config
+```
+
+The output includes your `versioning.config.ocm.software/v1alpha1` entry.
```
matthiasbruns: **The configuration-verification step does not show versioning.**

This step does not currently show the versioning configuration. A configured non-semver version can be published and retrieved with its custom constraint, proving the scheme is active, but `get config --output json` omits the versioning entry.

Could we include the new type in effective-config output, or adjust this verification step to match what the command exposes? This is an integration/documentation gap, not evidence that the configuration is ignored.

[CLI command test](https://github.com/matthiasbruns/open-component-model/blob/6bad2375f/bindings/go/cli/cmd/get/config/review_pr3635_5d3e672f_effective_versioning_test.go) · [CLI reproducer](https://github.com/matthiasbruns/open-component-model/blob/6bad2375f/review-pr3635-effective-config.sh)

matthiasbruns (commented):
Five verified findings attached to the relevant diff lines, with links to the tests and isolated reproducers. Submitted as comments, not a request for changes.

bindings/go/constructor/construct.go
jakobmoellerdev: The component *version* is now validated before any content is pushed: `validateComponentVersion` runs at construct.go:231, ahead of `processDescriptor` (which uploads resource/source blobs). So a genuinely invalid component version fails fast with no upload.

The `validateVersions` call you flagged at 243 covers *resource, source, and reference* versions specifically. Those are defaulted during `processDescriptor` (an input resource inherits the component version, etc.), so their final values do not exist until processing has run — hence the validation must follow it. It still runs before `repo.AddComponentVersion`, so the component version itself is never published with an invalid element version. The comment at 226-230 documents this split. Validating element versions before upload would require defaulting them ahead of blob processing, a larger restructuring outside this PR's scope.

bindings/go/runtime/versioning/versioning.go
jakobmoellerdev: Fixed in 6b7e5f0df. `parseConstraintTerms` now returns an error when a relational operator has no operand (e.g. a trailing `>=` or a bare `<`), propagated through `regexScheme.Satisfies`/`ValidateConstraint`, so an incomplete predicate is rejected instead of silently satisfied. Covered by `TestVersioning_ConstraintOperatorWithoutOperandRejected`.

bindings/go/runtime/versioning/versioning.go
jakobmoellerdev: Fixed in 6b7e5f0df. `Registry.Filter`/`Satisfies` now validate the constraint once against the whole registry, then treat any per-version scheme error on a registry-valid constraint as a foreign grammar: Filter retains the version, Satisfies returns false. So `>=build-100` filters the build history and retains semver entries symmetrically, while a constraint no scheme can parse (e.g. `not-a-constraint`) still errors. Covered by `TestVersioning_ForeignConstraintHandlingIsSymmetric`.

PR #3636, opened 2026-09-17
feat: add RFC 3161 TSA timestamping for OCM signatures

jakobmoellerdev:

Add back optional RFC 3161 timestamping to OCM signing/verification so that signatures can be validated after the signing certificate expires, as long as the timestamp proves the signature was created while the certificate was valid. This was present in v1 OCM but is not yet in v2, and brings traditional PKI signing closer to sigstore.

This **refreshes and supersedes #2154**: rebased onto the current `main`, ported from the old multi-module layout to the single `bindings/go` module (the CLI now lives under `bindings/go/cli`), and adapted to the new typed-credential signing/verification architecture. All CodeRabbit review comments from #2154 are addressed.

Implementation + ADR for https://github.com/open-component-model/ocm-project/issues/568.

#### Summary of changes

| Scope | Description |
|-------|-------------|
| `bindings/go/descriptor` | Add `TimestampSpec` type and optional `Timestamp` field to `Signature` (runtime + v2), conversion + deep copy |
| `bindings/go/signing/tsa` | New RFC 3161 TSA client: ASN.1 types, `RequestTimestamp()` (now cryptographically validates the returned token), `Verify()`, PEM encode/decode, credential-graph integration, URL redaction in errors; full unit tests |
| `bindings/go/rsa` | Use TSA-verified time for X.509 chain validation, threaded through a typed `DirectCredentials` helper (`VerifiedTimeFromCredentials`) rather than a raw map |
| `bindings/go/cli` (`sign`/`verify`) | `--tsa` / `--tsa-url` flags on sign; automatic TSA verification on verify with URL-specific credential-graph lookup; integration tests |
| `docs/adr/0015` | ADR: trust model, threat analysis, verification levels, credential-graph integration |

#### Changes vs. #2154 (review feedback addressed)

- **`RequestTimestamp` now cryptographically validates** the returned `TimeStampToken` (PKCS#7 `Verify`) before exposing it, instead of only parsing TSTInfo.
- **TSA URL is redacted** (userinfo/query/fragment stripped) in all error messages.
- **Default TSA URL uses HTTPS** (`https://timestamp.digicert.com`).
- **`json.Marshal`/`json.Unmarshal` errors are handled** (signed label marshal returns an error; verify logs unmarshal failures at debug).
- **Credentials error message** names the actual PEM source (`root_certs_pem` vs `root_certs_pem_file`).
- **Integration tests**: the generic-vs-URL credential cases now use distinct configs (generic identity has no URL attributes), and the dry-run case asserts **no TSA network request** via a request counter.
- **ADR markdown**: fenced code blocks have language specifiers.

#### Trust model (see [ADR 0030](docs/adr/0030_rfc3161_timestamping.md))

- Signer chooses the TSA via `--tsa-url`; the URL is stored as a **signed label** (`url.tsa.ocm.software/{sig}`) before digest computation, making it tamper-evident.
- TSA counter-signs the digest with a PKCS#7 timestamp token proving *when* the signature was created.
- Verifier controls trust: TSA root certificates come exclusively from the credential graph (`TSA/v1alpha1` identity with URL-specific matching).
- Graceful degradation: without TSA root certs, structural verification still runs with a warning.
- Expired-cert support: verified TSA time is passed to the RSA handler, which uses it instead of `time.Now()` for X.509 chain validation.

#### Testing

- `go test ./bindings/go/signing/tsa/...`
- `go test ./bindings/go/cli/integration/ -run Test_Integration_Signing_TSA` (requires Docker)

> [!NOTE]
> Opened as a **draft** refresh of #2154 for continued evaluation/discussion; no urgency to merge.


bindings/go/cli/cmd/sign/component-version/cmd.go
```diff
@@ -389,10 +415,26 @@ func SignComponentVersion(cmd *cobra.Command, args []string) error {
...
+	// TSA timestamp (optional). Requested after signing so it can cover the same
+	// digest. Never performed on a dry run (useTSA already excludes dryRun).
+	var tsSpec *descruntime.TimestampSpec
+	if useTSA {
+		httpConfig, err := httpv1alpha1.ResolveHTTPConfig(config)
+		if err != nil {
+			return fmt.Errorf("resolving HTTP configuration for TSA request failed: %w", err)
+		}
+		tsaClient := ocmhttp.New(ocmhttp.WithConfig(httpConfig))
+		tsSpec, err = requestTSATimestamp(ctx, logger, tsaClient, tsaURL, unsignedDigest)
```
matthiasbruns: _LLM-generated after reviewing and testing this PR at `d276ed06d`._

**[P1] Timestamp the signature, not only the descriptor digest**

Passing `unsignedDigest` to the TSA proves that the descriptor existed at the timestamped time, but does not prove that this signature existed. Verification nevertheless uses that time to validate the signing certificate historically.

A temporary regression test reproduced this by creating a trusted timestamp while a short-lived signing certificate was valid, waiting for the certificate to expire, and only then creating the component signature. The actual CLI timestamp-verification helpers and RSA handler rejected the signature without the timestamp override but accepted it with the earlier token. A second, different RSA-PSS signature was also accepted with the same token.

This requires the signing key and a historical token for the same descriptor, but remains possible with correctly configured TSA roots.

For certificate-lifetime extension, could we timestamp a defined cryptographic representation of the actual signature and verify that binding? Otherwise, this timestamp should only establish descriptor existence, not signature creation time.

bindings/go/cli/cmd/verify/component-version/cmd.go
```diff
@@ -375,6 +404,17 @@ func VerifyComponentVersion(cmd *cobra.Command, args []string) error {
...
 
+			// Verify an optional RFC 3161 timestamp attached to the signature.
+			if signature.Timestamp != nil {
+				verifiedTime, err := verifyTSATimestamp(egctx, logger, credentialGraph, desc, signature)
+				if err != nil {
+					return err
+				}
+				// Pass the verified time to the handler so it can validate the
+				// signing certificate chain as of the signing time rather than now.
+				creds = withVerifiedTime(creds, verifiedTime)
```
matthiasbruns: _LLM-generated after reviewing and testing this PR at `d276ed06d`._

**[P1] Only trusted TSA verification should override certificate-validation time**

`verifyTSATimestamp` can succeed without TSA roots: in that case, `tsa.Verify` checks the token against its embedded certificate without establishing trust in the TSA. This line still forwards the returned time to RSA certificate validation.

A temporary regression test reproduced a newly created signature under an expired, otherwise trusted signing certificate being accepted after attaching a backdated token from a self-signed TSA. Without the timestamp override, RSA verification rejected the expired certificate; with explicit TSA roots, the token was correctly rejected as untrusted. The test obtained the time through the actual `verifyTSATimestamp` helper rather than directly injecting the time property.

Could we distinguish structural verification from trusted timestamp verification and only apply the time override for the latter? A warning is not sufficient when an untrusted timestamp changes the signature acceptance result.

bindings/go/signing/tsa/tsa.go
```diff
@@ -0,0 +1,228 @@
...
+		return time.Time{}, err
+	}
+
+	p7, err := pkcs7.Parse(raw)
+	if err != nil {
+		return time.Time{}, fmt.Errorf("tsa: parsing PKCS#7 timestamp token: %w", err)
+	}
+
+	if roots != nil {
+		if err := p7.VerifyWithChain(roots); err != nil {
```
matthiasbruns: _LLM-generated after reviewing and testing this PR at `d276ed06d`._

**[P1] Enforce the TSA signer timestamping EKU**

The pinned PKCS#7 dependency defaults this chain-verification path to `ExtKeyUsageAny`, so chaining to a configured root is enough even when the signing certificate is not authorized for timestamping.

A temporary regression test verified that a timestamp signed by a TLS-server-auth-only leaf is accepted here with its issuing CA trusted. Explicit X.509 verification requiring `ExtKeyUsageTimeStamping` rejects the same leaf, and an unrelated root correctly rejects the token.

Could we verify the chain explicitly for timestamping and check that the signer EKU extension is critical and contains only `id-kp-timeStamping`, as RFC 3161 requires? Explicit timestamping chain verification alone does not enforce the critical/exclusive EKU requirement.

bindings/go/cli/cmd/verify/component-version/cmd.go
```diff
@@ -405,3 +445,97 @@ func loadVerifierConfig(config *genericv1.Config, signatureName string, logger *
...
+// as a DirectCredentials property, preserving any properties already present.
+// The RSA verification handler reads this property to validate the signing
+// certificate chain as of the signing time instead of the current time.
+func withVerifiedTime(creds runtime.Typed, verifiedTime time.Time) runtime.Typed {
+	properties := map[string]string{}
+	for k, v := range directCredentialProperties(creds) {
+		properties[k] = v
+	}
+	properties[tsa.VerifiedTimeKey] = verifiedTime.Format(time.RFC3339)
+	return &credconfigv1.DirectCredentials{
```
matthiasbruns: _LLM-generated after reviewing and testing this PR at `d276ed06d`._

**[P1] Preserve typed verification credentials when adding the timestamp time**

`directCredentialProperties` returns nil for typed credentials such as `*RSACredentials`, so this replaces them with a `DirectCredentials` containing only `tsa_verified_time`. The credential graph supports and returns those typed credentials.

Temporary regression tests reproduced plain RSA verification succeeding before wrapping and failing afterward with `missing public key`; PEM verification with an explicit private root similarly changed from success to `unknown authority`. The legacy `DirectCredentials` controls continued to pass.

Dropping an explicit root can also weaken trust policy: the RSA handler falls back to its configured roots when the credential anchor disappears. A controlled test confirmed that a chain rejected under explicit anchor A became accepted through fallback root B after wrapping. That experiment injected a private fallback root into the handler; it demonstrates the policy substitution, not an actual public-CA exploit.

Could we carry the verified time without discarding the original credential material or changing its trust policy? The current integration test uses legacy RSA credentials, so it misses this typed-credential path.

matthiasbruns (changes requested):
_LLM-generated after reviewing and testing this PR at `d276ed06d`._

Thanks for adding this, especially the credential integration and documentation. This LLM-assisted review identified four issues that need addressing before merge.

All four were verified with temporary regression tests and negative controls against `d276ed06d`. These exercised the actual CLI timestamp-verification helpers and RSA handler, plus direct TSA certificate-purpose checks; they were not full repository-backed CLI exploit tests. Existing affected-area tests and lint pass, but do not cover these cases. Temporary test files were removed and no production files were changed.

The main concern is that the implementation can accept signatures created after signing-certificate expiry: the timestamp covers the descriptor rather than the signature, and even an untrusted timestamp can override certificate-validation time. There are also issues with TSA certificate-purpose enforcement and preservation of typed verification credentials.

Requesting changes for the four inline comments.

bindings/go/cli/cmd/sign/component-version/cmd.go
jakobmoellerdev: Fixed in `153ea120a`: the token is now requested over the **signature value** (`sigBytes.Value`), not the descriptor digest, in `requestTSATimestamp`. Verification recomputes the imprint over `signature.Signature.Value`, so the timestamp attests to the signature's existence.

bindings/go/cli/cmd/verify/component-version/cmd.go
jakobmoellerdev: Fixed in `153ea120a`: `tsa.Verify` now returns a `trusted` flag; `verifyTSATimestamp` propagates it and the verify command applies `withVerifiedTime` **only when `trusted`**. A structural-only token logs a warning and never relaxes certificate validity.

bindings/go/signing/tsa/tsa.go
jakobmoellerdev: Fixed in `153ea120a`: `tsa.Verify` enforces the RFC 3161 signer profile — exactly one signer, critical `id-kp-timeStamping` as the sole EKU (`verifyTimestampingEKU`) — and validates the chain via `VerifyWithOpts` with `KeyUsages: [ExtKeyUsageTimeStamping]` at `GenTime`.

bindings/go/cli/cmd/verify/component-version/cmd.go
jakobmoellerdev: Fixed in `153ea120a`: `withVerifiedTime` now flattens the resolved typed credential via `credentialProperties` (JSON field names preserved) before adding `tsa_verified_time`, so RSA public keys / trust anchors survive re-wrapping instead of being dropped.

bindings/go/cli/cmd/sign/component-version/cmd.go
```diff
@@ -389,10 +414,28 @@ func SignComponentVersion(cmd *cobra.Command, args []string) error {
...
+	// digest) existed at the timestamped time. Never performed on a dry run
+	// (useTSA already excludes dryRun).
+	var tsSpec *descruntime.TimestampSpec
+	if useTSA {
+		httpConfig, err := httpv1alpha1.ResolveHTTPConfig(config)
+		if err != nil {
+			return fmt.Errorf("resolving HTTP configuration for TSA request failed: %w", err)
+		}
+		tsaClient := ocmhttp.New(ocmhttp.WithConfig(httpConfig))
+		tsSpec, err = requestTSATimestamp(ctx, logger, tsaClient, tsaURL, unsignedDigest.HashAlgorithm, []byte(sigBytes.Value))
```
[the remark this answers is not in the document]
jakobmoellerdev: Fixed in `cc2fd58`: the TSA client sets `CheckRedirect = tsa.RejectInsecureRedirect`, which rejects any redirect whose target scheme is not `https`. Plain-HTTP initial URLs for local TSA servers remain allowed.

bindings/go/cli/cmd/sign/component-version/cmd.go
```diff
@@ -472,3 +515,70 @@ func printSignature(cmd *cobra.Command, sig descruntime.Signature) error {
...
+		return l.Name == labelName
+	}); idx >= 0 {
+		desc.Component.Labels[idx] = label
+	} else {
+		desc.Component.Labels = append(desc.Component.Labels, label)
+	}
+	return nil
+}
+
+// requestTSATimestamp obtains an RFC 3161 timestamp over the signature value
```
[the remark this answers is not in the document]
jakobmoellerdev: Fixed in `cc2fd58`: `addSignedTSALabel` now stores `tsa.SanitizeURL(tsaURL)` (scheme/host/port/path only), dropping userinfo/query/fragment. Verification derives the credential identity from those components only, so nothing is lost.

bindings/go/cli/cmd/sign/component-version/cmd.go
```diff
@@ -472,3 +515,70 @@ func printSignature(cmd *cobra.Command, sig descruntime.Signature) error {
...
+		Value:   tsaURLJSON,
+		Signing: true,
+		Version: "v1alpha1",
+	}
+	if idx := slices.IndexFunc(desc.Component.Labels, func(l descruntime.Label) bool {
+		return l.Name == labelName
+	}); idx >= 0 {
+		desc.Component.Labels[idx] = label
+	} else {
+		desc.Component.Labels = append(desc.Component.Labels, label)
```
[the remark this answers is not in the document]
jakobmoellerdev: Addressed in `cc2fd58`: since the signing-relevant TSA label changes the normalized digest and foreign signatures cannot be re-signed here, signing now **refuses** with a clear error when a signature under another name already exists, rather than silently invalidating it. Covered by a new integration subtest.

bindings/go/signing/tsa/tsa.go
```diff
@@ -0,0 +1,297 @@
...
+}
+
+// RedactURL strips potentially sensitive components (userinfo, query, fragment)
+// from a URL so it can be safely included in error messages and logs. If the URL
+// cannot be parsed a fixed placeholder is returned, because a malformed URL may
+// still embed userinfo or query credentials that must never reach logs.
+func RedactURL(raw string) string {
+	u, err := url.Parse(raw)
+	if err != nil {
+		return "<invalid TSA URL>"
```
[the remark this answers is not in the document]
jakobmoellerdev: Fixed in `cc2fd58`: `RedactURL` now returns the placeholder when `u.Opaque != ""` (e.g. `user:pass@host`, where `url.Parse` keeps credentials in `Opaque` and `User` is nil). `SanitizeURL` rejects the same form.

bindings/go/signing/tsa/tsa.go
```diff
@@ -0,0 +1,334 @@
...
+	// Validate the signer chain against the trusted roots as of GenTime, again
+	// requiring the timestamping EKU across the chain. GenTime (not the optional
+	// CMS signing-time attribute or the current time) is the authoritative
+	// timestamp creation time per RFC 3161.
+	opts := x509.VerifyOptions{
+		Roots:       roots,
+		CurrentTime: info.GenTime,
+		KeyUsages:   []x509.ExtKeyUsage{x509.ExtKeyUsageTimeStamping},
+	}
+	if err := p7.VerifyWithOpts(opts); err != nil {
```
matthiasbruns: `VerifyWithOpts` does not load intermediate certificates from the token automatically. Valid root → intermediate → TSA chains therefore fail despite the correct root being trusted.

Could we populate `opts.Intermediates` from the embedded certificates? [Reproduction with passing control](https://github.com/matthiasbruns/open-component-model/blob/c9baec65e43e04321befde2869e7b7df8126ccaf/bindings/go/signing/tsa/review_regression_test.go#L71-L90).

bindings/go/cli/cmd/verify/component-version/cmd.go
```diff
@@ -405,3 +452,128 @@ func loadVerifierConfig(config *genericv1.Config, signatureName string, logger *
...
+	if err := json.Unmarshal(raw, &fields); err != nil {
+		return properties
+	}
+	for k, v := range fields {
+		if k == "type" {
+			continue
+		}
+		var s string
+		if err := json.Unmarshal(v, &s); err == nil {
+			properties[k] = s
```
matthiasbruns: Plugin-returned `*runtime.Raw` credentials miss the `*DirectCredentials` assertion. The JSON fallback then drops their nested `properties`, losing the public key or trust root.

Could we decode by runtime type before extracting properties? [Reproduction comparing concrete and raw credentials](https://github.com/matthiasbruns/open-component-model/blob/c9baec65e43e04321befde2869e7b7df8126ccaf/bindings/go/cli/cmd/verify/component-version/review_regression_test.go#L43-L74).

bindings/go/signing/tsa/tsa.go
```diff
@@ -0,0 +1,334 @@
...
+// TSA servers reachable only over plain HTTP keep working, but any redirect to a
+// non-HTTPS target is refused. This blocks an HTTPS-to-HTTP downgrade in which
+// net/http would forward the Authorization header (and other sensitive headers)
+// derived from URL userinfo over an unencrypted connection.
+func RejectInsecureRedirect(req *http.Request, _ []*http.Request) error {
+	if req.URL.Scheme != "https" {
+		return fmt.Errorf("tsa: refusing insecure redirect to %s", RedactURL(req.URL.String()))
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
jakobmoellerdev: Fixed in `5ce7839`: `RejectInsecureRedirect` now re-imposes net/http's default cap — it returns an error once `len(via) >= 10` (a custom `CheckRedirect` otherwise disables the built-in limit). Covered by an added case in `TestRejectInsecureRedirect`.

bindings/go/signing/tsa/tsa.go
jakobmoellerdev: Fixed in `5ce7839`: `Verify` now builds `opts.Intermediates` from `p7.Certificates` (excluding the signer leaf) before `VerifyWithOpts`, so a root->intermediate->TSA chain validates when only the root is trusted. New `TestVerify_RootIntermediateLeafChain` reproduces it (token embeds the intermediate; only the root is in the pool).

bindings/go/cli/cmd/verify/component-version/cmd.go
jakobmoellerdev: Fixed in `5ce7839`: `credentialProperties` now flattens a nested string->string object (e.g. the `properties` of a `*runtime.Raw` wrapping DirectCredentials) in addition to top-level string fields, so plugin-returned Raw credentials keep their public key / trust root. New `TestCredentialProperties_RawWithNestedProperties` and `TestWithVerifiedTime_PreservesMaterialAndAddsTime` cover both the Raw and typed-flat shapes.

bindings/go/signing/tsa/tsa_test.go
```diff
@@ -0,0 +1,1252 @@
...
+	return key, cert
+}
+
+// newChainMockTSAHandler serves timestamp tokens whose SignedData embeds the
+// signer leaf together with the supplied intermediate certificates.
+func newChainMockTSAHandler(t *testing.T, leaf *x509.Certificate, leafKey *rsa.PrivateKey, chain []*x509.Certificate) http.Handler {
+	t.Helper()
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		body, err := io.ReadAll(r.Body)
+		require.NoError(t, err)
```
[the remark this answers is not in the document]
jakobmoellerdev: Fixed in `6da40d8`: `newChainMockTSAHandler` no longer calls `require` on the server goroutine — it reports failures via `t.Errorf` and aborts with `http.Error`, matching `newMockTSAHandler`.

PR #3638, opened 2026-09-18
chore(deps): update dependency golangci/golangci-lint to v2.13.1 (fix #3637)

frewilhelm:
#3637 updates `golanci-lint` which deprecates `exhaustruct`, so it can be replaced by `exhaustruct_v5`. Since `exhauststruct` was ignored previously, I propose to just add `exhaustruct_v5` to the ignore list. 

~I don't want to remove `exhaustruct` from that ignore list, although its deprecated now, because it would force users to update their local version of `golangci-lint` if they run it on this project.~ 

~Removed `exhaustruct` from the ignore list.~~

> exhaustruct: deprecated and replaced by exhaustruct_v5

Apparently, `exhaustruct` is still active, so we need it in our ignore list

Fixes #3637


jakobmoellerdev:
> I don't want to remove `exhaustruct` from that ignore list, although its deprecated now, because it would force users to update their local version of `golangci-lint` if they run it on this project.

I think thats invalid because we ship our own version of golangci lint in our task runner. Its fine to ask for upgrade

frewilhelm:
> I think thats invalid because we ship our own version of golangci lint in our task runner. Its fine to ask for upgrade

Apparently, `exhaustruct` is still active, so we need it in our ignore list

matthiasbruns (approved):

jakobmoellerdev (approved):

PR #3639, opened 2026-09-18
chore!: Remove support for nested generic configs in v2

### LINKED ISSUE #1201: Revisit support for v1 nested ocm generic configs
**Description**
For context see https://github.com/open-component-model/open-component-model/pull/3118/changes#r3578761543

We support nested configurations, e.g. 
```yaml
configurations:
  - type: generic
    configurations:
      - type: transfer
        copyMode: localBlob
  - type: transfer
    copyMode: allResources
```
main reason we support this was because it was supported in v1.

The resolution order is unintuitive though. In the above example `localBlob` wins, because this was the behavior in v1.

#### Do we want to maintain this behavior for v2 going forward?
If yes, then likely no further action is required.

#### Do we want to support nested configs, but fix the precedence
* breaking change from v1
* config format stays as is
* migration guide should reflect the change

#### Do we want to drop support for nested configs?
* introduce new config format that does not accept nesting
* new format has 'intuitive' precedence and no edge-case due to nesting
* config@v1 evaluation stays in the code as is and is deprecated at some point
  * Or is removed and a migration script converts configs to the new format?

### Open questions
* Current behavior of nested config does not appear to be documented, whatever we decide on should be reflected in the documentation
* Was the v1 behavior intended or a bug from the start?
* What was the usecase for the nesting to begin with? 

**Done Criteria**
- [ ] ...

jneisener:
PR removes the support for nested generic configs that was possible, but undocumented, in v1. 

Without this logic `FlatMap` folds in on itself and could thus be eliminated.

Fixes https://github.com/open-component-model/ocm-project/issues/1201

##### Verification


bindings/go/kubernetes/controller/pkg/configuration/config.go
```diff
@@ -219,7 +219,7 @@ func LoadConfigurations(ctx context.Context, k8sClient client.Reader, namespace
 		return nil, err
 	}
 
-	var configs []*genericv1.Config
+	merged := &genericv1.Config{Configurations: []*runtime.Raw{}}
```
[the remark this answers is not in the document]
jneisener: Seems to me like this mirrors the preexisting FlatMap behavior

jneisener:
> > **Document the removed exported API and migration path.**

The v1 behavior was not documented, so I'm not convinced that is necessary

bindings/go/configuration/generic/v1/spec/config.go
```diff
@@ -16,6 +16,13 @@ const (
 	ConfigTypeV1 = Version
 )
 
+const NestedConfigIgnoredWarning = "ignoring nested configuration: nested generic configurations are not supported, " +
+	"move the nested entries to the top-level configurations list"
+
+func IsGenericConfig(t runtime.Type) bool {
+	return t.Name == ConfigType
+}
```
fabianburth: This feels off. Shouldn't we rather check on type?

bindings/go/configuration/generic/v1/spec/config.go
```diff
@@ -16,6 +16,13 @@ const (
 	ConfigTypeV1 = Version
 )
 
+const NestedConfigIgnoredWarning = "ignoring nested configuration: nested generic configurations are not supported, " +
+	"move the nested entries to the top-level configurations list"
```
fabianburth: this const is also suspicious

bindings/go/cli/cmd/configuration/ocm_config.go
```diff
@@ -113,9 +106,18 @@ func loadAndMergeConfigs(paths []string, strict bool) (*genericv1.Config, error)
...
+		for _, entry := range cfg.Configurations {
+			if genericv1.IsGenericConfig(entry.GetType()) {
+				slog.Warn(genericv1.NestedConfigIgnoredWarning,
+					slog.String("type", entry.GetType().String()),
+					slog.String("path", path),
+				)
+				continue
+			}
+			merged.Configurations = append(merged.Configurations, entry)
+		}
```
fabianburth: Would it make sense to make this an exported function? Looks identical to the code in the controller.

fabianburth (changes requested):

bindings/go/configuration/generic/v1/spec/config.go
jneisener: This is meant to not have the message duplicated in cli and controller

bindings/go/configuration/generic/v1/spec/config.go
jneisener: If by check the type you mean `Scheme.IsRegistered`, I think the type check would pass for `generic.config.ocm.software/v2`?

bindings/go/cli/cmd/configuration/ocm_config.go
jneisener: I had that before, my first draft had the `FlatMap` return `(merged *Config, ignored []*runtime.Raw)` and let the caller warn.
But that was not longer flattening anything and a simple for loop felt like it is better inline. Since the main content is the logging (which is different for CLI and controller) it felt better to merge here

bindings/go/configuration/generic/v1/spec/config.go
jneisener: So essentially this boils down to how we address the different logging in cli and controller (so related to https://github.com/open-component-model/open-component-model/pull/3639#discussion_r4045659582) and if we prefer the single exported string over having the string duplicated

bindings/go/configuration/generic/v1/spec/config.go
jneisener: Actually, nevermind I was overthinking it.

bindings/go/cli/cmd/configuration/ocm_config.go
jneisener: Warning is slightly less specific now, but code is not duplicated anymore

bindings/go/configuration/generic/v1/spec/config.go
```diff
@@ -16,6 +16,27 @@ const (
...
+// any encountered nested generic configs, which are not supported.
+func MergeConfigs(warnFn func(msg string, keysAndValues ...any), configs ...*Config) *Config {
+	merged := new(Config)
+	merged.Configurations = make([]*runtime.Raw, 0)
+	for _, config := range configs {
+		for _, entry := range config.Configurations {
+			if Scheme.IsRegistered(entry.GetType()) {
+				warnFn(
+					"ignoring nested configuration: nested generic configurations are not supported, move the nested entries to the top-level configurations list",
+					"type", entry.GetType().String(),
```
jneisener, line 30: This isn't as helpful as it was before https://github.com/open-component-model/open-component-model/pull/3639/commits/ca917cdc8a3483a4798495aa15e2e92722ace35c and https://github.com/open-component-model/open-component-model/pull/3639/commits/ca917cdc8a3483a4798495aa15e2e92722ace35c

I could change the function interface to supply the source to log here to help a user trace the error better
```go
type Source struct {
	Name   string
	Config *Config
}

func MergeConfigs(warnFn func(msg string, keysAndValues ...any), sources ...Source) *Config {
```

Opinions?

bindings/go/configuration/generic/v1/spec/config.go
fabianburth: was more thinking about something like this

 ```go
   func IsLocalBlob(access runtime.Typed) bool {
      if access == nil {
         return false
      }
      var local LocalBlob
      if err := Scheme.Convert(access, &local); err != nil {
         return false
      }
      return true
   }
 ```

bindings/go/configuration/generic/v1/spec/config.go
fabianburth: I think, I'd rather have duplicated code somewhere than this.

jakobmoellerdev (approved):

PR #3641, opened 2026-09-18
chore: enable digest verification across all technologies during download

Skarlso:

Approach taken.

Why not inmemory.New WithDigest. It ended up being very ugly trying to give it the resource where it's used and not everything would be using it so it's not a uniform solution.

Verifier during download is a good approach and kept the changes to a minimum.

There are bugs around and there are a couple of things I left out explicitly. Like, we have now the 6th parser for Digests, which is shit, I know, but I didn't want to refactor the entire thing in this PR that would muddle the review.

Then there is the OCI part in which you can see the coderabbit comment. Please think about this how to handle it nicely. We might download the content, we might fetvch by digest, or disallow this... 

Further, the read verifier might looks strange because it's a "hand rolled tee writer" because we are verifying on Read and not just on Close. Because we don't want an accidental verification ignore on having something like `_ = x.Close()`.

The other decisions were taken because they were the most convenient and least impactful overall.

##### Verification


Skarlso:
@coderabbitai review

bindings/go/helm/repository/resource/resource_repository.go
```diff
@@ -145,6 +153,13 @@ func (r *ResourceRepository) DownloadResource(ctx context.Context, resource *des
 		return nil, fmt.Errorf("error downloading helm chart %q: %w", helmURL, err)
 	}
 
+	if isOCI {
+		slog.WarnContext(ctx, "a helm chart from an OCI registry has the digest of its manifest rather than of its archive, so the downloaded chart cannot be verified against it",
+			slog.String("chartReference", helmURL))
+	} else if err := verifyChartArchive(ctx, resource, result.ChartBlob); err != nil {
+		return nil, fmt.Errorf("error verifying downloaded helm chart %q: %w", helmURL, err)
```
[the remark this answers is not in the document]
Skarlso: I think the declared oci manifest digest is not the same as the content digest of the helm chart. Can you verify that?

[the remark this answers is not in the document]
Skarlso: Exactly. Which decision is a follow up discussion that we are having internally.

bindings/go/repository/verify_blob.go
fabianburth, line 1: check whether we could reuse https://github.com/oras-project/oras-go/blob/v2.6.2/content/reader.go#L58-L129

bindings/go/github/repository/resource/resource_repository.go
```diff
@@ -130,6 +133,27 @@ func (r *ResourceRepository) DownloadResource(ctx context.Context, resource *des
...
+}
+
+// DownloadCommitArchive fetches the archive of the commit pinned in GitHub, which
+// [download.Download] rejects the access without.
+//
+// It takes an access rather than a resource because there is no resource here, and
+// so no digest to compare the archive to. The digest processor calls it while
+// establishing that digest, which cannot be verified before it exists. Callers
+// holding a resource want DownloadResource, which verifies.
+func (r *ResourceRepository) DownloadCommitArchive(ctx context.Context, gitHub *v1.GitHub, credentials runtime.Typed) (blob.ReadOnlyBlob, error) {
```
fabianburth: How would you feel about `DownloadResource` returning a VerifyReader? So, we do not need 2 interfaces and the reader can decide whether to verify or not? Would also solve your problem in the digest processor?

bindings/go/repository/verify.go
```diff
@@ -0,0 +1,72 @@
...
+	if err != nil || !ok {
+		return err
+	}
+
+	if expected != actual {
+		return fmt.Errorf("digest mismatch for resource %q: expected %s, got %s", res.ToIdentity(), expected, actual)
+	}
+
+	return nil
+}
```
fabianburth: I'm worried about increasing our public interface surface so much

Skarlso: This is now behind `internal` so no longer a "public interface surface" per say. WDYT?

bindings/go/repository/verify_blob.go
Skarlso: Applied. It looks nice, and we got a nice size check for free. 🎉

bindings/go/github/repository/resource/resource_repository.go
Skarlso: Tried this, but ended up reverting, but the work showed a couple of problems that I fixed in between. So hopefully this is now nicer. 

The main reason for the revert was that now everything, including our transformers, would have to remember to call `Verify` on the endresult. Including every potential library user. This ended up with a lot of code and a lot of verifications being called.

So it makes sense to have it internally done already by the reader. Also, still like that Read will ALSO error in case anyone forgets a Close..

Skarlso:
>+1,234 -33

Heh. Nice. :)

bindings/go/helm/repository/resource/resource_repository.go
```diff
@@ -145,6 +154,13 @@ func (r *ResourceRepository) DownloadResource(ctx context.Context, resource *des
 		return nil, fmt.Errorf("error downloading helm chart %q: %w", helmURL, err)
 	}
 
+	if isOCI {
+		slog.WarnContext(ctx, "a helm chart from an OCI registry has the digest of its manifest rather than of its archive, so the downloaded chart cannot be verified against it",
+			slog.String("chartReference", helmURL))
+	} else if err := verifyChartArchive(ctx, resource, result.ChartBlob); err != nil {
```
fabianburth: I think code rabbit is spot on. Let's assume the `helmURL` is something like `oci://my.domain/my/helm/chart:1.0.0`. We never verify whether the manifests digest of 1.0.0 matches our resource digest.

bindings/go/s3/repository/resource_repository.go
```diff
@@ -98,7 +102,7 @@ func (r *ResourceRepository) DownloadResource(ctx context.Context, resource *des
 		return nil, err
 	}
 
-	return result.Blob, nil
+	return verify.Download(ctx, resource, result.Blob)
```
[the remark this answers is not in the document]
Skarlso: This is a much larger discussion here: https://github.com/open-component-model/ocm-project/issues/1326

bindings/go/wget/repository/resource_repository.go
```diff
@@ -104,7 +105,7 @@ func (r *ResourceRepository) DownloadResource(ctx context.Context, resource *des
 	if err != nil {
 		return nil, err
 	}
-	return b, nil
+	return verify.Download(ctx, resource, b)
```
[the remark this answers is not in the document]
Skarlso: We usually let the caller handle that, because we don't know what they still might want to do with the result.

bindings/go/internal/verify/blob.go
```diff
@@ -0,0 +1,163 @@
+package verify
```
matthiasbruns: what now we have 
1. more blobs
2. an internal package in `bindings/go`?

I get the idea behind the blob - but this could have also been done by the caller side after the blob is fully downloaded

Skarlso: See this comment https://github.com/open-component-model/open-component-model/pull/3641#discussion_r4069571755

matthiasbruns (commented):
ok one maybe naive question - why didn't you just call/reuse the digest checks from `ProcessResourceDigest` after `DownloadResource` in each repository and be done with it?

Skarlso:
> ok one maybe naive question - why didn't you just call/reuse the digest checks from `ProcessResourceDigest` after `DownloadResource` in each repository and be done with it?

Errr... ProcessResourceDigest is the upload path, right...? not during download. So how would that be useful here? :thinking: Also, not every repository implements it I believe.

Skarlso:
I will scope this down to the readers only to avoid the OCI part for now as that's a separate discussion.

bindings/go/internal/verify/doc.go
```diff
@@ -0,0 +1,10 @@
+// Package verify compares downloaded content to the digest a component descriptor
+// declares for it. It is internal to bindings/go and not part of the public API.
+//
+// Two functionalities live here:
+//   - Download wraps a blob so that reading it hashes the content and compares.
+//     Use it where the download returns exactly the bytes the digest represents.
+//   - Digest compares a digest the repository already resolved. Use it where the
+//     digest covers something else, such as an OCI image resource whose digest is
+//     that of its manifest.
+package verify
```
matthiasbruns: I honestly still think this can be split in the digest and repository bindings

Skarlso: We didn't want to increase the public API more. :)

bindings/go/repository/verify_blob.go
jakobmoellerdev: Issue here is that we bind most modules to oras. we should have our own version I believe

jakobmoellerdev: please also get rid of the internal package. its an anti-pattern. either it should be part of runtime and be generic enough, or blob, or similar. dont introduce new top level packages!

Skarlso: Got it. 👍

bindings/go/internal/verify/doc.go
matthiasbruns: ok then listen to Jakob instead :-*

jakobmoellerdev:
@jneisener can you please check why @Skarlso didnt run into a failing job when the package was introduced? we should not simply allow a new package creation I think

Skarlso:
I did. I fixed it by updating the dependency checker. I thought that what we are supposed to do to make it explicit.

bindings/go/README.md
```diff
@@ -27,6 +27,7 @@ Go bindings for the Open Component Model.
 | **input/dir** | Provides functionality for handling directory-based inputs in the Open Component Model (OCM) constructor |
 | **input/file** | Provides functionality for handling file-based inputs in the Open Component Model (OCM) constructor |
 | **input/utf8** | Provides functionality for handling UTF8 string-based inputs in the Open Component Model (OCM) constructor |
+| **internal/verify** | Compares downloaded content to the digest a component descriptor declares for it |
```
Skarlso: @jakobmoellerdev here

jneisener: Also https://github.com/open-component-model/open-component-model/pull/3641/changes#diff-02f3aab86a2dfffc9cf61ee7bd633da2c63efcd7a7c5d668932b93d7612d9dfaR367-R376 (`golangci.yml`)

bindings/go/internal/verify/doc.go
Skarlso: I didn't deny this, I just told the reason why I did it like this. :D

bindings/go/README.md
Skarlso: Yep. Reverted all of this.

bindings/go/repository/verify_digest.go
```diff
@@ -0,0 +1,44 @@
...
+)
+
+// parseDigest returns d as a [digest.Digest] in canonical "algorithm:hex" form, so that
+// content can be verified against it.
+//
+// Note: There are several places in the code today in which we are parsing digests in
+// one way or another. It will be a separate issue to pull them all together. Not in this one.
+// And we aren't using those to avoid having to import OCI package or some other package
+// and dilute the dependency graph.
+func parseDigest(d *descriptor.Digest) (digest.Digest, error) {
```
jakobmoellerdev: this should move to descriptor runtime no? whats the reason for repository?

bindings/go/repository/verify.go
```diff
@@ -0,0 +1,49 @@
...
+	"ocm.software/open-component-model/bindings/go/blob"
+	descriptor "ocm.software/open-component-model/bindings/go/descriptor/runtime"
+)
+
+// VerifyDownload verifies content with digest `res` contains, as it is read.
+//
+// A resource that has no digest passes through unverified with a warning, but a
+// digest that is present and unusable errors. Content that cannot be checked
+// against the digest must not pass as verified.
+func VerifyDownload(ctx context.Context, res *descriptor.Resource, content blob.ReadOnlyBlob) (blob.ReadOnlyBlob, error) {
```
jakobmoellerdev: arguably this should also be in descriptor runtime? WDYT?

bindings/go/github/repository/resource/resource_repository.go
```diff
@@ -130,7 +133,12 @@ func (r *ResourceRepository) DownloadResource(ctx context.Context, resource *des
 		}
 	}
 
-	return download.Download(ctx, gitHub, gitHubCredentials, r.httpClient)
+	archive, err := download.Download(ctx, gitHub, gitHubCredentials, r.httpClient)
+	if err != nil {
+		return nil, err
+	}
+
+	return repository.VerifyDownload(ctx, resource, archive)
```
matthiasbruns: I think this morning we talked about "how to prevent that we forgot to verify again in the future"

While this is a good addition in general, it does not solve the problem that each access now has to do that itself.
I would want a follow up later that tries to find a better abstraction through our plugins and become part of the download resource lifecycle and where ever it might become useful

Skarlso: I don't know how you could avoid that as all of our technologies have specific Download implementation for a reason.

MAYBE if we completely restructure the download or the transfer or whatever that could insert a hook anywhere where there is a download call from the outside.. or shift the verification process or create some kind of framework that create tasks and runs them like ocmv1 runtime did (*shudder*) and insert a `verification` step so it never gets omitted...

bindings/go/repository/verify.go
Skarlso: @jakobmoellerdev Sorry Jakob, I'm really, really not trying to be nitpicking here, just want to be a 100% clear on your wording... by _also_ you don't mean _as well_ like, duplicate it there, right? But just move it there?

That would mean that we'd have to import `"ocm.software/open-component-model/bindings/go/blob"` there. Are you okay with that?

Skarlso: With regards to WDYT: I think it was fine in the repository because everything already imports repository and descriptor/runtime seemed like a very chill package with few imports. :D

bindings/go/github/repository/resource/resource_repository.go
matthiasbruns: Thought about this a bit

we can use GetResourcePlugin and return another implementation through the plugin manager that combines resourceRepo with a verifier as a facade 


so cli will never see a difference
access resource repos do not have to verify on download

the abstraction does

```text
CLI
 │
 │ resource.Repository
 ▼
Verification facade
 ├── Downloader: built-in implementation or external-plugin adapter
 └── Verifier: selected for the digest normalization / content representation
```

Skarlso: GetResourcePlugin is the external plugin getter. That does not work. And besides, at the point you don't have enough information or context even if you coerce it somehow into only dealing with the internal plugins.

Skarlso: I see your poc, I know because my llm also suggested that approach :P And I refuted it because it's a) ugly as hell b) the plugin system is not designed to be abused like that c) it's a boundary, it's a plugin manager and not a verification machinery d) library users DO NOT GO THROUGH THIS. They use Download and the Repository. This would actively kill verification for their use-case e) external plugins are not reached unless you rewrite that section completely f) this is not the right place because verification is actively a technology specific thing and making it global will hurt that place because now it needs to "switch" or otherwise make it opt-out.

It just doesn't work nicely like this does.

Skarlso: I'm so sorry. :D ❤️

PR #3642, opened 2026-09-18
feat(oci): chunked blob push with streaming for unknown digests

jakobmoellerdev:

Implements the OCI Distribution Spec [chunked blob upload](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pushing-a-blob-in-chunks) (`POST` session → `PATCH` chunks with `Content-Range` → `PUT` close with the whole-blob digest) on `remotestore.RemoteStore`.

- Blobs at or above a configurable threshold upload in chunks; manifests and small blobs keep the monolithic push (oras `remote.Repository`, which has no chunked path).
- On a **pre-consumption** failure (session `POST` fails, or the first `PATCH` not yet sent) the chunked path falls back to the embedded monolithic push, so no registry regresses. Post-consumption failures return a wrapped error and best-effort `DELETE` the session.
- Honors the registry-advertised `OCI-Chunk-Min-Length`; validates `Location` host/scheme to avoid credential leakage (same checks as oras).

Because the whole-blob digest is only needed to **close** the session, this also unlocks true streaming: `PushStreaming` uploads content whose digest and size are unknown up front, computing both during upload and returning the completed descriptor. The `pack` layer now **streams unknown-digest resource blobs instead of buffering them into memory** (`ArtifactBlob.Buffer` / `cache.Cache`) purely to precompute the descriptor.

Configuration: `urlresolver.WithChunkedPush(chunkSize, threshold)` (defaults 16 MiB). Chunking is bypassed when a blob/reference cache proxy is active (documented on the option).

##### Automated

- Unit (`remote_store_push_test.go`): exact `POST`/`PATCH`/`PUT` sequence + `Content-Range`; POST-rejection fallback to monolithic; post-consumption `PATCH`-rejection errors (no silent success); `ChunkSize=0` / sub-threshold / manifest stay monolithic (no `PATCH`); `OCI-Chunk-Min-Length` raises chunk size; streaming computes the digest; streaming-unavailable cases.
- Pack (`pack_test.go`): a fake `StreamingPusher` proves the streaming path is taken (not buffering `Push`), the digest is computed from streamed bytes, and the resource access is populated.
- Integration (`chunked_push_integration_test.go`, real `registry:3.0.0`):
  - `Test_Integration_OCIRepository_ChunkedPush`: multi-chunk `Push` (known digest) and `PushStreaming` (unknown digest) round-trip.
  - `Test_Integration_OCIRepository_StreamingDoesNotBuffer`: an interleaving reader blocks mid-stream until it observes a `PATCH` reach the registry — a buffering implementation would read the whole blob before any network call and time out, so passing proves **no buffering for missing digests**.

##### How to test the changes

```bash
task bindings/go:test -- -run 'TestRemoteStore_Push|TestRemoteStore_PushStreaming|TestResourceLocalBlobOCILayer_StreamsUnknownDigest' ./oci/...
task bindings/go/oci/integration:test/integration   # needs Docker
```


jakobmoellerdev:
Updated: chunked push now routes all requests (POST/PATCH/PUT/DELETE) through a `RemoteStore.do` helper that mirrors oras' internal `Repository.do` — same client resolution (`r.Client` else `auth.DefaultClient`, so retries/auth/TLS/proxy/headers are respected identically) and the same `HandleWarning` response-header handling. Added `TestRemoteStore_Push_InvokesHandleWarning` covering the warning parity.

jakobmoellerdev:
Expanded streaming to also cover **known-digest but unknown-size** blobs (e.g. lazily-loaded blobs that expose a digest). Previously these were buffered to obtain the Content-Length; now they stream via chunked PATCHes, discovering the size during upload and closing with the known digest. The streamed content is always hashed and verified against a caller-supplied digest (mismatch fails without closing the session). Blobs that already expose both size and digest keep the regular push path (which still chunks large blobs by threshold). Added unit tests (`PushStreaming_KnownDigestUnknownSize`, `PushStreaming_DigestMismatchErrors`) and a pack test (`StreamsKnownDigestUnknownSize`).

jakobmoellerdev:
### Benchmarks

Added benchmarks (Apple M4 Pro, `-benchmem`). Two levels: store-level push overhead against an in-process sink registry, and the pack-level streaming-vs-buffering memory profile.

#### Pack layer — streaming (no buffer) vs buffering an unknown-digest blob

`go test -bench=BenchmarkResourceLayer_StreamingVsBuffering -benchmem ./oci/internal/pack/...`

| Blob | Path | B/op | allocs/op | MB/s |
|------|------|------|-----------|------|
| 1 MiB  | streaming | **1,936**       | 23 | 2947 |
| 1 MiB  | buffering | 2,233,602       | 72 | 1809 |
| 16 MiB | streaming | **1,936**       | 23 | 2967 |
| 16 MiB | buffering | 36,231,306      | 82 | 2494 |
| 64 MiB | streaming | **1,937**       | 23 | 2997 |
| 64 MiB | buffering | 165,353,416     | 83 | 2467 |

Streaming allocates a **flat ~1.9 KB regardless of blob size**; buffering allocates on the order of the blob (≈2.5× for the 64 MiB case, since the in-memory cache grows). Streaming is also faster — no full pre-read pass to compute the digest.

#### Store layer — monolithic vs chunked (sink registry, zero network latency)

`go test -bench=BenchmarkRemoteStore_Push -benchmem ./oci/internal/remotestore/...`

Against a **local, zero-latency** sink, monolithic wins on raw throughput/allocs (e.g. 64 MiB: monolithic ~1224 MB/s, 50 KB/op vs chunked-16MiB ~846 MB/s, 16 MB/op — the chunk buffer dominates). This is expected: chunked push does not exist to beat a single PUT on a LAN; its value is (1) the pack-layer memory win above (no buffering when digest/size are unknown) and (2) resumability/large-blob reliability over real networks. The chunk buffer is sized to `ChunkSize`, so larger chunks trade memory for fewer round-trips.

Run all: `go test -bench=. -benchmem ./oci/internal/remotestore/... ./oci/internal/pack/...`

bindings/go/oci/internal/remotestore/remote_store_push_test.go
```diff
@@ -0,0 +1,490 @@
...
+	store := newTestStore(t, srv, 2, 1)
+	store.HandleWarning = func(w remote.Warning) { warnings = append(warnings, w.Text) }
+
+	data := []byte("hello")
+	dig := digest.FromBytes(data)
+	desc := ociImageSpecV1.Descriptor{MediaType: "application/octet-stream", Digest: dig, Size: int64(len(data))}
+	require.NoError(t, store.Push(t.Context(), desc, bytes.NewReader(data)))
+
+	require.Equal(t, []string{"this repository is deprecated"}, warnings)
+}
```
jneisener, line 494: sha-256 is always used in `uploadChunks` but if a sha-512 is declared, thats what `resolveFinalDigest` compares against, creating a mismatch:
```go
func TestRemoteStore_Push_ChunkedNonSHA256Digest(t *testing.T) {
	r := require.New(t)

	var algoParam, closedDigest string
	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
		switch req.Method {
		case http.MethodPost:
			algoParam = req.URL.Query().Get("digest-algorithm")
			w.Header().Set("Location", "/v2/test-repo/blobs/uploads/1")
			w.WriteHeader(http.StatusAccepted)
		case http.MethodPatch:
			w.Header().Set("Location", req.URL.Path)
			w.WriteHeader(http.StatusAccepted)
		case http.MethodPut:
			closedDigest = req.URL.Query().Get("digest")
			w.WriteHeader(http.StatusCreated)
		default:
			w.WriteHeader(http.StatusOK)
		}
	}))
	t.Cleanup(srv.Close)

	store := newTestStore(t, srv, 4, 1)

	data := bytes.Repeat([]byte("x"), 10)
	sum := sha512.Sum512(data)
	dig := digest.Digest(fmt.Sprintf("sha512:%x", sum))
	desc := ociImageSpecV1.Descriptor{MediaType: "application/octet-stream", Digest: dig, Size: int64(len(data))}

	r.NoError(store.Push(t.Context(), desc, bytes.NewReader(data)))
	r.Equal("sha512", algoParam)
	r.Equal(dig.String(), closedDigest, "session must be closed with the sha512 digest")
}
```

jakobmoellerdev: Good catch — fixed in 7444897.

The session digester was hardcoded to `digest.Canonical` (sha256) while `resolveFinalDigest` compared it against the declared digest, so any non-sha256 blob failed with a spurious mismatch (even though the `POST` already advertised `digest-algorithm=<algo>` per the distribution spec [end-4c](https://github.com/opencontainers/distribution-spec/blob/main/spec.md#pushing-a-blob-in-chunks)). `openUploadSession` now selects the digester from the declared digest's algorithm (canonical when none is declared) and rejects an unavailable algorithm before any content is consumed.

Added your `TestRemoteStore_Push_ChunkedNonSHA256Digest` verbatim — it fails without the fix (`content digest sha256:… does not match declared digest sha512:…`) and passes with it.

bindings/go/oci/internal/pack/pack.go
```diff
@@ -96,6 +111,82 @@ func ResourceLocalBlobOCILayer(ctx context.Context, storage content.Storage, b *
...
+
+	reader, err := b.ReadCloser()
+	if err != nil {
+		return ociImageSpecV1.Descriptor{}, false, fmt.Errorf("failed to get blob reader: %w", err)
+	}
+	defer func() {
+		_ = reader.Close()
+	}()
+
+	partial := ociImageSpecV1.Descriptor{MediaType: mediaType, Digest: knownDigest, Size: blob.SizeUnknown}
```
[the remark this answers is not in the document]
jakobmoellerdev: Fixed in 97a1577. `streamResourceLayer` now forwards `b.Size()` to `PushStreaming` when the size is known (falling back to `SizeUnknown` only when it isn't), so `pushChunked`'s size check catches a short/truncated reader. Empty blobs keep working since `PushStreaming` still maps `Size == 0` to unknown. Added `TestResourceLocalBlobOCILayer_ForwardsKnownSize` (fails pre-fix: forwards `-1` instead of the real size).

bindings/go/oci/internal/remotestore/chunked_push.go
```diff
@@ -0,0 +1,466 @@
...
+		return nil, parseErrorResponse(resp)
+	}
+	location, err := resolveUploadLocation(resp, req)
+	if err != nil {
+		return nil, err
+	}
+	chunk := r.ChunkSize
+	if minLen := parseChunkMinLength(resp); minLen > chunk {
+		chunk = minLen
+	}
```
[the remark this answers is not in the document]
jakobmoellerdev: Fixed in 97a1577. Added `MaxChunkSize` (128 MiB); `openUploadSession` now rejects an advertised `OCI-Chunk-Min-Length` above it before any byte is consumed, so `Push` still falls back to the monolithic upload and `PushStreaming` still returns `ErrStreamingUnavailable`. Regression tests `TestRemoteStore_Push_RejectsOversizedChunkMinLength` and `TestRemoteStore_PushStreaming_RejectsOversizedChunkMinLength` cover both paths.

bindings/go/oci/internal/remotestore/chunked_push.go
```diff
@@ -0,0 +1,466 @@
...
+// the session (the reader cannot be rewound for a monolithic fallback).
+func (r *RemoteStore) uploadChunks(ctx context.Context, up *chunkedUpload, content io.Reader) error {
+	hasher := up.digester.Hash()
+	buf := make([]byte, up.chunk)
+	for {
+		n, readErr := io.ReadFull(content, buf)
+		if readErr == io.EOF {
+			return nil
+		}
+		if readErr != nil && readErr != io.ErrUnexpectedEOF {
```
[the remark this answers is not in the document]
jakobmoellerdev: Fixed in 97a1577. `uploadChunks` now sets `up.consumed` as soon as a read yields bytes (`n > 0`), before any error branch, so an `io.ReadFull` returning `n>0` with a non-EOF error is treated as post-consumption: `Push` no longer replays the advanced reader through the monolithic fallback and `PushStreaming` surfaces the real read error instead of `ErrStreamingUnavailable`. The now-redundant `consumed` assignment before the PATCH and its guard were removed. Added `TestRemoteStore_PushStreaming_MidStreamReadErrorIsPostConsumption`, which fails pre-fix (error wrongly wraps `ErrStreamingUnavailable`).

bindings/go/oci/internal/pack/pack.go
```diff
@@ -96,6 +111,89 @@ func ResourceLocalBlobOCILayer(ctx context.Context, storage content.Storage, b *
...
+// unavailable — so the caller falls back to the buffering push path.
+//
+// A digest known in advance (from the blob or the access LocalReference) is
+// passed through and verified against the streamed bytes; an unknown digest is
+// computed from them.
+func streamResourceLayer(ctx context.Context, storage content.Storage, b *ociblob.ArtifactBlob, opts ResourceBlobOCILayerOptions) (ociImageSpecV1.Descriptor, bool, error) {
+	pusher, ok := storage.(remotestore.StreamingPusher)
+	if !ok {
+		return ociImageSpecV1.Descriptor{}, false, nil
+	}
```
fabianburth:
```suggestion
func streamResourceLayer(ctx context.Context, storage remotestore.StreamingPusher, b *ociblob.ArtifactBlob, opts ResourceBlobOCILayerOptions) (ociImageSpecV1.Descriptor, bool, error) {
```

bindings/go/oci/internal/remotestore/chunked_push.go
```diff
@@ -0,0 +1,480 @@
...
+// precomputed digest, streaming the content and computing the digest during
+// upload. Callers use this to avoid buffering blobs of unknown size/digest.
+type StreamingPusher interface {
+	// PushStreaming uploads content described only by partial (MediaType is
+	// used; Digest/Size are ignored and computed during upload) and returns the
+	// completed descriptor with the computed Digest and Size. If streaming is
+	// unavailable it returns an error wrapping ErrStreamingUnavailable and no
+	// content is consumed, so the caller may safely buffer and fall back.
+	PushStreaming(ctx context.Context, partial ociImageSpecV1.Descriptor, content io.Reader) (ociImageSpecV1.Descriptor, error)
+}
```
fabianburth, line 66: The names are pretty misleading if you are not very familiar with oras or read the entire doc string. the usual `Push` also streams, but only when the size and digest are known before upload, right?

bindings/go/oci/internal/remotestore/chunked_push.go
```diff
@@ -0,0 +1,480 @@
...
+
+// DefaultChunkThreshold is the smallest blob eligible for chunked upload when
+// no explicit threshold is configured (16 MiB): blobs below this go monolithic.
+const DefaultChunkThreshold int64 = 16 << 20
+
+// MaxChunkSize bounds the PATCH buffer the client will allocate. A registry can
+// force the buffer size up via its advertised OCI-Chunk-Min-Length; an
+// implausible value (up to math.MaxInt64) would otherwise panic make([]byte, n)
+// or exhaust memory. Advertised minimums above this are rejected (128 MiB).
+const MaxChunkSize int64 = 128 << 20
```
fabianburth: > MaxChunkSize bounds the PATCH buffer the client will allocate.

I think this is currently not true. It only ensures that 

> Advertised minimums above this are rejected (128 MiB)

I believe, we also need an upper bound for a user configured chunk size (see comment below)

bindings/go/oci/internal/remotestore/chunked_push.go
```diff
@@ -0,0 +1,480 @@
...
+	if err != nil {
+		return nil, err
+	}
+	chunk := r.ChunkSize
+	if minLen := parseChunkMinLength(resp); minLen > chunk {
+		if minLen > MaxChunkSize {
+			return nil, fmt.Errorf("chunked blob push: advertised OCI-Chunk-Min-Length %d exceeds the maximum supported chunk size %d", minLen, MaxChunkSize)
+		}
+		chunk = minLen
+	}
```
fabianburth, line 267: We could configure arbitrarily huge `r.ChunkSize` and crash the `buf := make([]byte, up.chunk)` in `uploadChunks`, I think.

bindings/go/oci/integration/chunked_push_integration_test.go
```diff
@@ -0,0 +1,389 @@
...
+}
+
+// Test_Integration_OCIRepository_AddComponentVersion_Streaming adds a component
+// version whose single resource is backed by a lazily-loaded blob with unknown
+// size and digest, then reads it back. This exercises the full add-cv transfer
+// path (AddLocalResource -> pack.ArtifactBlob -> ResourceLocalBlobOCILayer ->
+// RemoteStore.PushStreaming), proving the resource blob is streamed (not
+// buffered) and the component version round-trips.
+//
+// Behavior of the transfer as a function of the input blob (remote registry
```
matthiasbruns, line 283: should this be moved into a doc.go?

bindings/go/oci/internal/remotestore/chunked_push.go
```diff
@@ -0,0 +1,494 @@
...
+				r.cancelUpload(ctx, up.location)
+			}
+			return fmt.Errorf("chunked blob push: failed to read content after %d bytes: %w", up.offset, readErr)
+		}
+		if n == 0 {
+			return nil
+		}
+		_, _ = hasher.Write(buf[:n])
+
+		req, err := http.NewRequestWithContext(ctx, http.MethodPatch, up.location.String(), bytes.NewReader(buf[:n]))
```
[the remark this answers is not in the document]
jakobmoellerdev: Fixed in 6ddb59a. `RemoteStore.do` now strips `Request.GetBody` for body-carrying requests. Per `net/http` `redirectBehavior`, a `307`/`308` with a body but no `GetBody` is *not* followed — the stdlib returns the 3xx instead, and the upload's own status checks (`!= 202`/`201`) reject it, so no blob bytes are replayed to an unvalidated host or downgraded scheme. Legitimate upload-location handoffs still occur explicitly through the `Location` header, which `resolveUploadLocation` already validates for host/scheme.

Added `TestRemoteStore_Push_DoesNotFollowPATCHRedirectToOtherHost`: it drives a real redirect-following `http.Client` against a registry that `307`-redirects the PATCH to a separate host and asserts the target is never contacted and receives no body. It fails without the fix (attacker host hit, body leaked).

bindings/go/oci/integration/chunked_push_integration_test.go
jakobmoellerdev: I'd keep it here. `doc.go` is for package-level API documentation; this table is the behavior contract for the `AddComponentVersion` streaming transfer that the test immediately below it asserts, so it reads best next to that test. Moving it to `doc.go` would divorce the matrix from the case that exercises it. Happy to relocate if you'd rather it live at package scope.

matthiasbruns: keep it where it is then :)

fabianburth (dismissed):

bindings/go/oci/internal/remotestore/chunked_push.go
```diff
@@ -0,0 +1,552 @@
...
+// challenge/rewind flow is untouched. For a plain *http.Client the copy sets
+// CheckRedirect directly. For any other remote.Client the redirect policy is
+// out of reach, so GetBody is cleared as a last resort: the standard library
+// then declines to replay a 307/308 body (see net/http redirectBehavior).
+func noFollowRedirects(c remote.Client) remote.Client {
+	stop := func(*http.Request, []*http.Request) error { return http.ErrUseLastResponse }
+	switch client := c.(type) {
+	case *auth.Client:
+		copyClient := *client
+		inner := &http.Client{}
```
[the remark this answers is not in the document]
jakobmoellerdev: Fixed in d9e3c01. `noFollowRedirects` now bases the redirect-suppressing copy on `http.DefaultClient` when `auth.Client.Client` is nil (which is what `auth.Client.client()` resolves to), instead of a bare `&http.Client{}`. A customized default transport, proxy, TLS, timeout, or cookie jar is preserved.

Added `TestNoFollowRedirects_PreservesDefaultClientForNilAuthInner`: it sets a sentinel transport + timeout on `http.DefaultClient`, wraps a nil-inner `auth.Client`, and asserts the copy inherits both and does not mutate the original. Fails pre-fix (transport is nil).

fabianburth (approved):

matthiasbruns (approved):
ack

PR #3643, opened 2026-09-18
feat(transfer): add configurable uploader configurations

jakobmoellerdev:

Adds **configurable uploader configurations** to the transfer subsystem. An uploader routes matched resources through a custom upload transformer instead of the default download → local-blob path, driven by a new config entry carried alongside the transfer config in the central generic config:

```yaml
configurations:
  - type: transfer.config.ocm.software/v1alpha1
    copyMode: allResources
  - type: uploader.transfer.config.ocm.software/v1alpha1
    match:
      accessType: Wget/v1
    stream:
      type: HTTPStreaming/v1alpha1
      targetURL: '${"https://mytarget.registry.com/uploads" + resource.access.path}'
      method: PUT
```

Key pieces:

- **`uploader.transfer.config.ocm.software/v1alpha1`** — a sibling config type carrying independent `match` → `stream` rules. Multiple entries are order-preserving and **first match wins**, so more specific rules should be declared before broader ones. `match` selects by resource **access type** and, optionally, by resource **name** and **extraIdentity** key/value pairs, so several resources of the same access type can be routed to different targets.
- **Works with any source access type** — the source may be wget, OCI, S3, GitHub, … Its content is fetched through the access-type-specific downloader (dispatched by the source access type); only the **target** is always an HTTP endpoint. The `resource` alias exposes every field of the source access dynamically under `resource.access.<field>` (`resource.access.url` for wget, `resource.access.imageReference` for OCI, `resource.access.bucket` for S3, …), plus always-available `resource.name`/`version`/`type`/`extraIdentity`/`labels`. URL-bearing accesses additionally expose parsed `resource.access.path`/`host`/`scheme`.
- **Explicit `${…}` CEL target URLs** — `stream.targetURL` is a [CEL](https://cel.dev/) expression evaluated by the graph runtime, and it **must be wrapped in `${…}`** exactly like every other CEL field in the transfer graph (unwrapped expressions are rejected at build time). The uploader rewrites the `resource` alias to the concrete environment node path with a string-literal-aware identifier scan (so `resource` appearing inside a URL string literal is preserved), then the runtime resolves the final URL. Any static query string goes directly inside the expression.
- **`HTTPStreaming/v1alpha1` transformation** — carries an explicit **source** and **target** resource reference (mirroring `TransferOCIArtifact`'s `Resource`/`TargetResource`). The target resource is built at graph-build time with a `Wget/v1` access whose URL is the CEL expression, so the transfer plan stays declarative and the URL is resolved deterministically from the source resource at execution time.
- **Fused streaming transformer** — resolves source credentials from the resource consumer identity and target credentials independently from the credential graph (by the target URL's wget consumer identity), opens the source stream via the injected `ResourceRepository` (dispatched by access type; never built from the spec), and pipes the body straight into the HTTP request via `io.TeeReader`. The digest is computed **during** the stream, or verified as passthrough when the source already has one — never buffering the body in memory or on disk. On success the output is a new resource with `Wget/v1` access at the resolved target URL and the recorded digest.
- **Wget-native target mapping** — the `stream` block maps field-for-field onto the target `Wget/v1` access (`targetURL`→`URL`, `method`→`Verb`, `header`/`body`/`noRedirect`/`mediaType` pass through).
- **ADR 28 label** — the node renders as `"<component>@<version> [Stream <resource> to <host>]"`, consistent with the other transfer transformations.
- **Deterministic plan** — `fillGraphDefinitionWithPrefetchedComponents` iterates vertices in sorted key order.
- **Validation & error surfacing** — `UploaderConfig.Validate` rejects an empty match access type or a missing/untyped stream; `BuildGraphDefinition` validates every uploader before graph processing (so exported-API callers that bypass `LookupUploaderConfigs` cannot reach a nil-`Stream` dereference); `processUploader` rejects an empty or non-`${…}` `stream.targetURL` at decode time; and non-2xx upload failures include a bounded response-body excerpt. The replication controller looks up and passes uploader configs so they are honored during replication too.

#### Documentation

- Reference: `docs/reference/transfer-configuration.md` documents the uploader config, the explicit `${…}` CEL `resource` alias with dynamic `resource.access.*` fields, and the `stream`→`Wget/v1` target mapping.
- Tutorial: `docs/tutorials/configure-custom-uploads.md` walks through routing resources (any access type) to a custom HTTP upload target by rule.

##### How to test the changes

Configure an uploader as shown above and run `ocm transfer componentversion <src> <target>`; a resource whose access type (and optional name/extraIdentity) matches is streamed to the resolved target URL and re-described with a `Wget/v1` access.

##### Verification

- [x] I have added/updated tests for my changes (unit: uploader lookup + validation, graph shape + source/target refs, explicit-`${}` target-URL rewrite with string-literal preservation, unwrapped-URL rejection, non-wget (OCI) source access-field exposure, name/extraIdentity matching, deterministic plan, transformer digest passthrough + Wget field mapping; integration: full wget→HTTP-PUT roundtrip)
- [x] Tests pass locally (`task test` and `task test/integration` for the affected areas)
- [x] I have tested the changes locally


bindings/go/kubernetes/controller/internal/controller/replication/replication_controller.go
```diff
@@ -344,7 +344,7 @@ func (r *Reconciler) reconcile(ctx context.Context, replication *v1alpha1.Replic
 	// The process takes turns to complete: resolved descriptors are cache hits, each
 	// pass enqueues the next component version of the graph until all component versions are in the cache and
 	// accounted for.
-	tgd, err := transfer.BuildGraphDefinition(ctx, transferCfg, transfer.Mapping{
+	tgd, err := transfer.BuildGraphDefinition(ctx, transferCfg, nil, transfer.Mapping{
```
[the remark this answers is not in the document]
jakobmoellerdev: Fixed in c16835f8d. The replication controller now calls `transferspec.LookupUploaderConfigs(cfg.Config)` alongside the transfer-config lookup and passes the result to `transfer.BuildGraphDefinition` instead of `nil`, so uploader entries are honored during replication.

bindings/go/transfer/internal/uploader.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+// processUploader emits a single HTTPStreaming (or other stream-typed) transformation for
+// resource. The target resource is a copy of the source carrying a Wget access built from
+// the uploader's stream config, so the transfer plan is literal and deterministic.
+func processUploader(resource descriptorv2.Resource, u *transferv1alpha1.UploaderConfig, id string, val *discoveryValue, tgd *transformv1alpha1.TransformationGraphDefinition, resourceTransformIDs map[int]string, i int) error {
+	var cfg streamConfig
+	if err := json.Unmarshal(u.Stream.Data, &cfg); err != nil {
+		return fmt.Errorf("cannot decode uploader stream config: %w", err)
+	}
```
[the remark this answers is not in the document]
jakobmoellerdev: Fixed in c16835f8d. `processUploader` now rejects an empty `stream.targetURL` at decode time (`strings.TrimSpace(cfg.TargetURL) == ""`) before any graph nodes are built, so a Wget access with an empty URL can no longer be appended.

bindings/go/transfer/transfer.go
```diff
@@ -26,6 +26,7 @@ import (
 func BuildGraphDefinition(
 	ctx context.Context,
 	cfg *transferv1alpha1.Config,
+	uploaders []*transferv1alpha1.UploaderConfig,
 	mappings ...Mapping,
 ) (*transformv1alpha1.TransformationGraphDefinition, error) {
 	if err := cfg.Validate(); err != nil {
```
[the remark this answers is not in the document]
jakobmoellerdev: Fixed in c16835f8d. `BuildGraphDefinition` now validates every uploader (`u.Validate()`) before graph processing, returning `invalid uploader config at index %d`, so an exported-API caller that bypasses `LookupUploaderConfigs` cannot reach the nil-`Stream` dereference.

bindings/go/wget/transformation/http_streaming.go
```diff
@@ -0,0 +1,250 @@
...
+		if size := sizer.Size(); size != blob.SizeUnknown {
+			req.ContentLength = size
+		}
+	}
+
+	client := ocmhttp.New(ocmhttp.WithConfig(t.HTTPConfig))
+	if tw.NoRedirect {
+		client = download.CloneClientWithNoRedirect(client)
+	}
+	if err := download.ApplyCredentials(ctx, req, &client, dstCreds); err != nil {
```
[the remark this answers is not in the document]
jakobmoellerdev: This is a valid pre-existing limitation in the shared `wget/internal/download.ApplyCredentials` (and the `bindings/go/http` transport chain), not something introduced by this feature. `ocmhttp.New` wraps the base `*http.Transport` in retry/host-routing/user-agent RoundTrippers, and a `hostRouter` holds a *per-host* map of transports with no single inner `*http.Transport` to augment, so preserving host-routing while injecting an mTLS `tls.Config` requires building the client with the client certificate integrated into `ocmhttp` transport construction. That is a cross-cutting change to the HTTP config types and every wget caller (regular downloads included), out of scope for the uploader-configurations PR. Tracking as a follow-up.

bindings/go/wget/transformation/http_streaming.go
```diff
@@ -0,0 +1,250 @@
...
+		"targetURL", safeURL.String(),
+		"method", method)
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("failed uploading to %s: %w", safeURL.String(), err)
+	}
+	defer func() { _ = resp.Body.Close() }()
+	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
+		return nil, fmt.Errorf("upload to %s returned status %d", safeURL.String(), resp.StatusCode)
```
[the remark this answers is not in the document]
jakobmoellerdev: Fixed in c16835f8d. Non-2xx upload errors now include a bounded response-body excerpt (`io.LimitReader`, `maxErrorBodyBytes = 4KiB`, trimmed) so server validation/API error detail is retained without risking unbounded reads.

bindings/go/wget/transformation/http_streaming.go
```diff
@@ -0,0 +1,250 @@
...
+// resolveSourceCredentials resolves credentials for the source resource by its consumer
+// identity. A missing provider, identity, or ErrNotFound yields nil credentials.
+func (t *HTTPStreamingTransformer) resolveSourceCredentials(ctx context.Context, resource *descriptor.Resource) (runtime.Typed, error) {
+	if t.CredentialProvider == nil {
+		return nil, nil
+	}
+	consumerID, err := t.ResourceRepository.GetResourceCredentialConsumerIdentity(ctx, resource)
+	if err != nil {
+		// A resource without a resolvable identity simply has no credentials.
+		return nil, nil
```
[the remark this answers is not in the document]
jakobmoellerdev: Fixed in c16835f8d. `resolveSourceCredentials` now propagates the `GetResourceCredentialConsumerIdentity` error (`failed deriving source consumer identity: %w`) instead of swallowing it, so a nil/invalid access or unparseable URL is no longer masked by a later download failure.

bindings/go/transfer/internal/uploader.go
```diff
@@ -0,0 +1,277 @@
...
+// celTargetURLField turns a user targetURL CEL expression into a standalone
+// ${...} field value that the graph runtime evaluates. It rewrites the `resource`
+// alias to the concrete environment node path, appends any static query params as
+// a literal suffix, and validates the result is a single CEL expression.
+func celTargetURLField(rawTargetURL, nodePath string, queryParams map[string][]string) (string, error) {
+	trimmed := strings.TrimSpace(rawTargetURL)
+	if trimmed == "" {
+		return "", fmt.Errorf("stream.targetURL is required")
+	}
+	rewritten := aliasPattern.ReplaceAllString(trimmed, "${1}"+nodePath)
```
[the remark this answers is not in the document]
jakobmoellerdev: Fixed in 17ab26f7e. The `resource` alias is now rewritten with a string-literal-aware identifier scan (`rewriteAlias`) instead of a regex: it tracks quote state and only replaces `resource` when it is a bare identifier token (not preceded by an identifier char or `.`, not followed by an identifier char), so an occurrence inside a string literal such as `"https://uploads.example/resource/" + resource.name` keeps the literal `resource/` segment and only rewrites the identifier before `.name`. Added a regression test (`TestBuildGraphDefinition_UploaderPreservesResourceInStringLiteral`).

bindings/go/transfer/internal/uploader.go
```diff
@@ -0,0 +1,277 @@
...
+	if suffix := encodeQuerySuffix(queryParams); suffix != "" {
+		rewritten = "(" + rewritten + ") + " + strconvQuote(suffix)
```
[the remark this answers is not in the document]
jakobmoellerdev: Fixed in 17ab26f7e by removing `queryParams`. Appending a literal `?k=v` suffix could not know whether the runtime-resolved URL already carried a query string, producing the `?token=abc?part=1` double-delimiter you describe. Since `targetURL` is now an explicit CEL expression, a static query string belongs directly inside it (e.g. `${"https://host/path?token=abc&part=1"}` or built with `+`), which removes the entire fragile suffix-append path. Docs and schema updated.

website/content/docs/tutorials/configure-custom-uploads.md
```diff
@@ -0,0 +1,235 @@
...
+an uploader config that matches `Wget/v1` resources and streams them to your target:
+
+```yaml
+type: generic.config.ocm.software/v1
+configurations:
+  - type: transfer.config.ocm.software/v1alpha1
+    copyMode: allResources
+  - type: uploader.transfer.config.ocm.software/v1alpha1
+    match:
+      accessType: Wget/v1alpha1
```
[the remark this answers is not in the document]
jakobmoellerdev: Fixed in 17ab26f7e. The tutorial rule now matches `accessType: Wget/v1` (the registered wget access version), consistent with the prose. I also aligned the reference examples, which had the same `Wget/v1alpha1` mismatch — the version-tolerant matcher would not have selected a real `Wget/v1` resource against a `v1alpha1` rule.

bindings/go/transfer/internal/uploader.go
```diff
@@ -0,0 +1,343 @@
...
+		if match.AccessType.Name != accessType.Name {
+			continue
+		}
+		if match.AccessType.Version != "" && match.AccessType.Version != accessType.Version {
+			continue
+		}
+		if match.Name != "" && match.Name != resource.Name {
+			continue
+		}
+		if !identityContains(identity, match.ExtraIdentity) {
```
matthiasbruns: what about `semver` version constrains here?

bindings/go/transfer/v1alpha1/spec/uploader_test.go
```diff
@@ -0,0 +1,148 @@
...
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	genericv1 "ocm.software/open-component-model/bindings/go/configuration/generic/v1/spec"
+	"ocm.software/open-component-model/bindings/go/runtime"
+	"ocm.software/open-component-model/bindings/go/transfer/v1alpha1/spec"
+)
+
+func TestLookupUploaderConfigs(t *testing.T) {
```
matthiasbruns: could we get a move complex matching test here with names, version etc?

bindings/go/transfer/internal/uploader.go
matthiasbruns: (not in the ADR as far as I see but maybe worth thinking about)

bindings/go/transfer/v1alpha1/spec/uploader.go
```diff
@@ -0,0 +1,124 @@
...
+	Type runtime.Type `json:"type"`
+
+	// Match selects the resources this uploader applies to.
+	Match UploaderMatch `json:"match"`
+
+	// Stream is the transformer specification used to upload matched resources.
+	// It is kept generic so the uploader stays open to different transformer types;
+	// the transformer is selected by Stream's own type (e.g. HTTPStreaming/v1alpha1).
+	Stream *runtime.Raw `json:"stream"`
+}
```
fabianburth: didn't we discuss we'd want dedicated config types per uploader, so we can avoid another sub-type system? 
(so, `wget.uploader.transfer.config.ocm.software/v1alpha1`, `oci.uploader.transfer.config.ocm.software/v1alpha1`, ...)

also, do we even want to add match in our initial implementation? alternative suggestion would be, if `wget` uploader is specified, everything with a `wget` reference hint will be attempted to be uploaded.

bindings/go/transfer/internal/graph.go
```diff
@@ -254,6 +260,14 @@ func processResources(
...
 		}
 
+		// An uploader is an explicit instruction to move a matched resource, so it
+		// runs regardless of copy mode and takes precedence over the default handlers.
+		if u := matchUploader(uploaders, resource); u != nil {
+			if err := processUploader(resource, u, id, val, tgd, resourceTransformIDs, i); err != nil {
+				return nil, nil, fmt.Errorf("cannot process uploader for resource %v: %w", resource.ToIdentity(), err)
+			}
+			continue
+		}
```
fabianburth, line 270: I guess this is to be fixed? Currently, it looks like we would eventually do both, upload to remote and local blob upload?

fabianburth (commented):
Without reference hints, this implementation would only allow uploads for non local blob access.
Are you planning to add reference hints? Or are you planning to keep local blob out of scope for the initial implementation? 

EDIT: that assumption was wrong as the config contains an upload url

matthiasbruns:
I like the approach, not gonna lie
I will test this this evening - once Fabians comments are resolved, I would give it one more review

bindings/go/transfer/v1alpha1/spec/uploader.go
jakobmoellerdev: fixed

bindings/go/transfer/v1alpha1/spec/uploader_test.go
jakobmoellerdev: added

bindings/go/transfer/internal/uploader.go
jakobmoellerdev: probably something we should consider for the matcher behavior in runtime. WDYT?

bindings/go/transfer/internal/graph.go
jakobmoellerdev: Good to double-check, but this cannot do both. In `processResources`, a matched uploader runs `processUploader` and then `continue`s (graph.go:269), so the default download → local-blob path (`processResource`, graph.go:276) is unreachable for that resource — exactly one transformation node per matched resource.

Covered by two exclusive tests:
- `TestBuildGraphDefinition_UploaderMatch_EmitsHTTPStreaming` asserts `sawDownloadWget == false` ("uploader path must not emit a DownloadWgetResource node").
- `TestBuildGraphDefinition_NoUploader_KeepsDownloadWgetPath` asserts the inverse (download path taken, no HTTPStreaming node) without an uploader.

Resolving as already handled.

bindings/go/transform/graph/env/hex.go
```diff
@@ -0,0 +1,63 @@
...
+	"encoding/hex"
+
+	"cel.dev/cel-go/cel"
+	"cel.dev/cel-go/common/types"
+	"cel.dev/cel-go/common/types/ref"
+)
+
+// Hex returns a cel.EnvOption registering the "hex.encode" and "hex.decode"
+// functions, mirroring the base64.encode / base64.decode functions from the CEL
+// encoders extension.
```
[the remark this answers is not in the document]
jakobmoellerdev: Fixed by registering both functions in Hex(). Rather than pointing callers at a second HexEncode() option, Hex() now installs hex.decode and hex.encode together (mirroring how the encoders extension bundles base64.encode/base64.decode), so the doc comment is accurate and one option is enough. HexEncode() removed. Covered by TestHexFunctions (both directions + the full base64.encode(hex.decode(...)) chain).

matthiasbruns:
We might have a problem with digests

What I tried to do:
- download podinfo via oci
- upload via wget to my local http server
- digest mismatch

> Tested OCI → HTTP upload with ghcr.io/stefanprodan/podinfo:6.8.0 and hit a digest mismatch (reproducer below).
> The source resource carries the OCI manifest/index digest, but DownloadResource() returns an OCI-layout tar. HTTPStreamingTransformer hashes the uploaded tar bytes and compares them against that original digest—so it compares two different representations.

This might be related to https://github.com/open-component-model/open-component-model/issues/3674

```bash
#!/bin/sh
# Reproduce PR #3643's OCI -> HTTP uploader digest mismatch.
# Requires Go, Docker, and access to ghcr.io. Change this path to your checkout.
set -eu

ocm_main=/Users/mbruns/dev/sap/open-component-model/bindings/go/cli/main.go
cli_dir=$(dirname -- "$ocm_main")
work=$(mktemp -d)
container=
cleanup() {
  if [ -n "$container" ]; then docker rm -f "$container" >/dev/null; fi
  rm -rf "$work"
}
trap cleanup EXIT
trap 'exit 130' INT
trap 'exit 143' TERM

echo '== Building CLI (this may take a moment)...'
(cd "$cli_dir" && go build -o "$work/ocm" "$ocm_main")
cat > "$work/nginx.conf" <<'EOF'
events {}
http {
  server {
    listen 80;
    client_max_body_size 0;
    location / {
      root /tmp/uploads;
      dav_methods PUT;
      create_full_put_path on;
    }
  }
}
EOF
echo '== Starting HTTP upload target...'
container=$(docker run -d --rm -p 127.0.0.1::80 \
  -v "$work/nginx.conf:/etc/nginx/nginx.conf:ro" nginx:alpine)
# Validate the upload server configuration.
docker exec "$container" nginx -t
address=$(docker port "$container" 80/tcp)

cat > "$work/constructor.yaml" <<'EOF'
components:
- name: ocm.software/oci-upload-repro
  version: 1.0.0
  provider:
    name: ocm.software
  resources:
  - name: podinfo
    version: 6.8.0
    type: ociImage
    relation: external
    access:
      type: OCIImage/v1
      imageReference: ghcr.io/stefanprodan/podinfo:6.8.0
EOF

cat > "$work/config.yaml" <<EOF
configurations:
- type: transfer.config.ocm.software/v1alpha1
  copyMode: allResources
- type: http.uploader.transfer.config.ocm.software/v1alpha1
  match:
    accessType: OCIImage/v1
  targetURL: '\${"http://$address/" + resource.name + ".tar"}'
  method: PUT
EOF

"$work/ocm" add cv --repository "$work/source" --constructor "$work/constructor.yaml"
# Expected: successful upload and a target resource with Wget/v1 access.
# Actual: digest mismatch (OCI manifest/index digest vs. uploaded OCI-layout tar hash).
"$work/ocm" transfer cv "$work/source//ocm.software/oci-upload-repro:1.0.0" \
  "ctf::$work/target" --config "$work/config.yaml"

```

@jakobmoellerdev

matthiasbruns (dismissed):
as discussed - we will fix the bug in another pr 


lgtm

bindings/go/transfer/v1alpha1/spec/uploader.go
```diff
@@ -0,0 +1,140 @@
...
+		return fmt.Errorf("targetURL is required")
+	}
+	return nil
+}
+
+// LookupHTTPUploaderConfigs extracts all HTTP uploader configurations from a central
+// generic config. All entries of type [HTTPUploaderConfigType] are decoded and
+// validated, preserving their declaration order. Returns nil if cfg is nil or
+// contains no HTTP uploader entries.
+func LookupHTTPUploaderConfigs(cfg *genericv1.Config) ([]*HTTPUploaderConfig, error) {
```
fabianburth, line 115: I think the config placement is wrong. So far, I believe we have been completely consistent with having a `LookupConfig` per config type.

bindings/go/transfer/internal/uploader.go
```diff
@@ -0,0 +1,339 @@
...
+		if !accessTypeMatches(u.Match.AccessType, accessType) {
+			continue
+		}
+		if !runtime.IdentitySubset(matchIdentity(u.Match), identity) {
+			continue
+		}
+		return u
+	}
+	return nil
+}
```
fabianburth, line 43: With each uploader being its own config type now, I'm not sure whether this function is still correct. Wouldnt we want or need this to work with all uploaders?

bindings/go/transfer/v1alpha1/spec/uploader.go
```diff
@@ -0,0 +1,140 @@
...
+	// AccessType is the resource access type this uploader matches (e.g. Wget/v1).
+	AccessType runtime.Type `json:"accessType"`
+	// Name optionally restricts the match to resources with this exact name.
+	// When empty, resources of any name match.
+	Name string `json:"name,omitempty"`
+	// ExtraIdentity optionally restricts the match to resources whose identity
+	// contains all of these key/value pairs. When empty, no extra-identity
+	// constraint is applied.
+	ExtraIdentity runtime.Identity `json:"extraIdentity,omitempty"`
+}
```
fabianburth, line 86: Do we need to include version here as well?

bindings/go/transfer/internal/uploader.go
```diff
@@ -0,0 +1,339 @@
...
+		return err
+	}
+
+	// Header values may be CEL-templated against the source resource (e.g. RFC 9530
+	// Repr-Digest or x-checksum-* headers from resource.digest.value). A value wrapped
+	// in ${...} is rewritten and resolved by the graph runtime; a plain value is a literal.
+	header, err := templateHeader(u.Header, nodePath)
+	if err != nil {
+		return err
+	}
```
fabianburth, line 291: Could we evaluate targetURL and templated headers during graph construction instead? The source
 resource is already available here and passed through spec.resource, so environment.uploads
 duplicates it solely for deferred CEL evaluation. Evaluating CEL against resource here and writing
 the resolved values into targetResource would simplify the graph significantly.

bindings/go/transfer/v1alpha1/spec/uploader.go
jakobmoellerdev: Added in 79ea4ccd9. `UploaderMatch` now has a dedicated `match.version` field mirroring `match.name` (both map to the reserved identity attributes and feed the same `runtime.IdentitySubset` check in `matchIdentity`). Previously a version constraint was only expressible indirectly via `extraIdentity: {version: ...}` since `Resource.ToIdentity()` includes `version` — the dedicated field removes that asymmetry. Schema regenerated and covered by two new `TestMatchUploader` cases (matching + non-matching version); reference docs updated.

bindings/go/transfer/v1alpha1/spec/uploader.go
jakobmoellerdev: To confirm the intent so I address this the way you want: the current placement keeps `HTTPUploaderConfig` and its `LookupHTTPUploaderConfigs` together in `v1alpha1/spec/uploader.go`, which is per config type — each uploader config type owns its own lookup, so this doesn't reintroduce a sub-type system. The one aggregation point is `transfer.BuildGraphDefinition`, which today takes a concrete `[]*HTTPUploaderConfig`.

Are you asking to (a) move `uploader.go` out of the `transfer` config package entirely (e.g. next to the wget binding), or (b) keep the per-type `LookupX` functions but have `BuildGraphDefinition` accept a generic uploader abstraction so future `oci.uploader...`/`s3.uploader...` types flow through without a signature change? (b) is the larger change and ties into the `matchUploader` generality point below — happy to do it, but it defines the extension contract for every future uploader type, so I'd rather land the shape you intend. cc @jakobmoellerdev

bindings/go/transfer/internal/uploader.go
jakobmoellerdev: Right — `matchUploader` (and the whole pipeline: `graph.go`, `processUploader`, `BuildGraphDefinition`) is typed to `[]*HTTPUploaderConfig`, so it only matches HTTP uploaders. With one uploader config type today that's correct, but it's not generic over future `oci.uploader...`/`s3.uploader...` types.

Making it work for all uploaders means a shared interface — e.g. `Uploader { Match() UploaderMatch; Process(resource, ...) (transformation, error) }` — that each config type implements, with `matchUploader`/graph iterating `[]Uploader` and each type owning its own `Process`. That's the same decision as the config-placement thread above (the aggregation contract in `BuildGraphDefinition`). It's a clean extension point but a non-trivial refactor that fixes the shape for every future uploader, so I'd like your call on whether to introduce it now or keep the HTTP-only path until a second uploader type actually lands. cc @jakobmoellerdev

bindings/go/transfer/internal/uploader.go
jakobmoellerdev: Confirmed feasible: the CEL env is fully reconstructable at graph-construction time. `env.NewEnvBuilder(...).CurrentEnv()` gives an env with `url()`, `contentDigestAlgorithm()`, `hex.*` and the encoders extension (`builder.go:44-68`), and the source `resource` is already a concrete value here in `processUploader`. So we could bind `resource` to `buildResourceNode(resource)`, evaluate `targetURL` + each header eagerly, and write the resolved literal strings straight into `targetResource.Access`. That drops `environment.uploads` duplication, the `celTargetURLField`/`rewriteAlias` string-rewrite path, and `templateHeader`'s deferred branch — a real simplification.

The tradeoff is that resolution moves from graph-runtime to build-time: CEL errors surface earlier (arguably better), but the transformation spec then carries resolved literals rather than expressions, so it loses the ability to reference any value only known at graph execution. For uploads everything is known at build time, so I don't see a functional blocker.

@jakobmoellerdev this touches the transformation spec shape (`HTTPStreaming` carrying resolved URL/headers vs. CEL), so I'd like your sign-off before I refactor — it also overlaps with the generic-uploader-interface direction in the two threads above. If you're good with it I'll fold all three into one follow-up refactor commit.

fabianburth (dismissed):

bindings/go/transfer/internal/uploader.go
```diff
@@ -0,0 +1,344 @@
...
+		return fmt.Errorf("cannot convert target wget access: %w", err)
+	}
+
+	targetResource := *resource.DeepCopy()
+	targetResource.Access = targetAccessRaw
+
+	spec, err := runtime.UnstructuredFromMixedData(map[string]any{
+		"resource":       resource,
+		"targetResource": targetResource,
+	})
```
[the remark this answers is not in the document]
jakobmoellerdev: https://github.com/open-component-model/open-component-model/pull/3643#issuecomment-5790058875 is the root. we will address this in a separate PR

bindings/go/transfer/internal/uploader.go
```diff
@@ -0,0 +1,217 @@
...
+	mediaType := u.MediaType
+	if mediaType == "" {
+		mediaType = mediaTypeFromAccess(resource)
+	}
+	// Build the target access from the raw user strings; expression templating is applied
+	// generically to the whole object below rather than to hand-picked fields.
+	targetAccess := &wgetaccessv1.Wget{
+		Type:       wgetaccess.V1VersionedType,
+		URL:        u.TargetURL,
+		Verb:       u.Method,
```
matthiasbruns: The field mapping is intentional, but `method: PUT` also becomes the download method. A subsequent CLI download sends an empty PUT, overwrites the object, and exits successfully. Omitting `method` works.

Can we separate the upload request from the published download access?

Tested with [explicit-put.sh](https://github.com/matthiasbruns/open-component-model/blob/adabee73f0b52cdd8ea97df414239a8563886112/bindings/go/cli/integration/uploader-review/explicit-put.sh) and [default-method.sh](https://github.com/matthiasbruns/open-component-model/blob/adabee73f0b52cdd8ea97df414239a8563886112/bindings/go/cli/integration/uploader-review/default-method.sh).

bindings/go/wget/transformation/http_streaming.go
```diff
@@ -0,0 +1,260 @@
...
+	}
+
+	slog.InfoContext(ctx, "streaming resource to HTTP target",
+		"resource", srcResource.ToIdentity(),
+		"targetURL", safeURL.String(),
+		"method", method)
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("failed uploading to %s: %w", safeURL.String(), err)
```
matthiasbruns: `safeURL` sanitizes the outer message, but the wrapped `*url.Error` still includes the query token in CLI logs and the final error. Can we sanitize the nested error too?

Tested with [query-token.sh](https://github.com/matthiasbruns/open-component-model/blob/adabee73f0b52cdd8ea97df414239a8563886112/bindings/go/cli/integration/uploader-review/query-token.sh).

bindings/go/transfer/internal/uploader.go
```diff
@@ -0,0 +1,217 @@
...
+	keys := slices.Sorted(maps.Keys(identity))
+	predicates := make([]string, 0, len(keys))
+	for _, k := range keys {
+		switch k {
+		case descriptorv2.IdentityAttributeName:
+			predicates = append(predicates, fmt.Sprintf("%s.name == %s", resourceFilterVar, strconv.Quote(identity[k])))
+		case descriptorv2.IdentityAttributeVersion:
+			predicates = append(predicates, fmt.Sprintf("%s.version == %s", resourceFilterVar, strconv.Quote(identity[k])))
+		default:
+			predicates = append(predicates, fmt.Sprintf("%s.extraIdentity[%s] == %s", resourceFilterVar, strconv.Quote(k), strconv.Quote(identity[k])))
```
matthiasbruns: Selecting a resource with `extraIdentity` and referencing `resource.name` fails with `undefined field 'extraIdentity'` in the generated selector. A constant target URL works.

Could we bind the selected resource directly instead?

Tested with [extra-identity.sh](https://github.com/matthiasbruns/open-component-model/blob/adabee73f0b52cdd8ea97df414239a8563886112/bindings/go/cli/integration/uploader-review/extra-identity.sh).

matthiasbruns (dismissed):
Three priority findings from the CLI end-to-end runs: explicit-PUT readback overwrites the uploaded object, transport errors expose query tokens, and extra-identity resource expressions fail. Tested against 73bcddec1; I checked the current head (5812f62c5), and the relevant code paths remain unchanged. Focused shell scripts are linked in each comment.

fabianburth (dismissed):

bindings/go/wget/transformation/http_streaming.go
```diff
@@ -0,0 +1,260 @@
...
+		// Include a bounded excerpt of the response body to aid debugging without
+		// risking unbounded memory use on a hostile or verbose server.
+		excerpt, _ := io.ReadAll(io.LimitReader(resp.Body, maxErrorBodyBytes))
+		if len(excerpt) > 0 {
+			return nil, fmt.Errorf("upload to %s returned status %d: %s", safeURL.String(), resp.StatusCode, strings.TrimSpace(string(excerpt)))
+		}
+		return nil, fmt.Errorf("upload to %s returned status %d", safeURL.String(), resp.StatusCode)
+	}
+
+	// Finalize the digest only after the whole body has been streamed to the target.
```
Skarlso: Err. At that point didn't we deliver the potentially corrupted resource already to fail _now_ after it's been done?

bindings/go/examples/08_transfer_test.go
```diff
@@ -96,7 +96,7 @@ func TestExample_TransferCTFtoCTF(t *testing.T) {
 	// transfer.NewRepositoryResolver wraps the source repo directly, so no custom resolver is needed.
 	// A nil cfg uses the defaults; see TestExample_Transfer_WithTransferConfig
 	// below for driving transfer settings explicitly.
-	tgd, err := transfer.BuildGraphDefinition(ctx, nil,
+	tgd, err := transfer.BuildGraphDefinition(ctx, nil, nil,
```
Skarlso: Technically a breaking change.

bindings/go/transfer/internal/uploader.go
```diff
@@ -0,0 +1,217 @@
...
+	if mediaType == "" {
+		mediaType = mediaTypeFromAccess(resource)
+	}
+	// Build the target access from the raw user strings; expression templating is applied
+	// generically to the whole object below rather than to hand-picked fields.
+	targetAccess := &wgetaccessv1.Wget{
+		Type:       wgetaccess.V1VersionedType,
+		URL:        u.TargetURL,
+		Verb:       u.Method,
+		Header:     u.Header,
```
Skarlso: Errr. This puts this into the access, right? Does that mean that headers like `Authorization: Bearer` and such will end up in the access verbatim? I know we use creds through OCM but I think this still leaks any headers. But looking at what Matthias is saying I think that fix fixes this as well.

bindings/go/transfer/internal/uploader.go
Skarlso: I think this is even worse, I tried to modify the e2e test and it failed with this:

```
cannot compile expression "...filter(r, r.extraIdentity["architecture"] == "amd64" && ...)"
ERROR: found no matching overload for '_[_]' applied to
  '(__underscores__type_environment.<id>.component.resources.@idx.extraIdentity, string)'
```

bindings/go/wget/transformation/http_streaming.go
```diff
@@ -0,0 +1,260 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed creating upload request: %w", err)
+	}
+	for k, vals := range tw.Header {
+		for _, v := range vals {
+			req.Header.Add(k, v)
+		}
+	}
+	if contentType := targetContentType(tw, srcBlob); contentType != "" {
+		req.Header.Set("Content-Type", contentType)
```
Skarlso: I think we are overriding any potential user headers here, right?

Skarlso (changes requested):
Agreed with Matthias' findings and found a couple more things while running through the e2e tests and modifying it here and to see what happens. :D

bindings/go/wget/transformation/http_streaming.go
jakobmoellerdev: we cannot do it before because streaming means we cannot buffer unfortunately

bindings/go/examples/08_transfer_test.go
jakobmoellerdev: I think lib breaks we dont target as breaking here no?

bindings/go/transfer/internal/uploader.go
jakobmoellerdev: Correct — and this is now safe. The uploader `header` is applied to the **upload request** only (`spec.request`), consumed by the transformer for the outbound PUT. The **published** access recorded on the transferred resource (`spec.targetResource.access`) carries only `url` + `mediaType`, never verb/body/headers — so `Authorization: Bearer` / `Content-Digest` etc. never end up in the persisted access. Fixed by the request/published split in 79c2d1e (see `processHTTPUploader`: `requestAccess` vs `publishedAccess`).

bindings/go/wget/transformation/http_streaming.go
jakobmoellerdev: Good catch — that was a real bug. `targetContentType` derives a `Content-Type` from the media type and used `Set`, which overrode a `Content-Type` the user supplied via the uploader `header`. Fixed in 79c2d1e: the media-type default is now only applied when no `Content-Type` was set by the request headers, so an explicit user `Content-Type` reaches the target unchanged. Covered by `TestHTTPStreamingTransformer_UserContentTypeHeaderPreserved`.

bindings/go/transfer/internal/uploader.go
jakobmoellerdev: I think I need to change the r pointer to index so this isnt such a pain

bindings/go/transfer/internal/uploader.go
jakobmoellerdev: yeah changed to index access since the field cel expressions were super hard to get stable

bindings/go/wget/transformation/http_streaming.go
Skarlso: that's fair, yah :/

bindings/go/examples/08_transfer_test.go
Skarlso: OH WAIT! This is now one library. :O :O :O :O

Skarlso (approved):

fabianburth (approved):

PR #3644, opened 2026-09-18
feat(wget): add checksum policy for wget inputs

jakobmoellerdev:

Adds a **checksum policy** to the `Wget/v1` **input** type so that, when an input carries no digest of its own, an expected checksum for the downloaded content can be obtained and verified from different sources — modelled on [Maven's expected-checksum strategies](https://maven.apache.org/resolver/expected-checksums.html) and IETF standards:

```yaml
resources:
- name: release-archive
  type: blob
  version: 1.0.0
  input:
    type: Wget/v1
    url: https://repo1.maven.org/maven2/xom/xom/1.3.9/xom-1.3.9.jar
    checksumPolicy:
      onMissing: fail          # fail | compute (default: fail when a policy is set)
      sources:                 # tried in order; first that yields a checksum wins
        - type: httpHeader     # "Remote Included": checksum in the response headers
        - type: externalUrl    # "Remote External": checksum from <url>.<ext>
          algorithms: [sha256, sha1]
```

Key pieces:

- **`httpHeader` ("Remote Included")** — reads the checksum from the download response headers: the IETF-standard [RFC 9530](https://www.rfc-editor.org/rfc/rfc9530) `Content-Digest` field and the non-standard `x-checksum-{sha256,sha1,md5}` family (plus the `x-goog-meta-*`/`x-amz-meta-*` variants). Extra header names can be added via `headers`.
- **`externalUrl` ("Remote External")** — fetches a sibling checksum resource from a separate URL, by default `<url>.<ext>` (`.sha256`, `.sha1`, …), overridable with a `urlTemplate` using `{{.url}}`/`{{.ext}}`. The file may be a bare hex digest or GNU coreutils format (`<hex>  <name>`).
- **`stream`** — no expected checksum; the digest is computed from the download.
- **`onMissing`** — `fail` (default when a policy is set) or `compute` (fall back to a stream digest without verification). With no policy configured, behaviour is unchanged.
- **Verification and storage are decoupled** — a policy may verify the transferred bytes against any supported algorithm (Maven commonly ships SHA-1 or MD5, RFC 9530 may offer SHA-512), but the digest recorded on the resource is **always SHA-256** with the `genericBlobDigest/v1` normalisation. A non-SHA-256 transport checksum therefore never leaks a weak or non-canonical algorithm into the component descriptor or into OCI storage/signing. A mismatch fails construction before anything is stored.
- **Pinned-digest verification** — a resource that carries a `digest` (the standard OCM digest info) is now verified against the download on the wget **input** path, fulfilling the documented `resource.digest` contract that the input path previously did not enforce. A pinned digest and a checksum policy are each verified independently against the downloaded bytes (never against each other).

Implementation notes:

- New `bindings/go/wget/checksum` package: an algorithm registry (OCM name ↔ RFC 9530 key ↔ Maven file extension ↔ `crypto.Hash`), an RFC 9530 `Content-Digest` parser, a legacy `x-checksum-*` header parser, an external-URL checksum fetcher, and a transport-independent policy resolver.
- The shared `download` core computes every required hash in a single streaming pass (`io.MultiWriter`, no re-read) and exposes the response headers plus a precalculated digest on the returned blob.
- Scope is the wget **input** method only; the access/repository digest path is unchanged.

##### How to test the changes

Author a component constructor with a `Wget/v1` input and a `checksumPolicy` (as above), then run `ocm add componentversions --create ...` / `ocm transfer ...`. A resource whose download matches the obtained checksum is embedded as a local blob with a SHA-256 digest; a mismatch fails the build.

##### Verification

- [x] I have added/updated tests for my changes (unit: algorithm registry, RFC 9530 + legacy header parsers, external-file parser, Verify/Select, policy resolver; input method: RFC 9530 header, `x-checksum-sha1`, external sibling URL, transport mismatch, onMissing fail/compute, SHA-1-verify-stores-SHA-256, pinned-digest match/mismatch/unsupported-algorithm/combined-with-policy)
- [x] Tests pass locally (`task bindings/go:test` for the affected area)
- [x] I have tested the changes locally


bindings/go/wget/checksum/policy.go
```diff
@@ -0,0 +1,150 @@
...
+			}
+		case SourceExternalURL:
+			if in.Fetch == nil {
+				continue
+			}
+			algs := src.Algorithms
+			if len(algs) == 0 {
+				algs = All
+			}
+			exp, ok, ferr := in.Fetch(ctx, in.URL, algs)
```
jneisener: I think this path is missing the handling for `URLTemplate`

jneisener: Fetcher is not aware of the template either https://github.com/open-component-model/open-component-model/pull/3644/changes#diff-951b1cc99ed2bad2211e8658293474da44140f02e408fef5a897ed07c08bcd3aR163

bindings/go/wget/checksum/algorithm.go
```diff
@@ -0,0 +1,124 @@
...
+//     from a separate URL (e.g. <url>.sha256), as used by Maven repositories.
+//   - stream: no expected checksum; the digest is computed from the stream.
+//
+// Verification and storage are deliberately decoupled. A policy may verify the
+// transferred bytes against any supported algorithm (Maven commonly ships SHA-1
+// or MD5), but the digest recorded on the OCM resource is always SHA-256 with the
+// genericBlobDigest/v1 normalisation, so a non-SHA-256 transport checksum never
+// leaks a weak or non-canonical algorithm into the component descriptor or OCI
+// storage. See [Algorithm] for the supported set.
+package checksum
```
matthiasbruns, line 17: don't we want to use this for maven as well? might be worth making it reusable

bindings/go/wget/checksum/algorithm.go
```diff
@@ -0,0 +1,124 @@
...
+	_ "crypto/sha512"
+)
+
+// Algorithm identifies a checksum algorithm across the three naming schemes this
+// package bridges: the OCM digest name (as stored on descriptor.Digest), the
+// RFC 9530 structured-field key, and the Maven external-checksum file extension.
+type Algorithm struct {
+	// OCMName is the canonical OCM hash algorithm name (e.g. "SHA-256"), matching
+	// descriptor.Digest.HashAlgorithm.
+	OCMName string
```
matthiasbruns: why not just `Name`?

bindings/go/wget/checksum/algorithm.go
```diff
@@ -0,0 +1,124 @@
...
+	RFC9530Key string
+	// Extension is the file extension (without a dot) used by Maven "Remote
+	// External" checksum files (e.g. "sha256").
+	Extension string
+	// Hash is the crypto.Hash used to compute the checksum.
+	Hash crypto.Hash
+}
+
+// New returns a fresh hash.Hash for the algorithm.
+func (a Algorithm) New() interface{ Write([]byte) (int, error) } { //nolint:ireturn // returns hash.Hash intentionally.
```
matthiasbruns: That's the `io.Writer` interface - why aren't we using this again?

bindings/go/wget/checksum/external.go
```diff
@@ -0,0 +1,121 @@
...
+	return Expected{}, false, nil
+}
+
+// renderChecksumURL substitutes the {{.url}} and {{.ext}} tokens. It avoids
+// text/template so the extension cannot inject template actions and to keep the
+// substitution obvious.
+func renderChecksumURL(tmpl, baseURL, ext string) (string, error) {
+	rendered := strings.ReplaceAll(tmpl, "{{.url}}", baseURL)
+	rendered = strings.ReplaceAll(rendered, "{{.ext}}", ext)
+	if strings.Contains(rendered, "{{") {
```
matthiasbruns: wouldn't it make more sense to check if there is still a `{` or `}` in the string? or use https://pkg.go.dev/text/template

bindings/go/wget/input/method.go
```diff
@@ -109,7 +127,120 @@ func (i *InputMethod) ProcessResource(ctx context.Context, resource *constructor
...
+	// input path, which otherwise stores the blob without verifying it.
+	if provided != nil {
+		if err := verifyProvidedDigest(provided, data); err != nil {
+			_ = data.Close()
+			return nil, fmt.Errorf("provided digest verification failed for wget input from %q: %w", wget.URL, err)
+		}
+	}
+
+	if hasPolicy {
+		if err := i.verifyChecksum(ctx, client, &wget, policy, data); err != nil {
```
jneisener: This calls its own `Fetch` instance and doesn't use `opts` from above, so the digest download seems to happen without authentication

bindings/go/wget/input/method.go
```diff
@@ -109,7 +127,120 @@ func (i *InputMethod) ProcessResource(ctx context.Context, resource *constructor
...
+	}
+	for _, src := range spec.Sources {
+		s := checksum.Source{
+			Type:    checksum.SourceType(src.Type),
+			Headers: src.Headers,
+		}
+		// Unknown extensions are dropped here; RequiredAlgorithms/Resolve then fall
+		// back to the full supported set, and an unresolvable checksum surfaces via
+		// the onMissing behaviour.
+		if algs, err := checksum.AlgorithmsFromExtensions(src.Algorithms); err == nil {
```
jneisener: Errors due to invalid algorithms returned in `AlgorithmsFromExtensions` are swallowed here
```suggestion
   algs, err := checksum.AlgorithmsFromExtensions(src.Algorithms)                                                                                                                          
   if err != nil {                                                                                                                                                                         
       return checksum.Policy{}, false, fmt.Errorf("invalid checksum algorithm in source %d: %w", i, err)                                                                                  
   }                                                                                                                                                                                       
   s.Algorithms = algs  
```

bindings/go/wget/checksum/algorithm.go
jakobmoellerdev: Package is intentionally shaped so it could serve a Maven flow (RFC 9530 key, Extension, OCMName all bridged in one place). No refactor in this PR — the wget input is the only current caller — but it is set up for reuse if/when a Maven-specific access/input type lands.

bindings/go/wget/checksum/algorithm.go
jakobmoellerdev: `OCMName` is deliberately disambiguated from `RFC9530Key` and `Extension`, which are also names/keys for the algorithm. Renaming to `Name` alone loses that disambiguation. Happy to reconsider if the trade-off looks different from the consumer side.

bindings/go/wget/checksum/algorithm.go
jakobmoellerdev: Good catch — widened to `hash.Hash` in ffe6a69 to match `download.DigestAlgorithm.New`. Also wired `alg.New` into the digest-algorithms loop in `method.go` for consistency.

bindings/go/wget/checksum/external.go
jakobmoellerdev: The whole `{{.url}}/{{.ext}}` mini-language is gone — the field is now a CEL expression evaluated with `resource.url.*`, `ext`, `alg` variables (see the latest revision of `wget/input/checksum_url.go`), so the manual `{}` guard is no longer needed. Uses the same `${…}` shape as the uploader `stream.targetURL` in #3643.

bindings/go/wget/checksum/source.go
```diff
@@ -0,0 +1,206 @@
...
+	"x-checksum-sha512":           SHA512,
+	"x-checksum-sha1":             SHA1,
+	"x-checksum-md5":              MD5,
+	"x-goog-meta-checksum-sha256": SHA256,
+	"x-goog-meta-checksum-sha1":   SHA1,
+	"x-goog-meta-checksum-md5":    MD5,
+	"x-amz-meta-checksum-sha256":  SHA256,
+	"x-amz-meta-checksum-sha1":    SHA1,
+	"x-amz-meta-checksum-md5":     MD5,
+}
```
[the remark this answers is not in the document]
jakobmoellerdev: Fixed in b21225c — `legacyChecksumHeaders` is now an ordered slice; strongest-first per family, `x-checksum-*` before the `x-goog-meta-*` / `x-amz-meta-*` variants.

bindings/go/wget/input/method.go
```diff
@@ -109,7 +127,120 @@ func (i *InputMethod) ProcessResource(ctx context.Context, resource *constructor
...
+// verifyChecksum resolves the checksum policy against the completed download and
+// verifies the transferred bytes against the first source that yields an expected
+// checksum. Recording the resulting digest on the blob is handled by the caller.
+func (i *InputMethod) verifyChecksum(ctx context.Context, client *nethttp.Client, wget *v1.Wget, policy checksum.Policy, data *download.Blob) error {
+	fetcher := &checksum.ExternalFetcher{Client: client}
+	_, _, err := checksum.Resolve(ctx, policy, checksum.Input{
+		URL:      wget.URL,
+		Headers:  data.Headers(),
+		Computed: data.Digests(),
+		FetchURL: fetcher.FetchURL,
```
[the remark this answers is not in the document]
jakobmoellerdev: Fully addressed across a16d986 and this PR: `verifyChecksum` now uses one undecorated client for cross-origin/HTTP checksum URLs and a credential-decorated client (mTLS transport cloned once, redirects hard-disabled) only for same-origin HTTPS. `Authorization` never reaches a different origin. See `sameOriginHTTPS` in `wget/input/method.go`.

bindings/go/wget/internal/download/download.go
```diff
@@ -150,7 +152,21 @@ func Download(ctx context.Context, req Request, opts ...Option) (_ *Blob, err er
...
+		h := alg.New()
+		hashers[alg.Name] = h
+		writers = append(writers, h)
+	}
+	dst := io.Writer(file)
+	if len(writers) > 1 {
+		dst = io.MultiWriter(writers...)
+	}
+
+	written, err := io.Copy(dst, respBody)
```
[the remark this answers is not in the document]
jakobmoellerdev: Fixed in b21225c — when `WithDigestAlgorithms` is set, `Download` now sets `Accept-Encoding: identity` on the request, so Go`s transport does not auto-decompress and the computed digest matches an RFC 9530 `Content-Digest` computed over the representation data.

bindings/go/wget/input/method.go
jakobmoellerdev: Addressed in 9184612 — `toChecksumPolicy` now returns `fmt.Errorf("checksum policy source #%d: %w", i, err)` on both the algorithm-parse and CEL-compile paths, and `ProcessResource` surfaces it as `invalid checksum policy for wget input from %q: %w`. Same substance as your suggestion; the wording just says "policy source #" instead of "algorithm in source" so it also covers the CEL-URL error path.

bindings/go/wget/internal/download/download.go
```diff
@@ -98,12 +100,23 @@ func Download(ctx context.Context, req Request, opts ...Option) (_ *Blob, err er
...
+	// content-decoding: Go's default transport auto-adds Accept-Encoding: gzip
+	// and decompresses transparently, but RFC 9530 Content-Digest is computed
+	// over the representation-data (i.e. the transferred, content-coded bytes).
+	// Digesting decoded bytes while checking against a server-provided digest
+	// over encoded bytes causes a spurious mismatch. Setting the header
+	// explicitly (even to "identity") suppresses the transparent-decompression
+	// path in net/http.
+	if len(o.DigestAlgorithms) > 0 && httpReq.Header.Get("Accept-Encoding") == "" {
+		httpReq.Header.Set("Accept-Encoding", "identity")
+	}
```
jneisener: If a user specifies
```
header:                  
  Accept-Encoding: [gzip]
```
we don't set the header to `identity` and subsequently can get a spurious mismatch again

jneisener: I think we would have to check the encoding on the response to validate if it is digestable

jneisener: Though this seems like an edgecase

fabianburth:
the input type does not solve it for use cases that need an `Wget` access instead of `LocalBlob`, as we currently have the concept `input -> local blob`

**discussed options:**
- allow `input -> (wget/remote) access` 
  _technically, there is no reason why not, purely a conceptual thing we have taught our users so far and thus, might lead to confusion_
- put the digest processor config into the ocm config
   _we believe the constructor is the only place where the digest calculation has to happen, so UX wise, this feels like the better place to configure it_
- only implement the static behavior (might not be sufficient)

bindings/go/wget/internal/download/blob.go
```diff
@@ -21,17 +22,56 @@ import (
...
+
+// HasPrecalculatedDigest reports whether a precalculated digest was set.
+func (b *Blob) HasPrecalculatedDigest() bool {
+	return b.precalculated != ""
+}
+
+// SetPrecalculatedDigest sets the digest returned by [Blob.Digest] verbatim.
+func (b *Blob) SetPrecalculatedDigest(digest string) {
+	b.precalculated = digest
+}
```
jneisener: https://github.com/open-component-model/open-component-model/blob/be9a9cf296780f3de6c5f1aac1a9921c9128256c/bindings/go/blob/blob.go#L85-L87

Should be `atomic.Pointer[string]` like https://github.com/open-component-model/open-component-model/blob/be9a9cf296780f3de6c5f1aac1a9921c9128256c/bindings/go/blob/inmemory/blob.go#L35-L37

bindings/go/wget/input/method.go
```diff
@@ -109,7 +127,225 @@ func (i *InputMethod) ProcessResource(ctx context.Context, resource *constructor
...
+		return fmt.Errorf("unsupported provided hash algorithm %q: only %s is supported", provided.HashAlgorithm, checksum.StorageAlgorithm.OCMName)
+	}
+	if provided.NormalisationAlgorithm != "" && provided.NormalisationAlgorithm != genericBlobDigestV1 {
+		return fmt.Errorf("unsupported provided normalisation algorithm %q: only %s is supported", provided.NormalisationAlgorithm, genericBlobDigestV1)
+	}
+	computed, ok := data.Digests()[checksum.StorageAlgorithm.OCMName]
+	if !ok || computed == "" {
+		return fmt.Errorf("no computed %s digest available to verify the provided digest against", checksum.StorageAlgorithm.OCMName)
+	}
+	// The provided value may be bare hex or go-digest form (sha256:<hex>).
```
jneisener, line 178: Preexisting issue, but wanted to mention it: `NewDigestFromEncoded` prefixes with the `sha256:` regardless of whether its already prefixed

```
// in bindings/go/oci/blob
package blob

import (
	"strings"
	"testing"

	"github.com/stretchr/testify/assert"

	"ocm.software/open-component-model/bindings/go/blob/inmemory"
	descriptor "ocm.software/open-component-model/bindings/go/descriptor/runtime"
)

func TestPrefixedPinnedDigest(t *testing.T) {
	precalc := "sha256:2cf24dba5fb0a30e26e83b2ac5b9e29e1b161e5c1fa7425e73043362938b9824"

	res := &descriptor.Resource{}
	res.Digest = &descriptor.Digest{
		HashAlgorithm:          "SHA-256",
		NormalisationAlgorithm: "genericBlobDigest/v1",
		Value:                  "sha256:2cf24dba5fb0a30e26e83b2ac5b9e29e1b161e5c1fa7425e73043362938b9824",
	}

	b := inmemory.New(strings.NewReader("hello"))
	b.SetPrecalculatedDigest(precalc) // Stand-in for `ProcessResource` setting on the download blob ("sha256:" + sha).

	d, derr := digestSpecToDigest(res.Digest)
	assert.NoError(t, derr)
	assert.Equal(t, precalc, d) // Actual   :digest.Digest("sha256:sha256:2cf24dba5fb0a30e26e83b2ac5b9e29e1b161e5c1fa7425e73043362938b9824")
}
```

jakobmoellerdev: Nice catch — that's a pre-existing bug in `bindings/go/oci/blob.digestSpecToDigest`: it calls `digest.NewDigestFromEncoded(algo, dig.Value)` even when `dig.Value` already carries the algorithm prefix, so a user who pins `{value: "sha256:xxx"}` gets a double-prefixed OCI digest.

The wget input path in this PR side-steps that specific bug — `verifyProvidedDigest` normalises the pinned value with `strings.TrimPrefix(provided.Value, "sha256:")` before comparison, and `SetPrecalculatedDigest("sha256:" + sha)` only ever gets bare hex from `data.Digests()`. So no regression added here.

Fixing `digestSpecToDigest` itself belongs in a separate PR against `bindings/go/oci/blob/artifact_blob.go` — happy to file it, but keeping it out of this branch to keep the wget-checksum-policy scope tight.

bindings/go/wget/internal/download/download.go
jakobmoellerdev: Fixed by unconditionally forcing `Accept-Encoding: identity` when digests are requested, overriding any Accept-Encoding on the wget spec. Reasoning in the code comment: a user `Accept-Encoding: gzip` on a wget input either causes the RFC 9530 spurious mismatch you flagged or, worse, makes the stored OCM SHA-256 cover a compressed representation that diverges across mirrors serving different encodings. Simpler than sniffing `Content-Encoding` on the response, and it matches the ‘storage digest is over the canonical representation' contract.

bindings/go/wget/internal/download/blob.go
jakobmoellerdev: Fixed — `Blob.precalculated` is now `atomic.Pointer[string]` (`Load`/`Store`), matching `blob/inmemory.Blob`'s pattern. `Digest`/`HasPrecalculatedDigest`/`SetPrecalculatedDigest` all go through the atomic accessors.

bindings/go/wget/spec/config/v1alpha1/config.go
```diff
@@ -0,0 +1,183 @@
...
+//
+// Callers should treat the returned policy as read-only.
+func (c *Config) PolicyForURL(rawURL string) *inputv1.ChecksumPolicy {
+	if c == nil {
+		return nil
+	}
+	u, err := url.Parse(rawURL)
+	if err == nil && u.Host != "" && len(c.Hosts) > 0 {
+		for _, key := range hostKeys(u.Host) {
+			if hc := c.Hosts[key]; hc != nil && hc.ChecksumPolicy != nil {
```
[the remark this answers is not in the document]
jakobmoellerdev: Fixed in 64df6d1. `PolicyForURL` now lowercases both sides before the host-map lookup (RFC 3986 §3.2.2). Added `TestPolicyForURL_MixedCaseHostnameMatchesLowercasedConfigKey` covering both directions — mixed-case URL against lowercased config key and vice versa.

bindings/go/wget/repository/resource_repository.go
```diff
@@ -196,9 +234,67 @@ func (r *ResourceRepository) ProcessResourceDigest(ctx context.Context, resource
...
 		return nil, fmt.Errorf("unsupported normalisation algorithm: expected %s, got %s", genericBlobDigestV1, resource.Digest.NormalisationAlgorithm)
 	}
-	if resource.Digest.Value != dig.Encoded() {
-		return nil, fmt.Errorf("digest mismatch: expected %s, got %s", resource.Digest.Value, dig.Encoded())
+	if resource.Digest.Value != sha {
+		return nil, fmt.Errorf("digest mismatch: expected %s, got %s", resource.Digest.Value, sha)
 	}
 
 	return resource, nil
 }
```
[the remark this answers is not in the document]
jakobmoellerdev: Fixed in 64df6d1 — `ProcessResourceDigest` now strips the optional `sha256:` prefix on `resource.Digest.Value` before comparing against the bare-hex computed digest, matching `verifyProvidedDigest` on the input path. Regression test `TestProcessResourceDigest_AcceptsPrefixedPinnedDigest` exercises the go-digest-form pinned value.

bindings/go/wget/input/method.go
jakobmoellerdev: Fixed in 45c7574b1 — `digestSpecToDigest` now strips a matching algorithm prefix from `dig.Value` before handing it to `NewDigestFromEncoded`, and rejects a mismatched prefix rather than silently double-encoding. Your exact repro case (bare hex → prefixed `sha256:...` pin) is now a regression test in `artifact_blob_test.go` (`TestNewResourceBlobWithMediaType_DigestValidation/prefixed_pinned_digest_matches_blob`); it fails pre-fix with `sha256:sha256:...` and passes post-fix. A second subtest covers the mismatched-prefix case (`sha512:` value declared as `SHA-256`) — now an explicit error.

bindings/go/wget/internal/download/download.go
jakobmoellerdev: Already addressed after your original observation on `8a68ebb`. Line 116 now uses `Header.Set("Accept-Encoding", "identity")`, which *replaces* any user-supplied `Accept-Encoding: gzip` from the `wget` spec header block (verified: `http.Header.Set` clears prior values). So the digest scope stays over identity-encoded bytes regardless of what the input spec asked for.

bindings/go/wget/input/method.go
```diff
@@ -26,6 +34,12 @@ type InputMethod struct {
 	// HTTPConfig configures the HTTP client (timeouts, retries, TLS, routing) used for
 	// downloads. When nil, a default client is used.
 	HTTPConfig *httpv1alpha1.Config
+	// WgetConfig steers the wget behavioural knobs — today, the default
+	// [inputv1.ChecksumPolicy] that applies when a resource's spec does not
+	// carry its own checksumPolicy, and per-host overrides thereof. When nil,
+	// the input honours only spec-level policies; the access-side digest
+	// processor honours the same config.
+	WgetConfig *wgetconfigv1alpha1.Config
```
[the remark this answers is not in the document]
jakobmoellerdev: Stale — this observation predates the wiring fix. `builtin.Register` now takes a `*checksumhttpv1alpha1.Config`, the CLI setup resolves it via `checksumhttpv1alpha1.LookupConfig`, and `wget.Register` threads the same value into both `wgetinput.InputMethod.WgetConfig` and `wgetrepository.WithWgetConfig`. Confirmed in `bindings/go/cli/internal/plugin/builtin/wget/register.go:26,35,47` and `builtin.go:27,64`.

bindings/go/configuration/checksum/http/v1alpha1/spec/config.go
```diff
@@ -0,0 +1,197 @@
...
+// An entry keyed by the bare hostname applies to every port on that host, and
+// one keyed "host:port" applies to that port alone and wins where both are
+// present — the same rule http.config.ocm.software uses.
+func hostKeys(host string) []string {
+	if host == "" {
+		return nil
+	}
+	host = strings.ToLower(host)
+	if name := (&url.URL{Host: host}).Hostname(); name != host {
+		return []string{host, name}
```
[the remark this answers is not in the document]
jakobmoellerdev: Fixed. `hostKeys` and a new `normalizeHostKey` helper now strip one terminal dot from a DNS hostname component (preserving ports and IP literals) so `repo.example.com.` and `repo.example.com` map to the same config key both directions. Regression test in `config_test.go` (`TestPolicyForURL_TerminalDotHostnameMatchesConfigKey`). See commit on the branch.

bindings/go/wget/checksum/policy.go
```diff
@@ -0,0 +1,182 @@
...
+	return Expected{}, false, nil
+}
+
+// resolveExternalURL returns the checksum URL for src at alg: the explicit
+// [Source.URL] when set, otherwise the Maven default `<baseURL>.<alg.Extension>`.
+func resolveExternalURL(src Source, baseURL string, alg Algorithm) string {
+	if src.URL != "" {
+		return src.URL
+	}
+	return baseURL + "." + alg.Extension
```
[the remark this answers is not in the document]
jakobmoellerdev: Fixed. `resolveExternalURL` now parses `baseURL`, appends the extension to `URL.Path` (or `URL.Opaque` for opaque URLs), and keeps query & fragment via `URL.String()`. Regression test `TestResolve_ExternalDefaultURL_PreservesQuery` in `checksum_test.go` verifies `https://example.com/artifact.tar?download=1` yields `https://example.com/artifact.tar.sha256?download=1`.

matthiasbruns (commented):
Five correctness/security issues to address before merging; details inline.

bindings/go/configuration/checksum/http/v1alpha1/spec/config.go
```diff
@@ -0,0 +1,194 @@
...
+// Validate rejects an unknown [Config.Type]. Nested policies are validated by
+// their consumer.
+func (c *Config) Validate() error {
+	if c == nil {
+		return nil
+	}
+	if c.Type.Name != "" && c.Type.Name != ConfigType {
+		return fmt.Errorf("invalid config type %q, expected %s", c.Type.Name, ConfigType)
+	}
+	return nil
```
matthiasbruns: **[P1] Reject invalid modes without making the field required**

`mode` should remain optional, but explicitly supplied unknown values need validation. Currently a typo such as `Requre` silently disables verification, including when a host override overrides top-level `Require`.

Could we validate the top-level and host modes here? Allow empty for defaulting/inheritance, plus `Require`, `Prefer`, and `Skip`; reject everything else.

bindings/go/wget/checksum/httpverify/verify.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	client := baseClient
+	if credentials != nil {
+		credentialedClient := baseClient
+		if err := download.ApplyCredentials(ctx, req, &credentialedClient, credentials); err != nil {
+			return checksum.Expected{}, false, fmt.Errorf("cannot apply credentials for checksum peek: %w", err)
+		}
+		client = credentialedClient
+	}
+	if resp, herr := client.Do(req); herr == nil {
```
matthiasbruns: **[P1] Honor redirect restrictions on the HEAD path**

The HEAD path ignores the access spec’s `NoRedirect` and can forward authorization across an HTTPS→HTTP redirect. The ordinary download correctly refuses that redirect, but digest discovery follows it.

Please honor the access redirect setting and enforce credential-safe redirects. The documentation currently promises credentialed redirects are disabled.

bindings/go/wget/repository/resource_repository.go
```diff
@@ -94,99 +98,138 @@ func (r *ResourceRepository) GetResourceCredentialConsumerIdentity(ctx context.C
...
-// DownloadResource downloads a resource from the URL specified in the wget access spec.
-// The returned blob is backed by a file under the configured temp folder that outlives
-// this call. The blob owns that file: callers should close it (it implements
-// io.Closer) once they are done, and an unclosed blob has its file removed when it
-// becomes unreachable.
+// DownloadResource downloads the resource. The returned blob owns a temp file
+// callers should close; unclosed blobs have their file removed on GC.
 func (r *ResourceRepository) DownloadResource(ctx context.Context, resource *descriptor.Resource, credentials runtime.Typed) (blob.ReadOnlyBlob, error) {
-	b, err := r.download(ctx, resource, credentials)
+	b, _, err := r.download(ctx, resource, credentials)
```
matthiasbruns: **Apply checksum policy during by-value transfer**

Transfer calls `DownloadResource()` directly, which does not apply the checksum policy. An unpinned resource therefore transfers successfully under `Require` even with a missing or incorrect advertised checksum.

Target-side SHA-256 hashing does not replace source verification. Please enforce the policy on this path too; transfer does not currently “re-run the input-side rules” as documented.

bindings/go/wget/repository/resource_repository.go
```diff
@@ -196,9 +239,86 @@ func (r *ResourceRepository) ProcessResourceDigest(ctx context.Context, resource
...
+		pinnedHex := strings.ToLower(out.Digest.Value)
+		if idx := strings.IndexByte(pinnedHex, ':'); idx >= 0 {
+			pinnedHex = pinnedHex[idx+1:]
+		}
+		if !strings.EqualFold(pinnedHex, exp.Value) {
+			return nil, false, fmt.Errorf("pinned digest %s does not match source-advertised %s digest %s", out.Digest.Value, exp.Algorithm.OCMName, exp.Value)
+		}
+	}
+	out.Digest = &descriptor.Digest{
+		HashAlgorithm:          exp.Algorithm.OCMName,
```
matthiasbruns: **Resolve non-SHA-256 pins before introducing them**

This records SHA-512/SHA-1/MD5 resource digests, but OCI transfer preserves those pins and only supports SHA-256 in its digest mapping. Valid SHA-512 and SHA-1 pins consequently fail at `AddLocalResource`.

We need compatible resource/storage digest handling end to end. The documented automatic conversion to SHA-256 does not happen, and simply replacing the resource digest could affect signatures.

bindings/go/wget/repository/resource_repository.go
```diff
@@ -196,9 +239,86 @@ func (r *ResourceRepository) ProcessResourceDigest(ctx context.Context, resource
...
+	// The reduced configuration accepts every supported algorithm, strongest
+	// first; there is no operator-facing preference knob.
+	prefer := checksum.All
+	names := make([]string, 0, len(prefer))
+	for _, alg := range prefer {
+		names = append(names, alg.OCMName)
+	}
+	slog.DebugContext(ctx, "wget: peeking source-advertised checksum", "url", url, "prefer", names)
+
+	exp, ok, err := httpverify.Peek(ctx, r.client, credentials, url, policy, prefer)
```
matthiasbruns: **[P2] HEAD must describe the same resource as the download**

`Peek()` drops the access headers, verb, and body. A header-selected resource can therefore get the default representation’s digest rather than the digest of the bytes actually downloaded.

Please preserve representation-selecting headers and only use the HEAD fast path where it is equivalent to the configured download request.

bindings/go/wget/checksum/httpverify/verify.go
```diff
@@ -0,0 +1,169 @@
...
+	// the peek credential-safe and consistent with the body download.
+	if req.NoRedirect || credentials != nil {
+		clone := *client
+		clone.CheckRedirect = func(*http.Request, []*http.Request) error {
+			return http.ErrUseLastResponse
+		}
+		client = &clone
+	}
+	if resp, herr := client.Do(httpReq); herr == nil {
+		headers = resp.Header
```
Skarlso: These are the redirects headers, right?

bindings/go/wget/repository/resource_repository.go
```diff
@@ -197,13 +257,102 @@ func (r *ResourceRepository) ProcessResourceDigest(ctx context.Context, resource
...
+func (r *ResourceRepository) processDigestViaPeek(
+	ctx context.Context,
+	resource *descriptor.Resource,
+	credentials runtime.Typed,
+	policy checksum.Policy,
+) (*descriptor.Resource, bool, error) {
+	url := policyURL(resource)
+	// The reduced configuration accepts every supported algorithm, strongest
+	// first; there is no operator-facing preference knob.
+	prefer := checksum.All
```
Skarlso: Run a quick test on this and this will call `x-checksum-md5` and we don't support that further down the line. 🤔 

I will have to write a test to verify the blast radius though. But that's definitely not good here. :D

Skarlso (dismissed):

matthiasbruns (approved):

Skarlso (approved):

